Issue #5, opened 2025-02-27
A combined approach of Go compile time instrumentation


y1yang0:
Move the technical discussion on Slack to GitHub so more people can participate. Thanks.
----

I’m imagining a combined approach to leverage the benefits of both solutions. In this proposal, we would inject lib-side code and bridge the hook code and the target code via function pointers. An ideal workflow would look something like this:
```
func Target() {
    Trampoline()
    ....
}

func Trampoline() {
    Hook()
}

//go:linkname Hook datadog.gopkg/Foo
var Hook func()
```

We inject trampoline code into the Target (lib-side) function, which ultimately jumps to the actual monitoring code via the function pointer Hook.  What do you think?

----
Under this overall workflow, there are some known points that need to be discussed:
1. Why instrument the lib-side(rather than callsite)
    ○ When instrumenting on the library side for different scenarios, such as echo.New() or Go's native net/http, the instrumentation strategy is the same.
    ○ Other auto-instrumentation projects in OpenTelemetry(Such as OpenTelemetry Java Instrumentation, OpenTelemetry Python Instrumentation) use the lib-side as well.
    ○ Providing tracing capabilities by replacing the implementation on the calling side can be entirely achieved on the library side.
    ○ Lib-side instrumentations are much more flexible and powerful than call-side ones, especially when we need to capture some detailed information such as messaging.message.body.size, [network.peer.address](https://opentelemetry.io/docs/specs/semconv/attributes-registry/network/) and so on. It's easier for instrumentations developers to find a suitable aspect (or even multiple aspects) where we can capture all important details with a low cost.
2. Why use `//go:linkname`
By using function pointers along with `//go:linkname`, we dont need to introduce any new dependencies for given package, therefore we don't need to intercept the link-time process or modify the linking commands.
3. Performance overhead
    ○ At 1000 QPS, there will be an additional 1ms latency between applications A and B.

ralf0131:
We have discussed this issue in the bi-weekly meeting. The conclusions are:
1. Go with the library side instrumentation, and make sure we can extend it to support other scenarios like RASP. @RomainMuller Could you provide a example in the RASP that callsite instrumentation has advantage over library implementation?
2. Keep the trampoline.

RomainMuller:
Call-side instrumentation is probably not necessary for RASP (Runtime Application Self-Protection); but might be necessary for IAST (Interactive Application Self-Test) to implement taint tracking.

---

[...]

PR #7, opened 2025-02-28
feat: prototype injection

y1yang0:
@pdelewski @RomainMuller Please review this propotype. Related to #5 

Things are going more smoothly than I expected. Of course, it might be because I haven't encountered the hidden challenges yet.

In the package where the target function resides, we inject the following code:

```go
package p

func Target() {
    Trampoline()
    ....
}

func Trampoline() {
    Hook()
}

func Hook()
```

Since Hook does not have a function body, we must remove the -complete flag to pass compilation.

In the SDK, we use `//go:linkname` to link the actual hook code to the Hook point above:

```go
package sdk

import _ "unsafe"

//go:linkname RealHook p.Hook
func RealHook(){}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Run `sh -x build.sh` to show instrumentation example. In this example, we will i
 If you want to check the result of instrumentation, go to the directory location that appears as output when running build.sh, e.g., WORK=/var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287.
 
 ### Navigate to the WORK directory:
+```bash
 cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287
 
 ### Locate the main package under the b001 subdirectory:
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
 
 ## Getting Started
 
-TBD
+Run `sh -x build.sh` to show instrumentation example. In this example, we will inject a piece of code into the `main` function of the `main` package under the `demo` module to output the "Entering hook" string. This injected code comes from the `sdk` module.
 
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287
 ls -l b001
 
 ### Inspect files:
+```bash
 cat modified.go
 
 
--- a/README.md
+++ b/README.md
@@ -17,20 +17,22 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
 ## Getting Started
 
 Run `sh -x build.sh` to show instrumentation example. In this example, we will inject a piece of code into the `main` function of the `main` package under the `demo` module to output the "Entering hook" string. This injected code comes from the `sdk` module.
-If you want to check the result of instrumentation, go to the directory location that appears as output when running build.sh, e.g., WORK=/var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287.
+If you want to check the result of instrumentation, go to the directory location that appears as output when running build.sh, e.g., `WORK=/var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287`.
 
 ### Navigate to the WORK directory:
 ```bash
 cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287
+```
 
 ### Locate the main package under the b001 subdirectory:
 ```bash
 ls -l b001
+```
 
 ### Inspect files:
 ```bash
 cat modified.go
-
+```
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
 ## Getting Started
 
 Run `sh -x build.sh` to show instrumentation example. In this example, we will inject a piece of code into the `main` function of the `main` package under the `demo` module to output the "Entering hook" string. This injected code comes from the `sdk` module.
+If you want to check the result of instrumentation, go to the directory location that appears as output when running build.sh, e.g., WORK=/var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287.
+
+### Navigate to the WORK directory:
+cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287
+
+### Locate the main package under the b001 subdirectory:
+ls -l b001
+
+### Inspect files:
+cat modified.go
 
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ If you want to check the result of instrumentation, go to the directory location
 cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287
 
 ### Locate the main package under the b001 subdirectory:
+```bash
 ls -l b001
 
 ### Inspect files:
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,8 @@
+# build instrumentation tool
+go build -a -o otel
+TOOL=$(pwd)/otel
+# compile-time instrumentation via toolexec
+cd demo
+rm -rf save.go
+go build -a -toolexec=$TOOL -o demo .
+./demo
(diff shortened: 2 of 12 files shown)


internal/instrument.go
```diff
@@ -0,0 +1,146 @@
...
+					continue
+				}
+				// Instrument the function
+				rewriteAst(ast, fn)
+				// Update the compilation command
+				modified := filepath.Join(findOutputDir(args), "modified.go")
+				storeAst(modified, ast)
+				args[i] = modified
+				break
+				// saveAstToFile("save.go", ast)
```
pdelewski: Seems leftover

build.sh
```diff
@@ -0,0 +1,7 @@
+# build instrumentation tool
+go build -a -o otel
+TOOL=$(pwd)/otel
+# compile-time instrumentation via toolexec
+cd demo
+go build -a -toolexec=$TOOL -o demo .
```
pdelewski: What do you think about adding `work` switch to output compiler working directory location? Then, we could mention that in the Readme
```suggestion
go build -a -work -toolexec=$TOOL -o demo .
```

README.md
```diff
@@ -16,7 +16,7 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
 
 ## Getting Started
 
-TBD
+Run `sh -x build.sh` to show instrumentation example. In this example, we will inject a piece of code into the `main` function of the `main` package under the `demo` module to output the "Entering hook" string. This injected code comes from the `sdk` module.
```
pdelewski:
```suggestion
If you want to check the result of instrumentation, go to the directory location that appears as output when running build.sh, e.g., WORK=/var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287.

### Navigate to the WORK directory:
cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287

### Locate the main package under the b001 subdirectory:
ls -l b001

### Inspect files:
cat modified.go

```

y1yang0: This makes it easier to further explore the details of the injection. That does make sense.

README.md
```diff
@@ -16,7 +16,17 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
...
+If you want to check the result of instrumentation, go to the directory location that appears as output when running build.sh, e.g., WORK=/var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287.
```
pdelewski: Minor formatting improvements
```suggestion

If you want to check the result of instrumentation, go to the directory location that appears as output when running build.sh, e.g., `WORK=/var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287`.
```

README.md
```diff
@@ -16,7 +16,17 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
...
+
+### Navigate to the WORK directory:
+cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287
```
pdelewski, line 24:
```suggestion
```bash
cd /var/folders/x9/fddsvlt5363c0plvvw8_2mr80000gn/T/go-build2020695287
```
```

File: README.md [hunk 6]
```diff
@@ -16,7 +16,17 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
...
+
+### Locate the main package under the b001 subdirectory:
+ls -l b001
```
pdelewski, line 29:
```suggestion
```bash
ls -l b001
```
```

File: README.md [hunk 7]
```diff
@@ -16,7 +16,17 @@ OpenTelemetry-Go-Compile-Instrumentation provides compile time  [OpenTelemetry](
...
+
+### Inspect files:
+cat modified.go
```
pdelewski, line 34:
```suggestion
```bash
cat modified.go
```
```

**Comment by pdelewski** [comment]
Looks like good starting point!

**Review by pdelewski — APPROVED** [review]

**Review by NameHaibinZhang — APPROVED** [review]

File: internal/instrument.go [hunk 8]
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"fmt"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
```
123liuziming, line 15: I think we need ut for this

pdelewski: @123liuziming  Do you mean unit test?

123liuziming: > @123liuziming Do you mean unit test?

Yes

y1yang0: This is a prototype implementation. Let's focus on the most critical injection mechanism first, and then add unit tests.

demo/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/open-telemetry/opentelemetry-go-compile-instrumentation/demo
+
+go 1.22.0
```
RomainMuller: Not too strong an opinion here, but I'd say we probably want to support only the 2 most recent go majors (1.23 and 1.24 currently), especially at the inception... I kind of tend to consider this is community service (gently nudging people towards more recent runtimes allows more folks to use the newer features).

internal/instrument.go
```diff
@@ -0,0 +1,145 @@
...
+	file, err := os.Open(filePath)
+	if err != nil {
+		panic(err)
+	}
+	astFile, err := parser.ParseFile(fset, name, file, parser.ParseComments)
+	if err != nil {
+		panic(err)
+	}
+	dec := decorator.NewDecorator(fset)
+	dstFile, err := dec.DecorateFile(astFile)
```
RomainMuller, line 37: It would be neat if we could avoid taking the dependency on `dave/dst` (it does a lot beyond what is necessary to do our deed here; and adds a non-negligible allocations load).

main.go
```diff
@@ -0,0 +1,47 @@
...
+
+func isCompilePackage(args []string, pkg string) bool {
+	checks := []string{"compile", "-buildid", "-p " + pkg}
+	for _, check := range checks {
+		if !strings.Contains(strings.Join(args, " "), check) {
+			return false
+		}
+	}
+	return true
+}
```
RomainMuller: This I would do somewhat differently... In essence the only thing that is relevant here is:
- `argv[0]` ends in `compile` (or `compile.exe` on Windows)
- `argv[1:]` does not include `-V=full` (alternatively, there is `-buildid` and `-p`)

The repeated call to `strings.Join` here is not ideal (and joining should not be necessary at all)

RomainMuller (commented):
At the moment, this would result in tainting the go build cache with instrumented objects (and possibly not re-instrument them if the instrumentation instructions change).

internal/instrument.go
pdelewski: I agree with that. Personally, I'm fan of depending only on standard library, considering very carefully third party dependencies.

demo/go.mod
y1yang0: Changed to 1.23

internal/instrument.go
y1yang0: But without `dst`, how do we insert code at the correct position? As far as I know the `go/ast` package wasn't created with source manipulation as an intended use-case. Comments are stored by their byte offset instead of attached to nodes, so re-arranging nodes breaks the output and confuses the debugger.

pdelewski: True, that there is a problem with comments and in consequence it might confuse the debugger. That's strong argument.

y1yang0: So let's keep the status quo. If needed later, we can trim down to a minimal version of dst that doesn't rely on any packages outside the standard library.

RomainMuller: @y1yang0 you can modify the AST with `go/ast` in pretty much the same way that you do it with `dave/dst`; just by modifying nodes or inserting new ones, and then you use `go/printer` to turn it back into go source.

The main thing that `dave/dst` makes easier is managing comments (such as `//go:linkname` pragmas). The way you've set things up in this prototype actually only uses `//go:linkname` on the `sdk` side, which I reckon contains only synthetic nodes or code that we have laid out previously so we may not even need to deal with comment placement at all.

pdelewski: @RomainMuller Well, we can do everything with AST using `go/ast`, that's for sure. The challenge lies in building a mapping between the input source code and the output. When modifying AST nodes, tracking the relationship between the original and transformed positions can be difficult. Tools like `token.Pos` and `go/token` help but may not cover all cases. Actually, I'm not sure whether `dave/dst` offers any advantages in this regard, as I don't have experience with it.

y1yang0: @RomainMuller As @pdelewski  said, go/ast cannot handle spaces/comments correctly, generating incorrect line numbers/column numbers, which can confuse the debugger.

pdelewski: I think it's time to merge it. We can revisit some decisions later if needed.

PR #10, opened 2025-03-25
[WIP] SDK encapsulation and basic CI

123liuziming:


pdelewski:
This PR introduces a lot of new code. I envisioned that we could start with a minimal approach, providing only what the current implementation requires, and gradually expand it over time, evolving tests and CI as needed.

123liuziming:
> This PR introduces a lot of new code. I envisioned that we could start with a minimal approach, providing only what the current implementation requires, and gradually expand it over time, evolving tests and CI as needed.

We just want the project to follow OpenTelemetry Semconv, so I think we can encapsulate the OpenTelemetry SDK first so once the compilation framework is ready, it can just use the `instrumentation-api` and `instrumentation-api-semconv` to generate span and metric that follows the OpenTelemetry Semconv.

PR #11, opened 2025-03-26
docs: user experience design brief

RomainMuller:
Bootstrapping the UX design brief.


_docs/ux-design.md
```diff
@@ -0,0 +1,138 @@
...
+  a small, self-contained package that contains essential runtime functionality
+  used by instrumented applications (not intended for manual usage).
+
+### Getting Started
+
+The tool offers a wizard for getting started with the tool automatically by
+following a series of prompts:
+
+```console
+$ go run github.com/open-telemetry/opentelemetry-go-compile-instrumentation/cmd/gotel@latest setup
```
pdelewski: That's a similar approach to what we have done for PHP some time ago in [this repository](https://github.com/open-telemetry/opentelemetry-php-contrib/tree/main/src/AutoInstrumentationInstaller). Nice to see a similar thought process.

_docs/go.mod
```diff
@@ -0,0 +1,7 @@
+// This module is not intended to be published nor used. This go.mod file is
```
123liuziming, line 4:
```suggestion
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0
// This module is not intended to be published nor used. This go.mod file is
```

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+   package(s) requires it.
+🤖 Should I add gotel as a tool dependency?
+   [Yes]  No
+🆗 I will add a tool dependency
+
+ℹ️ You may enable one or more instrumentation packages for your application.
+   Most users need to choose only one. This can be changed at any time.
+🤖 What instrumentation do you want to enable for this project? (Select one or
+   more using space, then press enter to confirm your selection)
+   [X] OpenTelemetry (github.com/open-telemetry/opentelemetry-go)
```
123liuziming: I think we should only provide OpenTelemetry instrumentation here. Vendor-related content probably shouldn't be here

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+```console
+$ go run github.com/open-telemetry/opentelemetry-go-compile-instrumentation/cmd/gotel@latest setup
+╭──────────────────────────────────────────────────────────────────────╮
+│                                                                      │
+│           OpenTelemetry compile-time instrumentation tool            │
+│                    v1.2.3 (go1.24.1 darwin/arm64)                    │
+│                                                                      │
+╰──────────────────────────────────────────────────────────────────────╯
+✨ This setup assistant will guide you through the steps needed to properly
+   configure gotel for the github.com/example/demo application. After you've
```
pdelewski:
```suggestion
   configure gotel for your project. After you've
```

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+   answered all the prompts, it will present you with the list of actions it
+   will execute; and you will have an choice to apply those or not.
+🤖 Press enter to continue...
+
+ℹ️ Registering gotel as a tool dependency is recommended: it allows you to
+   manage the dependency on gotel like any other dependency of your application,
+   via the go.mod file. When using go 1.24 or newer, you can use `go tool gotel`
```
pdelewski: Just for reference
```suggestion
   via the go.mod file. When using [go 1.24](https://go.dev/doc/modules/managing-dependencies#tools) or newer, you can use `go tool gotel`
```

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+To allow for this, the `gotel` tool allows managing configuration in several
+ways:
+
+1. Using `tool` dependencies (requires `go1.24` or newer) allows the tool to be
+   used by the simpler invocation `go tool gotel`, and has all instrumentation
+   configuration made available by `tool` dependencies, without requiring the
+   addition of any new `.go` or `.yml` file;
+
+2. Using a `gotel.instrumentation.go` file is an alternate strategy that is
+   fully supported by `go1.23`, and allows more direct control over what
```
pdelewski, line 171: Not clear to me, why it's fully supported as of `1.23`?

_docs/ux-design.md
pdelewski: Yes, good point.

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+instrumentation, does not affort users the same amount of control over their
+instrumentation as manual instrumentation; it trades very granual control for
+significantly reduced implementation effort. As a result, compile-time
+instrumentation may not appeal to developers who have very specific requirements
+on what their instrumentation produces.
+
+The primary audience for the OpenTelementry Go compile-time instrumentation tool
+is composed of the following personas:
+
+- Application developers looking for a no-frills, turney instrumentation
```
kakkoyun:
```suggestion
- Application developers looking for a no-frills, turnkey instrumentation
```

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+
+The main value proposition is:
+- Very low effort required for holistic instrumentation
+- Ability to instrument within third-party dependencies
+- Keeping the codebase completely de-coupled from instrumentation
+
+## Target Audience
+
+Compile-time instrumentation, like other techniques of automatic
+instrumentation, does not affort users the same amount of control over their
```
kakkoyun:
```suggestion
instrumentation, does not afford users the same amount of control over their
```

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+### Building Applications
+
+Once the configuration has been created, either by the command-line assistant,
+or directly by the user (possibly through automated processes), the tool can be
+used directly to build, run, and test go applications directly:
+
+1. If the tool is installed as a Go `tool` dependency (`go1.24` and newer):
+
+   ```console
+   $ go tool gotel build -o bin/app .
```
y1yang0: Regarding the name, "go tool gootel" seems to redundantly repeat "go." Perhaps we can simply use "otel" (or another name)?

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+   $ go tool gotel test -shuffle=on ./...
+   ```
+
+2. Installing `gotel` in `$GOBIN`
+
+   ```console
+   $ go install github.com/open-telemetry/opentelemetry-go-compile-instrumentation/cmd/gotel
+   $ gotel build -o bin/app
+   $ gotel test -shuffle=on ./...
```
y1yang0: For users, a more intuitive approach might be to directly use the `otel` tool as a prefix to the original commands. For instance, if the original command is `go build`, we could now use `otel go build`. What do you think? Additionally, apart from the usage patterns already listed, supporting the use of the `otel` tool via the `GO_TOOLEXEC` parameter is also an appealing option.

y1yang0 (commented):
Generally looks good, I left two comments.

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+package all
+
+import (
+   _ "github.com/open-telemetry/opentelemetry-go/gotel/instrumentation/net-http"
+   _ "github.com/open-telemetry/opentelemetry-go/gotel/instrumentation/database-sql"
+   // ...
+)
+```
+
+Using a `.go` file with `import` declarations allows to make sure the surrouding
```
y1yang0, line 378: There is a relatively critical issue: in `gotel.instrumentation.go`, we introduced a dependency, for example, pkga v2. At the same time, if users proactively depend on pkga v1, this will cause their dependency to be forcibly updated (this is not a hypothetical situation—I actually encountered this issue in a brand-new implementation). As a result, the compilation will fail. Additionally, since many of our users are operations developers who only use this tool in their CI pipelines, they do not have the capability to fix such errors. This is a major concern for us because it would prevent a large number of users from being able to practically use this tool. However, to be honest, I don’t have a better solution at the moment. Perhaps the SDK code needs to vendor dependencies in advance to ensure that no external dependencies are introduced besides otel-sdk? I’m not sure. Maybe for now, we can only maintain the current state.

y1yang0:
The issue regarding dependency updates actually points to one of our design philosophies: **we want the automatic injection tool to be non-intrusive, fully automated, and not require users to perform any additional actions**—just as users themselves would expect.

1. We aim to simplify user operations as much as possible: the only thing users need to do is add `otel` as a prefix to their original `go build` command. Of course, I agree that we can provide an additional wizard for new users who want to try it out, but perhaps by default, we don’t have to display it.

2. We want the tool to be non-intrusive: we do not retain generated files like `gotel.instrument.go`, nor do we want to modify the user's `go.mod` file or require users to join a service (svc) to ensure reproducible builds. However, in the current implementation, we temporarily don’t have a better solution for this.

_docs/ux-design.md
RomainMuller: Yeah via `go tool` it's a little awkward I agree... I'd say `go tool otel` is fine by me, just would need to check if this is fine by OpenTelemetry's preferences (as this commingles the project & this specific tool).

_docs/ux-design.md
RomainMuller: Yah this is what we do for `orchestrion`. My rationale here was if the tool is named `gotel` then there is a stutter `gotel go`; but with a different name it works nicer.

_docs/ux-design.md
RomainMuller: I am guessing you're not referring to major versions here (as these can coexist); rather different minors/patches...

Since these changes are done coding-time, the user's local `go.mod` file shows the fully resolved closure and they can make decisions on their own based on that...

Vendoring dependencies could be a way to avoid pushing this on the user, but it'll probably cause more bloat (more packages need to be built = compile time overhead + binary becomes larger). Tradeoffs, tradeoffs...

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+- directly in the `go.mod` file by adding or removing `tool` dependencies,
+- in the `gotel.instrumentation.go` file by adding or removing `import`
+  declarations.
+
+### Custom Configuration
+
+Users may wish to add their own, application-specific automatic instrumentation
+configuration. This is achieved by adding a `.gotel.yml` file in the same
+directory as the application's `go.mod` file. Such instructions will be used
+only when building packages in this module's context (and not when themodule's
```
ralf0131:
```suggestion
only when building packages in this module's context (and not when the module's
```

_docs/ux-design.md
ralf0131: +1. We may provide a way for downstream projects to customize this process.

_docs/ux-design.md
ralf0131: I am also fine with `go tool otel`. It is easy to understand with `otel` following `go tool`.

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Instrumentation Packages
+
+A majority of users of the OpenTelemetry compile-time instrumentation tool will
+rely on instrumentation packages to instrument their application. These are
+standard Go packages that are part of a Go module and contain either (or both):
+
+- a `gotel.instrumentation.yml` file that declares all instrumentation
+  configuration that is vended by this package;
+- a `gotel.instrumentation.go` file that imports at least one valid
```
ralf0131: Could you elaborate more about how this file may looks like (Since you have introduced `gotel.instrumentation.yml`)?

ralf0131:
Also, this branch is out of date, could you please update with the base branch?

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+The main value proposition is:
+- Very low effort required for holistic instrumentation
+- Ability to instrument within third-party dependencies
+- Keeping the codebase completely de-coupled from instrumentation
+
+## Target Audience
+
+Compile-time instrumentation, like other techniques of automatic
+instrumentation, does not afford users the same amount of control over their
+instrumentation as manual instrumentation; it trades very granual control for
```
darccio: Did we mean `granular` or `gradual`?

_docs/ux-design.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Intended Purpose
+
+The OpenTelemetry Go compile-time instrumentation tool allows users to
+instrument their full application automatically, without having to perform any
+significant code change beyond adding simple, well-defined configuration that
+may be as simple as adding the relevant tool dependencies.
+
+The main value proposition is:
+- Very low effort required for holistic instrumentation
```
darccio: (nit)

```suggestion
- Very little effort is required for holistic instrumentation
```

darccio (changes requested):

_docs/ux-design.md
RomainMuller: This is an alternative to using `go tool` dependencies, which is 1.24+ only... So the non-tool dependencies strategy is useful before that.

y1yang0:
We summary two alternative user interaction approaches for code injection:

1. Single-Command Injection

```bash
otel go build
```

All workflow steps are encapsulated in a single command. Users only need to prepend the otel prefix (or use the method mentioned in documentation) to their original build command. e.g. :
1. Discover hook schema
2. Identify dependencies
3. Generate otel.instrumentation.go
4. Update go.mod
5. Execute real build
6. Perform file cleanup

All in one.

2. Multi-Stage Injection

```bash
otel setup
otel go build
otel cleanup
```
As discussed in the document, after otel setup, users have the option to manually add generated files (otel.instrumentation.go, go.mod etc) to their SVC and commit them to SVC.

Which interaction pattern do you prefer?

_docs/ux-design.md
pdelewski: Should not we add a comment about that?

RomainMuller:
I think my preference is to have the setup/build, and possibly an optional cleanup. The setup (+ ability to check in artifacts) is essential to enable fully reproductible builds... The cleanup may not be necessary: if you're doing setup/build in CI, the environment is inherently transient so you need not "clean up".

Reproductible builds are a compliance requirement in many highly regulated businesses (management of payment info, health, ...), and not enabling this would basically make it harder or impossible for these users to consume this tool.

y1yang0:
Thanks for your thoughtful perspective(especial regulated business). I gently suggest we consider supporting both the setup/build/cleanup mode and a single-command mode—since not all users are in CI workflows (many test locally before CI integration), and we’ve received feedback that our tool unexpectedly modifies users’ code(before automatically cleanup). Offering cleanup gives users with different needs a choice: those prioritizing reproducibility can skip cleanup, while those seeking non-intrusive code can use it, and the single command is just a streamlined fusion of the three steps.

ralf0131:
> Thanks for your thoughtful perspective(especial regulated business). I gently suggest we consider supporting both the setup/build/cleanup mode and a single-command mode—since not all users are in CI workflows (many test locally before CI integration), and we’ve received feedback that our tool unexpectedly modifies users’ code(before automatically cleanup). Offering cleanup gives users with different needs a choice: those prioritizing reproducibility can skip cleanup, while those seeking non-intrusive code can use it, and the single command is just a streamlined fusion of the three steps.

I agree that we do support both of the two modes.

ralf0131 (approved):

y1yang0 (approved):

_docs/ux-design.md
```diff
@@ -0,0 +1,383 @@
...
+            instrumentation package. These may be presented to the users when
+            they install this package for the first time.
+
+instrumentation: # Required with at least 1 item
+   foo: # A unique identifier for this instrumentation item within this file
+      description: #Optional
+         |-
+            A decription of this instrumentation configuration, intended for
+            end-users.
+      pointcut: # Required
```
y1yang0, line 308: I overlooked an important point: regarding this pointcut, I believe we should support Go source code rather than AST node manipulation. Maybe something like

```
      pointcut: github.com/open-telemetry/opentelemetry-go-compile-instrumentation/sdk/rules/gorm
```

Otherwise looks good to me.

y1yang0 (changes requested):

_docs/ux-design.md
RomainMuller: The pointcut is to determine _where_ to inject... Are you meaning to evaluate arbitrary Go code as part of rules evaluation? I'm not too sure how we'd pull this out (safely)?

PR #14, opened 2025-04-03
docs: add implementation detail document

y1yang0:
Add implementation detail document


_docs/implementation.md
```diff
@@ -0,0 +1,104 @@
+# Introduction
+This proposal outlines a method for injecting runtime hooks into target functions in Go programs, enabling dynamic monitoring and modification of function behavior. The approach leverages trampoline code injection and function pointer redirection to seamlessly integrate monitoring logic without requiring significant changes to the user's codebase.
+
+The goal is to provide a flexible and non-intrusive mechanism for instrumenting Go applications, particularly for use cases such as observability (e.g., OpenTelemetry), debugging, or performance profiling.
+
+# Core Principles
+## 1. Trampoline Code Injection
+We inject trampoline code into the Target (lib-side) function, which ultimately jumps to the actual monitoring code via the function pointer Hook.
```
pdelewski: To be precise, it's trampoline function call
```suggestion
We inject trampoline function call into the Target (lib-side) function, which ultimately jumps to the actual monitoring code via the function pointer Hook.
```

pdelewski: Just to avoid additional questions and confusions, I think it would be good to explain why are we doing two-level indirection, first `Trampoline` call, then hook, instead of let's say directly injecting opentelemetry code.

_docs/implementation.md
```diff
@@ -0,0 +1,104 @@
...
+
+//go:linkname Hook github.com/open-telemetry/opentelemetry-go-compile-instrumentation/sdk/hook.MyHook
+var Hook func()
+```
+
+## 2. Linkage via golinkname
+The `Hook` function is linked to the monitoring code using the `//go:linkname` directive. This allows us to dynamically associate the hook with the target function at compile time.
+
+# Implementation Details
+## 1. Adding Dependencies
```
pdelewski: From my perspective, go compile time instrumentation is two-phase process. First pass is about updating dependencies, second about building project (via some kind of proxy tool)

y1yang0: Using a two-phase narrative is a good idea. I revised the design doc based on the comments. Please help review it when you have time. Thanks!

_docs/implementation.md
RomainMuller, line 1: GitHub's UI is not so great for reviewing prose (you can't comment on just certain words; have to comment on the full line).

It can make things a little easier to flow documents at a fixed width, so long sentences are broken down on multiple lines...

I'd suggest (completely optional) flowing Markdown at 100 columns (or 80; but I usually do 100 myself).
This means just putting a hard-line-break before the 1st word that crosses the line.
Many IDEs have auto-formatters that'll help do this automatically; and we could enforce this with `markdownlint` or similar.

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
+# Introduction
+This proposal outlines a method for injecting runtime hooks into target functions in Go programs, enabling dynamic monitoring and modification of function behavior. The approach leverages trampoline code injection and function pointer redirection to seamlessly integrate monitoring logic without requiring significant changes to the user's codebase.
```
RomainMuller: I would phrase this not as "a proposal", but rather a description of what is (to be) done. I think of this kind of document as part of the project/product's documentation, and pre-implementation it serves as a guide (what we are doing? how? why?), and post-implementation it serves as documentation to someone new to the project or a curious user.

y1yang0:
Do you have any further comments? @pdelewski @RomainMuller

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+
+
+## 2. Linkage via golinkname
+The `Hook` function is linked to the monitoring code using the `//go:linkname` directive. This allows us to dynamically associate the hook with the target function at compile time.
+
+# Implementation Details
+Go compile-time instrumentation is a two-phase process: the first phase involves updating dependencies, and the second phase focuses on building the project using a custom toolchain.
+
+## Phase 1: Adding Dependencies
+To prepare the project for instrumentation, we first integrate the necessary dependencies into the user's codebase. This step ensures that the required hook code is available during the build process.
```
pdelewski: I think it would be good to clarify the meaning of 'we' in this description. Does it mean that the user should modify the sources manually, or will it be done automatically by the toolchain?

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+```go
+// otel_importer.go
+package main
+
+import _ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/sdk"
+```
+
+This import statement ensures that the SDK, which contains the hook code, is included in the project. The `_` prefix indicates that the package is imported only for its side effects (e.g., registration of hooks).
+
+After adding the dependency, we run `go mod tidy` to update the `go.mod` file. This step ensures that the dependency is properly recorded and downloaded into the module cache.
```
pdelewski: Same here — who is actually running `go mod tidy`?

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+By completing this phase, the project is now ready for the next step, where the instrumentation logic will be applied during the build process.
+
+## Phase 2: Building with Instrumentation
+The build process is invoked with a custom toolchain by specifying the `-toolexec=otel` flag
+
+```bash
+go build -toolexec=otel
+```
+
+The `-toolexec=otel` flag specifies a custom tool (e.g., otel) that intercepts compilation commands. We find the target function from compilation commands and inject trampoline code into the AST of these functions. Since the hook dependency was already imported in Phase 1, we can link the target function to the hook code via golinkname without any other modifications.
```
pdelewski: Maybe
```suggestion
The `-toolexec=otel` flag specifies a custom tool (e.g., otel) that intercepts compilation commands. The tool identifies the target function from the compilation commands and injects trampoline code into the AST of these functions. Since the hook dependency was already imported in Phase 1, the tool can link the target function to the hook code via //go:linkname without requiring any additional modifications.
```

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+	SetReturnVal(idx int, val interface{})
+	// Get the original function name
+	GetFuncName() string
+	// Get the package name of the original function
+	GetPackageName() string
+}
+```
+
+Key Features:
+- Parameter Access and Modification: Allows inspection and alteration of function arguments.
```
pdelewski:
```suggestion
- Parameter Access and Modification: Allows for the inspection and modification of function arguments.
```

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+- Return Value Manipulation: Enables modification of return values before they are passed back to the caller.
```
pdelewski:
```suggestion
- Return Value Manipulation: Enables the modification of return values before they are returned to the caller.
```

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+}
+func MyHookAfter(ctx Context) {
+	msg := ctx.GetData("msg")
+}
+```
+
+## 2. Full Context API
+The full context API is listed below.
+
+```go
```
pdelewski, line 147: Consider extracting this into a source file and linking to it instead.

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+import _ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/sdk"
+```
+
+This import statement ensures that the SDK, which contains the hook code, is included in the project. The `_` prefix indicates that the package is imported only for its side effects (e.g., registration of hooks).
+
+After adding the dependency, we run `go mod tidy` to update the `go.mod` file. This step ensures that the dependency is properly recorded and downloaded into the module cache.
+By completing this phase, the project is now ready for the next step, where the instrumentation logic will be applied during the build process.
+
+## Phase 2: Building with Instrumentation
+The build process is invoked with a custom toolchain by specifying the `-toolexec=otel` flag
```
pdelewski: In https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/11 we talk about `gotel`. I think we should be consistent about naming

_docs/implementation.md
y1yang0: Good catch! I will remove the first-person pronoun "we" and use a more neutral expression.

_docs/implementation.md
y1yang0: Commented in that PR regarding the name

_docs/implementation.md
y1yang0: Jumping to a new file here might disrupt the flow of thought for users while reading the documentation. Perhaps it would be more convenient to understand if the API code is displayed directly here. Once we have the actual API code, it makes sense(to me) to add a new link

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+
+//go:linkname Hook github.com/open-telemetry/opentelemetry-go-compile-instrumentation/sdk/hook.MyHook
+var Hook func()
+```
+The two-level indirection (trampoline call to hook) is a deliberate design choice that offers several key benefits:
+
+- Exception Handling: The trampoline catches panics and isolates exception handling, preventing them from affecting the target function or hook code.
+- Context Construction: The trampoline initializes and manages the necessary context before invoking the hook code.
+- Decoupling: The trampoline decouples the hook code from the target function, enabling flexibility and maintainability.
+- Debugging: The trampoline provides a centralized point for debugging and observability.
```
RomainMuller: Would be nice to move this up before the example...

I'd add the trampoline allows dynamic instrumentation (that can be turned on/off at run-time).

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+
+
+## 2. Linkage via golinkname
+The `Hook` function is linked to the monitoring code using the `//go:linkname` directive. This allows us to dynamically associate the hook with the target function at compile time.
```
RomainMuller: I'd clarify this is putting us in the "handshake" style of `//go:linkname` (this is using language from the maintainers), which is expected to continue being supported indefinitely (as opposed to "pull", which is blocked outside of grandfathered-in exceptions)

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Implementation Details
+Go compile-time instrumentation is a two-phase process: the first phase involves updating dependencies, and the second phase focuses on building the project using a custom toolchain.
```
RomainMuller: Updating dependencies is done as part of set-up, isn't it? It would be nice to perhaps have a schema explaining what happens at what time ("setup/coding time" versus "compilation time" versus "run time" [for illustration, since we do nothing at run time])

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+// otel_importer.go
+package main
+
+import _ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/sdk"
+```
+
+This import statement ensures that the SDK, which contains the hook code, is included in the project. The `_` prefix indicates that the package is imported only for its side effects (e.g., registration of hooks).
+
+After adding the dependency, `go mod tidy` is run to update the `go.mod` file. This step ensures that the dependency is properly recorded and downloaded into the module cache.
+By completing this phase, the project is now ready for the next step, where the instrumentation logic will be applied during the build process.
```
RomainMuller: This is a place where we can possibly cross-reference with #11.

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Phase 2: Building with Instrumentation
+The build process is invoked with a custom toolchain by specifying the `-toolexec=otel` flag
+
+```bash
+go build -toolexec=otel
+```
```
RomainMuller: I wonder if we should explicitly support folks specifying `-toolexec` "by hand" or if we should wrap `go build` entirely (and manage it behind the scenes). It can be useful to have a way to maintain state throughout the entire build, and that is made significantly easier if we control the root process...

I'm not entirely certain this is necessary, though...

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+	ctx.GetParam(1)
+	ctx.SetParam(1, "new value")
+	ctx.GetReturnValue(1)
+	ctx.SetReturnValue(1, "new value")
+	ctx.SetData("msg", "hello world")
+}
+func MyHookAfter(ctx Context) {
+	msg := ctx.GetData("msg")
+}
+```
```
RomainMuller, line 141: As mentioned when we discussed during the SIG call... I see why you've done it this way (more general, etc...); but I am somewhat bothered (not necessarily blocking) by how this API essentially relies heavily on `any` and manual type assertions in the hook code... I would like if we had a solution that did not require this...

I am thinking as we are generating code that'll be built with the app, we could produce a type-safe context object, that may implement the interface you describe above as well... so a hook that is used ONLY for a given function signature receives strongly typed inputs... and one that is polymorphic or signature-agnostic works off the interface?

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+
+## 2. Full Context API
+The full context API is listed below.
+
+```go
+type Context interface {
+	// Skip the original function call
+	SetSkipCall(bool)
+	// Check if the original function call should be skipped
+	IsSkipCall() bool
```
RomainMuller: I think `SetSkipCall(bool)` is a little odd... Observability use-cases most probably don't need this...

Security use-cases may need this to early-return with specific results instead of running the original code; and so at this point it may be more suitable to have a "return these values immediately" call on the BeforeHook (skip is meaningless in after hook) -- `ctx.ImmediatelyReturn(results...)`?

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+	// Set the data field, can be used to pass information between OnEnter & OnExit
+	SetData(interface{})
+	// Get the data field, can be used to pass information between OnEnter & OnExit
+	GetData() interface{}
+	// Get the map data field by key
+	GetKeyData(key string) interface{}
+	// Set the map data field by key
+	SetKeyData(key string, val interface{})
```
RomainMuller, line 156: I feel like these could be extracted out so it's "more obvious" they are not about accessing/manipulating the current call's information, rather about managing hook state.

_docs/implementation.md
```diff
@@ -0,0 +1,121 @@
...
+	// Has the map data field by key
+	HasKeyData(key string) bool
+	// Get the original function parameter at index idx
+	GetParam(idx int) interface{}
+	// Change the original function parameter at index idx
+	SetParam(idx int, val interface{})
+	// Get the original function return value at index idx
+	GetReturnVal(idx int) interface{}
+	// Change the original function return value at index idx
+	SetReturnVal(idx int, val interface{})
```
RomainMuller: This is missing accessors to inspect the number of available arguments/results. That could be useful for hooks that inspect everything for arbitrary signatures.

_docs/implementation.md
y1yang0: Let's remove it.

_docs/implementation.md
y1yang0: Yes, the approach I described here was only to simplify the initial implementation. In fact, there are at least two ways to avoid redundant type assertions(in my mind):

1. Declare the types in the hook function parameters:
```go
// plugin developers:
func MyHookBefore(ctx Context, param1 int, param2 Request) {
	...
}
func MyHookAfter(ctx Context, ret1 string, ret2 error) {
	...
}
```

This is the approach used in our current repository (alibaba-otel), and it is proven to be feasible.

2. Generate methods for the Context object:

```go
// otel generated:
func (ctx *MyHookContext) GetParam1() int {...}
func (ctx *MyHookContext) GetParam2() double {...}

func (ctx *MyHook2Context) GetParam1() double {...}
func (ctx *MyHook2Context) GetParam2() Request {...}


// plugin developers:
type MyHookContext {} // implements all methods of Context

func (ctx *MyHookContext) MyHookBefore() {
	ctx.GetParam1() // type of int
}
func (ctx *MyHook2Context)  MyHook2Before() {
	ctx.GetParam1() // type of double
}
```
This also seems feasible, but it requires plugin developers to write more code. Additionally, the IDE will complain that `ctx.GetParam1` cannot be found because this method is dynamically generated by the otel tool during compilation.

_docs/implementation.md
y1yang0: Not quite sure if the "extracted out" you mentioned means removing them as part of the context API? If that's the case, then that's okay to me

_docs/implementation.md
y1yang0: I prefer to support both `otel go build` and `go build -toolexec=otel`. Perhaps I should explicitly mention in the documentation that the motivation for supporting `go build -toolexec=otel` is that it allows us to directly append `-toolexec=otel` via the `GOFLAGS` environment variable, without having to modify the build command (`go build`).

_docs/implementation.md
```diff
@@ -0,0 +1,171 @@
...
+function to the hook code via //go:linkname without requiring any additional
+modifications.
+
+
+# Interface Design
+
+## 1. Context
+
+The Context interface is designed to provide a structured way to access and
+manipulate the parameters, return values, and other relevant data of the target
```
ralf0131, line 123: It is better to specify more scenarios for users to better understand better. For example: parameters, return values, exceptions, panics, goroutine local storage and etc.

_docs/implementation.md
ralf0131: @y1yang0 +1 to support both of the usage. Would you please add it to the doc?

_docs/implementation.md
```diff
@@ -0,0 +1,171 @@
...
+
+This document outlines a method for injecting runtime hooks into target functions
+in Go programs, enabling dynamic monitoring and modification of function behavior.
+The approach leverages trampoline code injection and function pointer redirection
+to seamlessly integrate monitoring logic without requiring significant changes to
+the user's codebase.
+
+The goal is to provide a flexible and non-intrusive mechanism for instrumenting
+Go applications, particularly for use cases such as observability (e.g., OpenTelemetry),
+debugging, or performance profiling.
```
ralf0131: Maybe we can mention security as well?

y1yang0: Done

_docs/implementation.md
y1yang0: > @y1yang0 +1 to support both of the usage. Would you please add it to the doc?

Done

_docs/implementation.md
y1yang0: the description already contains `parameters, return values`, `exceptions, panics, goroutine local storage` are unrelated stuff.

_docs/implementation.md
RomainMuller: That could possibly be worked around by wrapping the LSP (but that sounds like a lot of work)

_docs/implementation.md
RomainMuller: I mean something like isolating them in a dedicated interface, or just use a plain context.Context object for this? Either way I think you did get my point there.

_docs/implementation.md
```diff
@@ -0,0 +1,183 @@
...
+
+This document outlines a method for injecting runtime hooks into target functions
+in Go programs, enabling dynamic monitoring and modification of function behavior.
+The approach leverages trampoline code injection and function pointer redirection
+to seamlessly integrate monitoring logic without requiring significant changes to
+the user's codebase.
+
+The goal is to provide a flexible and non-intrusive mechanism for instrumenting
+Go applications, particularly for use cases such as observability (e.g., OpenTelemetry),
+debugging, security or performance profiling.
```
RomainMuller, line 11: Like this paragraph - very good overview of the target domain!

_docs/implementation.md
```diff
@@ -0,0 +1,183 @@
...
+Go compile-time instrumentation is a two-phase process: the first phase involves
+setup dependencies, and the second phase focuses on building the project using
+a custom toolchain.
+
+## Phase 1: Setup Dependencies
+
+### 1.1 Dependency Analysis
+The first step is to analyze the project's dependencies by collecting the list 
+of modules involved in the build. This is done using the `go build -n` command, 
+which prints the build plan without executing it.
```
RomainMuller, line 68: This feels a little heavy handed to me. Using go list is sufficient to determine the complete dependency closure of the project without involving as much toolchain overhead as running a dry-run build, and also does not require parsing the output of go build -n, which is not designed/intended for being parsed; and is not part of the go compatibility guarantee.

_docs/implementation.md
```diff
@@ -0,0 +1,183 @@
...
+the module cache.
+By completing this phase, the project is now ready for the next step, where the
+instrumentation logic will be applied during the build process.
+
+## Phase 2: Building with Instrumentation
+
+The build process can be integrated with custom toolchains in the following ways:
+
+1. Command Prefix: `otel go build` (simple but requires manual prefixing)
+2. Environment Variable: `GOFLAGS=-toolexec=otel go build` (global effect; no per-command setup)
```
RomainMuller: I think the command there would be different

```suggestion
2. Environment Variable: `GOFLAGS=-toolexec=otel toolexec` (global effect; no per-command setup)
```

_docs/implementation.md
```diff
@@ -0,0 +1,183 @@
...
+3. Direct flag: `go build -toolexec=otel`  (on-demand use; ideal for scripts/CI)
```
RomainMuller: Here as well... And I guess the command needs to be the same as that of the GOFLAGS setup.

```suggestion
3. Direct flag: `go build -toolexec=otel toolexec` (on-demand use; ideal for scripts/CI)
```

_docs/implementation.md
```diff
@@ -0,0 +1,183 @@
...
+
+Both of these leveraging the `-toolexec` flag, which allows users to specify a
```
RomainMuller:
```suggestion
All of these leverage the `-toolexec` flag, which allows users to specify a
```

_docs/implementation.md
```diff
@@ -0,0 +1,183 @@
...
+$ go build -n > build_plan.txt
+```
+
+From the build_plan.txt file, the tool extracts all third-party module paths 
+(e.g., `github.com/go-redis/redis`, `github.com/gin-gonic/gin`).
+
+## 1.2 Add Dependencies
+Once the third-party dependencies are identified, the tool generates a file 
+(e.g., otel_import.go) to import the SDK and corresponding hook packages for each 
+dependency:
```
RomainMuller: I think this forgets to explain how available hook packages/instrumentation opportunities are identified... Doesn't need to be too detailed, but I am curious what you have in mind.

_docs/implementation.md
y1yang0: Very good suggestion!

There is a technical detail here: go build supports various package patterns:
```
go build [packages]

Usually, [packages] is a list of import paths.

Example:

1.go build std
2.go build all
3.go build .
4.go build main.go
5.go build ./...
6.go build ./cmd1 ./cmd2
7.go build -modfile file ./cmd1 ./cmd2
```
When using the `go list -json [packages]` approach, we first need to identify from the user's go build command which parameters belong to `[packages]`. However, this is actually challenging. In example.7, how do we determine whether `{file, ./cmd1, ./cmd2}` all belong to `[packages]`, or only `{./cmd1, ./cmd2}` are part of `[packages]`?

When using the `go build -n` approach, we simply need to add an additional `-n` parameter to the user's original command.

Another minor issue is that if the `go build` command includes a parameter with the "test_" prefix (e.g., `go build test_basic_kitex.go`), replacing it with `go list -json test_basic_kitex.go` will not work.

123liuziming (approved):

ralf0131 (approved):

_docs/implementation.md
y1yang0: The first approached is practical, feasible, and convenient. If there are no further objections, I will proceed with the first one.

_docs/implementation.md
y1yang0: Done

y1yang0:
All comments resolved, please take a look again

_docs/implementation.md
```diff
@@ -0,0 +1,185 @@
...
+This step ensures that the dependency is properly recorded and downloaded into
+the module cache.
+By completing this phase, the project is now ready for the next step, where the
+instrumentation logic will be applied during the build process.
+
+## Phase 2: Building with Instrumentation
+
+The build process can be integrated with custom toolchains in the following ways:
+
+1. Command Prefix: `otel go build` (simple but requires manual prefixing)
```
pdelewski: Numbers might suggest ordering.

```suggestion
- Command Prefix: `otel go build` (simple but requires manual prefixing)
```

_docs/implementation.md
```diff
@@ -0,0 +1,185 @@
...
+2. Environment Variable: `GOFLAGS=-toolexec=otel toolexec` (global effect; no per-command setup)
```
pdelewski:
```suggestion
- Environment Variable: `GOFLAGS=-toolexec=otel toolexec` (global effect; no per-command setup)
```

_docs/implementation.md
```diff
@@ -0,0 +1,185 @@
...
+3. Direct flag: `go build -toolexec=otel toolexec` (on-demand use; ideal for scripts/CI)
```
pdelewski:
```suggestion
- Direct flag: `go build -toolexec=otel toolexec` (on-demand use; ideal for scripts/CI)
```

_docs/implementation.md
pdelewski: I understand @RomainMuller's concerns about plan parsing potentially being unstable. However, I don't see a better solution at the moment that would fully take @y1yang0's arguments into account — in other words, I don’t know how to reliably inspect the original arguments passed to go build.

That said, there's one more thing that’s unclear to me from the above description: how exactly will we inject the `-n` flag into the original `go build` command?

y1yang0: > That said, there's one more thing that’s unclear to me from the above description: how exactly will we inject the `-n` flag into the original `go build` command?

We can simply find the keyword "build"(left to right, e.g. "go build -o foo ./cmd/foo") and append `-n` right after it

pdelewski: > > That said, there's one more thing that’s unclear to me from the above description: how exactly will we inject the `-n` flag into the original `go build` command?
> 
> We can simply find the keyword "build"(left to right, e.g. "go build -o foo ./cmd/foo") and append `-n` right after it

That was not my question. My question was, how it will work from user perspective, additional wrapper command, etc...?

y1yang0: > That was not my question. My question was, how it will work from user perspective, additional wrapper command, etc...?

Apologies Przemyslaw, I still don’t fully understand your point. Let me summarize a workflow to see if I can address your concerns:

1. The user runs `otel go build -o foo ./cmd/...`
2. The otel tool extracts the original command: `go build -o foo ./cmd/...`
3. The otel tool appends -n to generate a build plan file: `go build -n -o foo ./cmd/... > build_plan.txt`
4. The subsequent steps remain consistent with the documentation.

From the user’s perspective, all they need to do is prefixing `otel` to the original build command, and the tool will handle the rest of the process automatically.

pdelewski: Ok, thank you!, that answers my question. Just to clarify — I wanted to know whether point `3` is executed by `otel`, or if it's triggered by some other command that we intend to introduce in between. I hope we're on the same page now.

123liuziming: > Ok, thank you!, that answers my question. Just to clarify — I wanted to know whether point `3` is executed by `otel`, or if it's triggered by some other command that we intend to introduce in between. I hope we're on the same page now.

I believe that it is executed by `otel` binary.

NameHaibinZhang (approved):

123liuziming:
Any more comments? Should we merge this doc first?

pdelewski (approved):

RomainMuller (approved):

PR #15, opened 2025-04-05
feat: support semconv for http

123liuziming:
Spans and metrics generated by this project need to follow OpenTelemetry spec, and this PR encapsulates the OpenTelemetry HTTP specifications. The attributes that need to be recorded in Span/Metric and the Metrics that need to be recorded are specified according to the specification in https://opentelemetry.io/docs/specs/semconv/http.

public_api_design

We have some key abstraction:
1. Extractor: Extracting attributes using multiple Getters according to OpenTelemetry spec.
2. Getter:  Getting attributes from the REQUEST object(For example, HttpRequest object that HTTP Server received).
3. AttrsShadower: An extension for Extractor to customize the attributes extracted by the extractor.
4. OperationListener: An hook for better extension(For example, aggregate the metrics)

The Instrumenter will hold a number of Extractors and OperationListener, the Extractor will hold a number of Getters
and AttrsShadowers.The Getter extracts the important attributes of OpenTelemetry and returns them to the Extractor,
which then makes the necessary cuts to these attributes through the AttrsShadower.After Instrumenter gets the necessary
attributes through Extractor, it will perform some callback operations through OperationListener, such as aggregation
of metrics etc.

api_workflow


y1yang0 (approved):

pkg/inst-api/hook.go
```diff
@@ -0,0 +1,32 @@
...
+	Shadow(attrs []attribute.KeyValue) (int, []attribute.KeyValue)
+}
+
+type NoopAttrsShadower struct{}
+
+func (n NoopAttrsShadower) Shadow(attrs []attribute.KeyValue) (int, []attribute.KeyValue) {
+	return len(attrs), attrs
+}
+
+type ContextCustomizer[REQUEST interface{}] interface {
```
kakkoyun:
```suggestion
type ContextCustomizer[REQUEST any] interface {
```

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"context"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst-api-semconv/instrumenter/utils"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.30.0"
+	"sync/atomic"
+)
+
+type HttpCommonAttrsExtractor[REQUEST any, RESPONSE any, GETTER1 HttpCommonAttrsGetter[REQUEST, RESPONSE]] struct {
```
kakkoyun: I'm not really fond of using type parameters this verbose.
Maybe instead we can use marker interfaces instead of `any` and keep these less verbose.

kakkoyun (commented):
I would appreciate if we can be more verbose in pull request description. I would help to provide more context for the reviewers outside the SIG.

123liuziming:
I will add a design description

123liuziming:
@kakkoyun I've added add some verbose description for this PR, could you please review it again, thanks~

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"context"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst-api-semconv/instrumenter/utils"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.30.0"
+	"sync/atomic"
+)
+
+type HttpRequest interface{}
```
pdelewski, line 14: Just to be consistent
```suggestion
type HttpRequest any
```

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+
+type HttpResponse interface{}
```
pdelewski, line 16: Same as above
```suggestion
type HttpResponse any
```

pkg/inst-api/extractor_test.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (ts testSpan) SetStatus(status codes.Code, desc string) {
+	*ts.status = status
+}
+
+func TestDefaultSpanStatusExtractor(t *testing.T) {
+	unset := codes.Unset
+	ts := testSpan{Span: noop.Span{}, status: &unset}
+	d := defaultSpanStatusExtractor[interface{}, interface{}]{}
```
pdelewski: Same 
```suggestion
	d := defaultSpanStatusExtractor[any, any]{}
```

123liuziming:
> This PR, as well as #16, follows a package structure like:
> 
> * `pkg/inst-api-semconv/instrumenter`
> * `pkg/inst-api`
> 
> I think this structure should be described and explained as part of #14.
> 
> Another thing that’s missing from my perspective is an explanation of how the mentioned abstractions fit into the overall workflow — the bigger picture, so to speak. (This should also be part of #14, in my opinion.)

Can I put these into a separate `project-structure.md` rather than putting all into `implementation.md`?

pdelewski:
> > This PR, as well as #16, follows a package structure like:
> > 
> > * `pkg/inst-api-semconv/instrumenter`
> > * `pkg/inst-api`
> > 
> > I think this structure should be described and explained as part of #14.
> > Another thing that’s missing from my perspective is an explanation of how the mentioned abstractions fit into the overall workflow — the bigger picture, so to speak. (This should also be part of #14, in my opinion.)
> 
> Can I put these into a separate `project-structure.md` rather than putting all into `implementation.md`?

Yes, absolutely — the most important thing is just to have it.

_docs/project-structure.md
```diff
@@ -0,0 +1,26 @@
+# Project Structure
```
kakkoyun: 👍

_docs/project-structure.md
```diff
@@ -0,0 +1,26 @@
...
+            rpc
+            db
+            messaging
+            network
+sdk ---> Instrumentation code for each plugin (e.g. http, grpc, ...)
+```
+For Public API, we have some key abstractions as follows:
+![Public API](./assets/public_api_design.png)
+1. Instrumenter: Unified entrance of instrumentation (generating span, metrics, ...)
+2. Extractor: Extracting attributes using multiple Getters according to [OpenTelemetry Semconv](https://opentelemetry.io/docs/specs/semconv/).
```
ralf0131: I have a few quick questions:
1. Why don't components like SpanKindExtractor have connections to AbstractExtractor in the diagram? From the naming it seems to have connections.
2. The instrumenter should be named as AbstractInstrumenter for consistency?
3. Getter, how about AttrsGetter, is there any other possible Getter? If so we can still have AttrsGetter as sub interface for Getter.

_docs/project-structure.md
```diff
@@ -0,0 +1,26 @@
...
+3. Getter: Getting attributes from the REQUEST object(For example, HttpRequest object that HTTP Server received).
+4. AttrsShadower: An extension for Extractor to customize the attributes extracted by the extractor.
+5. OperationListener: An hook for better extension(For example, aggregate the metrics)
```
ralf0131: It is appreciated to document all the components in your diagram and try to explain the connections between them.

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+
+type HttpCommonAttrsExtractor[REQUEST HttpRequest, RESPONSE HttpResponse, GETTER1 HttpCommonAttrsGetter[REQUEST, RESPONSE]] struct {
```
ralf0131: This is too long for a single line, should we have a spec to have fix length for a single line?

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+	HttpGetter       GETTER1
+	AttributesFilter func(attrs []attribute.KeyValue) []attribute.KeyValue
+}
+
+func (h *HttpCommonAttrsExtractor[REQUEST, RESPONSE, GETTER1]) OnStart(attributes []attribute.KeyValue, parentContext context.Context, request REQUEST) ([]attribute.KeyValue, context.Context) {
+	attributes = append(attributes, attribute.KeyValue{
+		Key:   semconv.HTTPRequestMethodKey,
+		Value: attribute.StringValue(h.HttpGetter.GetRequestMethod(request)),
```
ralf0131, line 36: What if h.HttpGetter.GetRequestMethod(request) throws exception or contains unexpected return value?

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+		attributes = append(attributes, attribute.KeyValue{Key: semconv.ErrorTypeKey, Value: attribute.StringValue(errorType)})
+	}
+	return attributes, context
+}
+
+type HttpClientAttrsExtractor[REQUEST HttpRequest, RESPONSE HttpResponse, GETTER1 HttpClientAttrsGetter[REQUEST, RESPONSE]] struct {
+	Base HttpCommonAttrsExtractor[REQUEST, RESPONSE, GETTER1]
+}
+
+func (h *HttpClientAttrsExtractor[REQUEST, RESPONSE, GETTER1]) OnStart(attributes []attribute.KeyValue, parentContext context.Context, request REQUEST) ([]attribute.KeyValue, context.Context) {
```
ralf0131: It is appreciated to document the behaviour of the function using comments for better readability.

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+type HttpRequest interface{}
+
+type HttpResponse interface{}
+
+type HttpCommonAttrsExtractor[REQUEST HttpRequest, RESPONSE HttpResponse, GETTER1 HttpCommonAttrsGetter[REQUEST, RESPONSE]] struct {
+	HttpGetter       GETTER1
+	AttributesFilter func(attrs []attribute.KeyValue) []attribute.KeyValue
+}
+
+func (h *HttpCommonAttrsExtractor[REQUEST, RESPONSE, GETTER1]) OnStart(attributes []attribute.KeyValue, parentContext context.Context, request REQUEST) ([]attribute.KeyValue, context.Context) {
```
ralf0131: Instead of GETTER1, could you use more specific name?

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+	attributes = append(attributes, attribute.KeyValue{
+		Key:   semconv.HTTPRequestMethodKey,
+		Value: attribute.StringValue(h.HttpGetter.GetRequestMethod(request)),
+	})
+	return attributes, parentContext
+}
+
+func (h *HttpCommonAttrsExtractor[REQUEST, RESPONSE, GETTER]) OnEnd(attributes []attribute.KeyValue, context context.Context, request REQUEST, response RESPONSE, err error) ([]attribute.KeyValue, context.Context) {
+	statusCode := h.HttpGetter.GetHttpResponseStatusCode(request, response, err)
```
ralf0131, line 44: I think it is good for this right now.

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (h *HttpClientAttrsExtractor[REQUEST, RESPONSE, GETTER1]) OnStart(attributes []attribute.KeyValue, parentContext context.Context, request REQUEST) ([]attribute.KeyValue, context.Context) {
+	attributes, parentContext = h.Base.OnStart(attributes, parentContext, request)
+	resendCount := parentContext.Value(utils.CLIENT_RESEND_KEY)
+	newCount := int32(0)
+	if resendCount != nil {
+		newCount = atomic.AddInt32(resendCount.(*int32), 1)
+		if newCount > 0 {
+			attributes = append(attributes, attribute.KeyValue{
```
ralf0131, line 69: I see you are using atomic, is this operation thread safe?

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	parentContext = context.WithValue(parentContext, utils.CLIENT_RESEND_KEY, &newCount)
+	if h.Base.AttributesFilter != nil {
+		attributes = h.Base.AttributesFilter(attributes)
+	}
+	return attributes, parentContext
+}
+
+func (h *HttpClientAttrsExtractor[REQUEST, RESPONSE, GETTER1]) OnEnd(attributes []attribute.KeyValue, context context.Context, request REQUEST, response RESPONSE, err error) ([]attribute.KeyValue, context.Context) {
```
ralf0131: Should Context be placed as the first parameter?

pkg/inst-api-semconv/instrumenter/http/http_metrics.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	d, err := newHttpServerRequestDurationMeasures(meter)
+	if err != nil {
+		return nil, err
+	}
+	m.serverRequestDuration = d
+	return m, nil
+}
+
+func newHttpServerRequestDurationMeasures(meter metric.Meter) (metric.Float64Histogram, error) {
```
ralf0131: This function has some duplicate logic with newHttpClientRequestDurationMeasures, we can extract it to a common function.

pkg/inst-api-semconv/instrumenter/http/http_metrics.go
```diff
@@ -0,0 +1,191 @@
...
+	"log"
+	"sync"
+	"time"
+)
+
+const http_server_request_duration = "http.server.request.duration"
+
+const http_client_request_duration = "http.client.request.duration"
+
+type HttpServerMetric struct {
```
ralf0131, line 26: Could you please add more comments?

pkg/inst-api-semconv/instrumenter/utils/metrics_shadower.go
```diff
@@ -0,0 +1,26 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package utils
+
+import (
+	"go.opentelemetry.io/otel/attribute"
+)
+
+func Shadow(attrs []attribute.KeyValue, metricsSemConv map[attribute.Key]bool) (int, []attribute.KeyValue) {
```
ralf0131, line 15: Could you explain why we need to do this?

_docs/project-structure.md
123liuziming: 1. Those extractors are using to extract different information for spans, so actually they may not have connections.
2. `Instrumenter` is a struct actually, rather than an interface.
3. Yes, I've renamed it to something like `AttributesGetter`

_docs/project-structure.md
123liuziming: I've added a sequence diagram

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
123liuziming: Getter should ensure that it does not panic and return any unexpected value.

pkg/inst-api-semconv/instrumenter/utils/metrics_shadower.go
123liuziming: User may need to customize the attributes in their own distribution, so we may need some extensions for this.

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
123liuziming: Yes

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
123liuziming: I've add limited the max columns in my local IDE, I will add the limitation in golang-lint check in a new PR

ralf0131 (approved):
LGTM.

pkg/inst-api-semconv/instrumenter/http/http_status_code_extractor.go
```diff
@@ -0,0 +1,57 @@
...
+	if statusCode >= 400 || statusCode < 100 {
+		if err != nil {
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+		} else {
+			span.SetStatus(codes.Error, invalidHttpStatusCode)
+		}
+		span.SetAttributes(attribute.KeyValue{Key: semconv.ErrorTypeKey, Value: attribute.StringValue(strconv.Itoa(statusCode))})
+	} else if statusCode >= 200 && statusCode < 300 {
+		span.SetStatus(codes.Ok, "success")
```
zacharycmontoya, line 36: Hi folks, I was taking a look at this as a colleague of @kakkoyun. I looked at the latest [Set Status API](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status) and it has the following guidance:

> Generally, Instrumentation Libraries SHOULD NOT set the status code to `Ok`, unless explicitly configured to do so. Instrumentation Libraries SHOULD leave the status code as `Unset` unless there is an error, as described above.

Can we remove this branch or put this modification behind an opt-in configuration?

EDIT: This additional [semconv page](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status) says that Span Status `MUST be left unset if HTTP status code was in the 1xx, 2xx or 3xx ranges, unless there was another error ... in which case status MUST be set to Error.`

pkg/inst-api-semconv/instrumenter/http/http_status_code_extractor.go
```diff
@@ -0,0 +1,57 @@
...
+	if statusCode >= 500 || statusCode < 100 {
+		if err != nil {
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+		} else {
+			span.SetStatus(codes.Error, invalidHttpStatusCode)
+		}
+		span.SetAttributes(attribute.KeyValue{Key: semconv.ErrorTypeKey, Value: attribute.StringValue(strconv.Itoa(statusCode))})
+	} else if statusCode >= 200 && statusCode < 300 {
+		span.SetStatus(codes.Ok, "success")
```
zacharycmontoya, line 55: My feedback from a [previous comment](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/15/files#r2122188846) also applies here

pkg/inst-api-semconv/instrumenter/http/http_status_code_extractor_test.go
```diff
@@ -0,0 +1,182 @@
...
+			code: 400,
+		},
+	}
+	u := codes.Code(0)
+	span := &testSpan{status: &u}
+	c.Extract(span, nil, nil, nil)
+	if *span.status != codes.Error {
+		panic("span status should be error!")
+	}
+	if span.Kvs == nil {
```
zacharycmontoya, line 95: It looks like this line is testing that we set the span attribute `error.type=<value>`. Can we change the assert to specifically test against the `error.type` key?

pkg/inst-api-semconv/instrumenter/http/http_status_code_extractor_test.go
```diff
@@ -0,0 +1,182 @@
...
+	c := HttpServerSpanStatusExtractor[any, any]{
+		Getter: customizedNetHttpAttrsGetter{
+			code: 400,
+		},
+	}
+	u := codes.Code(0)
+	span := &testSpan{status: &u}
+	c.Extract(span, nil, nil, nil)
+	if *span.status != codes.Unset {
+		panic("span status should be error!")
```
zacharycmontoya, line 153: I think this message should be updated to: `"span status should be unset!"`

kakkoyun:
Thanks @zacharycmontoya.

@123liuziming I asked Zach to have look at it as an expert (he works on SDK API).

Issue #29, opened 2025-06-09
Split implementation tasks


y1yang0:
Thank you all for efforts. The implementation document review is nearing completion, and it's time for us to start implementation. Based on the document description, we can roughly summarize the following workflow:

UX Design: Tool encapsulation, initiating the setup.Setup() call, etc.
Phase 1: Setup Dependencies: Implementing setup.Setup()
Phase 2: Building with Instrumentation: Implementing instrument.Instrument()

Please feel free to claim the implementation tasks.

y1yang0:
I would like to claim Phase2 task

RomainMuller:
My speciality is more towards instrumentation / AST manipulation, so I'm definitely up to contribute on this.

y1yang0:
@RomainMuller @kakkoyun @123liuziming @pdelewski 

I've updated the implementation tasks based on PR #30. Please feel free to claim any tasks of interest and leave a comment to notify us.

I'm unsure whether the current granularity level works for you all. If a more granular task breakdown is needed, please let me know. Thank you!

y1yang0:
For example, I will attempt to take ownership of `findDeps` and start implementing

pdelewski:
@y1yang0 I’d like to help, but at the moment I can’t commit to more than doing reviews.

y1yang0:
Claiming task refreshDeps. Thanks for looking into this.

PR #38, opened 2025-07-16
feat: introduce hook dependencies via otel.runtime.go

y1yang0:
Subtask of #29 

This patch creates the `otel.runtime.go` file, which introduces the previously matched hook dependencies and forces their import.

Since refreshDeps has not been implemented yet, I manually added the instrumentation/helloworld dependency in the demo project so that the demo app can run properly. Once refreshDeps is implemented, these manual imports can be fully removed.

The current output of `make demo` is as follows:
```
Building demo with instrumentation...
Running demo...
=setupOpenTelemetry=
```


tool/internal/setup/add.go
```diff
@@ -3,8 +3,52 @@
...
+	importDecls := make([]dst.Decl, 0, len(matched))
+	for _, m := range matched {
+		decl := ast.NewImportDecl("_", m.Path)
+		importDecls = append(importDecls, decl)
+	}
+	return importDecls
+}
+
+func buildOtelRuntimeAst(decls []dst.Decl) *dst.File {
+	comment := "// This file is generated by the opentelemetry-go-compile-instrumentation tool. DO NOT EDIT."
```
pdelewski:
```suggestion
	const comment = "// This file is generated by the opentelemetry-go-compile-instrumentation tool. DO NOT EDIT."
```

demo/main.go
```diff
@@ -3,6 +3,8 @@
 
 package main
 
+import _ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/instrumentation/helloworld"
```
pdelewski, line 6: I would add comment for now
```suggestion
// This dependency is added manually for now, as refreshDeps is not yet implemented. 
// TODO: remove it once implemented.
import _ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/instrumentation/helloworld"
```

pdelewski:
LGTM

demo/main.go
y1yang0: Sorry, I missed this one. Let me adopt it in the future patch

PR #41, opened 2025-07-22
feat: custom error handling

y1yang0:


docs/developing.md
```diff
@@ -0,0 +1,39 @@
...
+
+Exit the program with `ex.Fatal` or `ex.Fatalf`.
+
+```go
+// Exit the program with an error
+ex.Fatal(err)
+```
+
+```go
+// Exit the program with an error and a additional message
```
pdelewski:
```suggestion
// Exit the program with an error and an additional message
```

docs/developing.md
```diff
@@ -0,0 +1,39 @@
...
+return ex.Error(err)
+```
+
+```go
+// Wrap an existing error with additional message and return to the caller
+return ex.Errorf(err, "additional context %s", "some value")
+```
+
+```go
+// Create a new error with a additional message
```
pdelewski:
```suggestion
// Create a new error with an additional message
```

docs/developing.md
```diff
@@ -0,0 +1,39 @@
...
+
+Any error should be wrapped with `ex.Error` or `ex.Errorf` to include the stack trace.
+
+```go
+// Wrap an existing error and return to the caller
+return ex.Error(err)
+```
+
+```go
+// Wrap an existing error with additional message and return to the caller
```
pdelewski:
```suggestion
// Wrap the existing error with an additional message and return it to the caller.
```

.golangci.yml
```diff
@@ -483,16 +483,19 @@ linters:
       # If this set, it will override the default set of ignored signatures.
       # See https://github.com/tomarrell/wrapcheck#configuration for more information.
       # Default: [".Errorf(", "errors.New(", "errors.Unwrap(", "errors.Join(", ".Wrap(", ".Wrapf(", ".WithMessage(", ".WithMessagef(", ".WithStack("]
+      # We use our own error handling library, so we need to ignore the default ones.
```
pdelewski: I think that this comment would be more appropriate 
```suggestion
      # We use our own error handling library (`ex`), so we override the default
```

.golangci.yml
```diff
@@ -483,16 +483,19 @@ linters:
...
-        - .Errorf(
-        - errors.New(
-        - errors.Unwrap(
-        - errors.Join(
-        - .Wrap(
-        - .Wrapf(
-        - .WithMessage(
-        - .WithMessagef(
-        - .WithStack(
+        # - .Errorf(
```
pdelewski: I would suggest just to remove this list and leave only `ex` calls

tool/ex/wrap.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return ""
+}
+
+// Error wraps an error with stack trace information
+// If you don't want to decorate the existing error, use it.
+func Error(previousErr error) error {
+	e := &stackfulError{
+		message: previousErr.Error(),
+		frame:   currentFrame(3), //nolint:mnd // skip the Errorf caller
```
pdelewski: Thus `3` literal is used few times. I think this might be good candidate for const

y1yang0:
Thanks @pdelewski for the review. I'm thinking there might be another simpler pattern. Currently, we use ex.Error(err) to record the current stack trace and return the error to the upper-level caller:

```
func a() error {
        return ex.Errorf(nil, "new error") //stack0:a()
}
func b() error {
        err := a()
        return ex.Error(err)  //stack1:b()
}
func c() error {
        err := b()
        return ex.Error(err)  //stack2:c()
}
```
Perhaps we can go even further—we could record the entire stack trace at the root and then simply return the error without wrapping it with ex.Error:
```
func a() error {
        return ex.Errorf(nil, "new error") //stack0:a() stack1:b() stack2:c()
}
func b() error {
        err := a()
        return err
}
func c() error {
        err := b()
        return err
}
```
This way is simpler. Anyway, this is another task, a potential enhancement.

PR #46, opened 2025-08-05
feat: match the rules with compile command in instrument phase

y1yang0:
Subtask of #29 

This patch checks which rules match the package being compiled by the current build command. In the next step, we will perform the actual instrumentation!

### DIFF
--- a/tool/internal/instrument/toolexec.go
+++ b/tool/internal/instrument/toolexec.go
@@ -5,6 +5,7 @@ package instrument
 
 import (
 	"log/slog"
+	"strings"
 
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
 )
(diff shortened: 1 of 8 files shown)


NameHaibinZhang (approved):

tool/util/go.go
```diff
@@ -0,0 +1,87 @@
...
+package util
+
+import (
+	"strings"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/ex"
+)
+
+// IsCompileCommand checks if the line is a compile command.
+func IsCompileCommand(line string) bool {
```
ralf0131, line 13: Shall we add some unit test to functions like `IsCompileCommand` and `SplitCompileCmds` ?

y1yang0: This could consider as a follow up patch #47

ralf0131 (approved):

tool/internal/instrument/toolexec.go
```diff
@@ -22,23 +23,29 @@ func Toolexec(logger *slog.Logger, args []string) error {
...
+		// current command with list of matched rules
+		matchedRules := ip.match(args, rules)
+		if len(matchedRules) > 0 {
+			ip.Info("Instrumenting package", "rules", matchedRules, "args", args)
+			// Okay, this package should be instrumented.
+			err = ip.instrument(args)
+			if err != nil {
+				return err
+			}
+			// return nil
```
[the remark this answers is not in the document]
ralf0131: This line is commented out because it is not implemented yet.

PR #50, opened 2025-09-05
feat: instrument the package

y1yang0:
This patch introduces the real implementation of the instrumentation

It looks something like this
```bash

$ make demo
Packaging instrumentation code into binary...
otel version v0.0.0-20250905015142-26f1d64fe8e5+26f1d64fe8e5ca7434b6fc32c95d126b84f2a5fa (2025-09-05T01:51:42Z)
Building demo with instrumentation...
go: found github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld in github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld v0.0.0-00010101000000-000000000000
Running demo...
=setupOpenTelemetry=
[MyHook] start to instrument hello world!
{"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"ScopeMetrics":[]}
{"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"ScopeMetrics":[]}
{"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"ScopeMetrics":[]}
{"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"ScopeMetrics":[]}
{"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"ScopeMetrics":[]}
{"Name":"hello-world","SpanContext":{"TraceID":"e7eda03eb4bd4d1543b0a55274c6eef2","SpanID":"b4bc176cf9ac82fc","TraceFlags":"01","TraceState":"","Remote":false},"Parent":{"TraceID":"00000000000000000000000000000000","SpanID":"0000000000000000","TraceFlags":"00","TraceState":"","Remote":false},"SpanKind":1,"StartTime":"2025-09-05T09:52:25.273622243+08:00","EndTime":"2025-09-05T09:52:30.274611445+08:00","Attributes":[{"Key":"url.scheme","Value":{"Type":"STRING","Value":"http"}},{"Key":"url.path","Value":{"Type":"STRING","Value":"/a"}},{"Key":"url.query","Value":{"Type":"STRING","Value":"a=5"}}],"Events":null,"Links":null,"Status":{"Code":"Unset","Description":""},"DroppedAttributes":0,"DroppedEvents":0,"DroppedLinks":0,"ChildSpanCount":0,"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"InstrumentationScope":{"Name":"hello-world","Version":"0.0.1","SchemaURL":"","Attributes":null},"InstrumentationLibrary":{"Name":"hello-world","Version":"0.0.1","SchemaURL":"","Attributes":null}}
[MyHook] hello world is instrumented!
{"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"ScopeMetrics":[{"Scope":{"Name":"hello-world","Version":"","SchemaURL":"","Attributes":null},"Metrics":[{"Name":"http.client.request.duration","Description":"Duration of HTTP client requests.","Unit":"ms","Data":{"DataPoints":[{"Attributes":[{"Key":"url.scheme","Value":{"Type":"STRING","Value":"http"}}],"StartTime":"2025-09-05T09:52:25.273612571+08:00","Time":"2025-09-05T09:52:31.274312009+08:00","Count":1,"Bounds":[0,5,10,25,50,75,100,250,500,750,1000,2500,5000,7500,10000],"BucketCounts":[1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"Min":-9223372036854776000,"Max":-9223372036854776000,"Sum":-9223372036854776000,"Exemplars":[{"FilteredAttributes":null,"Time":"2025-09-05T09:52:30.274844931+08:00","Value":-9223372036854776000,"SpanID":"tLwXbPmsgvw=","TraceID":"5+2gPrS9TRVDsKVSdMbu8g=="}]}],"Temporality":"CumulativeTemporality"}}]}]}
{"Resource":[{"Key":"service.name","Value":{"Type":"STRING","Value":"unknown_service:demo"}},{"Key":"telemetry.sdk.language","Value":{"Type":"STRING","Value":"go"}},{"Key":"telemetry.sdk.name","Value":{"Type":"STRING","Value":"opentelemetry"}},{"Key":"telemetry.sdk.version","Value":{"Type":"STRING","Value":"1.37.0"}}],"ScopeMetrics":[{"Scope":{"Name":"hello-world","Version":"","SchemaURL":"","Attributes":null},"Metrics":[{"Name":"http.client.request.duration","Description":"Duration of HTTP client requests.","Unit":"ms","Data":{"DataPoints":[{"Attributes":[{"Key":"url.scheme","Value":{"Type":"STRING","Value":"http"}}],"StartTime":"2025-09-05T09:52:25.273612571+08:00","Time":"2025-09-05T09:52:32.274525757+08:00","Count":1,"Bounds":[0,5,10,25,50,75,100,250,500,750,1000,2500,5000,7500,10000],"BucketCounts":[1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"Min":-9223372036854776000,"Max":-9223372036854776000,"Sum":-9223372036854776000,"Exemplars":[{"FilteredAttributes":null,"Time":"2025-09-05T09:52:30.274844931+08:00","Value":-9223372036854776000,"SpanID":"tLwXbPmsgvw=","TraceID":"5+2gPrS9TRVDsKVSdMbu8g=="}]}],"Temporality":"CumulativeTemporality"}}]}]}
[MyHook] after hook executed!
```


tool/cmd/cmd-go.go
```diff
@@ -34,12 +42,12 @@ var commandGo = cli.Command{
 
 		err = setup.Setup(ctx)
 		if err != nil {
-			return cli.Exit(err, exitCodeFailure)
+			return ex.Errorf(err, "failed to build with toolexec with exitcode %d", exitCodeFailure)
```
y1yang0: Side note: cli.exit was changed to ex.error to preserve the error chain. Otherwise, ex.fatal wouldn't be able to print the correct error stack.

tool/data/export.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/ex"
 )
 
-//go:embed *.yaml
```
kakkoyun, line 12: Is there a reason to remove the suffix?

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,321 @@
...
+				pos = ip.parser.FindPosition(stmt)
+				if !pos.IsValid() {
+					continue
+				}
+				tag := fmt.Sprintf("//line %s", pos.String())
+				stmt.Decorations().Before = dst.NewLine
+				stmt.Decorations().Start.Append(tag)
+			}
+		} else {
+			pos = ip.parser.FindPosition(funcDecl.Body)
```
pdelewski: This line seems redundant as pos is computed before if statement 
```suggestion
```

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,321 @@
...
+	// hook code. By inserting trampoline-jump-if at the target function entry,
+	// we can intercept the original function and execute before/after hooks.
+	tjump := createTJumpIf(t, funcDecl, args, retVals)
+
+	// Find if there is already a trampoline-jump-if, insert new tjump if so,
+	// otherwise prepend to block body.
+	ip.insertToFunc(funcDecl, tjump)
+
+	// Generate corresponding trampoline code
+	err := ip.generateTrampoline(t)
```
pdelewski: Can we simplify that to just 
```suggestion
	return ip.generateTrampoline(t)
```

pdelewski: and remove 4 below lines

tool/internal/setup/add.go
```diff
@@ -13,19 +15,97 @@ const (
...
+	{"runtime/debug", "_otel_debug"}, // The getstack function depends on runtime/debug
+	{"log", "_otel_log"},             // The printstack function depends on log
+	{"unsafe", "_"},                  // The golinkname tag depends on unsafe
+}
+
+func genImportDecl(matched []*rule.InstFuncRule) []dst.Decl {
 	importDecls := make([]dst.Decl, 0, len(matched))
 	for _, m := range matched {
-		decl := ast.NewImportDecl("_", m.Path)
+		decl := ast.ImportDecl("_", m.Path)
```
pdelewski: What about using `requiredImports`
```suggestion
		decl := ast.ImportDecl(requiredImports[2][1], m.Path)
```
or extract const to avoid consitency problems

tool/internal/setup/add.go
```diff
@@ -13,19 +15,97 @@ const (
...
 	return importDecls
 }
 
+func genVarDecl(matched []*rule.InstFuncRule) []dst.Decl {
+	decls := make([]dst.Decl, 0, len(matched))
+	for i, m := range matched {
+		// First variable declaration
+		// //go:linkname _getstatck%d %s.OtelGetStackImpl
+		// var _getstatck%d = _otel_debug.Stack
+		value := ast.SelectorExpr(ast.Ident("_otel_debug"), "Stack")
```
pdelewski, line 42: Similar as above

tool/internal/setup/extract.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
+)
+
+const (
+	EmbeddedInstPkgGzip = "otel-pkg.gz"
+)
+
+func normalizePath(name string) string {
+	cleanName := filepath.ToSlash(filepath.Clean(name))
+	if strings.HasPrefix(cleanName, "pkg_temp/") {
```
pdelewski: Using const for "pkg_temp" might be safer and consistent

tool/internal/ast/shared.go
```diff
@@ -8,16 +8,52 @@ import (
...
+		return nil, err
+	}
+	if len(decls) == 0 {
+		//nolint:nilnil // no function declaration found is not an error
+		return nil, nil
+	}
+	return decls[0], nil
+}
+
+func FindFuncDeclWithRecv(root *dst.File, funcName string) ([]*dst.FuncDecl, error) {
```
pdelewski: The current name is misleading IMO: it returns both free functions and methods. Maybe rename it to:
```suggestion
func FindFuncDecl(root *dst.File, funcName string) ([]*dst.FuncDecl, error) {
```

tool/internal/ast/shared.go
```diff
@@ -8,16 +8,52 @@ import (
...
+	found := make([]*dst.FuncDecl, 0)
+	for _, funcDecl := range funcDecls {
+		if lambda(funcDecl) {
+			found = append(found, funcDecl)
+		}
+	}
+	return found, nil
+}
+
+func FindFuncDecl(root *dst.File, funcName string) (*dst.FuncDecl, error) {
```
pdelewski: Rename it to (look my comment below):
```suggestion
func FindFuncDeclWithoutRecv(root *dst.File, funcName string) (*dst.FuncDecl, error) {
```

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/dave/dst"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/ast"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/rule"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
+)
+
+const (
+	TJumpLabel         = "/* TRAMPOLINE_JUMP_IF */"
+	OtelTrampolineFile = "otel.trampoline.go"
+	OtelAPIFile        = "otel.api.go"
```
pdelewski: This const is currently unused. What do you plan to store/generate in this file?

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,321 @@
...
+
+	"github.com/dave/dst"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/ast"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/rule"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
+)
+
+const (
+	TJumpLabel         = "/* TRAMPOLINE_JUMP_IF */"
+	OtelTrampolineFile = "otel.trampoline.go"
```
pdelewski: `OtelTrampolineFile` at the moment contains `HookContext` interface, do you plan to move/generate some stuff related to trampolines there?

y1yang0: You're right. It's generated alongside the trampoline, but doesn't actually contain the trampoline itself. Instead, it holds some variables and hook context APIs. Perhaps we should rename it to be more accurate.

y1yang0: Maybe `otel.prelude.go`?`otel.hookcontext.go`?`otel.context.go`?`otel.decls.go`?`otel.globals.go`?

tool/internal/setup/add.go
y1yang0: I'm not quite sure I understand the distinction here. It seems that both requiredImports and genImportDecl are used to generate imports. Could you clarify their respective roles?

pdelewski: What I meant is that the values from requiredImports are repeated as hardcoded literals. I was thinking we could just use the requiredImports variable here, or maybe extract them as constants. Anyway, it’s not critical.

tool/util/sys.go
```diff
@@ -68,3 +80,49 @@ func CopyFile(src, dst string) error {
 	}
 	return nil
 }
+
+func Crc32(s string) string {
```
kakkoyun: nit:
```suggestion
func CRC32(s string) string {
```

tool/internal/instrument/template_api.go
```diff
@@ -0,0 +1,34 @@
+//go:build ignore
+
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package instrument
+
+// !!!Any modification MUST be synced with pkg/inst/context.go
```
kakkoyun, line 8: Maybe we should automate this.

kakkoyun (commented):
I did an initial pass and I hope to have another pass soon.

tool/internal/instrument/template_api.go
pdelewski: That's good idea, however I would postpone that to another PR

kakkoyun: 💯

pdelewski:
I’d like to merge this soon, as I have follow-up changes depending on it and more planned.

tool/data/export.go
y1yang0: Because we need both *.yaml and *.gzip

kakkoyun:
@RomainMuller was on PTO and just returned. Let's give him a chance to review it and then we can proceed.

pkg/instrumentation/helloworld/helloworld_hook.go
```diff
@@ -37,7 +37,8 @@ func setupOpenTelemetry() {
 
 var helloWorldInstrumenter = BuildNetHttpClientOtelInstrumenter()
 
-func MyHookBefore(ictx inst.Context) {
+//go:linkname MyHookBefore main.MyHookBefore
```
RomainMuller, line 40: Not necessarily something to fix right in this PR, but I'm a little concerned if this needs to be done manually by hook authors, because the symbol namespace is flat, shared -- and so collisions will be a problem.

Would surely be nicer if the tool was allocating names for these entities so it cna then make sure collisions are not an issue.

tool/cmd/cmd-go.go
```diff
@@ -5,12 +5,19 @@ package main
...
 
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/ex"
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/setup"
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
 	"github.com/urfave/cli/v3"
 )
 
+// cleanupGenerated cleans up the generated files during the build process
+func cleanupGenerated() {
+	_ = os.RemoveAll(setup.OtelRuntimeFile)
```
RomainMuller: I would like to do away with the practice of silently ignoring errors... At least, these should be logged (even if debug), so they can be surfaced/reviewed when something weird happens as part of troubleshooting.

tool/cmd/cmd-go.go
RomainMuller: The previous code was asking the CLI framework to exit with the `exitCodeFailure` code, whereas the new code logs as if the build failed with `exitCodeFailure` and aborts with a generic status code.

If `ex.Errorf` wraps `err`, then we could in principle achieve the same behavior (+ expected stack trace) with


```suggestion
			return ex.Errorf(cli.Exit(err, exitCodeFailure), "failed to build with toolexec")
```

I'd also note that in this case the stack trace I'd be interested in is that of the error returned by `setup.Setup`, not "here"?

---

This applies to the other occurrences of this change.

tool/cmd/cmd-version.go
```diff
@@ -25,32 +26,32 @@ var commandVersion = cli.Command{
...
 		if err != nil {
-			return cli.Exit(err, exitCodeFailure)
+			return ex.Errorf(err, "failed to print version with exitcode %d", exitCodeFailure)
 		}
 
 		if cmd.Bool("verbose") {
 			_, err = fmt.Fprintf(cmd.Writer, "%s\n", runtime.Version())
 			if err != nil {
-				return cli.Exit(err, exitCodeFailure)
+				return ex.Errorf(err, "failed to print version with exitcode %d", exitCodeFailure)
```
RomainMuller: All these I feel are trying to augment errors that ought to be entirely sufficient by themselves already; and changes the exit code semantics.

tool/data/export.go
```diff
@@ -9,7 +9,7 @@ import (
...
+//go:embed *
```
RomainMuller, line 12: That'll embed _everything_, including the `.go` files, isn't it? That'll weight on the final binary size...


```suggestion
//go:embed *.yaml *.gzip
```

tool/internal/ast/parser.go
```diff
@@ -48,8 +48,23 @@ func (ap *AstParser) ParseFile(filePath string, mode parser.Mode) (*dst.File, er
 	return dstFile, nil
 }
 
-func (ap *AstParser) ParseFileFast(filePath string) (*dst.File, error) {
-	return ap.ParseFile(filePath, parser.SkipObjectResolution)
+// ParseSource parses the AST from complete source code.
+func (ap *AstParser) ParseSource(source string) (*dst.File, error) {
+	util.Assert(source != "", "empty source")
```
RomainMuller: I would expect the parser to fail if that is the case... I'm not entirely sure this is necessary (but it also does not hurt, so...)

tool/internal/ast/parser.go
```diff
@@ -48,8 +48,23 @@ func (ap *AstParser) ParseFile(filePath string, mode parser.Mode) (*dst.File, er
...
+	ap.dec = decorator.NewDecorator(ap.fset)
```
RomainMuller: This will create a new `ap.dec` each time this function is called, which strikes me as possibly not intended (especially since they all use the same `ap.fset`).

Perhaps the decorator should be created once when `ap.fset` is provisionned, and kept afterwards? Otherwise, I would say it should not be stored on `ap`.

tool/internal/ast/parser.go
```diff
@@ -48,8 +48,23 @@ func (ap *AstParser) ParseFile(filePath string, mode parser.Mode) (*dst.File, er
...
+	dstRoot, err := ap.dec.Parse(source)
+	if err != nil {
+		return nil, ex.Error(err)
+	}
+	return dstRoot, nil
+}
+
+func (ap *AstParser) FindPosition(node dst.Node) token.Position {
+	astNode := ap.dec.Ast.Nodes[node]
```
RomainMuller, line 66: As mentioned above -- this becomes invalid for a given node once `ParseSource` is called again...

Either the decorator needs to be stored & re-used, or it needs to be returned together with the AST nodes so it can be used to look up positions...

This current design is going to lead to problems if/when we look into parallelizing processing.

tool/internal/ast/primitives.go
```diff
@@ -0,0 +1,221 @@
...
+	IdentTrue   = "true"
+	IdentFalse  = "false"
+	IdentIgnore = "_"
+)
+
+func Ident(name string) *dst.Ident {
+	return &dst.Ident{
+		Name: name,
+	}
+}
```
RomainMuller, line 35: How does this differ from [`dst.NewIdent`](https://pkg.go.dev/github.com/dave/dst#NewIdent) ?

tool/internal/ast/primitives.go
```diff
@@ -0,0 +1,221 @@
...
+	return &dst.CallExpr{
+		Fun:  &dst.Ident{Name: name},
+		Args: args,
+	}
+}
+
+func StringLit(value string) *dst.BasicLit {
+	return &dst.BasicLit{
+		Kind:  token.STRING,
+		Value: fmt.Sprintf("%q", value),
```
RomainMuller: Maybe use [`strconv.Quote`](https://pkg.go.dev/strconv#Quote) instead?

tool/internal/ast/primitives.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+func Exprs(exprs ...dst.Expr) []dst.Expr {
+	return exprs
+}
+
+func Stmts(stmts ...dst.Stmt) []dst.Stmt {
+	return stmts
+}
```
RomainMuller, line 84: Are these really necessary?

tool/internal/ast/primitives.go
RomainMuller, line 1: Those feel rather trivial (basically functions that replace equally expressive struct initialization expressions. I wonder if that is actually necessary or what benefits this brings?

tool/internal/ast/shared.go
```diff
@@ -29,14 +66,32 @@ func ListFuncDecls(file string) ([]*dst.FuncDecl, error) {
...
+
+func IsUnusedIdent(ident *dst.Ident) bool {
+	return ident.Name == IdentIgnore
+}
+
+func IsStringLit(expr dst.Expr, val string) bool {
+	lit, ok := expr.(*dst.BasicLit)
+	return ok &&
+		lit.Kind == token.STRING &&
+		lit.Value == fmt.Sprintf("%q", val)
```
RomainMuller: There are multiple equivalent string representations for a given string, so this is not guaranteed to always return the correct value.

It would be more accurate to un-quote `lit.Value` with `strconv.Unquote` then match the un-quoted values.

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,339 @@
...
+
+func makeName(r *rule.InstFuncRule,
+	funcDecl *dst.FuncDecl, isBefore bool,
+) string {
+	prefix := TrampolineAfterName
+	if isBefore {
+		prefix = TrampolineBeforeName
+	}
+	return fmt.Sprintf("%s_%s%s",
+		prefix, funcDecl.Name.Name, util.CRC32(r.String()))
```
RomainMuller, line 30: CRC32 might not be a great collision-avoidance mechanism... It's designed so that two _similar_ values have a different checksum, but very different values may be more likely than desired to have the same checksum.

I'd recommend using `FNV-1a` instead, possibly the 256 or 512-bit variants, encoding the output as hex (easier) or identifier-safe Base64 (harder).

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,339 @@
...
+	// name them manually for further use
+	if retList := funcDecl.Type.Results; retList != nil {
+		retVals = make([]dst.Expr, 0)
+		for i, field := range retList.List {
+			if field.Names != nil {
+				for _, name := range field.Names {
+					retVals = append(retVals, ast.Ident(name.Name))
+				}
+			} else {
+				retValIdent := ast.Ident(fmt.Sprintf("retVal%d", i))
```
RomainMuller: I would recommend naming these with (at minimum) an `_` prefix to reduce the likelihood of collision with other variables in the same namespace. Ideally, they'd not look like something one would want to write by hand, and also somewhat unlikely to be generated by some (other) generator.

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,339 @@
...
+func collectArguments(funcDecl *dst.FuncDecl) []dst.Expr {
+	args := make([]dst.Expr, 0)
+	if ast.HasReceiver(funcDecl) {
+		if recv := funcDecl.Recv.List; recv != nil {
+			receiver := recv[0].Names[0].Name
+			args = append(args, ast.AddressOf(ast.Ident(receiver)))
+		} else {
+			util.Unimplemented()
+		}
+	}
```
RomainMuller, line 85: `ast.HasReceiver` already checked that `funcDecl.Recv.List` is not nil?

This makes it so that using the helper which is supposed to cut boilerplate actually forces more boilerplate (the `else unimplemented` is actually `else unreachable` here, isn't it?).

TBH it might be more elegant to have `ast.Receiver(decl *dst.FuncDecl) (*dst.Expr, bool)` return the receiver expression (the 1 value from the list) and `true`, or `nil` and `false` if there is no receiver.

RomainMuller: I'm also not quite sure I consider the receiver to be an argument in the traditional sense, so this is a little surprising to me.

tool/internal/instrument/template_api.go
RomainMuller, line 1: Since this is `//go:build ignore`, maybe that should be named `.tmpl.go` or `.go.tmpl`? It'd make it more obvious. Same applies to other such files.

tool/internal/instrument/template_impl.go
```diff
@@ -0,0 +1,119 @@
...
+
+//line <generated>:1
+type HookContextImpl struct {
+	Params      []interface{}
+	ReturnVals  []interface{}
+	SkipCall    bool
+	Data        interface{}
+	FuncName    string
+	PackageName string
+}
```
RomainMuller, line 16: I am quite partial to using `any` instead of empty interfaces (it is more semantically clear).

```suggestion
type HookContextImpl struct {
	Params      []any
	ReturnVals  []any
	SkipCall    bool
	Data        any
	FuncName    string
	PackageName string
}
```

tool/internal/instrument/template_impl.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (c *HookContextImpl) SetSkipCall(skip bool)    { c.SkipCall = skip }
+func (c *HookContextImpl) IsSkipCall() bool         { return c.SkipCall }
```
RomainMuller: We have writers + readers for everything... but all the fields are exported... This is a little weird (the accessors are uncalled for if the fields are all exported -- but since this backs an interface, I'd argue the fields should be unexported).

tool/internal/instrument/toolexec.go
```diff
@@ -6,51 +6,112 @@ package instrument
...
-	ip := &InstrumentPhase{
-		logger: logger,
+func stripCompleteFlag(args []string) []string {
+	for i, arg := range args {
+		if arg == "-complete" {
+			return append(args[:i], args[i+1:]...)
+		}
 	}
+	return args
+}
```
RomainMuller, line 80: Why would we need to do that? Are we adding C or ASM files to packages?

tool/internal/instrument/toolexec.go
```diff
@@ -6,51 +6,112 @@ package instrument
...
+func Toolexec(ctx context.Context, args []string) error {
 	// Check if the command is a compile command.
 	if util.IsCompileCommand(strings.Join(args, " ")) {
-		// Load matched hook rules from setup phase
-		rules, err := ip.load()
-		if err != nil {
-			return err
+		// Read compilation output directory
+		target := util.FindFlagValue(args, "-o")
+		util.Assert(target != "", "why not otherwise")
```
RomainMuller: I don't get the assertion message here?

tool/internal/instrument/toolexec.go
```diff
@@ -6,51 +6,112 @@ package instrument
...
 			// Okay, this package should be instrumented.
-			err = ip.instrument(args)
+			err = ip.instrument(matchedRules, args)
 			if err != nil {
 				return err
 			}
-			// return nil
+
+			// Strip -complete flag as we may insert some hook points that are
+			// not ready yet, i.e. they don't have function body
```
RomainMuller: Complete only states the package contains no C or ASM fragments... It essentially controls whether it'll pack extra objects than the ones produced by `gc` or not... So I don't think we actually need to remove it -- if the package was complete previously, it should still be (and vice-versa).

tool/internal/instrument/trampoline.go
```diff
@@ -0,0 +1,756 @@
...
+	TrampolineGetReturnValName      = "GetReturnVal"
+	TrampolineValIdentifier         = "val"
+	TrampolineCtxIdentifier         = "c"
+	TrampolineParamsIdentifier      = "Params"
+	TrampolineFuncNameIdentifier    = "FuncName"
+	TrampolinePackageNameIdentifier = "PackageName"
+	TrampolineReturnValsIdentifier  = "ReturnVals"
+	TrampolineHookContextImplType   = "HookContextImpl"
+	TrampolineBeforeNamePlaceholder = "\"OtelBeforeNamePlaceholder\""
+	TrampolineAfterNamePlaceholder  = "\"OtelAfterNamePlaceholder\""
```
RomainMuller:
```suggestion
	TrampolineBeforeNamePlaceholder = `"OtelBeforeNamePlaceholder"`
	TrampolineAfterNamePlaceholder  = `"OtelAfterNamePlaceholder"`
```

tool/internal/instrument/trampoline.go
RomainMuller, line 1: That file is pretty large, might benefit from being split into smaller units.

tool/internal/setup/add.go
```diff
@@ -13,19 +15,99 @@ const (
 	OtelRuntimeFile = "otel.runtime.go"
 )
 
-func genImportDecl(matched []*rule.InstRule) []dst.Decl {
-	importDecls := make([]dst.Decl, 0, len(matched))
+//nolint:gochecknoglobals // This is a constant
+var requiredImports = [][]string{
+	{"runtime/debug", "_otel_debug"}, // The getstack function depends on runtime/debug
+	{"log", "_otel_log"},             // The printstack function depends on log
```
RomainMuller, line 21: Are these always injected though?

tool/internal/setup/extract.go
```diff
@@ -0,0 +1,143 @@
...
+
+			_, err = io.CopyN(file, tarReader, header.Size)
+			if err != nil {
+				return ex.Error(err)
+			}
+			err = file.Close()
+			if err != nil {
+				return ex.Error(err)
+			}
+		}
```
RomainMuller: Why is this all wrapped in a block?

tool/internal/setup/extract.go
```diff
@@ -0,0 +1,143 @@
...
+	gzReader, err0 := gzip.NewReader(strings.NewReader(string(data)))
+	if err0 != nil {
+		return ex.Error(err0)
+	}
+	defer func() {
+		err0 = gzReader.Close()
+		if err0 != nil {
+			ex.Fatal(err0)
+		}
+	}()
```
RomainMuller, line 81: Why are these not named `err` but `err0`? That looks unidiomatic?

tool/internal/setup/extract.go
```diff
@@ -0,0 +1,143 @@
...
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return ex.Error(err)
+		}
+
+		// Skip AppleDouble files (._filename) and other hidden files
+		if strings.HasPrefix(filepath.Base(header.Name), "._") ||
+			strings.HasPrefix(filepath.Base(header.Name), ".") {
```
RomainMuller: Strings that start with `._` all start with `.` so it is pointless to check both.

```suggestion
		if strings.HasPrefix(filepath.Base(header.Name), ".") {
```

RomainMuller: We should probably also ensure that the embedded files do not contain any unnecessary data, as this is dead weight.

tool/util/go.go
```diff
@@ -85,3 +85,7 @@ func SplitCompileCmds(input string) []string {
 	}
 	return args
 }
+
+func IsGoFile(path string) bool {
+	return strings.HasSuffix(path, ".go")
```
RomainMuller: That is susceptible to false negatives on case-insensitive file systems.

.golangci.yml
```diff
@@ -197,7 +197,7 @@ linters:
       # Report about not checking of errors in type assertions: `a := b.(MyStruct)`.
       # Such cases aren't reported by default.
       # Default: false
-      check-type-assertions: true
+      check-type-assertions: false
```
RomainMuller: Why so? Not doing it and making a mistake results in a panic... this is kind of dangerous?

.markdownlint.yaml
```diff
@@ -10,7 +10,7 @@ MD010: false
 
 # line-length
 MD013:
-  line_length: 100
+  line_length: 160
```
RomainMuller, line 13: This is perhaps a little too generous for folks with smaller screens or reviewing side-by-side difffs. Usually 120 is a decent compromise. No hard feelings tho.

Makefile
```diff
@@ -15,9 +17,20 @@ BUILD_TIME := $(shell date -u '+%Y-%m-%d')
...
+# Package the instrumentation code into binary
+.PHONY: package
+package:
+	@echo "Packaging instrumentation code into binary..."
+	@rm -rf $(INST_PKG_TMP)
+	@cp -a pkg $(INST_PKG_TMP)
+	@cd $(INST_PKG_TMP) && go mod tidy
+	@tar -czf $(INST_PKG_GZIP) --exclude='*.log' $(INST_PKG_TMP)
+	@mv $(INST_PKG_GZIP) tool/data/
+	@rm -rf $(INST_PKG_TMP)
```
RomainMuller, line 29: The need for this makes it impossible to use `go get -tool` to manage use of this tool in `go.mod`, which ought to be the "idiomatic" way to go in go 1.25 and newer. I think that is a usability problem.

tool/util/sys.go
```diff
@@ -41,6 +44,15 @@ func IsUnix() bool {
 	return runtime.GOOS == "linux" || runtime.GOOS == "darwin"
 }
 
+func PathExists(path string) bool {
+	_, err := os.Stat(path)
+	return err == nil
+}
```
RomainMuller: This is mapping ALL errors to "does not exist" whereas only `fs.ErrNotExist` should. All file operations should be fail-able, always.

tool/internal/instrument/apply_func.go
```diff
@@ -0,0 +1,339 @@
...
+)
+
+const (
+	TJumpLabel      = "/* TRAMPOLINE_JUMP_IF */"
+	OtelGlobalsFile = "otel.globals.go"
+)
+
+func makeName(r *rule.InstFuncRule,
+	funcDecl *dst.FuncDecl, isBefore bool,
+) string {
```
RomainMuller: Instead of `isBefore bool`, you could pass a newtype constant, so that call-sites are easier to understand:

```go
makeName(r, decl, true)

// versus

makeName(r, decl, trampolineBefore)
```

RomainMuller (changes requested):
I was going to mention that we should probably to to work without `dave/dst` but I realize it makes it difficult to manage the `//go:linkname` directives that we need to output. That being said, I remain convinced we'd be better off without `dave/dst` in the long run because that dependency is not super well maintained anymore (Dave admitted to not having a lot of time to allocate to it since a long while). It also has a number of fairly bad bugs when dealing with import lists that cause hard failures when used against arbitrary code.

---

Second, this PR introduces large chunks of functionality without introducing corresponding unit tests. We need to improve our test coverage game or we will end up with future maintainability hazards as the product evolves.

---

Finally, the codebase will be more approachable to new maintainers/contributors if - at minimum - all exported functions had adequate documentation strings. This would at least make it easier to understand some of the intents without having to reason over the entire code.

kakkoyun:
> I was going to mention that we should probably to to work without `dave/dst` but I realize it makes it difficult to manage the `//go:linkname` directives that we need to output. That being said, I remain convinced we'd be better off without `dave/dst` in the long run because that dependency is not super well maintained anymore (Dave admitted to not having a lot of time to allocate to it since a long while). It also has a number of fairly bad bugs when dealing with import lists that cause hard failures when used against arbitrary code.
> 
> Second, this PR introduces large chunks of functionality without introducing corresponding unit tests. We need to improve our test coverage game or we will end up with future maintainability hazards as the product evolves.
> 
> Finally, the codebase will be more approachable to new maintainers/contributors if - at minimum - all exported functions had adequate documentation strings. This would at least make it easier to understand some of the intents without having to reason over the entire code.

Thanks, Romain, for raising the bar here. Starting with a high standard of quality pays dividends down the line. If we need to go deeper on any of the points, we can always set up an ad-hoc chat since our weekly sync was already this morning.

pdelewski:
@y1yang0 I extracted `compileCommand` from `toolexec` and added `instrument_test.go`. I pushed these changes to your branch so we don’t lose them and to keep it in sync with the current implementation. Feel free to tweak or revert if you prefer.

tool/internal/instrument/instrument_test.go
```diff
@@ -0,0 +1,125 @@
+//go:build !windows
```
pdelewski, line 1: Disabled temporarily on `windows` as it's failing, mostly due to path differences.

tool/internal/ast/primitives.go
pdelewski: I was thinking about the same, however did not mention during review :)

tool/internal/instrument/apply_func.go
pdelewski: That might be good idea

kakkoyun:
> I was going to mention that we should probably to to work without `dave/dst` but I realize it makes it difficult to manage the `//go:linkname` directives that we need to output. That being said, I remain convinced we'd be better off without `dave/dst` in the long run because that dependency is not super well maintained anymore (Dave admitted to not having a lot of time to allocate to it since a long while). It also has a number of fairly bad bugs when dealing with import lists that cause hard failures when used against arbitrary code.

Also, not using dave/dst is news to me. I agree it’s a good direction in the long run, however I think it’s more of a stretch goal at this stage. For now, We can keep the focus on getting the basics right first. There’s also an upstream Go issue that might help resolve this in the future: https://github.com/golang/go/issues/20744.

Besides that we should be any architectural decision that would tight our hands in the future.

pkg/instrumentation/helloworld/helloworld_hook.go
y1yang0: Great question. I'm planning to auto-generate the `go:linkname` tags with otel tool later on, instead of depending on hook author. Tracked in #29

tool/cmd/cmd-version.go
y1yang0: Let me clarify my point with a concrete example. Let's assume an error occurs during the YAML parsing process within `setup.Setup`. Here's a comparison of the outputs for the different approaches:

**1. `return cli.Exit(err, exitCodeFailure)`**

This approach preserves the exit code but loses the stack trace entirely.

```
Building demo with instrumentation...
failed to decode yaml file: yaml: line 14: could not find expected ':'
```

**2. `return ex.Errorf(cli.Exit(err, exitCodeFailure), "failed to setup")`**

This approach provides a stack trace, but it's the **truncated**. It starts from the call site (`cmd-go.go`), not from the original point of failure.

```
Building demo with instrumentation...
Error:
[0] failed to setup: failed to decode yaml file: yaml: line 14: could not find expected ':'

Stack:
[0]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/cmd-go.go:43 main.init.func1
[1]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:344 github.com/urfave/cli/v3.(*Command).run
[2]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:278 github.com/urfave/cli/v3.(*Command).run
[3]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:94 github.com/urfave/cli/v3.(*Command).Run
[4]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/main.go:52 main.main
[5]/usr/local/go/src/runtime/proc.go:283 runtime.mainmake: *** [Makefile:44: demo] Error 1
```

**3. `return ex.Errorf(err, "failed to setup")`**

This is the desired behavior. It provides the **full, original stack trace**, starting from the actual point of failure inside `setup/match.go`, allowing for much easier debugging.

```
Building demo with instrumentation...
Error:
[0] failed to decode yaml file: yaml: line 14: could not find expected ':'
[1] failed to setup

Stack:
[0]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/internal/setup/match.go:24 /tool/internal/setup.parseEmbeddedRule
[1]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/internal/setup/match.go:37 /tool/internal/setup.materalizeRules
[2]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/internal/setup/match.go:54 /tool/internal/setup.(*SetupPhase).matchedDeps
[3]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/internal/setup/setup.go:60 /tool/internal/setup.Setup
[4]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/cmd-go.go:41 main.init.func1
[5]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:344 github.com/urfave/cli/v3.(*Command).run
[6]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:278 github.com/urfave/cli/v3.(*Command).run
[7]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:94 github.com/urfave/cli/v3.(*Command).Run
[8]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/main.go:52 main.main
[9]/usr/local/go/src/runtime/proc.go:283 runtime.mainmake: *** [Makefile:44: demo] Error 1
```

The fundamental issue is that we need to **preserve and propagate the original `err` from its source all the way up to `main`**, instead of creating a new error midway(`ex.Errorf(cli.Exit(err)` creates a new one, sorry for misleading). Creating a new error, or wrapping it in a way that generates a new stack, breaks the error chain and loses the crucial information about where the problem actually originated.

And as you pointed out, this highlights that I need to add better documentation for our ex.Errorf framework to make this behavior clearer. I will address that.

tool/cmd/cmd-go.go
y1yang0: See below

tool/internal/ast/parser.go
y1yang0: Good catch

tool/internal/ast/primitives.go
y1yang0: I kept it all ast.XX for consistency. A dst.NewIdent call here would look out of place.

```
func setValue(field string, idx int, t dst.Expr) *dst.CaseClause {
	// *(c.Params[idx].(*int)) = val.(int)
	// c.Params[idx] = val iff type is interface{}
	se := ast.SelectorExpr(ast.Ident(TrampolineCtxIdentifier), field)
	ie := ast.IndexExpr(se, ast.IntLit(idx))
	te := ast.TypeAssertExpr(ie, ast.DereferenceOf(t))
	pe := ast.ParenExpr(te)
	de := ast.DereferenceOf(pe)
	val := ast.Ident(TrampolineValIdentifier)
	assign := ast.AssignStmt(de, ast.TypeAssertExpr(val, t))
	if ast.IsInterfaceType(t) {
		assign = ast.AssignStmt(ie, val)
	}
	caseClause := ast.SwitchCase(
		ast.Exprs(ast.IntLit(idx)),
		ast.Stmts(assign),
	)
	return caseClause
}
```

tool/internal/ast/primitives.go
y1yang0: For consistency see above

tool/internal/instrument/template_impl.go
y1yang0: Let's stick to using interface{} in the generated code for better backward compatibility. This ensures the code will compile correctly in environments where any is not yet defined

tool/internal/instrument/toolexec.go
y1yang0: We will insert function with no body for further golinkname, see its comment
```
			// Strip -complete flag as we may insert some hook points that are
			// not ready yet, i.e. they don't have function body
			ip.compileArgs = stripCompleteFlag(ip.compileArgs)

			args = ip.compileArgs
```

tool/internal/instrument/toolexec.go
y1yang0: This ensures that we can always parse the -o flag from the compile command. The goal is to prevent any unexpected behavior or nasty surprises from surfacing later in the code.

tool/internal/instrument/toolexec.go
y1yang0: Removing `-complete` flag is required when there is no function body. Keeping it will cause an error
```
Building demo with instrumentation...
go: found github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld in github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld v0.0.0-00010101000000-000000000000
WORK=/tmp/go-build1863349319
# github.com/open-telemetry/opentelemetry-go-compile-instrumentation/demo
<generated>:112: missing function body
<generated>:113: missing function body
Error:
[0] failed to run command /usr/local/go/pkg/tool/linux_amd64/compile: exit status 2
[1] failed to run toolexec with exitcode -1

Stack:
[0]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/util/sys.go:30 /tool/util.RunCmdWithEnv
[1]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/util/sys.go:36 /tool/util.RunCmd
[2]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/internal/instrument/toolexec.go:116 /tool/internal/instrument.Toolexec
[3]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/cmd-toolexec.go:22 main.init.func3
[4]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:344 github.com/urfave/cli/v3.(*Command).run
[5]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:278 github.com/urfave/cli/v3.(*Command).run
[6]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:94 github.com/urfave/cli/v3.(*Command).Run
[7]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/main.go:52 main.main
[8]/usr/local/go/src/runtime/proc.go:283 runtime.main
Error:
[0] failed to run command go: exit status 1
[1] failed to build with toolexec with exitcode -1

Stack:
[0]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/util/sys.go:30 /tool/util.RunCmdWithEnv
[1]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/internal/setup/setup.go:117 /tool/internal/setup.BuildWithToolexec
[2]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/cmd-go.go:46 main.init.func1
[3]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:344 github.com/urfave/cli/v3.(*Command).run
[4]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:278 github.com/urfave/cli/v3.(*Command).run
[5]/home/y1yang0/go/pkg/mod/github.com/urfave/cli/v3@v3.4.1/command_run.go:94 github.com/urfave/cli/v3.(*Command).Run
[6]/home/y1yang0/opentelemetry-go-compile-instrumentation/tool/cmd/main.go:52 main.main
[7]/usr/local/go/src/runtime/proc.go:283 runtime.mainmake: *** [Makefile:44: demo] Error 1
```

tool/internal/instrument/trampoline.go
y1yang0: I'll do this in separate PR.

tool/internal/setup/add.go
y1yang0: Yes, always

tool/internal/setup/extract.go
y1yang0: I renamed the first error to err0 because the name err is used for multiple subsequent checks in this function. This was primarily for brevity, so I could keep using the simple err variable. My apologies for the slightly awkward err0 name!

.markdownlint.yaml
y1yang0: README  badge exceeds 120, I have to use 160

tool/internal/ast/primitives.go
y1yang0: For consistency see above

tool/internal/instrument/apply_func.go
y1yang0: A name conflict here only occurs in one specific scenario: when the same function is instrumented by multiple rules. In that case, we need to add a suffix to the before and after functions for each rule to prevent the generated trampoline function names from colliding.

However, I believe a simple and short CRC32 hash is sufficient to prevent this, for two main reasons:

- It's a rare edge case: The situation where a single function is targeted by multiple instrumentation rules is expected to be very uncommon in practice.
- Hashes will differ: The CRC32 hash is calculated based on the rule's content, so different rules will naturally produce different hash values.

Given these points, the probability of a collision is extremely low, and a short CRC32 provides an adequate and efficient solution.

tool/internal/instrument/apply_func.go
y1yang0: Good catch

Makefile
y1yang0: There are several viable solutions for this, but let's defer addressing it for now, maybe go-lize the Makefile in the long run

y1yang0:
Thanks, everyone, for the incredibly helpful feedback and thorough review.

---

> Second, this PR introduces large chunks of functionality without introducing corresponding unit tests. We need to improve our test coverage game or we will end up with future maintainability hazards as the product evolves.

@RomainMuller  Yes, this is a high-priority item on my to-do list, and I'll be adding more tests in follow-up PRs.

> Finally, the codebase will be more approachable to new maintainers/contributors if - at minimum - all exported functions had adequate documentation strings. This would at least make it easier to understand some of the intents without having to reason over the entire code.

@RomainMuller I'll add the documentation to this PR to make the codebase more approachable for new contributors.

---

> For some reason I was thinking that we discussed that before and there is an internal agreement to use dave/dst. Personally, I would like to use standard library, however I know that this library has some advantages.

@pdelewski  Yes, we have discussed this topic before. Let's agree to defer the dave/dst discussion for now and revisit it later.

> I agree with you and we discussed this topic on last SIG meeting. My idea was to merge this sooner than later and then focus on testability.

@pdelewski  Yes, agreed. Adding more tests is a high priority for me as well.

> I extracted compileCommand from toolexec and added instrument_test.go. I pushed these changes to your branch so we don’t lose them and to keep it in sync with the current implementation. I originally planned to do that in separate PR. Feel free to tweak or revert if you prefer.

@pdelewski  Thank you for adding this code! I think it's valuable and highly relevant to the scope of this PR. Let's definitely keep it.

---

> Also, not using dave/dst is news to me. I agree it’s a good direction in the long run, however I think it’s more of a stretch goal at this stage. For now, We can keep the focus on getting the basics right first.

@kakkoyun Yes. Let's focus on getting the basics right in this PR and park the dave/dst discussion for a future effort.

tool/internal/instrument/template_api.go
y1yang0: I tend to use template_api.go and template_impl.go for aesthetic purposes.

tool/internal/instrument/template_impl.go
y1yang0: I agree. Let me do this in separate PR. Tracked in #29

tool/data/export.go
y1yang0: The otel-pkg.gz is generated before compile-time, and fixing this is non-trivial. I'll address it in a follow-up PR. Tracked in #29

pkg/inst/context.go
```diff
@@ -3,7 +3,12 @@
 
 package inst
 
-type Context interface {
+// !!!Any modification MUST be synced with tool/internal/instrument/template_api.go
```
ralf0131:
```suggestion
// !!! Any modification MUST be synced with tool/internal/instrument/template_api.go
```

tool/cmd/cmd-go.go
```diff
@@ -34,12 +40,12 @@ var commandGo = cli.Command{
 
 		err = setup.Setup(ctx)
 		if err != nil {
-			return cli.Exit(err, exitCodeFailure)
+			return ex.Errorf(err, "failed to build with toolexec with exitcode %d", exitCodeFailure)
```
ralf0131:
```suggestion
			return ex.Errorf(err, "failed to build with toolexec with exit code %d", exitCodeFailure)
```

tool/cmd/cmd-go.go
```diff
@@ -34,12 +40,12 @@ var commandGo = cli.Command{
...
 		}
 
 		err = setup.BuildWithToolexec(ctx, cmd.Args().Slice())
 		if err != nil {
-			return cli.Exit(err, exitCodeFailure)
+			return ex.Errorf(err, "failed to build with toolexec with exitcode %d", exitCodeFailure)
```
ralf0131:
```suggestion
			return ex.Errorf(err, "failed to build with toolexec with exit code %d", exitCodeFailure)
```

tool/cmd/cmd-setup.go
```diff
@@ -18,7 +19,7 @@ var commandSetup = cli.Command{
 	Action: func(ctx context.Context, _ *cli.Command) error {
 		err := setup.Setup(ctx)
 		if err != nil {
-			return cli.Exit(err, exitCodeFailure)
+			return ex.Errorf(err, "failed to setup with exitcode %d", exitCodeFailure)
```
ralf0131:
```suggestion
			return ex.Errorf(err, "failed to setup with exit code %d", exitCodeFailure)
```

pkg/inst/context.go
```diff
@@ -3,7 +3,12 @@
...
 
-type Context interface {
+// !!! Any modification MUST be synced with tool/internal/instrument/template_api.go
+type HookContext interface {
+	// Set the skip call flag, can be used to skip the original function call
+	SetSkipCall(bool)
+	// Get the skip call flag, can be used to skip the original function call
+	IsSkipCall() bool
 	// Set the data field, can be used to pass information between Before and After hooks
 	SetData(interface{})
```
kakkoyun: nit: We should migrate using `any` instead of `interface{}`.

tool/data/helloworld.yaml
```diff
@@ -1,13 +1,12 @@
...
+  path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld"
   # Where we want to inject the hook code
   pointcut: main.Example
-  # What hook code we want to inject
-  advice:
-    - before: MyHookBefore
-    - after: MyHookAfter
+  # What hook code we want to inject at target func entry or exit
+  before: MyHookBefore
+  after: MyHookAfter
```
kakkoyun, line 12: We plan to have different types of advices and injecting before and after hook is an advice for me.

So let's have:
```suggestion
  advice:
  - hook:    
    # What hook code we want to inject at target func entry or exit
    before: MyHookBefore
    after: MyHookAfter
```

tool/internal/rule/def.go
```diff
@@ -1,32 +0,0 @@
-// Copyright The OpenTelemetry Authors
```
kakkoyun, line 1: I think we should have this abstraction.

Adding a hook is just a type of "advice". Advice should be even more generic.

tool/data/helloworld.yaml
y1yang0: Could you please provide some specific examples? My motivation for this change is to create a generic model: `path + pointcut + before/after ` = `we inject the {before} and {after} functions at the start and end of the {pointcut} function, respectively. The code to be injected can be found at {path}`

 This description should be generic enough.

pkg/inst/context.go
y1yang0: Let's stick to using interface{} in the generated code for better backward compatibility. This ensures the code will compile correctly in environments where any is not yet defined. In normal tool code, `any` is definitively better and preferable

tool/data/helloworld.yaml
kakkoyun: Even though, you can do a lot of things with "before" and "after", it is just now enough. Our experience on writing integrations (https://github.com/search?q=repo%3ADataDog%2Fdd-trace-go%20orchestrion.yml&type=code), showed us you need a lit more flexibility. You can check the abstractions we had to invent previously https://datadoghq.dev/orchestrion/contributing/aspects/join-points

So I don't want us to tie our hands here. In the OpenTelemetry community, I'm sure we will have many more integrations and having the flexibility to expand on this space is quite valuable.

kakkoyun (commented):
More comments.

pkg/inst/context.go
kakkoyun: I understand the reasoning here. However, then we need to specify the minimum supported version then.

`any` is added Go 1.18 and any release of this tool should follow the Go support policy and it's the last two major versions. Supporting =< 1.18 is quite a farfetched goal to be honest.

tool/data/helloworld.yaml
y1yang0: Got it, thanks for explaining. I see where we differ now. What you call "advice" is basically what I call different "rules" in my design.

My InstRule interface is designed for exactly this, to support multiple rule types. We already have InstFuncRule, and I'm planning to add InstStructRule (for adding struct fields) and InstFileRule (as the most flexible option).

So, while we might disagree on the naming, it seems we're on the same page about the overall direction.

kakkoyun: Sorry, the request was specifically about the "advice"s, so this is more relevant https://datadoghq.dev/orchestrion/contributing/aspects/advice/

What we have right now could be solved with "prepend statements" in this jargon.

pkg/inst/context.go
y1yang0: Perhaps our strategy could be this: We officially support code injection for Go 1.22 and later (this is just an example and we can discuss it in a meeting). Projects on older Go versions can still use the OTel tool, but we don't guarantee that the generated code will compile.

In practice, supporting this on a tool level has zero cost, the only potential trade-off is using the non-modern interface{} instead of any.

The more fundamental issue, in my view, is that changing interface{} to any in the generated code brings zero benefits while potentially alienating a small number of users on older versions. For non-generated code, however, I fully agree with using any.

kakkoyun: If we decide to keep the integrations within the same repo, believe me supporting old versions will be a lot problem. But this is another discussion.

There are linters and a modern go initiative as part of `go vet` will update these sort of patterns. So I don't want to make this a bikeshedding discussion :)

y1yang0: ~~I don't have a firm conclusion on this point myself. However, considering that both you and Romain have recommended using any, and since I found the cost of change to be zero(search + repalce = done), I've decided to adopt the suggestion.~~

~~After all, it's not a critically important issue, and we'll have the flexibility to switch back on an as-needed basis later on. :)~~

.. sorry, spoke too soon. I'm retracting my last message. I realized this change also touches ast.InterfaceType and ast.IsInterfaceType, so it's not as simple as I thought.

pdelewski:
> I don't want to nitpick each and every thing, not to block the PR, however, I just wanted to note, we will have to change too many things afterwards, especially when it comes to style.
> 
> Some of the methods names are not idiomatic to Go, some of the file names are not very idiomatic to go. We are introducing too many indirections (helper functions, util packages etc.), with good intentions. to make the code more readable, however (again) this is not idiomatic Go. Some of the functions makes me feel like we try to invent a domain-specific language around AST.
> 
> We should also have to be conscience about allocations.

It’s true that some filenames are non-idiomatic and inconsistent (e.g., cmd-go.go vs. func_rule.go). We should use underscores (_) instead of hyphens (-). Some files could be combined to reduce the number of files—for example, instrument.go contains only one function. That said, we should all agree on what needs to be fixed in this PR and what can be deferred as follow-up tasks. I’m okay with merging this as is and making improvements afterward.

ralf0131 (approved):

ralf0131:
I am also okay with merging it first and improve it later, I think right now it is important for us to make progress rather than making it perfect.

y1yang0:
I've made the following changes:

- Replaced hyphens with underscores in all relevant filenames for consistency.
- Consolidated load.go and match.go into a single match.go file.

The structure of other files remains unchanged to allow for potential future feature additions.
I plan to merge this patch on Friday, Sep 26th if there is no further feedback. Thank you for the review!

tool/cmd/cmd-version.go
RomainMuller: Supposedly we can use `errors.As` to determine if we have an error-with-stacktrace somewhere below, and use that stacktrace? In this case, wrapping an error that comes from `ex.Errorf` with something else would allow the `return ex.Errorf(cli.Exit(err, exitCodeFailure), "failed to setup")` form to have the expected behavior?

tool/internal/ast/primitives.go
RomainMuller: It could then be (but this is not a big deal)

```suggestion
func Ident(name string) *dst.Ident {
	return dst.NewIdent(name)
}
```

tool/internal/ast/primitives.go
RomainMuller: I see the above comment about consistency, but what I am claiming here is that _none_ of these are necessary... Their presence cannot be justified by "consistency with themselves"?

tool/internal/instrument/apply_func.go
RomainMuller: Maybe we should then make it opaque, so we can chang it later on (if needed) without breaking untold assumptions... I.e, rename it to `util.Checksum` or `util.Hash` or something?

tool/internal/instrument/template_api.go
RomainMuller: My concern here is really that there are both templates and regular go code in this package, and it's not entirely obvious from looking at the file tree... You might be implying that the `template_` prefix here is how you know what is a template versus code that is built as part of the tool? Then I guess maybe that's fine... although from an aesthetics standpoint, I tend to prefer differentiating based on the file extension rather than prefix :)

tool/internal/instrument/template_impl.go
RomainMuller: That is supporting `go`s before 1.18... Which then also means strictly no use of generics anywhere in the injected code (for `go` before 1.18 will not be able to compile those), etc... And also this is `go` runtimes that long-since out of support...

I don't care all that much, but I personally thing explicitly trying to support these is making a disservice to the Go community by encouraging continued use of dangerously old releases.

tool/internal/instrument/toolexec.go
RomainMuller: Yeah but if I see a breakage with an assertion error that states:
> why not otherwise

I would not know what this is about without having to grep the source code to find the assert call...

That could be `util.Asser(target != "", "missing -o flag value")` so it's more helpful, isn't it?

tool/internal/instrument/toolexec.go
RomainMuller: Oh so this is a case where the documentation of `go tool compile` is lying. Great.

tool/internal/setup/extract.go
RomainMuller: I'd have rather addressed this by re-factoring the code to reduce the occurrences of `err0`... Also -- this is why I usually prefer the "in-line" style (`if err := call(); err != nil`) as it scopes the variable to just where it's useful instead of polluting the surrounding namespace.

y1yang0:
@RomainMuller Thanks for the final review, Romain. I was out of office for past few days. I'll address your comments in follow-up PRs.

tool/internal/setup/extract.go
y1yang0: Sounds good, I have no preference here. Feel free to refactor it as you see fit.

PR #67, opened 2025-09-28
refactor: differentiate between Wrapf and Newf

y1yang0:


docs/developing.md
```diff
@@ -1,39 +0,0 @@
-# Developing Details
-
-## Project Structure
-
-See the [project structure](./api-design-and-project-structure.md).
-
-## Error Handling
-
-We use the `ex` package to handle errors as we want to include the stack trace
```
pdelewski, line 9: Should not be this file updated instead of removing?

y1yang0: The idea is to focus our efforts on maintaining the documentation that lives closer to the code, instead of dealing with the overhead of keeping two redundant files in sync.

tool/ex/error.go
```diff
@@ -0,0 +1,148 @@
...
+	message []string
+	frame   []string
+	wrapped error
+}
+
+func (e *stackfulError) Error() string { return strings.Join(e.message, "\n") }
+func (e *stackfulError) Unwrap() error { return e.wrapped }
+
+func getFrames() []string {
+	const initFrames = 30
```
ralf0131, line 74: Could you add some comments why the initFrames is set to 30?

y1yang0: Admittedly, this isn't based on logic, but my intuition tells me a 30-layer stack is enough for debugging.

tool/ex/error.go
```diff
@@ -0,0 +1,148 @@
...
+
+func Wrap(previousErr error) error {
+	return wrapOrCreate(previousErr, "")
+}
+
+func Wrapf(previousErr error, format string, args ...any) error {
+	return wrapOrCreate(previousErr, format, args...)
+}
+
+func Newf(format string, args ...any) error {
```
kakkoyun, line 134: I think we should have `Errorf` to mimic `fmt.Errorf` and `New` for `errors.New`. It will be more familiar.

tool/ex/error.go
```diff
@@ -0,0 +1,148 @@
...
+func getFrames() []string {
+	const initFrames = 30
+	frameList := make([]string, 0)
+	pcs := make([]uintptr, initFrames)
+	n := runtime.Callers(numSkipFrame, pcs)
+	if n == 0 {
+		return frameList
+	}
+	pcs = pcs[:n]
+	frames := runtime.CallersFrames(pcs)
```
kakkoyun, line 82: One think, I would prefer to defer the symbolication. It's more costly compare to capturing the stack traces. So let's just capture the PCs when we create the error.

And then have a `Format` function where we symbolicate the PCs and have the frames (we can cache them after this point).

tool/ex/error.go
```diff
@@ -0,0 +1,148 @@
...
+type stackfulError struct {
+	message []string
+	frame   []string
+	wrapped error
+}
+
+func (e *stackfulError) Error() string { return strings.Join(e.message, "\n") }
+func (e *stackfulError) Unwrap() error { return e.wrapped }
+
+func getFrames() []string {
```
kakkoyun:
```suggestion
func captureStack() []string {
```

tool/ex/error.go
```diff
@@ -0,0 +1,148 @@
...
+		panic("Fatal error: unknown")
+	}
+	e := &stackfulError{}
+	if errors.As(err, &e) {
+		em := ""
+		for i, m := range e.message {
+			em += fmt.Sprintf("[%d] %s\n", i, m)
+		}
+		msg := fmt.Sprintf("Error:\n%s\nStack:\n%s", em, strings.Join(e.frame, "\n"))
+		_, _ = fmt.Fprint(os.Stderr, msg)
```
kakkoyun:
```suggestion
		_, _ = fmt.Fprintf(os.Stderr, "Error:\n%s\nStack:\n%s", em, strings.Join(e.frame, "\n"))
```

tool/ex/error.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	pcs = pcs[:n]
+	frames := runtime.CallersFrames(pcs)
+	cnt := 0
+	for {
+		frame, more := frames.Next()
+		if !more {
+			break
+		}
+		const prefix = "github.com/open-telemetry/opentelemetry-go-compile-instrumentation"
```
kakkoyun: Let's have this const at package level, easier to discover.

tool/ex/error.go
y1yang0: Okay, I add a New function to align with errors.New. I would still recommend keeping the name Newf because it's consistent with our other API designs, concise, and clearly expresses its intent.

tool/ex/error.go
y1yang0: This doesn't make much of a difference at the moment, as all errors propagate up to main and cause the program to exit. In other words, all errors are one-off and fatal.

kakkoyun: I understood that, however we design a package here and I'm sure there will be other use cases pretty soon. And adding the change is really a low effort. I can do it in a subsequent PR if you want to merge this immediately.

y1yang0: Ok, if you don’t mind, I hope this PR can be merged as soon as possible to fix the CI failure issue. The symbolization can be addressed later in a follow-up PR as an enhancement.

PR #81, opened 2025-10-09
chore: allow renovatebot to update all deps at once

y1yang0:
Configure Renovate to group all dependency updates into a single Pull Request, rather than creating an individual PR for each one.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -7,7 +7,7 @@
   "packageRules": [
     {
       "groupName": "all patch versions",
-      "schedule": ["before 8am every weekday"]
+      "schedule": ["before 8am on Monday"]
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -7,12 +7,7 @@
   "packageRules": [
     {
       "groupName": "all patch versions",
-      "matchUpdateTypes": ["patch"],
       "schedule": ["before 8am every weekday"]
```
kakkoyun: We should have this weekly, rather than daily.

y1yang0: Good catch! I missed this..

And FYI: CI failures is unrealted and fixed in #67

PR #98, opened 2025-10-16
feat: propagate otel trace context from caller goroutine

y1yang0:
- Add file rule to introduce new file to target package
- Add raw rule to inject raw Go source code
- Use file rule + raw rule to implement trace context propagation through goroutines

### DIFF
--- a/tool/data/helloworld.yaml
+++ b/tool/data/helloworld.yaml
@@ -13,18 +13,21 @@ hook_helloworld:
   path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld"
 
 add_new_field:
-  # Where we want to inject the hook code
   target: main
-  # The struct name to inject the field into
   struct: MyStruct
   new_field:
     - name: NewField
       type: string
 
 raw_helloworld:
-  # Where we want to inject the raw code
   target: main
-  # What function we want to inject the raw code into
   func: Example
-  # What raw code we want to inject into the target function
   raw: "go func(){ println(\"Hello, World!\") }()"
+
+hook_recv:
+  target: main
+  func: Example
+  recv: "*MyStruct"
+  before: MyHook1Before
+  after: MyHook1After
+  path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld"
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -142,8 +142,8 @@ func runMatch(dependency *Dependency, allRules []rule.InstRule) (*rule.InstRuleS
 			// Let's match with the rule precisely
 			switch rt := available.(type) {
 			case *rule.InstFuncRule:
-				funcDecl := ast.FindFuncDecl(tree, rt.Func)
-				if len(funcDecl) > 0 {
+				funcDecl := ast.FindFuncDecl(tree, rt.Func, rt.Recv)
+				if funcDecl != nil {
 					set.AddFuncRule(source, rt)
 				}
 			case *rule.InstStructRule:
@@ -152,8 +152,8 @@ func runMatch(dependency *Dependency, allRules []rule.InstRule) (*rule.InstRuleS
 					set.AddStructRule(source, rt)
 				}
 			case *rule.InstRawRule:
-				funcDecl := ast.FindFuncDecl(tree, rt.Func)
-				if len(funcDecl) > 0 {
+				funcDecl := ast.FindFuncDecl(tree, rt.Func, rt.Recv)
+				if funcDecl != nil {
 					set.AddRawRule(source, rt)
 				}
 			case *rule.InstFileRule:
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@755f44910c12a3d7ca0d8c6e42c048b3362f7cec # v3.30.8
+        uses: github/codeql-action/upload-sarif@16140ae1a102900babc80a33c44059580f687047 # v4.30.9
         with:
           sarif_file: results.sarif
--- a/demo/main.go
+++ b/demo/main.go
@@ -26,6 +26,8 @@ func (tc *traceContext) Clone() interface{} {
 
 type MyStruct struct{}
 
+func (m *MyStruct) Example() { println("MyStruct.Example") }
+
 // Example demonstrates how to use the instrumenter.
 func Example() {
 	// Output:
@@ -49,4 +51,5 @@ func main() {
 	m := &MyStruct{}
 	// Add a new field to the struct
 	println(m.NewField)
+	m.Example()
 }
--- a/pkg/instrumentation/helloworld/helloworld_hook.go
+++ b/pkg/instrumentation/helloworld/helloworld_hook.go
@@ -64,3 +64,13 @@ func MyHookAfter(ictx inst.HookContext) {
 	// This is the after hook, we can do some clean up work here if needed
 	fmt.Println("[MyHook] after hook executed!")
 }
+
+//go:linkname MyHook1Before main.MyHook1Before
+func MyHook1Before(ictx inst.HookContext, recv interface{}) {
+	println("Before MyStruct.Example()")
+}
+
+//go:linkname MyHook1After main.MyHook1After
+func MyHook1After(ictx inst.HookContext) {
+	println("After MyStruct.Example()")
+}
--- a/tool/internal/ast/shared.go
+++ b/tool/internal/ast/shared.go
@@ -4,6 +4,7 @@
 package ast
 
 import (
+	"fmt"
 	"go/token"
 	"strconv"
 
@@ -42,15 +43,44 @@ func FindFuncDeclWithoutRecv(root *dst.File, funcName string) *dst.FuncDecl {
 	return decls[0]
 }
 
-func FindFuncDecl(root *dst.File, funcName string) []*dst.FuncDecl {
-	const maxMatchDecls = 2
+func FindFuncDecl(root *dst.File, funcName string, recv string) *dst.FuncDecl {
 	decls := findFuncDecls(root, func(funcDecl *dst.FuncDecl) bool {
-		return funcDecl.Name.Name == funcName
+		// Receiver type is empty? Match func name only
+		name := funcDecl.Name.Name
+		if recv == "" {
+			return name == funcName && !HasReceiver(funcDecl)
+		}
+		if !HasReceiver(funcDecl) {
+			return false
+		}
+
+		// Receiver type is specified, Match both func name and receiver type
+		switch recvTypeExpr := funcDecl.Recv.List[0].Type.(type) {
+		case *dst.StarExpr: // func (*Recv)T
+			tn, ok := recvTypeExpr.X.(*dst.Ident)
+			if !ok {
+				// This is a generic type, we don't support it yet
+				return false
+			}
+			t := "*" + tn.Name
+			return t == recv
+		case *dst.Ident: // func (Recv)T
+			t := recvTypeExpr.Name
+			return t == recv
+		case *dst.IndexExpr:
+			// This is a generic type, we don't support it yet
+			return false
+		default:
+			msg := fmt.Sprintf("unexpected receiver type: %T", recvTypeExpr)
+			util.Unimplemented(msg)
+		}
+		return false
 	})
 
-	// one with receiver and one without receiver, at most two
-	util.Assert(len(decls) <= maxMatchDecls, "sanity check")
-	return decls
+	if len(decls) == 0 {
+		return nil
+	}
+	return decls[0]
 }
 
 func ListFuncDecls(root *dst.File) []*dst.FuncDecl {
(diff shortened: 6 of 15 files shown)


tool/internal/ast/parser.go
```diff
@@ -49,6 +49,21 @@ func (ap *AstParser) Parse(filePath string, mode parser.Mode) (*dst.File, error)
...
+func (ap *AstParser) ParseSnippet(source string) ([]dst.Stmt, error) {
+	if source == "" {
+		return nil, ex.New("empty source")
+	}
+	snippet := "package main; func _() {" + source + "}"
+	file, err := decorator.ParseFile(ap.fset, "", snippet, 0)
+	if err != nil {
+		return nil, ex.Wrap(err)
+	}
+	funcDecl, ok := file.Decls[0].(*dst.FuncDecl)
```
ralf0131, line 62: Is there a chance to trigger the index out of bounds exception?

tool/internal/ast/parser.go
```diff
@@ -49,6 +49,21 @@ func (ap *AstParser) Parse(filePath string, mode parser.Mode) (*dst.File, error)
...
+	util.Assert(ok, "must be a func decl")
```
ralf0131, line 63: I think it is better to elaborate more about the error message.

tool/internal/instrument/apply_file.go
```diff
@@ -0,0 +1,76 @@
...
+	"slices"
+	"strings"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/ex"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/ast"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/rule"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
+)
+
+func listRuleFiles(p string) ([]string, error) {
```
ralf0131: The input parameters should be more meaningful.

tool/internal/ast/parser.go
y1yang0: No, because our snippet explicitly defines a function named _, so it's guaranteed to exist.

tool/internal/ast/parser.go
y1yang0: This is an assertion, not an error. It shouldn't happen in a user's application. It's a guarantee for other developers to ensure that future changes don't break this contract. So, the current message is probably sufficient.

tool/data/runtime.yaml
```diff
@@ -1,5 +1,25 @@
...
+gls_linker:
+  target: "runtime"
+  file: "runtime_gls.go"
+  path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/runtime"
+
+goroutine_propagate:
+  target: "runtime"
+  func: "newproc1"
+  raw: |
+    defer func(){
```
pdelewski, line 19: If my understanding is correct, we’re also modifying the newproc1 signature to add the named return variable `_unnamedRetVal0`; however, I don’t see that anywhere and I’m not sure where it happens.

tool/internal/instrument/apply_file.go
```diff
@@ -0,0 +1,78 @@
...
+	// If the path is a local path, use it directly, otherwise, it's a module
+	// path, we need to convert it to a local path and find the module files
+	var p string
+	if util.PathExists(path) {
+		p = path
+	} else {
+		p = strings.TrimPrefix(path, util.OtelRoot)
+		p = filepath.Join(util.GetBuildTempDir(), p)
+	}
+	files, err := util.ListFiles(p)
```
pdelewski: I think this can be simplified to 
```suggestion
	return util.ListFiles(p)
```

tool/internal/setup/match.go
```diff
@@ -41,15 +43,35 @@ func parseEmbeddedRule(path string) ([]rule.InstRule, error) {
 			r.Target, ok = fields["target"].(string)
 			util.Assert(ok, "target is not a string")
 			rules = append(rules, &r)
-		} else if _, ok1 := fields["func"]; ok1 {
+		} else if _, ok2 := fields["file"]; ok2 {
```
pdelewski: What do you think about refactoring this and extract factory method:
```suggestion
// createRuleFromFields creates a rule instance based on the field type present in the YAML
func createRuleFromFields(raw []byte, name string, fields map[string]any) (rule.InstRule, error) {
	target, ok := fields["target"].(string)
	util.Assert(ok, "target is not a string")

	switch {
	case fields["struct"] != nil:
		var r rule.InstStructRule
		if err := yaml.Unmarshal(raw, &r); err != nil {
			return nil, ex.Wrap(err)
		}
		r.Name = name
		r.Target = target
		return &r, nil
	case fields["file"] != nil:
		var r rule.InstFileRule
		if err := yaml.Unmarshal(raw, &r); err != nil {
			return nil, ex.Wrap(err)
		}
		r.Name = name
		r.Target = target
		return &r, nil
	case fields["raw"] != nil:
		var r rule.InstRawRule
		if err := yaml.Unmarshal(raw, &r); err != nil {
			return nil, ex.Wrap(err)
		}
		r.Name = name
		r.Target = target
		return &r, nil
	case fields["func"] != nil:
		var r rule.InstFuncRule
		if err := yaml.Unmarshal(raw, &r); err != nil {
			return nil, ex.Wrap(err)
		}
		r.Name = name
		r.Target = target
		return &r, nil
	default:
		util.ShouldNotReachHere()
		return nil, nil
	}
}

// parseEmbeddedRule parses the embedded yaml rule file to concrete rule instances
//
//nolint:nestif // It has many if statements, but it's straightforward
func parseEmbeddedRule(path string) ([]rule.InstRule, error) {
	yamlFile, err := data.ReadEmbedFile(path)
	if err != nil {
		return nil, err
	}
	var h map[string]map[string]any
	err = yaml.Unmarshal(yamlFile, &h)
	if err != nil {
		return nil, ex.Wrap(err)
	}
	rules := make([]rule.InstRule, 0)
	for name, fields := range h {
		raw, err1 := yaml.Marshal(fields)
		if err1 != nil {
			return nil, ex.Wrap(err1)
		}

		r, err2 := createRuleFromFields(raw, name, fields)
		if err2 != nil {
			return nil, err2
		}
		rules = append(rules, r)
	}
	return rules, nil
}
```

.github/workflows/ci.yaml
```diff
@@ -29,8 +28,6 @@ jobs:
     strategy:
       matrix:
         platform:
-          - os: ubuntu-latest
```
pdelewski, line 32: Why these platforms were removed?

tool/data/helloworld.yaml
```diff
@@ -15,8 +15,18 @@ hook_helloworld:
 add_new_field:
   # Where we want to inject the hook code
   target: main
+  # What type of struct we want to inject the field into
```
pdelewski: What about:
```suggestion
  # The struct name to inject the field into
```

tool/data/helloworld.yaml
```diff
@@ -15,8 +15,18 @@ hook_helloworld:
...
   struct: MyStruct
+  # Which field we want to inject the code into
```
pdelewski: Am I correct that at the moment to add more than one field we have to duplicate `struct rule`?

.github/workflows/ci.yaml
y1yang0: The 386 architecture is quite old, and supporting it might not be worth the effort. It's also slowing down our CI runs. However, I can't make a final call on this. I'll leave the decision up to the maintainers.

tool/data/helloworld.yaml
y1yang0: Yes, but I agree that allowing a rule to apply to multiple fields would be a better approach.

tool/data/runtime.yaml
y1yang0: The constant `unnamedRetValNam` is used in two places to rename unnamed function return values. You can find the renamed code in `demo/.otel-build/debug/runtime/proc.go` after `make demo`.

pdelewski: Yes, I found it there. My question is: where are we making this signature change? I’d expect it to be explicit.

y1yang0: I'm not sure if I fully understand your question, but I'll try to answer it. The modification of the function signature happens in two main scenarios:

The first one is related to hooking, with the key logic located here:
```
instrument.collectReturnValues
instrument.(*InstrumentPhase).insertTJump
instrument.(*InstrumentPhase).applyFuncRule
// ... (and so on)
```
The reason for this modification is that we need to pass the arguments and return values to the HookContext. This allows the return values to be modified within the hook function via the HookContext.

The second scenario is for raw code injection, which happens here:
```
instrument.insertRaw
instrument.(*InstrumentPhase).applyRawRule
// ... (and so on)
```
The reason we modify the signature here is that if the raw code we inject needs to use a return value, we must give that return value a name (similar to how it's done in runtime.yaml).

Additionally, it's worth noting that if a function's return value already has a name, it will be left unchanged. We only assign names to unnamed return values.

Does this answer your question? Thanks!

pdelewski: I will look again at this code, but I'm talking about this:
image

pdelewski: The rule that we have in `runtime.yaml` is about injecting `defer` statement, but there is also a change in `newproc1` signature regarding return argument and my question is about that

y1yang0: First, the otel tool reads this rule from the configuration:
```yaml
goroutine_propagate:
  target: "runtime"
  func: "newproc1"
  raw: |
    defer func(){
      _unnamedRetVal0.otel_trace_context = propagateOtelContext(callergp.otel_trace_context);
      _unnamedRetVal0.otel_baggage_container = propagateOtelContext(callergp.otel_baggage_container);
    }()
```
After the setup phase, the execution flow reaches instrument.Toolexec. It identifies this as a raw rule and consequently calls `InstrumentPhase.applyRawRule`

```go
// Naming unnamed return values: (*g) => (_unnamedRetVal0 *g)
func renameReturnValues(funcDecl *dst.FuncDecl) {
  if retList := funcDecl.Type.Results; retList != nil {
    idx := 0
    for _, field := range retList.List {
      if field.Names == nil {
        name := fmt.Sprintf("%s%d", unnamedRetValName, idx)
        field.Names = []*dst.Ident{ast.Ident(name)}
        idx++
      }
y1yang0 marked this conversation as resolved.
    }
  }
}

func insertRaw(r *rule.InstRawRule, decl *dst.FuncDecl) error {
  util.Assert(r.Raw != "", "sanity check")
  // Rename the unnamed return values so that the raw code can reference them
  // THIS IS EXACTLY WHERE THE RENAMING HAPPENS!
  renameReturnValues(decl)
  // ... code to inject the raw snippet follows ...
  return nil
}

func (ip *InstrumentPhase) applyRawRule(rule *rule.InstRawRule, root *dst.File) error {
  for _, funcDecl := range funcDecls {
    err := insertRaw(rule, funcDecl) // inject the raw code snippet (i.e., the defer func(){...} block).
  }
  return nil
}

```

y1yang0: I see your point now. Let's imagine for a moment what would happen if we didn't rename the return values. How would we write this rule?
```yaml
goroutine_propagate:
  target: "runtime"
  func: "newproc1"
  raw: |
    defer func(){
      // How do we reference the first return value here?
      ??.otel_trace_context = propagateOtelContext(callergp.otel_trace_context);
      // And here?
      ??.otel_baggage_container = propagateOtelContext(callergp.otel_baggage_container);
    }()
```
We simply can't. There would be no way to refer to the unnamed return value in the raw code snippet.

.github/workflows/golangci-lint.yaml
```diff
@@ -19,7 +19,6 @@ jobs:
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
       - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
-          cache-dependency-path: "**/go.mod"
```
kakkoyun, line 22: Why are we removing this?

tool/data/runtime.yaml
pdelewski: Thank you! That was my fault — I was on a different branch and didn’t see `apply_raw.go`. I agree that we wouldn’t be able to do that; however, there’s currently some inconsistency, as the rule is partially defined in JSON and partially in the Go code.

.github/workflows/ci.yaml
pdelewski: I'm ok with it

tool/data/runtime.yaml
y1yang0: Here is my thinking on the use cases for InstFuncRule and InstRawRule:

- InstFuncRule (The Standard Rule) should be used when a developer needs to inject formal, structured code or wants to leverage the HookContext. With this approach, the injection details are defined in a configuration file (e.g., JSON), but the actual hook logic resides in a separate .go file. This provides a superior development experience, as developers can take full advantage of IDE features like syntax highlighting, auto-completion, and automatic imports.

- InstRawRule (The Raw Rule) is the better choice when a developer needs to inject small, flexible, or even incomplete code snippets (for instance, a quick defer statement or a debug print). It offers maximum flexibility by allowing the code to be written as a simple string directly within the configuration file.

In summary, my view is:

- InstFuncRule is the safe, standard, and type-checked option.
- InstRawRule is the unsafe, flexible, and non-type-checked option, trading safety for convenience.

This is my current perspective, and I am open to discussing better alternatives if they exist. Thank you!

.github/workflows/golangci-lint.yaml
y1yang0: The cache-dependency-path configuration tells the caching system that the current cache should be invalidated whenever the specified files change.

However, in practice, we often encounter a scenario where the go.mod file remains unchanged, but the code for our tools is modified. This frequently causes our CI linters to use a stale cache and fail, because the cache was not invalidated as expected.

PR #105, opened 2025-10-20
feat: instrument func with receiver type

y1yang0:
Allow to instrument `func (*Recv) Func()`

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@755f44910c12a3d7ca0d8c6e42c048b3362f7cec # v3.30.8
+        uses: github/codeql-action/upload-sarif@16140ae1a102900babc80a33c44059580f687047 # v4.30.9
         with:
           sarif_file: results.sarif


tool/internal/ast/shared.go
```diff
@@ -42,15 +43,45 @@ func FindFuncDeclWithoutRecv(root *dst.File, funcName string) *dst.FuncDecl {
 	return decls[0]
 }
 
-func FindFuncDecl(root *dst.File, funcName string) []*dst.FuncDecl {
-	const maxMatchDecls = 2
+func FindFuncDecl(root *dst.File, funcName string, recv string) *dst.FuncDecl {
```
pdelewski, line 46: hmm, maybe `FindFuncDecl` should just take `*rule.InstFuncRule` as argument

tool/internal/ast/shared.go
```diff
@@ -42,15 +43,45 @@ func FindFuncDeclWithoutRecv(root *dst.File, funcName string) *dst.FuncDecl {
...
+			return name == funcName && !HasReceiver(funcDecl)
+		}
+		if !HasReceiver(funcDecl) {
+			return false
+		}
+
+		// Receiver type is specified, Match both func name and receiver type
+		switch recvTypeExpr := funcDecl.Recv.List[0].Type.(type) {
+		case *dst.StarExpr: // func (*Recv)T
+			if _, ok := recvTypeExpr.X.(*dst.Ident); !ok {
```
pdelewski: I think that this can be simplified to:
```suggestion
 tn, ok := recvTypeExpr.X.(*dst.Ident)
  if !ok {
      // This is a generic type, we don't support it yet
      return false
  }
  t := "*" + tn.Name
  return t == recv
```

tool/internal/ast/shared.go
y1yang0: I initially thought this was a good idea, but after trying it out, I found that in some call sites, the function name itself is passed directly. This means we're not always looking up the target Func for a given rule; it's also possible that we need to find the corresponding Hook function. So perhaps it's best to keep the current implementation.

https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/bdef73a5f61240fada5b649b0efb7665964d640a/tool/internal/instrument/trampoline.go#L155-L159

PR #107, opened 2025-10-23
feat: add version to instrumentation rules

y1yang0:
This patch adds version-aware rules. For example, we can write one rule for **redis@[v1.2.3, v1.3.5)** and another for **redis@[v1.3.6, v1.4.0)** . This allows us to handle situations where a Redis version update might remove or alter the behavior of a target function, ensuring our instrumentation remains stable and correct across different versions.

Before this patch, we could only instrument standard library code. Now, with this patch, we can truly instrument any function in any third-party module. It paves the way for instrumenting gRPC.

### DIFF
--- a/test/infra.go
+++ b/test/infra.go
@@ -12,7 +12,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func RunCmd(dir string, args ...string) (string, error) {
+// -----------------------------------------------------------------------------
+// E2E Test Infrastructure
+// This infrastructure is used to actually build the application, execute the
+// compiled binary program, and verify that the output of the binary program
+// is as expected.
+
+func runCmd(dir string, args ...string) (string, error) {
 	path := args[0]
 	args = args[1:]
 	cmd := exec.Command(path, args...)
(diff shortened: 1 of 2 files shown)


docs/rules.md
```diff
@@ -0,0 +1,128 @@
...
+
+### 3. Raw Code Injection Rule
+
+This rule injects a string of raw Go code at the beginning of a target function. This offers great flexibility but should be used with caution as the injected code is not checked for correctness at definition time.
+
+**Use Cases:**
+
+- Injecting complex logic that cannot be expressed with a simple function call.
+- Quick and dirty debugging or logging.
+- Prototyping new instrumentation strategies.
```
ralf0131, line 86: Another use case may be custom instrumentation for traces and metrics.

docs/rules.md
```diff
@@ -0,0 +1,128 @@
...
+
+**Fields:**
+
+- `func` (string, required): The name of the target function.
+- `recv` (string, optional): The receiver type for a method.
+- `raw` (string, required): The raw Go code to be injected. The code will be inserted at the beginning of the target function.
```
ralf0131, line 93: What about adding the raw Go code to the end of the target function?

y1yang0: Semantically, `defer func(){ ....code }()` is equivalent to inserting code at the end of the function.

README.md
```diff
@@ -1,41 +1,71 @@
-# OpenTelemetry Go Compile Instrumentation
+<div align="center">
```
pdelewski: These changes are out of scope of this PR IMO

test/infra.go
```diff
@@ -0,0 +1,42 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package test
```
pdelewski, line 4: Both this file and the one above belong to tool. It might make more sense to move the entire test package into the tool directory and treat them as tests of tool.

y1yang0: These are E2E tests. In the future, once the functionality is more complete, they will also validate that traces and spans are generated as expected. Therefore, they aren't just for testing the tool itself.

pdelewski: ok, then, maybe it would be good to add some comments or description. I was a bit confused at the beginning. I had to read the code to understand what they are doing and that they invoke the `tool`.

PR #119, opened 2025-10-31
chore: optimize performance-critical tjump

y1yang0:
Simplify trampoline-jump-if as it's performance-critical

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,13 @@ build: package
 	@go build -a -ldflags "-X main.Version=$(VERSION) -X main.CommitHash=$(COMMIT_HASH) -X main.BuildTime=$(BUILD_TIME)" -o $(BINARY_NAME)$(EXT) ./$(TOOL_DIR)
 	@./$(BINARY_NAME)$(EXT) version
 
+
+.PHONY: build-demo-grpc
+build-demo-grpc:
+	@echo "Building gRPC demo..."
+	@cd demo/grpc/server && go generate && go build -o server .
+	@cd demo/grpc/client && go build -o client .
+
 # Run the test with instrumentation
 .PHONY: test
 test: build
@@ -58,3 +65,8 @@ clean:
 	rm -f $(BINARY_NAME)$(EXT)
 	rm -f demo/basic/basic
 	rm -rf demo/basic/.otel-build
+	rm -f demo/grpc/server/server
+	rm -rf demo/grpc/server/pb
+	rm -f demo/grpc/client/client
+	rm -rf demo/grpc/server/.otel-build
+	rm -rf demo/grpc/client/.otel-build
--- a/demo/grpc/README.md
+++ b/demo/grpc/README.md
@@ -1,0 +1,127 @@
+# gRPC Demo
+
+This directory contains a simple gRPC server and client implementation for demonstrating OpenTelemetry compile-time instrumentation.
+
+## Structure
+
+- `server/` - gRPC server implementation
+  - `main.go` - Server code with unary and streaming RPC handlers (includes go:generate directives)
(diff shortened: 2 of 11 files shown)


ralf0131:
The linter CI is failing, which needs to be fixed.

tool/internal/instrument/apply_func.go
```diff
@@ -297,6 +301,9 @@ func (ip *InstrumentPhase) parseFile(file string) (*dst.File, error) {
 		return nil, err
 	}
 	ip.target = root
+	// Every time we parse a file, we need to reset the trampoline jumps
+	// because they are associated with one certain file
+	ip.tjumps = make([]*TJump, 0)
```
RomainMuller: You could probably truncate it instead, as this would reduce re-allocations (nb -- off the top of my head I have a doubt whether this is `nil`-safe or not):


```suggestion
	ip.tjumps = ip.tjumps[:0]
```

tool/internal/instrument/optimize.go
```diff
@@ -0,0 +1,223 @@
...
+// The compiler responsible for hoisting the initialization statement out of the
+// if skeleton, and the dce and sccp passes will remove the whole then block. All
+// these trampoline functions looks as if they are executed sequentially, i.e.
+//
+//	ctx,_ := otel_trampoline_before(&arg);
+//	defer otel_trampoline_after(ctx, &retval)
+//
+// Note that this optimization pass is fragile as it really heavily depends on
+// the structure of trampoline-jump-if and trampoline functions. Any change in
+// tjump should be carefully examined.
```
RomainMuller, line 79: Given this, there should be a comprehensive test suite to validate behaviors here... otherwise it's just good intention & wishful thinking...

tool/internal/instrument/trampoline.go
```diff
@@ -453,8 +453,8 @@ func assignSliceLiteral(assignStmt *dst.AssignStmt, vals []dst.Expr) bool {
 	return false
 }
 
-// replenishHookContext replenishes the hook context before hook invocation
-func (ip *InstrumentPhase) replenishHookContext(before bool) bool {
+// populateHookContext populates the hook context before hook invocation
+func (ip *InstrumentPhase) populateHookContext(before bool) bool {
```
RomainMuller, line 457: Definitely like the new name better 👍🏻

tool/internal/instrument/optimize.go
pdelewski: I agree that this should be carefully tested. We can begin with a minimal set of test cases based on what we currently know, then broaden coverage.

y1yang0:
@pdelewski  Thank you for your very valuable contribution on the unit tests. I really appreciate it! I may have some minor changes to make, which I'll go ahead and commit directly to the branch. Thanks again for your work!

I'd also like to share my thoughts on the testing strategy for the instrument package.

For methods that operate on a receiver, like (*InstrumentPhase).createTrampoline, optimizeTJump, etc., I'm thinking that writing individual unit tests for them might not be the most effective approach. The main difficulty is that we would need to meticulously construct a specific receiver state just to call and validate them in isolation.

Perhaps a more practical strategy would be to enhance instrument_test.go with more comprehensive, higher-level tests that naturally exercise these code paths. This way, we test their behavior within a more realistic context.

On the other hand, for standalone (non-receiver) functions like stripCompleteFlag, having dedicated unit tests is absolutely reasonable and the right approach, as they can be easily tested in isolation.

pdelewski:
@y1yang0 I agree we need to find the right balance for testing. Early on, I prefer a minimal set of unit tests, since interfaces change often and refactoring tests adds maintenance overhead. That said, we should cover critical behaviors.

PR #120, opened 2025-11-03
feat(tool): automatically generate go:linkname directives for hook fu…

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/29

> Auto-generate go:linkname tags instead of depending on hook author

### DIFF
--- a/pkg/instrumentation/nethttp/server_hook.go
+++ b/pkg/instrumentation/nethttp/server_hook.go
@@ -6,12 +6,10 @@ package nethttp
 import (
 	"fmt"
 	"net/http"
-	_ "unsafe"
 
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst"
 )
 
-//go:linkname BeforeServeHTTP net/http.BeforeServeHTTP
 func BeforeServeHTTP(ictx inst.HookContext, _ interface{}, w http.ResponseWriter, r *http.Request) {
 	fmt.Println("BeforeServeHTTP")
 	// TODO: Implement the real server hook logic here


tool/internal/instrument/trampoline.go
```diff
@@ -68,6 +69,23 @@ func (ip *InstrumentPhase) addDecl(decl dst.Decl) {
...
+// ensureUnsafeImport ensures that the unsafe package is imported in the target file.
+// This is required when using //go:linkname directives.
+func (ip *InstrumentPhase) ensureUnsafeImport() {
+	for _, decl := range ip.target.Decls {
+		genDecl, ok := decl.(*dst.GenDecl)
+		if !ok || genDecl.Tok != token.IMPORT {
+			continue
+		}
+		for _, spec := range genDecl.Specs {
+			if importSpec, ok2 := spec.(*dst.ImportSpec); ok2 && importSpec.Path.Value == `"unsafe"` {
```
pdelewski: nitpick: this `unsafe` literal might be extracted as const

tool/internal/instrument/trampoline.go
```diff
@@ -332,11 +350,22 @@ func (ip *InstrumentPhase) addHookFuncVar(t *rule.InstFuncRule,
 			Func:   false,
 			Params: paramTypes,
 		},
+		Decs: dst.FuncDeclDecorations{
```
pdelewski, line 355: Please, add unit test for these changes.

y1yang0:
Thanks for your contribution!

The current approach is to add a go:linkname directive to the hook function declarations generated in the target package. However, if the target package doesn't originally import unsafe, simply adding import "unsafe" at the source level will fail to compile. We would also need to add the unsafe package to the importcfg file.

For example, consider the following compile command for `sort` package:

`usr/local/go/pkg/tool/linux_amd64/compile -o $WORK/b042/_pkg_.a -trimpath "$WORK/b042=>" -p slices -lang=go1.24 -std -complete -buildid nluHZE85vwBGLnUYqyu_/nluHZE85vwBGLnUYqyu_ -goversion go1.24.2 -c=4 -nolocalimports -importcfg $WORK/b042/importcfg -pack /usr/local/go/src/slices/iter.go /usr/local/go/src/slices/slices.go /usr/local/go/src/slices/sort.go /usr/local/go/src/slices/zsortanyfunc.go /usr/local/go/src/slices/zsortordered.go **wow.go**`

If we add a new file like **wow.go** that imports `unsafe` (which wasn't used by the `sort` package before), we must also add an entry for `unsafe` to `$WORK/b042/importcfg `. This dependency then needs to be propagated, which would require us to modify the linking process directly. This is a very challenging task

A simpler solution would be to handle this during the SetupPhase. For instance, when we extract the `pkg` from embedded data into the `.otel-build `directory (in `SetupPhase.extract`), we could directly modify the hook files to add the go:linkname directive at that stage.

tool/internal/instrument/trampoline.go
txabman42: After reading [this](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/119#issuecomment-3483461223), I've some doubts regarding the testing strategy for this package.

In the end, this logic should be covered by e2e tests, as there is no func instrumentation without linknames.

Should we continue pushing unit testing here?

tool/internal/instrument/trampoline.go
txabman42: Agree, moved to const

tool/internal/instrument/trampoline.go
```diff
@@ -693,6 +725,8 @@ func (ip *InstrumentPhase) callHookFunc(t *rule.InstFuncRule, before bool) error
 }
 
 func (ip *InstrumentPhase) createTrampoline(t *rule.InstFuncRule) error {
+	// Ensure unsafe package is imported since we use //go:linkname directives
+	ip.ensureUnsafeImport()
```
y1yang0: Move to `insertTJump` right before calling `ip.createTrampoline`

tool/internal/instrument/trampoline.go
```diff
@@ -566,6 +597,7 @@ func getValue(field string, idx int, t dst.Expr) *dst.CaseClause {
 	de := ast.DereferenceOf(pe)
 	ret := ast.ReturnStmt(ast.Exprs(de))
 	if ast.IsInterfaceType(t) {
+		//nolint:ireturn // we don't know the type of the parameter
```
y1yang0: After merging main branch, this could be eliminated

tool/internal/instrument/trampoline.go
```diff
@@ -332,11 +352,22 @@ func (ip *InstrumentPhase) addHookFuncVar(t *rule.InstFuncRule,
...
 		},
+		Decs: dst.FuncDeclDecorations{
+			NodeDecs: dst.NodeDecs{
+				Before: dst.NewLine,
+				Start: dst.Decorations{
+					fmt.Sprintf("//go:linkname %s %s.%s",
+						fnName, t.Path, fnName),
+				},
+			},
+		},
```
y1yang0, line 363: This part could be abstracted into a function like ast.NewLineComment in tool/internal/ast/primitive.go, as there are multiple places with a nearly identical pattern.

y1yang0 (commented):
Great work. I have a few comments

tool/internal/instrument/trampoline.go
```diff
@@ -68,6 +70,24 @@ func (ip *InstrumentPhase) addDecl(decl dst.Decl) {
...
+			continue
+		}
+		for _, spec := range genDecl.Specs {
+			if importSpec, ok2 := spec.(*dst.ImportSpec); ok2 &&
+				importSpec.Path.Value == strconv.Quote(unsafePackageName) {
+				return
+			}
+		}
+	}
+	ip.target.Decls = append([]dst.Decl{ast.ImportDecl("_", unsafePackageName)}, ip.target.Decls...)
```
y1yang0, line 88:
```suggestion
	ip.addDecl(ast.ImportDecl("_", unsafePackageName))
```

txabman42: This would add unsafe import after other declarations. This would don't work if it goes after other type of declarations.

It'd be better to keep it as the first declaration as it's right now.

tool/internal/instrument/trampoline.go
txabman42: Good catch, removed

tool/internal/instrument/trampoline.go
txabman42: Thats a good idea. Let's keep it as it's in this PR and I'll create a new one for this change.

y1yang0: Tracked in #136

y1yang0:
It looks like inserting //go:linkname into the target package for VictoriaMetrics prevents it from compiling successfully.
(https://github.com/VictoriaMetrics/VictoriaMetrics)

e.g., otel go build -a ./app/vmctl fails.

Maybe we should take another look at this solution.

txabman42:
I think `otel go build -a ./app/vmctl` in VictoriaMetrics fails due to these two different problems:
- https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/195
- https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/196

PR #121, opened 2025-11-03
feat: Adding gRPC demo app

pdelewski:
The goal of this PR is to add minimal gRPC demo app that will be further use as a source for instrumentation.


y1yang0:
"Perhaps we could keep the generated .pb.go files in the repository? They are platform-agnostic, and this would allow users to try out the demo without needing to install protoc and the Go plugins.

kakkoyun (approved):
LGTM 🚀

PR #132, opened 2025-11-06
chore: add http integration test

y1yang0:
add http integration test

### DIFF
--- a/test/http_test.go
+++ b/test/http_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/test/app"
 	"github.com/stretchr/testify/require"
 )
 
@@ -57,15 +58,15 @@ func TestHttp(t *testing.T) {
 	clientDir := filepath.Join("..", "demo", "http", "client")
 
 	// Build the server and client applications with the instrumentation tool.
-	BuildApp(t, serverDir, "go", "build", "-a")
-	BuildApp(t, clientDir, "go", "build", "-a")
+	app.Build(t, serverDir, "go", "build", "-a")
+	app.Build(t, clientDir, "go", "build", "-a")
 
 	// Start the server and wait for it to be ready.
-	serverApp, outputPipe := StartApp(t, serverDir)
+	serverApp, outputPipe := app.Start(t, serverDir)
 	waitUntilDone := waitUntilReady(t, serverApp, outputPipe)
 
 	// Run the client, it will send a shutdown request to the server.
-	RunApp(t, clientDir, "-shutdown")
+	app.Run(t, clientDir, "-shutdown")
 
 	// Wait for the server to exit and return the output.
 	output := waitUntilDone()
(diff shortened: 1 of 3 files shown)


test/http_test.go
```diff
@@ -0,0 +1,74 @@
+// Copyright The OpenTelemetry Authors
```
kakkoyun, line 1: Let's add a build flag like `e2e` and we run these seperately. It would be helpful to a helper action to `Makefile` to run e2e tests.

Also let's add a CI step for it.

test/infra.go
```diff
@@ -37,15 +35,40 @@ func BuildApp(t *testing.T, appDir string, args ...string) {
...
+	cmd := newCmd(t.Context(), appDir, args...)
+	out, err := cmd.CombinedOutput()
 	require.NoError(t, err, out)
 }
 
 // RunApp runs the application and returns the output.
-func RunApp(t *testing.T, dir string) string {
-	out, err := runCmd(t.Context(), dir, "./"+filepath.Base(dir))
+// It waits for the application to complete.
+func RunApp(t *testing.T, dir string, args ...string) string {
```
kakkoyun: nit: No need to export the method.

pkg/instrumentation/nethttp/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/nethttp
```
kakkoyun, line 1: I think we should create a top-level `contrib` directory to the repo to aggregate the instrumentation now. Probably we will have a lot of them. We can keep `pkg` for the inner workings of the app.

And `contrib` can be all the integrations we plan to provide. It would make easier to reason about and maybe have outside codeowners in the future.

kakkoyun (commented):
Thanks for kicking this of. 

I have several comments.

test/http_test.go
y1yang0: All tests under the test directory appear to be E2E tests. I'm not quite sure what the +tag is used for, especially since I noticed the Makefile already clearly distinguishes between unit tests (test-unit) and E2E tests (test-integration).

pkg/instrumentation/nethttp/go.mod
y1yang0: Good idea. Since the changes are large (embedding the contrib package, etc.), it could be a separate task 

Tracked in #29

test/http_test.go
kakkoyun: This is quite a common practices. We will keep adding relatively long-running tests to under `e2e`. Yes, you can run them isolation when you want to run them, however you are missing the point of running unit tests in a bulk with `go test ./...`. Unit tests tend to be all over the place and it's harder to isolate the individual packages. I believe having a dedicated build tag for e2e tests would be cleaner.

y1yang0: Okay, let' me do it

kakkoyun: 🙏 Thank you so much

kakkoyun: @y1yang0 We can merge this, and I can also handle it in subsequent PRs if you prefer?

y1yang0: > @y1yang0 We can merge this, and I can also handle it in subsequent PRs if you prefer?

Absolutely ok!

PR #137, opened 2025-11-07
chore(ci): Add Go license header checker

kakkoyun:

### DIFF
--- a/pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
+++ b/pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
@@ -7,9 +7,10 @@ import (
 	"context"
 	"sync/atomic"
 
-	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst-api-semconv/instrumenter/utils"
 	"go.opentelemetry.io/otel/attribute"
 	semconv "go.opentelemetry.io/otel/semconv/v1.30.0"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst-api-semconv/instrumenter/utils"
 )
 
 type (
--- a/pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor_test.go
+++ b/pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor_test.go
@@ -7,10 +7,11 @@ import (
 	"context"
 	"testing"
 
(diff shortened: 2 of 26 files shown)


Makefile
```diff
@@ -275,3 +285,9 @@ checkmake: ## Install checkmake if not present
 		echo "Installing checkmake..."; \
 		go install github.com/checkmake/checkmake/cmd/checkmake@latest; \
 	fi
+
+go-license: ## Install go-license if not present
+	@if ! command -v go-license >/dev/null 2>&1; then \
+		echo "Installing go-license..."; \
+		go install github.com/palantir/go-license@latest; \
```
yuluo-yx, line 292: I think we don't need to add this action. We can do it like in the main database or use apache/skywalking-eyes instead.

.github/workflows/check-license-headers.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - main
+    paths:
+      - "**/*.go"
+      - "license.yml"
+      - ".github/workflows/check-license-headers.yaml"
+  pull_request:
+    paths:
+      - "**/*.go"
+      - "license.yml"
+      - ".github/workflows/check-license-headers.yaml"
```
yuluo-yx, line 15: Is this change really necessary?

kakkoyun: I liked to run trigger the CI when the configuration files change. What is the downside?

Makefile
kakkoyun: Care to give some links to the examples you mentioned? I didn't actually get what you are suggesting here.

kakkoyun:
@yuluo-yx I will address your comments in the subsequent PRs if necessary. Or feel free to open a PR, we can discuss in there.

.github/workflows/check-license-headers.yaml
yuluo-yx: The triggering condition should be all files in the project, not just Go files and those two files. I also think the Makefile and other YAML files should need license checks.

Furthermore, triggering CI when the check-license-headers.yaml file changes seems pointless, since it's also a YAML file, and YAML files also require a license header.

Makefile
yuluo-yx: Okay, what I mean is that this repository has very few stars. I think github.com/palantir/go-license could be replaced with a more stable library.

Or, why should we use a library? I think a shell script would be perfectly sufficient for this simple function.

for example: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/6c8f96efe133a5b00309e57f9919dfadca8a7c9f/Makefile#L235

Issue #115, opened 2025-10-30
Parallelize SetupPhase.matchDeps


y1yang0:
Parallelize `SetupPhase.matchDeps`

y1yang0:
Assigned @txabman42 
Thanks.

PR #138, opened 2025-11-07
refactor: parallelize dependency matching in matchDeps

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/115

This PR parallelize the `matchDeps` execution. It also adds a pre-indexing to avoid transversing rules per dependency.

No business logic change.

Goroutines have been limites to CPU*2, to prevent resource exhaustion and excessive context switching overhead when processing large dependency sets.

Note that matchDeps performance could still be improved, but I think no need that extra complexity/refactor in the current project stage.

Some local quick performance tests against:

- `Demo basic app` --> **From ~450ms to ~220ms**
- `Linkerd2` --> **From ~7.6s to ~270ms** (just to test performance VS a bigger project with more dependencies)

### DIFF
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -149,16 +149,31 @@ func matchVersion(dependency *Dependency, rule rule.InstRule) bool {
 
 // runMatch performs precise matching of rules against the dependency's source code.
 // It parses source files and matches rules by examining AST nodes
-func (sp *SetupPhase) runMatch(dependency *Dependency, rules []rule.InstRule) (*rule.InstRuleSet, error) {
-	set := rule.NewInstRuleSet(dependency.ImportPath)
+func (sp *SetupPhase) runMatch(dep *Dependency, rulesByTarget map[string][]rule.InstRule) (*rule.InstRuleSet, error) {
+	set := rule.NewInstRuleSet(dep.ImportPath)
+
+	// Filter rules by target
+	relevantRules := rulesByTarget[dep.ImportPath]
+	if len(relevantRules) == 0 {
+		return set, nil
+	}
+
+	// Filter rules by version
+	filteredRules := make([]rule.InstRule, 0)
+	for _, r := range relevantRules {
+		if !matchVersion(dep, r) {
+			continue
+		}
+		filteredRules = append(filteredRules, r)
+	}
 
 	// Separate file rules from rules that need precise matching
 	preciseRules := make([]rule.InstRule, 0)
-	for _, r := range rules {
+	for _, r := range filteredRules {
 		// If the rule is a file rule, it is always applicable
 		if fr, ok := r.(*rule.InstFileRule); ok {
 			set.AddFileRule(fr)
-			sp.Info("Match file rule", "rule", fr, "dep", dependency)
+			sp.Info("Match file rule", "rule", fr, "dep", dep)
 			continue
 		}
 		// We can't decide whether the rule is applicable yet, add it to the
@@ -171,7 +186,7 @@ func (sp *SetupPhase) runMatch(dependency *Dependency, rules []rule.InstRule) (*
 	}
 
 	// Precise matching
-	for _, source := range dependency.Sources {
+	for _, source := range dep.Sources {
 		// Parse the source code. Since the only purpose here is to match,
 		// no node updates, we can use fast variant.
 		tree, err := ast.ParseFileFast(source)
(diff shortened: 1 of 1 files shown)


tool/internal/setup/match.go
```diff
@@ -216,17 +232,41 @@ func (sp *SetupPhase) matchDeps(deps []*Dependency) ([]*rule.InstRuleSet, error)
...
-		// TODO: Parallelize this
-		m, err1 := sp.runMatch(dep, allRules)
-		if err1 != nil {
-			return nil, err1
-		}
-		if !m.IsEmpty() {
-			matched = append(matched, m)
-		}
+		g.Go(func() error {
+			relevantRules := rulesByTarget[dep.ImportPath]
```
y1yang0: matchTarget does the same thing. Perhaps we can move some of the code from runMatch (the matchVersion && matchTarget part) here.

Maybe something like
```
func quickMatch() { matchTarget and matchVersion}
// Quick filtering relevant rules
revelantRules := filterRules()
if len(revelantRules)==0{
return
}
// Precious matching
sp.matchRules() 
```

txabman42: Good idea, updated (9a32b1af73907df4f3fb694df0eba4b91773bde7)

Now, it works like this:

```go
func matchDeps {
  rulesByTarget // pre-indexed map by target
  go(func() {
    filteredRules = quickFilter(dep, rulesByTarget) // by target and version
    runMatch(dep, filteredRules) // precious matching
  })
}
```

Removed `rulesByTarget` func, as now pre-indexed map `rulesByTarget` is used directly. I haven't moved to an object with `getByTarget` method, as the code is quite encapsulated under `matchDeps`.

y1yang0: After the refactoring, we could potentially take it a step further by removing quickFilter and moving its logic into runMatch, as it is conceptually part of the matching process.

Consequently, the signature of runMatch would be changed from:

func (sp *SetupPhase) runMatch(dependency *Dependency, rules **[]rule.InstRule**)
=>
func (sp *SetupPhase) runMatch(dependency *Dependency, rules **map[string][]rule.InstRule**)

txabman42: Agree, updated (3a5e5b79e93c4cf4a350da8d213f929700f88755).

That way:
- `matchDeps`  prepares and run runMatch per dependency
- `runMatch` does all the filtering/matching logic

tool/internal/setup/match_test.go
```diff
@@ -163,58 +163,6 @@ func TestMatchVersion(t *testing.T) {
 	}
 }
 
-func TestMatchTarget(t *testing.T) {
```
kakkoyun, line 166: Probably I'm missing some context here. Why are we removing the tests?

tool/internal/setup/match.go
```diff
@@ -217,17 +241,36 @@ func (sp *SetupPhase) matchDeps(deps []*Dependency) ([]*rule.InstRuleSet, error)
...
+	rulesByTarget := make(map[string][]rule.InstRule)
+	for _, r := range allRules {
+		target := r.GetTarget()
+		rulesByTarget[target] = append(rulesByTarget[target], r)
+	}
+
 	// Match the default rules with the found dependencies
 	matched := make([]*rule.InstRuleSet, 0)
+	var mu sync.Mutex
+	g, _ := errgroup.WithContext(ctx)
```
kakkoyun, line 254: I would prefer creating the group in `Setup` and pass it down here (using maybe as a field to `SetupPhase`, instead of creating it for each `matchDeps` call. 

It's a dependency and it is cleaner to initialize it in a single place.

kakkoyun (commented):
Thanks for the PR. I have some minor suggestions.

tool/internal/setup/match_test.go
txabman42: `matchTarget` func has been removed.

I've added a pre-indexed rules by target, so instead of checking the target, we just get the rules by target.

I added that map to avoid iterating all rules per dep.

tool/internal/setup/match.go
txabman42: > instead of creating it for each matchDeps call

mmmm IMO that shouldn't be a good idea.

errgroups are stateful and they are not made for being reused, from their doc:

> A Group is a collection of goroutines working on subtasks that are part of the same overall task. A Group should not be reused for different tasks.

I think the idea of initializing things at Setup factory is good, but in this case, I see errgroups coupled just to matchDeps due to its stateful design. They need to be instantiated per run.

kakkoyun: Then the real question is do we wait for multiple calls to `matchDeps` at the same time?

kakkoyun: By all means, this is not a blocker but it would be nice to clarify.

txabman42: > Then the real question is do we wait for multiple calls to matchDeps at the same time?

Regardless running once or twice, I still consider they are part of `matching`, rather than `setup`. errorgroup are coupled to a specific context. But I would see ok if it was a working pool.

But I agree that having this in the middle of a function could be a bit 🧐

```go
	g, _ := errgroup.WithContext(ctx)
	g.SetLimit(runtime.NumCPU() * matchDepsConcurrencyMultiplier)
```

PR #144, opened 2025-11-10
chore(ci): optimize license header

yuluo-yx:
- For https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/137#issuecomment-3510778046

@kakkoyun pls review, thx

The following points need clarification:

1. Using Makefile for checks without third-party tools is faster and more suitable for local execution by developers.
2. A fix command has been added to automatically add the license header.
3. License headers have been added to YAML files and some Go files that lack them.

### DIFF
--- a/tool/internal/ast/shared.go
+++ b/tool/internal/ast/shared.go
@@ -46,16 +46,19 @@ func FindFuncDeclWithoutRecv(root *dst.File, funcName string) *dst.FuncDecl {
 
 func FindFuncDecl(root *dst.File, funcName string, recv string) *dst.FuncDecl {
 	decls := findFuncDecls(root, func(funcDecl *dst.FuncDecl) bool {
-		// Receiver type is empty? Match func name only
+		// Receiver type is ignored, match func name only
 		name := funcDecl.Name.Name
 		if recv == "" {
 			return name == funcName && !HasReceiver(funcDecl)
 		}
+		// Receiver type is specified, but target function does not have receiver
+		// That's not what we want
 		if !HasReceiver(funcDecl) {
 			return false
 		}
 
-		// Receiver type is specified, Match both func name and receiver type
+		// Receiver type is specified, and target function has receiver
+		// Match both func name and receiver type
 		switch recvTypeExpr := funcDecl.Recv.List[0].Type.(type) {
 		case *dst.StarExpr: // func (*Recv)T
 			tn, ok := recvTypeExpr.X.(*dst.Ident)
@@ -64,10 +67,10 @@ func FindFuncDecl(root *dst.File, funcName string, recv string) *dst.FuncDecl {
 				return false
 			}
 			t := "*" + tn.Name
-			return t == recv
+			return t == recv && name == funcName
 		case *dst.Ident: // func (Recv)T
 			t := recvTypeExpr.Name
-			return t == recv
+			return t == recv && name == funcName
 		case *dst.IndexExpr:
 			// This is a generic type, we don't support it yet
 			return false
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -18,6 +18,19 @@ INST_PKG_TMP = pkg_temp
 API_SYNC_SOURCE = pkg/inst/context.go
 API_SYNC_TARGET = tool/internal/instrument/api.tmpl
 
+# License header check exclusions
+LICENSE_EXCLUDE_PATHS := -path '**/vendor/*' \
```
kakkoyun: Let's exclude .github/workflows and the configuration files. We don't actually need license headers for configuration files.

tool/internal/instrument/instrument_test.go
```diff
@@ -1,3 +1,6 @@
+// Copyright The OpenTelemetry Authors
```
kakkoyun, line 1: 👍

Makefile
```diff
@@ -129,10 +137,62 @@ lint/makefile: checkmake
...
 
-lint/license: ## Check license headers
-lint/license: go-license
+# License header targets
+
+.PHONY: license-header-check
+license-header-check: ## Check license headers in source files
 	@echo "Checking license headers..."
-	go-license --config=license.yml --verify .
+	@licRes=$$(for pattern in $(LICENSE_FILE_PATTERNS); do \
```
kakkoyun: I know I'm the first on to introduce complete makefile actions but this is getting out of hand.

Let's create a `scripts` directory and add a `check-license-header.sh` in there with an additional `--fix` flag.

And add modify to make actions to be simpler.

Makefile
```diff
@@ -99,15 +112,10 @@ format/yaml: yamlfmt
...
-format/license: ## Apply license headers to Go files
-format/license: go-license
-	@echo "Applying license headers..."
-	go-license --config=license.yml .
-
 # Lint targets
 
 lint: ## Run all linters (Go, YAML, GitHub Actions, Makefile)
-lint: lint/go lint/yaml lint/action lint/makefile
+lint: lint/go lint/yaml lint/action lint/makefile license-header-check
```
```suggestion
lint: lint/go lint/yaml lint/action lint/makefile lint/license-header
```

Makefile
```diff
@@ -129,10 +137,62 @@ lint/makefile: checkmake
 	@echo "Linting Makefile..."
 	checkmake --config .checkmake Makefile
 
-lint/license: ## Check license headers
-lint/license: go-license
+# License header targets
+
+.PHONY: license-header-check
+license-header-check: ## Check license headers in source files
```
```suggestion
lint/license-header: ## Check license headers in source files
```

Makefile
```diff
@@ -129,10 +137,62 @@ lint/makefile: checkmake
...
+	    fi; \
+	done); \
+	if [ -n "$${licRes}" ]; then \
+	    echo "Missing license header in the following files:"; echo "$${licRes}"; \
+	    exit 1; \
+	fi
+	@echo "All files have proper license headers"
+
+.PHONY: license-header-fix
+license-header-fix: ## Add missing license headers to source files
```
kakkoyun: lint/license-header/fix: ## Add missing license headers to source file

Makefile
yuluo-yx: I want to complete this together with https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/145.

WDYT 👀

Makefile
kakkoyun: We can actually stick with the existing `lint/license`

Makefile
kakkoyun: ditto `format/license`

yuluo-yx: done

Makefile
yuluo-yx: done

Makefile
kakkoyun: I think that can wait. However, moving these to a script file already would be an improvement because we can benefit from some additional shell formatters and linters. And also we don't need to make sure the `\` correctly put.

Makefile
kakkoyun: hmm, I still see the license header is being added the those files. We should also remove those changes.

yuluo-yx: Sorry, please give me a moment while I submit the code. 😬

Makefile
yuluo-yx: got

yuluo-yx: done, ptal

Issue #65, opened 2025-09-23
test: Add snapshot testing facilities


kakkoyun:
Snapshot testing (golden files) helps immensely for testing expected output of AST manipulation and help to catch any regressions. Let's have snapshot testing as part of our integration testing strategy.

xref: https://pkg.go.dev/gotest.tools/v3@v3.5.2/golden

PR #146, opened 2025-11-10
chore(tests): Improve instrument integration tests

txabman42:
I was reading instrumentation code and I took the opportunity to add some integration tests for this phase.

Related to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/119#issuecomment-3483461223 and I also took some inspiration about golden files from https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/65.

Just a first attempt to cover some of the most basic cases. Having also the resulting testing files helped a lot to understand the results from the integration phase.

Not sure if you have in mind some testing like this for instrumentation phase, feel free to discard it if not.

btw, right now they run as unit tests, although they are not. The integration tests are defined in `./test/integration/...`, so I had some doubts about letting these as unit tests.

### DIFF
--- a/tool/internal/instrument/instrument_test.go
+++ b/tool/internal/instrument/instrument_test.go
@@ -46,7 +46,8 @@ const (
 )
 
 func TestInstrumentation_Integration(t *testing.T) {
-	entries, _ := os.ReadDir(filepath.Join(testdataDir, goldenDir))
+	entries, err := os.ReadDir(filepath.Join(testdataDir, goldenDir))
+	require.NoError(t, err)
 	for _, entry := range entries {
 		if !entry.IsDir() {
 			continue
@@ -68,7 +69,7 @@ func runTest(t *testing.T, testName string) {
 	sourceFile := filepath.Join(tempDir, mainGoFileName)
 	util.CopyFile(filepath.Join(testdataDir, sourceFileName), sourceFile)
 
-	ruleSet := loadRulesYAML(testName, sourceFile)
+	ruleSet := loadRulesYAML(t, testName, sourceFile)
 	writeMatchedJSON(ruleSet)
 
 	args := compileArgs(tempDir, sourceFile)
@@ -84,8 +85,9 @@ func runTest(t *testing.T, testName string) {
 	verifyGoldenFiles(t, tempDir, testName)
 }
 
-func loadRulesYAML(testName, sourceFile string) *rule.InstRuleSet {
-	data, _ := os.ReadFile(filepath.Join(testdataDir, goldenDir, testName, rulesFileName))
+func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet {
+	data, err := os.ReadFile(filepath.Join(testdataDir, goldenDir, testName, rulesFileName))
+	require.NoError(t, err)
 
 	var rawRules map[string]map[string]any
 	yaml.Unmarshal(data, &rawRules)


pdelewski:
I think that having unit tests for instrumentation phase is useful. Not sure yet if that's the right way to go, however developer experience right now is not perfect. What I would like to have is ability to reproduce artificial use case as unit test.

kakkoyun (commented):
This is awesome! I was meaning to add golden files and snapshot testing as part integration tests! (as I described in the issue).

I find this super useful and it helps us to iterate really fast! Asserting against the expected source code modification is way to go. We should have more of this.

I haven't reviewed it in-depth yet, I just wanted to comment to share my enthusiasm.

txabman42:
Agree that they should be marked and run as integration tests. Tests are even marked as integration (e.g. `TestInstrumentWithDifferentRuleTypes_Integration`).

With this:

> btw, right now they run as unit tests, although they are not. The integration tests are defined in ./test/integration/..., so I had some doubts about letting these as unit tests.

I wanted to show my doubts about where those tests should belong to. IMO, although they are integration tests, these tests shouldn't be put under `./test/integration/...` as they are just targeting instrumentation phase. `./test/integration/...` should target some app scope integration tests, although they end up usually being e2e. Similar to the project that @kakkoyun shared (https://github.com/DataDog/orchestrion/tree/main/internal/injector/testdata/injector) where they only have `/test/e2e`.

I usually like to start simple and keep this kind of tests under unit-test execution until the feedback loop is not fast enough. Just to keep things simple at the beginning. However, I'm ok with adding them under integration tests.

I see 3 main options:
- Put them under `./test/integration/...`
- Keep them under `/tool/internal/instrument` and run them as test-unit for now
- Keep them under `/tool/internal/instrument` and try to make them run as test-integration

I can adapt to any decision, as I don't have too much context about some design decisions.

btw, I'll also try to refactor tests to have rule yaml definition and golden file under the same folder in `testdata`, just like https://github.com/DataDog/orchestrion/tree/main/internal/injector/testdata/injector. I like that approach, much cleaner and visual than creating the rules directly in the tests.

kakkoyun:
> Similar to the project that @kakkoyun shared ([DataDog/orchestrion@`main`/internal/injector/testdata/injector](https://github.com/DataDog/orchestrion/tree/main/internal/injector/testdata/injector?rgh-link-date=2025-11-12T08%3A22%3A48Z)) where they only have `/test/e2e`

We actually have both of them. We call them with different names, and we run the snapshot tests as part of unit tests. They are quite fast.

> I usually like to start simple and keep this kind of tests under unit-test execution until the feedback loop is not fast enough.

👍 💯 

> * Put them under `./test/integration/...`

My vote is on this one. It is straight forward, and easier to discover.



> btw, I'll also try to refactor tests to have rule yaml definition and golden file under the same folder in `testdata`, just like [DataDog/orchestrion@`main`/internal/injector/testdata/injector](https://github.com/DataDog/orchestrion/tree/main/internal/injector/testdata/injector?rgh-link-date=2025-11-12T08%3A22%3A48Z). I like that approach, much cleaner and visual than creating the rules directly in the tests.

Yes, let's try to have a declarative approach as we have in Orchestrion, and have a single directory for each scenario. It is quite scalable and easy to read. It makes development a joy.

y1yang0:
> From that path we don't have access to `internal` pkgs. Where should we put all `rule.yaml` for testing different scenarios? AFAIU, it's not possible to provide external rules. Should that feature be added first? 🤔

How about we put the test rule rule.yaml into tool/data for now, and then add the feature to load external rules later?

txabman42:
I though about that idea, but it would complicate all testing. Right now, a test just add a rule and verify the output.
If we move to tool/data without allowing loading external rules, we would need to load all testing rules at the same time. This would involve generating different apps for each case, which in reality should be the opposite, keep a simple app and test different approaches against it.

tool/internal/instrument/instrument_test.go
```diff
@@ -3,236 +3,172 @@
...
-			err := Toolexec(ctx, args)
-			if tt.wantErr {
-				require.Error(t, err)
-			} else {
-				require.NoError(t, err)
-			}
-			// TODO: Link the instrumented binary and run it and further check
-			// output content
+func TestInstrumentation_Integration(t *testing.T) {
+	entries, _ := os.ReadDir(filepath.Join(testdataDir, goldenDir))
```
y1yang0: explicit err and require.NoErrr

tool/internal/instrument/instrument_test.go
```diff
@@ -3,236 +3,172 @@
...
-	matchedFile := filepath.Join(tempDir, util.BuildTempDir, matchedJSONFile)
-	err = os.MkdirAll(filepath.Dir(matchedFile), 0o755)
-	require.NoError(t, err)
-	err = util.WriteFile(matchedFile, string(matchedJSON))
 	require.NoError(t, err)
+	verifyGoldenFiles(t, tempDir, testName)
+}
+
+func loadRulesYAML(testName, sourceFile string) *rule.InstRuleSet {
+	data, _ := os.ReadFile(filepath.Join(testdataDir, goldenDir, testName, rulesFileName))
```
y1yang0: Ditto

y1yang0 (approved):
Otherwise looks good

tool/internal/instrument/instrument_test.go
txabman42: Agree, added in 44fbcc94671e6c6e046779984f5453daad7760e7

tool/internal/instrument/instrument_test.go
txabman42: Agree, added in 44fbcc94671e6c6e046779984f5453daad7760e7

PR #151, opened 2025-11-12
chore: simplify command actions

y1yang0:
simplify command actions

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,36 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+### Describe the bug
+
+A clear and concise description of what the bug is.
+
+### Environment
+
+ - OS: [e.g. iOS, Windows]
+ - Go Version: [e.g. 1.6]
+ - Version: [e.g. 1295520, v0.1.0]
+
+### To Reproduce
+
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+### Expected behavior
+
+A clear and concise description of what you expected to happen.
+
+### Additional context
+
+Add any other context about the problem here.
+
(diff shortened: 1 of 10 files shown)


tool/internal/setup/setup.go
```diff
@@ -120,3 +120,18 @@ func BuildWithToolexec(ctx context.Context, args []string) error {
 
 	return util.RunCmdWithEnv(ctx, env, newArgs...)
 }
+
+func GoBuild(ctx context.Context, args []string) error {
+	backupFiles := []string{"go.mod", "go.sum", "go.work", "go.work.sum"}
+	_ = util.BackupFile(backupFiles)
+	defer func() {
+		_ = os.RemoveAll(otelRuntimeFile)
```
kakkoyun: nit: Having debug logs for these errors won't hurt. It would make easier to troubleshoot if there is something wrong with the development environment.

tool/cmd/cmd_version.go
```diff
@@ -27,32 +27,32 @@ var commandVersion = cli.Command{
 	Action: func(_ context.Context, cmd *cli.Command) error {
 		_, err := fmt.Fprintf(cmd.Writer, "otel version %s", Version)
 		if err != nil {
-			return ex.Wrapf(err, "failed to print version with exit code %d", exitCodeFailure)
+			return ex.Wrap(err)
```
kakkoyun: Why do we remove the additional context?

tool/internal/setup/setup.go
y1yang0: The current logic attempts to back up and restore a fixed set of files: go.mod, go.sum, go.work, and go.work.sum. An error is returned if any of these operations fail.

However, since the vast majority of Go applications do not use go.work and go.work.sum, the backup and restore operations for these two files will almost always fail with a "no such file or directory" error.

As a result, the debug.log is consistently populated with two error messages like the ones below. This is confusing and concerning for new contributors, who are left wondering what actually failed and whether it's a critical issue.

msg="failed to back up... error=": open go.work: no such file or directory..."
msg="failed to restore... error=": open .../backup/go.work: no such file or directory..."

kakkoyun: It sounds like it is matter of adding more context to the message and changing the log level to debug.

What do you think?

tool/cmd/cmd_version.go
y1yang0: My intention was to remove exitCodeFailure, not the entire msg. Let me restore it.

kakkoyun (approved):
LGTM :shipit: 

Thank you so much!

PR #156, opened 2025-11-12
feat(instrumentation): Add HTTP instrumentation

kakkoyun:
Adds HTTP instrumentation.

## Key changes

- a8992f609a5228f8a73ea36143a1009366b71eba adds the `net/http` client and server implementation. It mimics the upstream `opentelemetry-go-contrib`
- 6b4f80407e536ad625ac8c57607463ff5b6f596c Removes the existing abstractions around semantic conventions and instrumenter for a leaner approach with more flexibility. 

**Screenshots (from `demo/insfrastructure/docker-compose`:**
Screenshot 2025-11-25 at 16 13 44

Screenshot 2025-11-25 at 16 13 57

### DIFF
--- a/demo/infrastructure/docker-compose/Makefile
+++ b/demo/infrastructure/docker-compose/Makefile
@@ -1,7 +1,9 @@
 # Makefile for OpenTelemetry Demo Infrastructure
 # Provides convenient commands for managing the observability stack
 
-.PHONY: help start stop restart reload status logs clean build load-test urls health
+.PHONY: help start stop restart reload status logs clean build load-test urls health \
+        start-http stop-http build-http build-http-nocache clean-docker-cache \
+        verify-traces test-propagation debug-http rebuild-http package
 
 # Default target - show help
 .DEFAULT_GOAL := help
@@ -13,18 +15,39 @@ YELLOW := \033[0;33m
 RED := \033[0;31m
 NC := \033[0m # No Color
 
+# HTTP-only services (excludes gRPC which has known build issues)
+HTTP_SERVICES := prometheus jaeger otel-collector grafana http-server http-client
+
+# Path to main project Makefile
+PROJECT_ROOT := ../../..
+
 help: ## Show this help message
 	@echo "$(BLUE)OpenTelemetry Demo Infrastructure$(NC)"
 	@echo "$(BLUE)====================================$(NC)"
 	@echo ""
 	@echo "Available targets:"
-	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "  $(GREEN)%-15s$(NC) %s\n", $$1, $$2}'
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "  $(GREEN)%-20s$(NC) %s\n", $$1, $$2}'
 	@echo ""
-	@echo "Examples:"
-	@echo "  $(YELLOW)make start$(NC)          - Start all services"
-	@echo "  $(YELLOW)make logs$(NC)           - Follow all logs"
-	@echo "  $(YELLOW)make reload$(NC)         - Reload configurations"
-	@echo "  $(YELLOW)make load-test$(NC)      - Run load tests"
+	@echo "$(BLUE)Quick Start Examples:$(NC)"
+	@echo "  $(YELLOW)make quickstart-http$(NC)   - Start HTTP demo (recommended)"
+	@echo "  $(YELLOW)make dev-http$(NC)          - Development mode with logs"
+	@echo "  $(YELLOW)make rebuild-http$(NC)      - Force full rebuild"
+	@echo ""
+	@echo "$(BLUE)Development:$(NC)"
+	@echo "  $(YELLOW)make package$(NC)           - Repackage instrumentation code"
+	@echo "  $(YELLOW)make build-http$(NC)        - Build HTTP apps (auto-packages)"
+	@echo "  $(YELLOW)make rebuild-http$(NC)      - Full rebuild with cache purge"
+	@echo ""
+	@echo "$(BLUE)Debugging:$(NC)"
+	@echo "  $(YELLOW)make verify-traces$(NC)     - Check traces in Jaeger"
+	@echo "  $(YELLOW)make test-propagation$(NC)  - Test context propagation"
+	@echo "  $(YELLOW)make debug-http$(NC)        - Start with debug logging"
+	@echo ""
+	@echo "$(BLUE)Note:$(NC) All build targets auto-run 'make package' to include latest code changes."
+	@echo ""
+	@echo "$(BLUE)Documentation:$(NC)"
+	@echo "  See CONTEXT_PROPAGATION.md for trace linking documentation"
+	@echo "  See VERIFICATION.md for manual verification steps"
 
 start: ## Start all observability services
 	@echo "$(BLUE)Starting observability stack...$(NC)"
@@ -77,6 +100,9 @@ logs-grafana: ## Follow Grafana logs
 logs-grpc: ## Follow gRPC server and client logs
 	@docker-compose logs -f grpc-server grpc-client
 
+logs-http: ## Follow HTTP server and client logs
+	@docker-compose logs -f http-server http-client
+
 health: ## Check health of all services
 	@echo "$(BLUE)Health Check:$(NC)"
 	@echo ""
@@ -105,12 +131,17 @@ urls: ## Show URLs for all UIs
 	@echo "  $(GREEN)Services: APM$(NC)             http://localhost:3000/d/opentelemetry-apm"
 	@echo ""
 
-build: ## Rebuild demo applications
+package: ## Package instrumentation code (run after changing pkg/ files)
+	@echo "$(BLUE)Packaging instrumentation code...$(NC)"
+	@cd $(PROJECT_ROOT) && make package
+	@echo "$(GREEN)✓ Package complete$(NC)"
+
+build: package ## Rebuild demo applications (auto-packages instrumentation)
 	@echo "$(BLUE)Building demo applications...$(NC)"
 	@docker-compose build
 	@echo "$(GREEN)✓ Build complete$(NC)"
 
-build-nocache: ## Rebuild demo applications without cache
+build-nocache: package ## Rebuild demo applications without cache (auto-packages)
 	@echo "$(BLUE)Building demo applications (no cache)...$(NC)"
 	@docker-compose build --no-cache
 	@echo "$(GREEN)✓ Build complete$(NC)"
@@ -243,3 +274,111 @@ quickstart: prereqs ports start health urls ## Complete setup: check prereqs, po
 dev: start logs ## Start services and follow logs (development mode)
 
 dev-rebuild: build-nocache start logs ## Rebuild without cache, start, and follow logs
+
+# =============================================================================
+# HTTP-Only Targets (for when gRPC demo has issues)
+# =============================================================================
+
+start-http: ## Start HTTP demo only (no gRPC)
+	@echo "$(BLUE)Starting HTTP demo stack...$(NC)"
+	@docker-compose up -d $(HTTP_SERVICES)
+	@echo "$(GREEN)✓ HTTP services started$(NC)"
+	@echo ""
+	@$(MAKE) urls
+
+stop-http: ## Stop HTTP demo services
+	@echo "$(BLUE)Stopping HTTP services...$(NC)"
+	@docker-compose stop http-server http-client
+	@echo "$(GREEN)✓ HTTP services stopped$(NC)"
+
+build-http: package ## Build only HTTP demo applications (auto-packages)
+	@echo "$(BLUE)Building HTTP demo applications...$(NC)"
+	@docker-compose build http-server http-client
+	@echo "$(GREEN)✓ HTTP build complete$(NC)"
+
+build-http-nocache: package ## Build HTTP apps without Docker cache (auto-packages)
+	@echo "$(BLUE)Building HTTP applications (no cache)...$(NC)"
+	@docker-compose build --no-cache http-server http-client
+	@echo "$(GREEN)✓ HTTP build complete$(NC)"
+
+rebuild-http: package clean-docker-cache ## Full rebuild: package, purge cache, rebuild HTTP apps
+	@echo "$(BLUE)Full HTTP rebuild...$(NC)"
+	@docker-compose build --no-cache http-server http-client
+	@echo "$(GREEN)✓ Full HTTP rebuild complete$(NC)"
+
+# =============================================================================
+# Docker Cache Management
+# =============================================================================
+
+clean-docker-cache: ## Purge Docker builder cache
+	@echo "$(BLUE)Purging Docker builder cache...$(NC)"
+	@docker builder prune -af
+	@echo "$(GREEN)✓ Docker cache purged$(NC)"
+
+clean-images: ## Remove demo Docker images
+	@echo "$(BLUE)Removing demo images...$(NC)"
+	@docker rmi docker-compose-http-server docker-compose-http-client 2>/dev/null || true
+	@docker rmi docker-compose-grpc-server docker-compose-grpc-client 2>/dev/null || true
+	@echo "$(GREEN)✓ Demo images removed$(NC)"
+
+clean-full: clean clean-images clean-docker-cache ## Full cleanup: containers, images, and cache
+	@echo "$(GREEN)✓ Full cleanup complete$(NC)"
+
+# =============================================================================
+# Trace Verification and Debugging
+# =============================================================================
+
+verify-traces: ## Verify traces are being collected in Jaeger
+	@echo "$(BLUE)Verifying trace collection...$(NC)"
+	@echo ""
+	@echo "$(YELLOW)HTTP Server traces:$(NC)"
+	@curl -s 'http://localhost:16686/api/traces?service=http-server&limit=3' 2>/dev/null | \
+		jq -r '.data[] | "  TraceID: \(.traceID) | Spans: \(.spans | length) | Op: \(.spans[0].operationName)"' 2>/dev/null || \
+		echo "$(RED)  ✗ Could not fetch traces (is Jaeger running?)$(NC)"
+	@echo ""
+	@echo "$(YELLOW)HTTP Client traces:$(NC)"
+	@curl -s 'http://localhost:16686/api/traces?service=http-client&limit=3' 2>/dev/null | \
+		jq -r '.data[] | "  TraceID: \(.traceID) | Spans: \(.spans | length) | Op: \(.spans[0].operationName)"' 2>/dev/null || \
+		echo "$(RED)  ✗ Could not fetch traces$(NC)"
+	@echo ""
+
+test-propagation: ## Test context propagation with a specific trace ID
+	@echo "$(BLUE)Testing trace context propagation...$(NC)"
+	@TRACE_ID=$$(openssl rand -hex 16); \
+	SPAN_ID=$$(openssl rand -hex 8); \
+	echo "$(YELLOW)Sending request with:$(NC)"; \
+	echo "  Trace ID: $$TRACE_ID"; \
+	echo "  Span ID:  $$SPAN_ID"; \
+	curl -s -H "Traceparent: 00-$$TRACE_ID-$$SPAN_ID-01" 'http://localhost:8080/greet?name=propagation-test' > /dev/null; \
+	sleep 3; \
+	echo ""; \
+	echo "$(YELLOW)Checking Jaeger for trace:$(NC)"; \
+	RESULT=$$(curl -s "http://localhost:16686/api/traces/$$TRACE_ID" 2>/dev/null | jq -r '.data[0].traceID // "NOT_FOUND"'); \
+	if [ "$$RESULT" = "$$TRACE_ID" ]; then \
+		echo "  $(GREEN)✓ Context propagation WORKING - trace found in Jaeger$(NC)"; \
+		curl -s "http://localhost:16686/api/traces/$$TRACE_ID" | jq -r '.data[0] | "  Services: \([.processes | to_entries[] | .value.serviceName] | unique | join(", "))\n  Span count: \(.spans | length)"'; \
+	else \
+		echo "  $(RED)✗ Context propagation NOT WORKING - trace not found$(NC)"; \
+		echo "  $(YELLOW)Note: Server created its own trace instead of using the provided one$(NC)"; \
+	fi
+
+debug-http: ## Start HTTP services with debug logging
+	@echo "$(BLUE)Starting HTTP services with DEBUG logging...$(NC)"
+	@docker-compose stop http-server http-client 2>/dev/null || true
+	@OTEL_LOG_LEVEL=debug docker-compose up -d http-server http-client
+	@echo "$(GREEN)✓ HTTP services started with debug logging$(NC)"
+	@echo "$(YELLOW)View logs with: make logs-http$(NC)"
+
+# =============================================================================
+# Quick Development Workflow
+# =============================================================================
+
+dev-http: start-http logs-http ## Start HTTP demo and follow logs
+
+dev-http-rebuild: rebuild-http start-http logs-http ## Full rebuild and start HTTP demo with logs
+
+quickstart-http: prereqs ports start-http health urls ## HTTP-only quickstart
+	@echo ""
+	@echo "$(GREEN)✓ HTTP observability stack is ready!$(NC)"
+	@echo "$(YELLOW)Note: gRPC demo excluded due to known build issues$(NC)"
+	@echo ""
--- a/demo/infrastructure/docker-compose/docker-compose.yml
+++ b/demo/infrastructure/docker-compose/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3.8'
-
 # OpenTelemetry Demo Infrastructure
 # Complete observability stack with Jaeger, Prometheus, OTel Collector, Grafana, and k6
 


pkg/instrumentation/nethttp/server/server_hook_test.go
```diff
@@ -0,0 +1,576 @@
...
+		{
+			name: "request with trace context propagation",
+			setupEnv: func() {
+				t.Setenv("OTEL_GO_AUTO_INSTRUMENTATION_ENABLED", "true")
+				t.Setenv("OTEL_GO_AUTO_INSTRUMENTATION_NETHTTP_ENABLED", "true")
+			},
+			setupRequest: func() *http.Request {
+				req := httptest.NewRequest("GET", "http://example.com/path", nil)
+				// Add traceparent header to simulate incoming trace
+				req.Header.Set("traceparent", "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01")
```
pdelewski: That should be enough:
```suggestion
				req.Header.Set("traceparent", "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0bb902b7-01")
```

pkg/instrumentation/nethttp/client/propagation_test.go
```diff
@@ -0,0 +1,237 @@
...
+	tracer := provider.Tracer("test")
+
+	// Create test server
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer ts.Close()
+
+	// Create parent span
+	ctx, parentSpan := tracer.Start(context.Background(), "parent_operation")
```
pdelewski: Extract parentSpanName as const

pkg/instrumentation/nethttp/client/propagation_test.go
```diff
@@ -0,0 +1,237 @@
...
+
+	// Create child span (simulating what instrumentation does)
+	ctx, childSpan := tracer.Start(ctx, "http_client_request")
```
pdelewski: Same as above

pkg/instrumentation/nethttp/client/propagation_test.go
```diff
@@ -0,0 +1,237 @@
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		receivedTraceparent = r.Header.Get("Traceparent")
+		// Baggage header can be checked if needed
+		_ = r.Header.Get("Baggage")
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer ts.Close()
+
+	// Create context with trace and baggage
+	traceID := trace.TraceID{
```
pdelewski: Seems that these values are duplicated

pkg/instrumentation/nethttp/client/propagation_test.go
```diff
@@ -0,0 +1,237 @@
...
+		0x09,
+		0x0a,
+		0x0b,
+		0x0c,
+		0x0d,
+		0x0e,
+		0x0f,
+		0x10,
+	}
+	spanID := trace.SpanID{0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08}
```
pdelewski: Same here

pdelewski:
> * [6b4f804](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/commit/6b4f80407e536ad625ac8c57607463ff5b6f596c) Removes the existing abstractions around semantic conventions and instrumenter for a leaner approach with more flexibility.

I like that

tool/internal/instrument/instrument.go
```diff
@@ -28,6 +31,47 @@ func groupRules(rset *rule.InstRuleSet) map[string][]rule.InstRule {
 	return file2rules
 }
 
+// findActualFile finds the actual file in compile arguments that matches the rule file.
+// This handles cases where file paths differ between setup and build phases
+// (e.g., different module versions in the cache).
+func (ip *InstrumentPhase) findActualFile(ruleFile string) (string, error) {
```
y1yang0, line 37: If the file versions used in the setup and instrument phases are inconsistent, it should actually be considered an error.

This is because the setup phase's matching process precisely locates a specific function in a specific version. If the instrument phase subsequently proceeds with a different version, it indicates that something went wrong in a process after the initial match was completed.

I will fix this.

pkg/instrumentation/grpc/go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/grpc
 
-go 1.23
+go 1.23.0
 
 replace github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg => ../..
+
+require github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg v0.0.0-00010101000000-000000000000
```
darccio, line 7: Is this required? I'm kinda surprised that it wasn't before.

pkg/instrumentation/helloworld/go.mod
```diff
@@ -5,12 +5,13 @@ go 1.23.0
 replace github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg => ../..
 
 require (
-	github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg v0.0.0-20251124021638-b2f0cf1e96a8
+	github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg v0.0.0-20251014023539-a5ba44c6591d
```
darccio: Why the downgrade?

tool/internal/instrument/instrument.go
kakkoyun: Thanks for that. I wasn't quite sure if this is the correct approach.
I can rebase the branch when we merged your PR.

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
123liuziming, line 1: I agree we can reuse some of the instrumentation points from `https://github.com/open-telemetry/opentelemetry-go-contrib`. However, I believe the actual instrumentation logic requires further encapsulation, much like what is done in `https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/spring/spring-kafka-2.7/library/src/main/java/io/opentelemetry/instrumentation/spring/kafka/v2_7/SpringKafkaTelemetry.java` . Such encapsulation proves highly beneficial for subsequent plugin development, reducing redundant code logic while enforcing strict adherence to OpenTelemetry Semantic Conventions for each plugin. The plugins supported in `https://github.com/open-telemetry/opentelemetry-go-contrib` remain limited; we shall need to support numerous additional plugins later, making relevant abstraction essential.

123liuziming, line 1: In my view, the instrumentation in `https://github.com/open-telemetry/opentelemetry-go-contrib` should be refactored using our encapsulation approach before being integrated into our trampoline.

kakkoyun: There are several things with that approach that doesn't sit right with me. 

First of all, it is really premature do any kind of abstraction. The best abstractions emerge naturally from the code itself. I don't believe introducing such abstractions provide any benefits.

And as we discussed in the Slack channel, we SHOULD not in the business of instrumentation. We will maintain certain instrumentations and maybe we will maintain more of them at the beginning but I envision these integrations will be provided by the community. That's why we should be the enablers of those libraries, such as openteletry-go-contrib.

Another argument is that by introducing these, we are cornering the integration authors. We don't have to do this. We don't have standardize anything around this.

The last argument, I have is the code you shared is a Java code. And replicating the patterns from another ecosystem is just cargo-culting, and it's not idiomatic to Go. It makes harder to understand what is going on. We should keep the code more idiomatic, more approachable to Go developers. It is better to keep it leaner. It took some time for me to understand what those packages do. And the initial version of this PR was based on those abstractions and the code was too convoluted. Generics are not ubiquitous to be used in this fashion in Go. My proposed patterns are more familiar for the Go developers and more familiar to the people who works on OpenTelemetry.

test/integration/http_client_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/test/app"
+)
+
+func TestHTTPClientIntegration(t *testing.T) {
+	clientDir := filepath.Join("..", "..", "demo", "http", "client")
```
txabman42, line 18: Couple of comments regarding integration tests (out of this PR scope):
- I prefer to have self-contained tests. Specially thinking on an scenario supporting multiple libraries. We can make the app service to accept a func defined in directly in the tests (instantiating the http client, etc.).
- At the beginning, I was a bit confused between the scope of these e2e and integration tests. AFAIU, these integration tests would prepare the scenario, build test/app, start test/app and verify. WDYT about making this a bit more explicit? Maybe adding a test fixture. This would also help adding some common steps like verifying the golden files (once migrated to this repo)

test/e2e/http_test.go
```diff
@@ -74,6 +74,14 @@ func TestHttp(t *testing.T) {
...
+	// Verify that server instrumentation was initialized
+	// Note: Client instrumentation is lazily initialized on first use,
+	// so it won't appear in the server output unless the server makes HTTP client requests
+	require.Contains(t, output, "HTTP server instrumentation initialized")
+
+	// Verify that the server ran successfully
+	require.Contains(t, output, "server started")
+
+	// Note: Full trace validation would require parsing OTel spans
+	// For now, we verify that instrumentation hooks are in place and executing
```
txabman42, line 83: e2e should verify those spans, now that we support http and aligned with otel semconv. Worth an issue?

Makefile
```diff
@@ -207,6 +207,9 @@ test-unit: package gotestfmt
 	@echo "Running unit tests..."
 	set -euo pipefail
 	go test -json -v -shuffle=on -timeout=5m -count=1 ./tool/... 2>&1 | tee ./gotest-unit.log | gotestfmt
+	@echo "Running pkg unit tests (semconv only - hook tests require full instrumentation)..."
+	set -euo pipefail
+	cd pkg/instrumentation/nethttp/semconv && go test -v -shuffle=on -timeout=5m -count=1 ./... 2>&1 | tee ../../../gotest-unit-pkg.log
```
txabman42:
```suggestion
	cd pkg/instrumentation/nethttp/semconv && go test -json -v -shuffle=on -timeout=5m -count=1 ./... 2>&1 | tee ../../../gotest-unit-pkg.log | gotestfmt
```

txabman42: Same for other tests

pkg/instrumentation/shared/otel_setup.go
```diff
@@ -0,0 +1,62 @@
...
+			InstrumentationName:    "github.com/open-telemetry/opentelemetry-go-compile-instrumentation",
+			InstrumentationVersion: "0.1.0",
+		})
+	})
+	return nil
+}
+
+// IsInstrumentationEnabled checks if instrumentation is enabled via environment variable
+// Default is enabled unless explicitly set to "false"
+func IsInstrumentationEnabled(instrumentationName string) bool {
```
darccio:
```suggestion
func Instrumented(instrumentationName string) bool {
```

Shorter and more clear, I think.

darccio (changes requested):

pkg/instrumentation/nethttp/server/server_hook.go
```diff
@@ -0,0 +1,149 @@
...
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/nethttp/semconv"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/shared"
+)
+
+const (
+	instrumentationName    = "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/nethttp"
+	instrumentationVersion = "0.1.0"
```
txabman42, line 23: Should be aligned with the otel_setup.go instrumentationVersion?

pkg/otelsetup/setup.go
```diff
@@ -0,0 +1,295 @@
...
+// Config holds configuration for OpenTelemetry setup
+type Config struct {
+	ServiceName            string
+	ServiceVersion         string
+	InstrumentationName    string
+	InstrumentationVersion string
+}
+
+// Initialize sets up OpenTelemetry with defensive error handling
+// This function is safe to call multiple times - it will only initialize once
```
txabman42, line 48: Is it? SetupOTelSDK is, but this one is not ensuring sync once. AFAIU, this would initialize everything again. Could have sense to keep it as private func in shared folder?

pkg/instrumentation/otelsetup/setup.go
```diff
@@ -0,0 +1,295 @@
...
+	if endpoint == "" {
+		logger.Debug("no OTLP endpoint configured, skipping meter provider setup")
+		return nil
+	}
+
+	// Strip http:// or https:// prefix for gRPC (gRPC expects host:port only)
+	grpcEndpoint := stripScheme(endpoint)
+
+	// Create OTLP metric exporter
+	metricExporter, err := otlpmetricgrpc.New(ctx,
```
txabman42: Why couple it to GRPC? WDYT about using [autoexport.NewMetricReader](https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport#NewMetricReader)? it would decide to use http or grpc one based on OTEL_EXPORTER_OTLP_PROTOCOL.

pkg/instrumentation/otelsetup/setup.go
```diff
@@ -0,0 +1,295 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	// Create trace provider with batch span processor
+	tracerProvider = sdktrace.NewTracerProvider(
+		sdktrace.WithResource(res),
+		sdktrace.WithBatcher(traceExporter,
+			sdktrace.WithBatchTimeout(5*time.Second),
```
txabman42: I'd move these values to const. Same for default serviceName, etc.

pkg/instrumentation/nethttp/README.md
```diff
@@ -0,0 +1,462 @@
...
+
+### Planned Features
+
+- 🔄 **Filter Support**: Skip instrumentation for specific paths/endpoints
+- 🔄 **Custom Span Names**: Configurable span name formatting
+- 🔄 **Enhanced Metrics**: Request/response body sizes, connection pool stats
+- 🔄 **HTTP/2 & HTTP/3**: Protocol-specific attributes
+- 🔄 **Public Endpoint Detection**: Differentiate internal vs external traffic
+
+See [nethttp-hybrid-design.md](../../../docs/nethttp-hybrid-design.md) for detailed roadmap.
```
txabman42: This file doesn't exist

pkg/instrumentation/nethttp/README.md
```diff
@@ -0,0 +1,462 @@
+# net/http Compile-Time Instrumentation
```
txabman42, line 1: This doc is ❤️ 

Could be simplified a bit just routing to this repo and otel docs? I'm thinking on trying to keep a consistency between the readmes of different instrumentation packages. The idea is to generate a readme like this per pkg?

pkg/inst-api-semconv/instrumenter/http/http_attrs_extractor.go
pdelewski: Let me add my 2 cents on that
> There are several things with that approach that doesn't sit right with me.
> 
> First of all, it is really premature do any kind of abstraction. The best abstractions emerge naturally from the code itself. I don't believe introducing such abstractions provide any benefits.

Totally agree with that. I’d prefer to start with something really small that can grow organically as our needs evolve, and avoid over-engineering from the beginning.
 
> And as we discussed in the Slack channel, we SHOULD not in the business of instrumentation. We will maintain certain instrumentations and maybe we will maintain more of them at the beginning but I envision these integrations will be provided by the community. That's why we should be the enablers of those libraries, such as openteletry-go-contrib.
> 
> Another argument is that by introducing these, we are cornering the integration authors. We don't have to do this. We don't have standardize anything around this.
> 
> The last argument, I have is the code you shared is a Java code. And replicating the patterns from another ecosystem is just cargo-culting, and it's not idiomatic to Go. It makes harder to understand what is going on. We should keep the code more idiomatic, more approachable to Go developers. It is better to keep it leaner. It took some time for me to understand what those packages do. And the initial version of this PR was based on those abstractions and the code was too convoluted. Generics are not ubiquitous to be used in this fashion in Go. My proposed patterns are more familiar for the Go developers and more familiar to the people who works on OpenTelemetry.

This is also something I agree with. I think we started from the wrong direction. One of the first decisions that was made—the one about how the integration/plugin architecture should work (https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/api-design-and-project-structure.md)—is
 something that should have emerged together with the core, as we gained a better understanding of our needs. It could have been postponed until now.

That said, I’m wondering whether providing specific instrumentations (support for specific frameworks or libraries) is within the scope of this SIG, or whether we should focus solely on delivering the baseline architecture. Of course, we still need to do some validation work to ensure that this architecture is the right one.

y1yang0: > And as we discussed in the Slack channel, we SHOULD not in the business of instrumentation. We will maintain certain instrumentations and maybe we will maintain more of them at the beginning but I envision these integrations will be provided by the community. That's why we should be the enablers of those libraries, such as openteletry-go-contrib.

I disagree with the idea that we shouldn't maintain instrumentation code ourselves. Separating the tool repository + instrumentation code repository might seem like a clean architecture in theory, but it introduces significant practical problems.

Look at go-contrib: in five years, it has only produced 8 instrumentation libraries. We cannot expect them to suddenly have a burst of inspiration to rapidly fix issues, resolve problems, or add support for more libraries to meet the demands of commercial users. Furthermore, this separation creates coordination problems between the two teams.

> That said, I’m wondering whether providing specific instrumentations (support for specific frameworks or libraries) is within the scope of this SIG, or whether we should focus solely on delivering the baseline architecture. Of course, we still need to do some validation work to ensure that this architecture is the right one.

Regarding our scope: our primary goal is to deliver zero-touch code instrumentation. Users should be able to enable perfect tracing simply by prefixing their existing commands with otel.

kakkoyun: As we had a consensus in SIG meeting, we will keep instrumentation and injector separate eventually. We will add instrumentation to this repo at first.

> **CONSENSUS:**
> This repo will eventually ONLY be responsible for injector code.
> We will try to enable other instrumentation libraries as much as possible in their own repos or modules

kakkoyun: https://docs.google.com/document/d/1XkVahJfhf482d3WVHsvUUDaGzHc8TO3sqQlSS80mpGY/edit?tab=t.0#bookmark=id.1weox9j2tkt5

test/e2e/http_test.go
kakkoyun: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/214

pkg/instrumentation/nethttp/README.md
kakkoyun: >  I'm thinking on trying to keep a consistency between the readmes of different instrumentation packages. The idea is to generate a readme like this per pkg?

I would say when it comes to instrumentation, we should have a README per package. As we discussed in the previous meeting, we will move them to somewhere else eventually so this would make our lives easier.

> Could be simplified a bit just routing to this repo and otel docs?
I will refine the document a little bit.

pkg/instrumentation/grpc/go.mod
kakkoyun: gRPC demo test is quite trivial. But you have a point maybe it shouldn't be here.

pkg/instrumentation/nethttp/README.md
txabman42: Agree on having those READMEs.

Lets move the pkg README format discussion to the future. For now ❤️

pkg/instrumentation/helloworld/go.mod
kakkoyun: Handled.

test/integration/http_client_test.go
kakkoyun: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/215

.gitignore
```diff
@@ -14,3 +14,6 @@ otel.runtime.go
 otel-pkg.gz
 pkg_temp
 coverage.txt
+
+go.work
```
darccio, line 18: Nit: as we have multiple nested modules, shouldn't we check in these files? I'm ok both ways, but from experience these seem to be useful in the repository so everybody has the same versions around.

tool/internal/instrument/instrument.go
y1yang0: This week I’ve been dealing with some public matters. Next week I’ll look into and fix this issue. Sorry for blocking this patch.

pkg/instrumentation/nethttp/server/server_hook.go
```diff
@@ -0,0 +1,149 @@
...
+	if route != "" {
+		span.SetAttributes(semconv.HTTPServerRoute(route))
+	}
+
+	// Wrap ResponseWriter to capture status code
+	wrapper := &writerWrapper{
+		ResponseWriter: w,
+		statusCode:     http.StatusOK,
+	}
+	ictx.SetParam(1, wrapper)
```
pdelewski: nitpick: just for readability this `1` literal could be extracted as const
```suggestion
    const responseWriterIndex = 1
	ictx.SetParam(responseWriterIndex, wrapper)
```

pkg/instrumentation/nethttp/client/client_hook.go
```diff
@@ -0,0 +1,153 @@
...
+		trace.WithSpanKind(trace.SpanKindClient),
+		trace.WithAttributes(attrs...),
+	)
+
+	// Inject trace context into request headers
+	propagator.Inject(ctx, propagation.HeaderCarrier(req.Header))
+
+	// Update request with new context
+	newReq := req.WithContext(ctx)
+	ictx.SetParam(1, newReq)
```
pdelewski: nitpick: same as below, `1` can be const
```suggestion
    const requestParamIndex = 1
	ictx.SetParam(requestParamIndex, newReq)
```

pkg/instrumentation/nethttp/client/client_hook.go
```diff
@@ -0,0 +1,153 @@
...
+		"ctx":   ctx,
+		"span":  span,
+		"req":   req,
+		"start": time.Now(),
+	})
+}
+
+func AfterRoundTrip(ictx inst.HookContext, res *http.Response, err error) {
+	if !clientEnabler.Enable() {
+		return
```
pdelewski, line 111: Not sure about that. Maybe that's redundant, but for consistency purposes, it would be good to log this case:
```suggestion
		logger.Debug("HTTP client instrumentation disabled")
		return
```

pkg/instrumentation/nethttp/client/client_hook.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	span, ok := data["span"].(trace.Span)
+	if !ok || span == nil {
+		logger.Debug("AfterRoundTrip: no span from before hook")
+		return
+	}
+	defer span.End()
+
+	startTime, _ := data["start"].(time.Time)
```
pdelewski: This is only used in `if res != nil {` branch, so it would be better to move it there
```suggestion
if res != nil {
	startTime, _ := data["start"].(time.Time)
```

pkg/instrumentation/nethttp/server/server_hook.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	span, ok := data["span"].(trace.Span)
+	if !ok || span == nil {
+		logger.Debug("AfterServeHTTP: no span from before hook")
+		return
+	}
+	defer span.End()
+
+	startTime, _ := data["start"].(time.Time)
```
pdelewski: nitpick: I think `startTime` closer to its usage, just before `logger.Debug` call, would be better

pkg/instrumentation/nethttp/server/server_hook.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// Extract status code from wrapped ResponseWriter
+	statusCode := http.StatusOK
+	if p, ok := ictx.GetParam(1).(http.ResponseWriter); ok {
```
pdelewski: `GetParam` could use const mentioned above

pkg/instrumentation/nethttp/client/client_hook.go
```diff
@@ -0,0 +1,153 @@
...
+		propagator = otel.GetTextMapPropagator()
+		logger.Info("HTTP client instrumentation initialized")
+	})
+}
+
+// netHttpClientEnabler controls whether client instrumentation is enabled
+type netHttpClientEnabler struct{}
+
+func (n netHttpClientEnabler) Enable() bool {
+	return shared.Instrumented("NETHTTP")
```
pdelewski: I see this `NETHTTP` literal is used several times, good candidate for const, to be more safe

pdelewski:
@kakkoyun Overall, LGTM and I like the direction.

Since many things have been simplified and several abstractions removed, the document
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/api-design-and-project-structure.md

has become outdated. Do you plan to update it as part of this PR, or will that be done afterward?

.gitignore
kakkoyun: Let's introduce it in another PR. BTW, feel free to send that PR, since you have some experience on it.

kakkoyun: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/217

txabman42 (approved):

kakkoyun:
> > has become outdated. Do you plan to update it as part of this PR, or will that be done afterward?
> 
> I'll update it in this PR. Thanks for pointing it out.

@pdelewski  https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/156/commits/6c7db5f7c2014de120613515dd3b446c23e1ffc9

pkg/instrumentation/nethttp/server/server_hook.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// Wrap ResponseWriter to capture status code
+	wrapper := &writerWrapper{
+		ResponseWriter: w,
+		statusCode:     http.StatusOK,
+	}
+	ictx.SetParam(responseWriterIndex, wrapper)
+
+	// Store data for after hook
+	ictx.SetData(map[string]interface{}{
```
y1yang0, line 106: We can also
ictx.SetKeyData("ctx", ctx)
ictx.SetKeyData("span", span)
...

and ictx.GetKeyData("span").(trace.Span)

docs/api-design-and-project-structure.md
```diff
@@ -2,179 +2,255 @@
...
-        }
+2. **Semantic Conventions (semconv)**: Functions that extract OpenTelemetry attributes
+   - Direct attribute extraction from request/response objects
+   - Span status computation based on response codes
+   - Error type extraction
+
+3. **HookContext**: Interface for accessing and modifying function parameters and storing hook data
+   - `GetParam(index)`: Retrieve function parameter
+   - `SetParam(index, value)`: Modify function parameter (e.g., inject context)
+   - `GetData()` / `SetData(data)`: Pass data between Before/After hooks
```
y1yang0, line 41: And GetKeyData/SetKeyData

Makefile
```diff
@@ -207,6 +207,9 @@ test-unit: package gotestfmt
 	@echo "Running unit tests..."
 	set -euo pipefail
 	go test -json -v -shuffle=on -timeout=5m -count=1 ./tool/... 2>&1 | tee ./gotest-unit.log | gotestfmt
+	@echo "Running pkg unit tests (semconv only - hook tests require full instrumentation)..."
+	set -euo pipefail
+	cd pkg/instrumentation/nethttp/semconv && go test -json -v -shuffle=on -timeout=5m -count=1 ./... 2>&1 | tee ../../../gotest-unit-pkg.log | gotestfmt
```
y1yang0:
```
go test -C pkg/instrumentation/nethttp/semconv  -json ...
```

test/e2e/http_test.go
```diff
@@ -74,6 +74,14 @@ func TestHttp(t *testing.T) {
...
 
-	// Verify that the server hook was called.
-	require.Contains(t, output, "BeforeServeHTTP")
+	// Verify that server instrumentation was initialized
+	// Note: Client instrumentation is lazily initialized on first use,
+	// so it won't appear in the server output unless the server makes HTTP client requests
+	require.Contains(t, output, "HTTP server instrumentation initialized")
+
+	// Verify that the server ran successfully
+	require.Contains(t, output, "server started")
```
y1yang0: Remove it as it already verified in waitUntilDone

tool/internal/instrument/instrument.go
kakkoyun: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/218

Issue #117, opened 2025-10-31
Introduce a `Verify()` method for all InstRule types to centralize validation


y1yang0:
To avoid checking for the existence of certain fields in a rule at every usage point (e.g., checking if InstRawRule.Raw is empty, or if at least one of Before or After exists in InstFuncRule), we can introduce a `Verify()` method. We would require all InstRule types to implement it. This allows us to perform centralized validation right after they are created, such as when they are unmarshaled from embedded data or a JSON file.

PR #167, opened 2025-11-13
chore(rules): Add validate() to rules

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/117

Main changes:
1. Move `InstRule` loading functions from `setup/match.go` to `rule/loader.go`. This way `match.go` just match.
2. Add `validate()` method to `InstRule` types. I have also created a constructor which uses the `valitate()` internal method. The idea is to just let creating valid objects, instead of creating objects and verify afterwards. It's quite easy to forget validating them in the future.

No business logic change


kakkoyun (approved):
LGTM

tool/internal/rule/loader.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/data"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/ex"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
+	"gopkg.in/yaml.v3"
+)
+
+// MaterializeRules materializes all available rules from the embedded data
+func MaterializeRules() ([]InstRule, error) {
```
y1yang0: These are single-purpose helper functions used only for pre-match loading. It makes no sense to expose them as a public API in the rule module, so let's keep them where they are.

tool/internal/instrument/apply_file.go
```diff
@@ -30,7 +30,6 @@ func listRuleFiles(path string) ([]string, error) {
 
 // applyFileRule introduces the new file to the target package at compile time.
 func (ip *InstrumentPhase) applyFileRule(rule *rule.InstFileRule, pkgName string) error {
-	util.Assert(rule.File != "", "sanity check")
```
y1yang0, line 33: I recall we also have assertions like this that can be removed:

util.Assert(rule.Before != "" || rule.After != "", "hook must be set")

tool/internal/rule/loader.go
txabman42: Moved back to match.go

tool/internal/instrument/apply_file.go
txabman42: Good catch, I missed that one. Removed and added that validation to `InstFuncRule`

tool/internal/rule/loader.go
y1yang0: ISorry for the confusion. What I meant was that we should move all the functions from loader.go to match.go. They are unlikely to be reused in the foreseeable future, so it doesn't make much sense to expose them as public functions of the rules package.

txabman42: Sorry I misunderstood it. Moved them back to match.go

txabman42:
@y1yang0 note that now `r.Verify()` not is not needed after `createRuleFromFields`.

`createRuleFromFields` uses directly the constructor which validates if the rule is correct or not. We just avoid in this way to verify manually, as we only allow to create valid rules.

> The idea is to just let creating valid objects, instead of creating objects and verify afterwards. It's quite easy to forget validating them in the future (in case new rules are created somewhere else).

Issue #183, opened 2025-11-19
Flatten trampoline-jump-if as AST-level optimization


y1yang0:
The current standard tjump pattern is as follows:

```go
	if ctx, skip := otel_trampoline_before(&arg); skip {
	    otel_trampoline_after(ctx, &retval)
	    return ...
	} else {
	    defer otel_trampoline_after(ctx, &retval)
	    ...
	}
```

A key optimization opportunity lies in `skip`, which is highly likely to be `false`. In this scenario, the tjump can be optimized into the following form:

```go
	ctx,_ := otel_trampoline_before(&arg);
	defer otel_trampoline_after(ctx, &retval)
```

Consider the following hook code (the Hook is ultimately called by `otel_trampoline_before`):

```go
func Hook(ictx inst.HookContext, arg1....) {
	ictx.SetSkipCall()
	passTo(ictx)
        var escape interface{} = ictx
}
```

We can assume that:
1. If `“SetSkipCall"` string never appears in the Hook code,
2. and `ictx` is never used for purposes other than method calls (e.g., assignment, pass as args),

then `ictx` does not escape, and the tjump represents a valid candidate for optimization. This would significantly boost performance.

PR #184, opened 2025-11-19
chore: flatten trampoline-jump-if at AST-level

y1yang0:
related to #183

### DIFF
--- a/.github/workflows/check-conventional-commit.yml
+++ b/.github/workflows/check-conventional-commit.yml
@@ -71,4 +71,3 @@ jobs:
           }}" -pr="${{ github.event.pull_request.number }}"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
--- a/.github/workflows/check-license-headers.yaml
+++ b/.github/workflows/check-license-headers.yaml
@@ -26,3 +26,18 @@ jobs:
     steps:
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955  # v4.3.0
       - run: make lint/license-header
+
+  done:
+    name: Done (Check License Headers)
+    runs-on: ubuntu-latest
+    needs: [check-license-headers]
+    if: always()
+    steps:
+      - name: Check result
+        run: |
+          echo ${{ needs.check-license-headers.result }}
+          if [ "${{ needs.check-license-headers.result }}" = "success" ] || [ "${{ needs.check-license-headers.result }}" = "skipped" ]; then
+            exit 0
+          else
+            exit 1
+          fi
--- a/.github/workflows/lint-golang.yaml
+++ b/.github/workflows/lint-golang.yaml
@@ -32,3 +32,18 @@ jobs:
         with:
           version: ${{ env.GOLANGCI_LINT_VERSION }}
           args: --config .config/golangci.yml
+
+  done:
+    name: Done (Check Go Code)
+    runs-on: ubuntu-latest
+    needs: [golangci-lint]
+    if: always()
+    steps:
+      - name: Check result
+        run: |
+          echo ${{ needs.golangci-lint.result }}
+          if [ "${{ needs.golangci-lint.result }}" = "success" ] || [ "${{ needs.golangci-lint.result }}" = "skipped" ]; then
+            exit 0
+          else
+            exit 1
+          fi
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -3,7 +3,27 @@ on:
   push:
     branches:
       - main
+    paths:
+      - '**/*.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - 'Makefile'
+      - 'tool/**'
+      - 'pkg/**'
+      - 'test/**'
+      - 'demo/**'
+      - '.github/workflows/test-e2e.yaml'
   pull_request:
+    paths:
+      - '**/*.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - 'Makefile'
+      - 'tool/**'
+      - 'pkg/**'
+      - 'test/**'
+      - 'demo/**'
+      - '.github/workflows/test-e2e.yaml'
 # Declare default permissions as read only.
 permissions: read-all
 jobs:
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -3,7 +3,25 @@ on:
   push:
     branches:
       - main
+    paths:
+      - '**/*.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - 'Makefile'
+      - 'tool/**'
+      - 'pkg/**'
+      - 'test/**'
+      - '.github/workflows/test-integration.yaml'
   pull_request:
+    paths:
+      - '**/*.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - 'Makefile'
+      - 'tool/**'
+      - 'pkg/**'
+      - 'test/**'
+      - '.github/workflows/test-integration.yaml'
 # Declare default permissions as read only.
 permissions: read-all
 jobs:
--- a/.github/workflows/test-unit.yaml
+++ b/.github/workflows/test-unit.yaml
@@ -3,7 +3,25 @@ on:
   push:
     branches:
       - main
+    paths:
+      - '**/*.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - 'Makefile'
+      - 'tool/**'
+      - 'pkg/**'
+      - 'test/**'
+      - '.github/workflows/test-unit.yaml'
   pull_request:
+    paths:
+      - '**/*.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - 'Makefile'
+      - 'tool/**'
+      - 'pkg/**'
+      - 'test/**'
+      - '.github/workflows/test-unit.yaml'
 # Declare default permissions as read only.
 permissions: read-all
 jobs:


tool/internal/instrument/optimize.go
```diff
@@ -178,6 +180,111 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+	// 2. and "ictx" is never used for purposes other than method calls (e.g.,
+	// assignment, pass as args, etc.),
+	// then "ictx" does not escape, and the tjump represents a valid candidate
+	// for optimization. This would significantly boost performance.
+	hookFunc, err := getHookFunc(tjump.rule, true)
+	if err != nil {
+		return err
+	}
+	// Check if the hook function contains any "SetSkipCall" string
+	foundPoison := false
```
kakkoyun: nit:
```suggestion
	found := false
```

tool/internal/instrument/apply_func.go
```diff
@@ -95,7 +95,7 @@ func createHookArgs(names []string) []dst.Expr {
 	// If we find "a type" in target func, we pass "&a" to trampoline func,
 	// if we find "_ type" in target func, we pass "nil" to trampoline func,
 	for _, name := range names {
-		if name == "_" {
+		if name == ast.IdentIgnore {
```
kakkoyun, line 96: 👍

tool/internal/instrument/optimize.go
```diff
@@ -178,6 +180,116 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+	}
+
+	// Flatten the trampoline-jump-if to sequential function calls
+	ifStmt := tjump.ifStmt
+	initStmt, ok := ifStmt.Init.(*dst.AssignStmt)
+	const twoLhs = 2
+	util.Assert(ok, "init statement is not an AssignStmt")
+	util.Assert(len(initStmt.Lhs) == twoLhs, "must have two lhs")
+	ifStmt.Cond = ast.BoolFalse()
+	ifStmt.Body = ast.Block(ast.EmptyStmt())
```
kakkoyun, line 300: Linter errors actually shows us an oppurtunity to add some helper methods.

Instead of pulling inline trick like this we can have. `util.TypeAssert` and similar for checking number of nodes.


What do you think?

tool/internal/instrument/optimize.go
```diff
@@ -178,6 +180,116 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+	dst.Inspect(hookFunc, func(node dst.Node) bool {
+		if ident, ok := node.(*dst.Ident); ok {
+			if strings.Contains(ident.Name, trampolineSetSkipCallName) {
+				found = true
+				return false
+			}
+		}
+		if found {
+			return false
+		}
```
txabman42:
```suggestion
```

tool/internal/instrument/optimize.go
```diff
@@ -195,16 +307,36 @@ func (ip *InstrumentPhase) optimizeTJumps() error {
...
+			}
+		}
+
+		// No SkipCall used in Before hook? Rewrite cond of trampoline-jump-if
+		// to always false, and remove return statement in then block, they are
+		// memory aware and may generate memory SSA values during compilation.
+		// This further simplifies the trampoline-jump-if and gives more chances
+		// for optimization passes to kick in.
+		if rule.Before != "" {
+			err := flattenTJump(tjump, removedOnExit)
```
txabman42: What do you think about splitting `flattenTJump` into two private funcs? sth like:

```
if canFlattenTJump {
  flattenTJump
}
```

tool/internal/instrument/optimize.go
```diff
@@ -178,6 +180,116 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+	//
+	// Consider the following hook code
+	//
+	//	func hookFunc(ictx HookContext, arg1....) {
+	// 		ictx.SetSkipCall()
+	// 		passTo(ictx)
+	// 		var escape interface{} = ictx
+	//	}
+	//
+	// We can assume that:
```
pdelewski: If my understanding is correct, then this optimization can be applied only when `ictx` is an receiver. If so, then maybe we could rephrase this comment a bit:
```suggestion
	// This optimization can be applied when the HookContext parameter meets these conditions:
	// 1. SetSkipCall is never called (so skip is always false)
	// 2. The HookContext is ONLY used as a receiver for method calls
	//
	// Allowed usage (optimization can proceed):
	//	func hookFunc(ictx HookContext, arg1....) {
	// 		ictx.GetParam(0)      // ✓ Method calls are allowed
	// 		ictx.SetParam(1, val) // ✓ Method calls are allowed
	//	}
	//
	// Disallowed usage (prevents optimization):
	//	func hookFunc(ictx HookContext, arg1....) {
	// 		ictx.SetSkipCall()           // ✗ SetSkipCall prevents optimization
	// 		passTo(ictx)                 // ✗ Cannot pass as argument
	// 		var escape interface{} = ictx // ✗ Cannot assign to variable
	// 		_ = ictx                     // ✗ Cannot use in any assignment
	//	}
	//
	// When both conditions are met, the HookContext doesn't escape and we can
	// safely flatten the trampoline-jump-if pattern, significantly boosting performance.
```

tool/internal/instrument/optimize.go
```diff
@@ -178,6 +180,116 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+	// 1. If "SetSkipCall" string never appears in the Hook code,
+	// 2. and HookContext parameter is never used for purposes other than method
+	// calls (e.g. assignment, pass as args, etc.), then the HookContext parameter
+	// does not escape, and the tjump represents a valid candidate for optimization.
+	// This would significantly boost performance.
+	hookFunc, err := getHookFunc(tjump.rule, true)
```
pdelewski: we could extract function to check whether optimization is allowed
```suggestion
// isFlattenTJumpAllowed checks if the tjump can be safely flattened based on
// the hook function's usage of HookContext. Returns true if:
// 1. SetSkipCall is never called (so skip is always false)
// 2. The HookContext parameter is only used as a receiver for method calls
func isFlattenTJumpAllowed(tjump *TJump) (bool, error) {
	hookFunc, err := getHookFunc(tjump.rule, true)
	if err != nil {
		return false, err
	}

	// Check if the hook function contains any "SetSkipCall" string
	found := false
	dst.Inspect(hookFunc, func(node dst.Node) bool {
		if ident, ok := node.(*dst.Ident); ok {
			if strings.Contains(ident.Name, trampolineSetSkipCallName) {
				found = true
				return false
			}
		}
		return true
	})
	if found {
		return false, nil
	}

	// Check if the hook context parameter escapes (used for non-method calls)
	hookContextParam := hookFunc.Type.Params.List[0].Names[0].Name
	if hookContextParam == ast.IdentIgnore {
		// If the parameter is ignored, it doesn't escape
		return true, nil
	}

	escape := false
	dst.Inspect(hookFunc.Body, func(n dst.Node) bool {
		if escape {
			return false
		}
		switch n := n.(type) {
		case *dst.SelectorExpr:
			// Check if ictx is used as method receiver: ictx.Method()
			if id, ok := n.X.(*dst.Ident); ok && id.Name == hookContextParam {
				// Valid usage.
				// Return false to stop visiting children, so we don't visit
				// the Ident "ictx", which would be caught by the case below.
				return false
			}
		case *dst.Ident:
			// If we encounter ictx here, it means it wasn't part of a method
			// call receiver (because we returned false above). So it is an
			// invalid usage.
			if n.Name == hookContextParam {
				escape = true
				return false
			}
		}
		return true
	})

	return !escape, nil
}
```

tool/internal/instrument/optimize.go
pdelewski: That's good suggestion. Look above

tool/internal/instrument/optimize.go
```diff
@@ -178,6 +180,116 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+				}
+			}
+			return true
+		})
+	}
+	if escape {
+		return nil
+	}
+
+	// Flatten the trampoline-jump-if to sequential function calls
```
pdelewski: Extracting function
```suggestion
	// Flatten the trampoline-jump-if to sequential function calls
	// rewriteTJumpAST transforms the trampoline-jump-if AST to a flattened form.
// It sets the condition to false and empties the then block, effectively
// converting the branching pattern to sequential execution.
func rewriteTJumpAST(tjump *TJump, removedOnExit bool) {
	ifStmt := tjump.ifStmt
	initStmt, ok := ifStmt.Init.(*dst.AssignStmt)
	const twoLhs = 2
	util.Assert(ok, "init statement is not an AssignStmt")
	util.Assert(len(initStmt.Lhs) == twoLhs, "must have two lhs")
	
	// Set condition to false and empty the then block
	ifStmt.Cond = ast.BoolFalse()
	ifStmt.Body = ast.Block(ast.EmptyStmt())
	
	if removedOnExit {
		// We removed the last reference to hook context after nulling out body
		// block, at this point, all lhs are unused, replace assignment to simple
		// function call
		ifStmt.Init = ast.ExprStmt(initStmt.Rhs[0])
		// TODO: Remove After declaration as well
	} else {
		// Otherwise, mark skipCall identifier as unused
		skipCallIdent, ok1 := initStmt.Lhs[1].(*dst.Ident)
		util.Assert(ok1, "skipCall identifier is not an Ident")
		ast.MakeUnusedIdent(skipCallIdent)
	}
}
```

tool/internal/instrument/optimize.go
y1yang0: Good suggestion.

tool/internal/instrument/optimize.go
y1yang0: Good suggestion.

Tracked in #186

Issue #169, opened 2025-11-14
Implement generic function instrumentation


y1yang0:
The current instrumentation logic does not support generic functions

```
func F[T any](value T) {}
```

However, this is a real-world and frequently requested feature, especially for instrumenting libraries like those used in Kubernetes. We need to implement this functionality, but we anticipate it will be a very challenging task.

txabman42:
`//go:linknames` and generic functions don't match well.

From[ Linkname Directive](https://pkg.go.dev/cmd/compile#hdr-Linkname_Directive)

> //go:linkname directive determines the object-file symbol used for a Go var or func declaration, allowing two Go symbols to alias the same object-file symbol.

As they are the alias of the object-file symbol, we assume that both functions (hook and target) must share the same signature.

Thats a problem with generic functions, as the generic function instantiation doesn't exist, but a function instantiation per shape ([generics-implementation-dictionaries-go1.18.md](https://github.com/golang/proposal/blob/master/design/generics-implementation-dictionaries-go1.18.md)). So, `//go:linknames` directive won't know how to link hook and target generic functions.

They also comment [here](https://github.com/golang/go/issues/60425#issuecomment-1562253823) that:

> //go:linkname is not supported for generic code. There's no reasonable way to make this work given how generics is implemented.

However, from [checklinkname Analyzer](https://pkg.go.dev/gvisor.dev/gvisor/tools/checklinkname#section-readme) we can read:

> Since //go:linkname only affects linker behavior, there is no built-in type safety and it is the programmer's responsibility to ensure the types on either side are **compatible**.

Not same signature, but compatible signatures!

So, my question is: could have sense from the hook design POV to test the following idea?


Example:
```go
func GenericExample[K comparable, V any](value T) map[K]V { ... }

func GenericHookBefore(ictx inst.HookContext, k interface{}, v interface{}) {
  // Can perform type assertion if needed
  ...
}

func GenericHookAfter(ictx inst.HookContext, r interface{}) {
  // Can perform type assertion if needed
  ...
}
```

y1yang0:
> Could that not be an acceptable limitation for now?
> 
> If its documented, although not being user friendly, devs interested in instrumenting a generic function wouldn't be blocked.

Yes, I agree this is an acceptable approach for now. It's not perfect, but it's a workable solution, and the need for generic injection, while it exists, isn't very common.

PR #185, opened 2025-11-19
feat: support generic funcs

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/169

More info in ☝️ issue about why Hooks use `interface{}` as argument type:

```go
func GenericExample[K comparable, V any](value T) map[K]V { ... }

func GenericHookBefore(ictx inst.HookContext, k interface{}, v interface{}) { ... }

func GenericHookAfter(ictx inst.HookContext, r interface{}) { ... }
```

This PR also add the support for parameter grouping in hooks, example:

```go
func MyHookGenericBefore(ictx inst.HookContext, _, _ interface{}) { ... }
```

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/semantic-router/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to semantic-router! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>Pass all linter checks. Please use <code>pre-commit</code> to format your code. See <code>README.md</code> for installation.</li>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>DCO and Signed-off-by</h3>
<p>When contributing changes to this project, you must agree to the <a href="https://github.com/vllm-project/vllm/blob/main/DCO">DCO</a>. Commits must include a <code>Signed-off-by:</code> header which certifies agreement with the terms of the DCO.</p>
<p>Using <code>-s</code> with <code>git commit</code> will automatically add this header.</p>

<h3>What to Expect for the Reviews</h3>

### DIFF
--- a/.github/workflows/check-license-headers.yaml
+++ b/.github/workflows/check-license-headers.yaml
@@ -35,9 +35,12 @@ jobs:
     steps:
       - name: Check result
         run: |
-          echo ${{ needs.check-license-headers.result }}
-          if [ "${{ needs.check-license-headers.result }}" = "success" ] || [ "${{ needs.check-license-headers.result }}" = "skipped" ]; then
+          result="${{ needs.check-license-headers.result }}"
+          echo "Test result: $result"
+          if [[ "$result" == "success" || "$result" == "skipped" ]]; then
+            echo "✅ Tests passed or were skipped"
             exit 0
           else
+            echo "❌ Tests failed"
             exit 1
           fi
--- a/.github/workflows/lint-golang.yaml
+++ b/.github/workflows/lint-golang.yaml
@@ -41,9 +41,12 @@ jobs:
     steps:
       - name: Check result
         run: |
-          echo ${{ needs.golangci-lint.result }}
-          if [ "${{ needs.golangci-lint.result }}" = "success" ] || [ "${{ needs.golangci-lint.result }}" = "skipped" ]; then
+          result="${{ needs.golangci-lint.result }}"
+          echo "Test result: $result"
+          if [[ "$result" == "success" || "$result" == "skipped" ]]; then
+            echo "✅ Tests passed or were skipped"
             exit 0
           else
+            echo "❌ Tests failed"
             exit 1
           fi
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -66,8 +66,16 @@ jobs:
     name: Done (E2E Tests)
     runs-on: ubuntu-latest
     needs: [test-e2e]
+    if: always()
     steps:
-      - name: Success
+      - name: Check result
         run: |
-          echo ${{ needs.test-e2e.result }}
-          test ${{ needs.test-e2e.result }} == "success"
+          result="${{ needs.test-e2e.result }}"
+          echo "Test result: $result"
+          if [[ "$result" == "success" || "$result" == "skipped" ]]; then
+            echo "✅ Tests passed or were skipped"
+            exit 0
+          else
+            echo "❌ Tests failed"
+            exit 1
+          fi
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -67,8 +67,16 @@ jobs:
     name: Done (Integration Tests)
     runs-on: ubuntu-latest
     needs: [test-integration]
+    if: always()
     steps:
-      - name: Success
+      - name: Check result
         run: |
-          echo ${{ needs.test-integration.result }}
-          test ${{ needs.test-integration.result }} == "success"
+          result="${{ needs.test-integration.result }}"
+          echo "Test result: $result"
+          if [[ "$result" == "success" || "$result" == "skipped" ]]; then
+            echo "✅ Tests passed or were skipped"
+            exit 0
+          else
+            echo "❌ Tests failed"
+            exit 1
+          fi
--- a/.github/workflows/test-unit.yaml
+++ b/.github/workflows/test-unit.yaml
@@ -71,8 +71,16 @@ jobs:
     name: Done (Unit Tests)
     runs-on: ubuntu-latest
     needs: [test-unit]
+    if: always()
     steps:
-      - name: Success
+      - name: Check result
         run: |
-          echo ${{ needs.test-unit.result }}
-          test ${{ needs.test-unit.result }} == "success"
+          result="${{ needs.test-unit.result }}"
+          echo "Test result: $result"
+          if [[ "$result" == "success" || "$result" == "skipped" ]]; then
+            echo "✅ Tests passed or were skipped"
+            exit 0
+          else
+            echo "❌ Tests failed"
+            exit 1
+          fi


tool/internal/ast/primitives_test.go
```diff
@@ -0,0 +1,316 @@
...
+	assert.Equal(t, expectedFuncName, funcIdent.Name)
+	require.Len(t, indexListExpr.Indices, len(expectedTypeParams))
+	for i, expectedParam := range expectedTypeParams {
+		paramIdent, _ := indexListExpr.Indices[i].(*dst.Ident)
+		assert.Equal(t, expectedParam, paramIdent.Name)
+	}
+	assert.Len(t, expr.Args, expectedArgCount)
+}
+
+func TestCallToGeneric(t *testing.T) {
```
pdelewski: I'm wondering whether operating on source code text would not be cleaner
```suggestion
// Helper function to parse a complete function and extract its type parameters
func parseFuncTypeParams(t *testing.T, funcSource string) *dst.FieldList {
	parser := NewAstParser()
	file, err := parser.ParseSource("package main\n" + funcSource)
	require.NoError(t, err)
	require.Len(t, file.Decls, 1)
	funcDecl, ok := file.Decls[0].(*dst.FuncDecl)
	require.True(t, ok)
	return funcDecl.Type.TypeParams
}

func TestCallToGeneric(t *testing.T) {
	tests := []struct {
		name       string
		funcName   string
		funcSource string // Source code for parsing type params
		args       []dst.Expr
		validate   func(*testing.T, *dst.CallExpr)
	}{
		{
			name:       "nil type params returns simple call",
			funcName:   "Foo",
			funcSource: "func Foo(x, y int) {}", // No type params
			args:       []dst.Expr{Ident("x"), Ident("y")},
			validate: func(t *testing.T, expr *dst.CallExpr) {
				assertSimpleCall(t, expr, "Foo", 2)
			},
		},
		{
			name:       "single type parameter creates IndexExpr",
			funcName:   "GenericFunc",
			funcSource: "func GenericFunc[T any](value T) {}",
			args:       []dst.Expr{Ident("value")},
			validate: func(t *testing.T, expr *dst.CallExpr) {
				assertIndexExprCall(t, expr, "GenericFunc", "T", 1)
			},
		},
		{
			name:       "multiple type parameters creates IndexListExpr",
			funcName:   "MultiGeneric",
			funcSource: "func MultiGeneric[T any, U comparable](x T, y U) {}",
			args:       []dst.Expr{Ident("x"), Ident("y")},
			validate: func(t *testing.T, expr *dst.CallExpr) {
				assertIndexListExprCall(t, expr, "MultiGeneric", []string{"T", "U"}, 2)
			},
		},
		{
			name:       "field with multiple names creates multiple indices",
			funcName:   "MultiNameGeneric",
			funcSource: "func MultiNameGeneric[T, U any](value T) {}",
			args:       []dst.Expr{Ident("value")},
			validate: func(t *testing.T, expr *dst.CallExpr) {
				assertIndexListExprCall(t, expr, "MultiNameGeneric", []string{"T", "U"}, 1)
			},
		},
		{
			name:       "no arguments with type parameters",
			funcName:   "NoArgsGeneric",
			funcSource: "func NoArgsGeneric[T any]() {}",
			args:       []dst.Expr{},
			validate: func(t *testing.T, expr *dst.CallExpr) {
				assertIndexExprCall(t, expr, "NoArgsGeneric", "T", 0)
			},
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			typeParams := parseFuncTypeParams(t, tt.funcSource)
			result := CallToGeneric(tt.funcName, typeParams, tt.args)
			require.NotNil(t, result)
			tt.validate(t, result)
		})
	}
}

```

tool/internal/ast/primitives.go
```diff
@@ -289,3 +319,43 @@ func StructLit(typeName string, fields ...*dst.KeyValueExpr) dst.Expr {
...
+		return nil
+	}
+	cloned, ok := dst.Clone(typeParams).(*dst.FieldList)
+	util.Assert(ok, "typeParams is not a FieldList")
+	return cloned
+}
+
+// SplitMultiNameFields splits fields that have multiple names into separate fields.
+// For example, a field like "a, b int" becomes two fields: "a int" and "b int".
+func SplitMultiNameFields(fieldList *dst.FieldList) *dst.FieldList {
```
pdelewski: I think that the body inside for loop can be generalized
```suggestion
func SplitMultiNameFields(fieldList *dst.FieldList) *dst.FieldList {
	if fieldList == nil {
		return nil
	}
	result := &dst.FieldList{List: []*dst.Field{}}
	for _, field := range fieldList.List {
		// Handle unnamed fields (e.g., embedded types) or fields with single/multiple names
		namesToProcess := field.Names
		if len(namesToProcess) == 0 {
			// For unnamed fields, create one field with no names
			namesToProcess = []*dst.Ident{nil}
		}
		
		for _, name := range namesToProcess {
			clonedType, ok := dst.Clone(field.Type).(dst.Expr)
			util.Assert(ok, "field.Type is not an Expr")
			
			var names []*dst.Ident
			if name != nil {
				clonedName, ok := dst.Clone(name).(*dst.Ident)
				util.Assert(ok, "name is not an Ident")
				names = []*dst.Ident{clonedName}
			}
			
			newField := &dst.Field{
				Names: names,
				Type:  clonedType,
			}
			result.List = append(result.List, newField)
		}
	}
	return result
}

```

tool/internal/ast/primitives.go
```diff
@@ -50,6 +50,36 @@ func CallTo(name string, args []dst.Expr) *dst.CallExpr {
 	}
 }
 
+// CallToGeneric creates a call expression to a generic function with explicit type arguments.
```
pdelewski:
```suggestion
// CallToGeneric creates a call expression to a generic function with an explicit type arguments.
```

demo/basic/main.go
```diff
@@ -31,6 +31,11 @@ type MyStruct struct{}
 
 func (m *MyStruct) Example() { println("MyStruct.Example") }
 
+func GenericExample[K comparable, V any](key K, value V) V {
+	println("Hello, Generic World!", key, value)
+	return value
+}
+
```
y1yang0, line 47: Add test case func (recv *Recv[T]) GenericExample

tool/internal/ast/primitives.go
```diff
@@ -289,3 +319,43 @@ func StructLit(typeName string, fields ...*dst.KeyValueExpr) dst.Expr {
 		X:  CompositeLit(Ident(typeName), exprs),
 	}
 }
+
```
y1yang0: Let's use primitive.go for basic AST building blocks. Utility functions like CloneXXX(), IsXXX(), and SplitXXX() should be placed in shared.go.

tool/internal/instrument/trampoline.go
```diff
@@ -676,37 +687,81 @@ func (ip *InstrumentPhase) rewriteHookContext() {
...
-		}
+	splitParams := ast.SplitMultiNameFields(ip.targetFunc.Type.Params)
+	for _, param := range splitParams.List {
+		paramType := replaceTypeParamsWithAny(desugarType(param), ip.targetFunc.Type.TypeParams)
+		clause := setParamClause(idx, paramType)
+		methodSetParamBody.List = append(methodSetParamBody.List, clause)
+		clause = getParamClause(idx, paramType)
+		methodGetParamBody.List = append(methodGetParamBody.List, clause)
+		idx++
 	}
```
y1yang0, line 805: The SetParam/SetReturnVal function is intended to modify the parameters of the original function, but the current implementation doesn't allow this. We should rewrite it to panic (or something similar) to reject user calls and make it clear this operation is unsupported.

tool/internal/instrument/apply_func.go
```diff
@@ -115,8 +115,8 @@ func createTJumpIf(t *rule.InstFuncRule, funcDecl *dst.FuncDecl,
 	argsToAfter := createHookArgs(retVals)
 	argHookContext := ast.Ident(trampolineHookContextName + funcSuffix)
 	argsToAfter = append([]dst.Expr{argHookContext}, argsToAfter...)
-	beforeCall := ast.CallTo(makeName(t, funcDecl, true), argsToBefore)
-	afterCall := ast.CallTo(makeName(t, funcDecl, false), argsToAfter)
+	beforeCall := ast.CallToGeneric(makeName(t, funcDecl, true), funcDecl.Type.TypeParams, argsToBefore)
+	afterCall := ast.CallToGeneric(makeName(t, funcDecl, false), funcDecl.Type.TypeParams, argsToAfter)
```
y1yang0: Let's keep the name CallTo and just add the logic to handle generics.

docs/rules.md
```diff
@@ -46,6 +46,34 @@ hook_helloworld:
 
 This rule will inject `MyHookBefore` at the start of the `Example` function in the `main` package, and `MyHookAfter` at the end. The hook functions are located in the specified `path`.
 
+**Generic Function Support:**
```
y1yang0: The documentation needs to be updated to state that SetParam and SetReturnVal cannot be used with generic hook functions.

pkg/instrumentation/helloworld/helloworld_hook.go
```diff
@@ -71,4 +71,12 @@ func MyHook1After(ictx inst.HookContext) {
 	println("After MyStruct.Example()")
 }
 
+func MyHookGenericBefore(ictx inst.HookContext, _, _ interface{}) {
```
y1yang0, line 87: Add usages of  HookContext methods inside Hook function to exercise their functionalities

tool/internal/instrument/trampoline.go
```diff
@@ -676,37 +687,81 @@ func (ip *InstrumentPhase) rewriteHookContext() {
...
+				return true
 			}
 		}
 	}
+	return false
+}
+
+// replaceTypeParamsWithAny replaces type parameters with interface{} for use in
+// non-generic contexts like HookContextImpl methods
+func replaceTypeParamsWithAny(t dst.Expr, typeParams *dst.FieldList) dst.Expr {
```
y1yang0: replaceTypeParamsWithAny and rectifyAnyType seem to have some similarities. It might be possible to refactor them into a single function, but this can be handled in a follow-up PR.

tool/internal/instrument/trampoline.go
```diff
@@ -676,37 +687,81 @@ func (ip *InstrumentPhase) rewriteHookContext() {
...
+	case *dst.ArrayType:
+		// []T -> []interface{}
+		return ast.ArrayType(replaceTypeParamsWithAny(tType.Elt, typeParams))
+	case *dst.MapType:
+		// map[K]V -> map[interface{}]interface{}
+		return &dst.MapType{
+			Key:   replaceTypeParamsWithAny(tType.Key, typeParams),
+			Value: replaceTypeParamsWithAny(tType.Value, typeParams),
+		}
+	}
```
y1yang0, line 905: Does this cover all possible syntax cases? Should we add a default case with util.Unimplemented() or util.ShouldNotReachHere() to be safe?

docs/rules.md
```diff
@@ -46,6 +46,34 @@ hook_helloworld:
...
+
+Generic functions are supported. The instrumentation tool automatically handles type parameters, propagating them to the generated trampoline functions.
```
y1yang0: The primary focus of this document is to detail the types and usage of rules. Perhaps we should create a separate document on how to write hooks, and this section could be moved there. However, this can be handled in a follow-up PR.

tool/internal/ast/primitives.go
```diff
@@ -50,6 +50,36 @@ func CallTo(name string, args []dst.Expr) *dst.CallExpr {
...
+			indices = append(indices, Ident(ident.Name))
+		}
+	}
+	var fun dst.Expr
+	if len(indices) == 1 {
+		fun = &dst.IndexExpr{
+			X:     Ident(name),
+			Index: indices[0],
+		}
+	} else {
```
y1yang0, line 66: ast.IndexExpr

tool/internal/instrument/trampoline.go
```diff
@@ -340,6 +341,11 @@ func (ip *InstrumentPhase) addHookFuncVar(t *rule.InstFuncRule,
 	if err != nil {
 		return err
 	}
+	// For generic target functions, replace type parameters with interface{}
+	// in the linkname declaration (since linkname doesn't support generics)
+	for _, field := range paramTypes.List {
+		field.Type = replaceTypeParamsWithAny(field.Type, ip.targetFunc.Type.TypeParams)
+	}
```
y1yang0, line 334: rectifyAnyType() already does this. Why do we need to repeat it here?

tool/internal/instrument/trampoline.go
```diff
@@ -429,19 +435,22 @@ func (ip *InstrumentPhase) buildTrampolineType(before bool) *dst.FieldList {
 	paramList := &dst.FieldList{List: []*dst.Field{}}
```
y1yang0: Could you please add a note to the comment above, clarifying that all grouped parameters (like a, b int) are expanded into separate parameters?
```
    // Build parameter list of trampoline function.
	// For before trampoline, it's signature is:
	// func S(h* HookContext, recv type, arg1 type, arg2 type, ...)
	// For after trampoline, it's signature is:
	// func S(h* HookContext, arg1 type, arg2 type, ...)
    // <comment>
```

tool/internal/ast/primitives_test.go
txabman42: I like this idea, much cleaner. Test refactored. c318de690837f39e8c4856e8fbb57ea3efe10cae

tool/internal/ast/primitives.go
txabman42: Updated in c318de690837f39e8c4856e8fbb57ea3efe10cae

demo/basic/main.go
txabman42: Test added.

There were some extra changes needed to support this case:
- Extract receiver type parameters
- Set hook ctx params as a combination of receiver and function types
- Preserve generics in trampolines

This new code would require some new 👀 63a9420ba8917e596b0f9b0e6efd80ea269e5b54

tool/internal/ast/primitives.go
txabman42: Agree, moved to shared.go 6ffc5d15f04b5d32f545e4096232437e6c968755

tool/internal/instrument/apply_func.go
txabman42: Agree, updated in 6ffc5d15f04b5d32f545e4096232437e6c968755

tool/internal/instrument/trampoline.go
txabman42: Updated in 58aa8e364a31f81d6b304529992d7a1c6ae7d74e

tool/internal/instrument/trampoline.go
txabman42: They are not exactly the same.

For example:
func (t *T) Func1(p1 string, p2 int)

The receiver trait is marked as IsInterfaceAny and replaced here, so it ends up in:

func H3Before(ctx inst.HookContext, recv interface{}, p1 string, p2 int)


TestInstrumentation_Integration/method-receiver covers this case.

However, it could be simplified 58aa8e364a31f81d6b304529992d7a1c6ae7d74e

tool/internal/ast/primitives.go
txabman42: Good catch, updated in 4f3724fb84b0ad10474d64831d01ec2b28275ea1. Also added ast.IndexListExpr

tool/internal/instrument/trampoline.go
txabman42: really good catch. I have added support for chans and set a default case with util.Unimplemented() 561c4675afaa16c62b029c88fbdace9b1059cc1b. Right now not supported funcs and uncommon types. They could be supported in the future.

docs/rules.md
txabman42: Good idea, removed from rules.md 73bafcdfb972578a2309945e78f0414e7f62d3e0

I have created this issue https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/194 to track the creation of hooks.md

docs/rules.md
txabman42: We can document these limitations as part of https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/194. They should be more related to hooks than rules itself.

pkg/instrumentation/helloworld/helloworld_hook.go
txabman42: Good idea, done in 74ff92a8502d94741e8cc24ce52ba725b02bd876

These integration tests should be refactored once golden file approach is moved to /test/integration

tool/internal/instrument/trampoline.go
txabman42: This would require a extra bit of code. Do you see any direct approach?

WDYT about moving this to a new issue? This PR is starting to get big

tool/internal/ast/primitives.go
```diff
@@ -275,3 +305,14 @@ func StructLit(typeName string, fields ...*dst.KeyValueExpr) dst.Expr {
 		X:  CompositeLit(Ident(typeName), exprs),
 	}
 }
+
+// CloneTypeParams safely clones a type parameter field list for generic functions.
+// Returns nil if the input is nil.
+func CloneTypeParams(typeParams *dst.FieldList) *dst.FieldList {
```
y1yang0: Move CloneTypeParams to shared.go

tool/internal/ast/primitives.go
```diff
@@ -275,3 +305,14 @@ func StructLit(typeName string, fields ...*dst.KeyValueExpr) dst.Expr {
...
+	if typeParams == nil {
+		return nil
+	}
+	cloned, ok := dst.Clone(typeParams).(*dst.FieldList)
+	util.Assert(ok, "typeParams is not a FieldList")
+	return cloned
```
y1yang0: And you can use `cloned := util.AssertType[*dst.FieldList](...)` after merging main branch

tool/internal/ast/shared.go
```diff
@@ -44,6 +44,48 @@ func FindFuncDeclWithoutRecv(root *dst.File, funcName string) *dst.FuncDecl {
...
 }
 
+// extractBaseReceiverType extracts the base type name from a receiver expression,
+// handling both generic and non-generic types.
+// For example:
+// - *MyStruct -> *MyStruct
+// - MyStruct -> MyStruct
+// - *GenStruct[T] -> *GenStruct
+// - GenStruct[T] -> GenStruct
+func extractBaseReceiverType(recvTypeExpr dst.Expr) string {
```
y1yang0: How about renaming it to "stripGenericTypes"? It might be more general. But it's up to you; the current name is fine as well.

tool/internal/instrument/trampoline.go
```diff
@@ -315,31 +316,27 @@ func (ip *InstrumentPhase) callAfterHook(t *rule.InstFuncRule, traits []ParamTra
 	return nil
 }
 
-func rectifyAnyType(paramList *dst.FieldList, traits []ParamTrait) error {
-	if len(paramList.List) != len(traits) {
+func (ip *InstrumentPhase) addHookFuncVar(t *rule.InstFuncRule,
+	traits []ParamTrait, before bool,
+) error {
+	paramTypes := ip.buildTrampolineType(before)
```
y1yang0: I think we could refactor buildTrampolineType to return two dst.FieldLists: one for the generic type parameters and the other for the normal function parameters.
```
	paramTypes, genericTypes := ip.buildTrampolineType(before)
```

Additionally, the logic for handling typeParame could be extracted into a separate function. The final structure might look something like this:
```
func replaceTypeWithAny(traits []ParamTrait, paramTypes, genericTypes *dst.FieldList) {
	if len(paramTypes.List) != len(traits) {
		return ex.New("hook func signature can not match with target function")
	}

	for i, field := range paramTypes.List {
		trait := traits[i]
		if trait.IsInterfaceAny {
			// Hook explicitly uses interface{} for this parameter
			field.Type = ast.InterfaceType()
		} else {
			// Replace type parameters with interface{} (for linkname compatibility)
			field.Type = replaceTypeParamsWithAny(field.Type, genericTypes)
		}
	}
}

func (ip *InstrumentPhase) addHookFuncVar(t *rule.InstFuncRule,
	traits []ParamTrait, before bool,
) error {
	// Build trampoline function signature, including both concrete and generic types
	paramTypes, genericTypes := ip.buildTrampolineType(before)
	addHookContext(paramTypes)
	replaceTypeWithAny(traits, paramTypes, genericTypes)

	// Generate var decl and append it to the target file, note that many target
	// functions may match the same hook function, it's a fatal error to append
```

tool/internal/instrument/trampoline.go
y1yang0: To clarify, what I meant is that we should generate code like the following:

```go
func (c *HookContextImpl3580597051) SetParam/SetReturnVal(idx int, val interface{}) {
	panic("Unsupported method")
}
```

This shouldn't be too difficult, I think. Couldn't we just restore the original code and then add an if branch to handle this case?

tool/internal/instrument/trampoline.go
```diff
@@ -464,6 +465,9 @@ func (ip *InstrumentPhase) buildTrampolineTypes() {
 		}
 	}
 	addHookContext(afterHookFunc.Type.Params)
+	trampolineTypeParams := combineTypeParams(ip.targetFunc)
```
y1yang0: Ditto. This can be removed once buildTrampolineType returning two field list

tool/internal/ast/shared.go
```diff
@@ -154,3 +185,39 @@ func AddStructField(decl dst.Decl, name, t string) {
...
+	for _, field := range fieldList.List {
+		// Handle unnamed fields (e.g., embedded types) or fields with single/multiple names
+		namesToProcess := field.Names
+		if len(namesToProcess) == 0 {
+			// For unnamed fields, create one field with no names
+			namesToProcess = []*dst.Ident{nil}
+		}
+
+		for _, name := range namesToProcess {
+			clonedType, ok := dst.Clone(field.Type).(dst.Expr)
```
y1yang0: util.AssertType

tool/internal/ast/shared.go
```diff
@@ -154,3 +185,39 @@ func AddStructField(decl dst.Decl, name, t string) {
...
+			util.Assert(ok, "field.Type is not an Expr")
+
+			var names []*dst.Ident
+			if name != nil {
+				clonedName, okC := dst.Clone(name).(*dst.Ident)
```
y1yang0: util.AssertType

tool/internal/ast/shared.go
```diff
@@ -154,3 +185,39 @@ func AddStructField(decl dst.Decl, name, t string) {
...
+				util.Assert(okC, "name is not an Ident")
+				names = []*dst.Ident{clonedName}
+			}
+
+			newField := &dst.Field{
```
y1yang0, line 213: ast.Field

tool/internal/instrument/trampoline.go
txabman42: Added panics to SetParam/SetReturnVal.

Tested via e2e (basic_test.go) and integration test (golden file).

tool/internal/ast/primitives.go
txabman42: Moved

tool/internal/ast/primitives.go
txabman42: Updated

tool/internal/ast/shared.go
txabman42: Both sounds good, I let stripGenericTypes

tool/internal/instrument/trampoline.go
txabman42: I like this idea, refactored

tool/internal/instrument/trampoline.go
txabman42: Nice catch, removed

tool/internal/ast/shared.go
txabman42: Updated

tool/internal/ast/shared.go
txabman42: Updated

tool/internal/ast/shared.go
txabman42: Updated

tool/internal/instrument/trampoline.go
```diff
@@ -464,6 +475,8 @@ func (ip *InstrumentPhase) buildTrampolineTypes() {
 		}
 	}
 	addHookContext(afterHookFunc.Type.Params)
+	beforeHookFunc.Type.TypeParams = ast.CloneTypeParams(beforeGenericTypes)
+	afterHookFunc.Type.TypeParams = ast.CloneTypeParams(afterGenericTypes)
```
y1yang0: Why not:
```
beforeHookFunc.Type.Params, beforeHookFunc.Type.Params = ip.buildTrampolineType(true)
```
But instead, it's separated into:
```
beforeHookFunc.Type.Params, beforeGenericTypes = ip.buildTrampolineType(true)
beforeHookFunc.Type.TypeParams = ast.CloneTypeParams(beforeGenericTypes)
```
Couldn't buildTrampolineType just do the cloning internally?

tool/internal/instrument/trampoline.go
```diff
@@ -426,32 +431,38 @@ func (ip *InstrumentPhase) buildTrampolineType(before bool) *dst.FieldList {
...
+		splitResults := ast.SplitMultiNameFields(ip.targetFunc.Type.Results)
+		for _, field := range splitResults.List {
 			retField := util.AssertType[*dst.Field](dst.Clone(field))
 			renameField(retField, "arg")
-			paramList.List = append(paramList.List, retField)
+			paramTypes.List = append(paramTypes.List, retField)
 		}
 	}
-	return paramList
+	genericTypes := combineTypeParams(ip.targetFunc)
```
y1yang0, line 459:
```suggestion
    // Build type parameter list of trampoline function according to the target 
	// function's type parameters and receiver type parameters
	genericTypes := combineTypeParams(ip.targetFunc)
```

tool/internal/instrument/trampoline.go
```diff
@@ -629,6 +642,108 @@ func setReturnValClause(idx int, t dst.Expr) *dst.CaseClause {
...
+//	  ) { ... }
+func combineTypeParams(targetFunc *dst.FuncDecl) *dst.FieldList {
+	var trampolineTypeParams *dst.FieldList
+	if ast.HasReceiver(targetFunc) {
+		receiverTypeParams := extractReceiverTypeParams(targetFunc.Recv.List[0].Type)
+		if receiverTypeParams != nil {
+			trampolineTypeParams = receiverTypeParams
+		}
+	}
+	if targetFunc.Type.TypeParams != nil {
```
y1yang0: ast/internal/shared.go
```go
func IsGeneric(fn *dst.FuncDecl) bool {
	return fn.Type.TypeParams != nil
}
```
Replace all `targetFunc.Type.TypeParams != nil` with `ast.IsGeneric()`

tool/internal/instrument/trampoline.go
```diff
@@ -629,6 +642,108 @@ func setReturnValClause(idx int, t dst.Expr) *dst.CaseClause {
...
+			Value: replaceGenericInstantiations(tType.Value),
+		}
+	case *dst.IndexExpr:
+		// GenStruct[T] -> interface{}
+		return ast.InterfaceType()
+	case *dst.IndexListExpr:
+		// GenStruct[T, U] -> interface{}
+		return ast.InterfaceType()
+	}
+	return t
```
y1yang0: default: util.Unimplemented()? (i.e. channel)

y1yang0 (approved):
LGTM otherwise. I have some ideas for some harmless refactoring, I'll open a PR for it.

tool/internal/instrument/trampoline.go
txabman42: Added

tool/internal/instrument/trampoline.go
txabman42: Even better, I have just removed `replaceGenericInstantiations` and replaced with `replaceTypeParamsWithAny` (56d991576c77ca3ea55deddb115b934f2bc33e8a)

tool/internal/instrument/trampoline.go
txabman42: Added c57660b1c47a2b107a3183bb816f4a59f3b6bcad

Issue #186, opened 2025-11-20
Refactor repeated type assertions into a util.TypeAssert helper


y1yang0:
inspired from 

There's a lot of value, ok := ...; util.Assert(ok, ...) boilerplate. We should abstract this into a util.TypeAssert helper.

This would clean up functions like SelectorExpr and IndexExpr:
```
// Before
e, ok := dst.Clone(x).(dst.Expr)
util.Assert(ok, "x is not an Expr")

// After (suggestion)
e := util.TypeAssert[dst.Expr](dst.Clone(x))
```

PR #187, opened 2025-11-20
refactor: extract repeated type assertions into util.AssertType

y1yang0:
related to #186

### DIFF
--- a/.config/golangci.yml
+++ b/.config/golangci.yml
@@ -314,7 +314,8 @@ linters:
         - os.WriteFile
         - prometheus.ExponentialBuckets.*
         - prometheus.LinearBuckets
-        - util.Assert # assertion message already explains the magic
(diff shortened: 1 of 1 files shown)


tool/internal/instrument/toolexec.go
```diff
@@ -71,7 +71,6 @@ func stripCompleteFlag(args []string) []string {
 }
 
 func interceptCompile(ctx context.Context, args []string) ([]string, error) {
-	util.Assert(util.IsCompileCommand(strings.Join(args, " ")), "sanity check")
```
kakkoyun, line 74: This change seems unrelated?

y1yang0: Yes, but it was just a one-line change and it didn't seem to warrant a separate PR. So I just removed it in this patch.

PR #191, opened 2025-11-21
chore(tool): format generated debugging files

y1yang0:
This patch formats the generated debug code. It also includes fixes for a few minor issues.e.g.

- exercise all functionalities of HookContext
- use const var instead of string literal

before:
```
func Example() {
	go func() { println("RawCode") }()
	//line <generated>:1
	if false {
	} else {
		defer OtelAfterTrampoline_Example466439887(&HookContextImpl466439887{params: []interface {
		}{}, returnVals: []interface {
		}{}})
		if /* __TRAMPOLINE_JUMP_IF__ */ hookContext21567240, skip21567240 := OtelBeforeTrampoline_Example21567240(); skip21567240 {
			OtelAfterTrampoline_Example21567240(hookContext21567240)
			return
		} else {
			defer OtelAfterTrampoline_Example21567240(hookContext21567240)
		}
	}
	//line main.go:35:16
}
```
after:
```
func Example() {
	go func() { println("RawCode") }()
	//line <generated>:1
	if false {
	} else {
		defer OtelAfterTrampoline_Example466439887(&HookContextImpl466439887{params: []interface{}{}, returnVals: []interface{}{}})
		if hookContext21567240, skip21567240 := OtelBeforeTrampoline_Example21567240(); skip21567240 {
			OtelAfterTrampoline_Example21567240(hookContext21567240)
			return
		} else {
			defer OtelAfterTrampoline_Example21567240(hookContext21567240)
		}
	}
	//line main.go:35:16
}
```

### DIFF
--- a/tool/internal/instrument/optimize_test.go
+++ b/tool/internal/instrument/optimize_test.go
@@ -303,7 +303,7 @@ func TestStripTJumpLabel(t *testing.T) {
 			tjump := &TJump{ifStmt: ifStmt}
 
 			stripTJumpLabel(tjump)
-			assert.Empty(t, len(ifStmt.Decs.If))
+			assert.Empty(t, ifStmt.Decs.If)
 		})
 	}
 }


tool/internal/instrument/optimize.go
```diff
@@ -316,7 +316,7 @@ func flattenTJump(tjump *TJump, removedOnExit bool) error {
 
 func stripTJumpLabel(tjump *TJump) {
 	ifStmt := tjump.ifStmt
-	ifStmt.Decs.If = ifStmt.Decs.If[1:]
+	ifStmt.Decs.If = nil
```
kakkoyun, line 319: Hmm, this change seems out of place. What's the reason of this change?

y1yang0: This is a bug fix. The original implementation intended to remove the `/* __TRAMPOLINE_JUMP_IF__ */` comment but failed to do so. This patch corrects that behavior.

```
before:
if /* __TRAMPOLINE_JUMP_IF__ */ hookContext21567240, skip21567240 ...

after:
if hookContext21567240, skip21567240 ...
```

PR #207, opened 2025-11-25
chore(demo): Improve demo convinience target

kakkoyun:

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -70,10 +70,11 @@ jobs:
         with:
           go-version-file: go.mod
           cache-dependency-path: "**/go.mod"
+      - name: Install protoc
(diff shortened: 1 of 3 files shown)


kakkoyun:
> @kakkoyun I invoked `make start http`, everything seems to be in-place and I would expect to see some traces on `jaeger` or `grafana`, but don't see them. Is my expectation correct?

You should be able to see things from prometheus, jaeger, etc. `make -C demo/infrastructure/docker-compose quickstart` best way to invoke.

You would not see anything from demo HTTP app until we merge #156

pdelewski:
> > @kakkoyun I invoked `make start http`, everything seems to be in-place and I would expect to see some traces on `jaeger` or `grafana`, but don't see them. Is my expectation correct?
> 
> You should be able to see things from prometheus, jaeger, etc. `make -C demo/infrastructure/docker-compose quickstart` best way to invoke.
> 
> You would not see anything from demo HTTP app until we merge #156

Ok, thank you, you are right, clear!

PR #208, opened 2025-11-26
refactor: consolidate trampoline signature creation

y1yang0:
consolidate trampoline signature creation

simplify trampoline generation overall

### DIFF
--- a/.github/workflows/check-license-headers.yaml
+++ b/.github/workflows/check-license-headers.yaml
@@ -23,7 +23,7 @@ jobs:
       code: ${{ steps.filter.outputs.code }}
     steps:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3  # v6.0.0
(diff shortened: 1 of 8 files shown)


tool/internal/instrument/testdata/golden/generic-functions/generic_functions.main.go.golden
```diff
@@ -202,29 +190,15 @@ func (c *HookContextImpl3378413897) HasKeyData(key string) bool {
 }
 
 func (c *HookContextImpl3378413897) GetParam(idx int) interface{} {
```
txabman42, line 192: Why are we unsupporting this?

y1yang0: The c.params field stores pointers to pointers. Given this, it's better to fail completely rather than return partially correct and partially incorrect data.

For example, if c.params[1] is bad but *(c.params[2].(*string)) is good, the inconsistent result could lead to subtle and hard-to-debug issues. A complete failure is safer and more predictable.

Issue #196, opened 2025-11-23
Support non-root building path


txabman42:
Coming from https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/120#issuecomment-3545920119

Let's move `demo/basic/main.go` to `demo/basic/cmd/main.go` and execute `go build -a cmd/` from `basic` folder. This would work without problem, however, when instrumenting it with otel (`../../otel go build -a cmd/`), it fails:

<details>
<summary>Error logs</summary>

```bash
# cmd/trace
net/http.OtelBeforeTrampoline_ServeHTTP802697909: relocation target github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/nethttp.BeforeServeHTTP not defined
Error:
[0] failed to run command "/usr/local/go/pkg/tool/darwin_arm64/link" with args: [-o $WORK/b661/exe/a.out -importcfg $WORK/b661/importcfg.link -X=cmd/internal/objabi.buildID=d3PxNBJKoXj5HhChQTjx/PqTyxeY2x3r8w-sZE2K0/JDzRK1jzYRrptT4B3Kcy/d3PxNBJKoXj5HhChQTjx -buildmode=pie -buildid=d3PxNBJKoXj5HhChQTjx/PqTyxeY2x3r8w-sZE2K0/JDzRK1jzYRrptT4B3Kcy/d3PxNBJKoXj5HhChQTjx -extld=clang $WORK/b661/_pkg_.a]: exit status 2
```

</details>

Same would happen for https://github.com/VictoriaMetrics/VictoriaMetrics --> `otel go build -a ./app/vmctl`

This is mainly because `otel.runtime.go` is not generated in the target folder, but in the root one. Linker wont be able to resolve linkname directives.

It could be also interesting to test the instrumentation of a library.

PR #209, opened 2025-11-27
feat: support non-root build paths

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/196

Support building non-root paths e.g. `otel go build -a ./app/vmctl`


tool/util/go.go
```diff
@@ -89,3 +90,50 @@ func SplitCompileCmds(input string) []string {
...
+}
+
+// GetBuildTarget extracts the build target path from the go build command.
+// For example:
+//   - "go build -a cmd/" returns "cmd"
+//   - "go build -a ./app/vmctl" returns "app/vmctl"
+//   - "go build -a ." returns ""
+//   - "go build" returns ""
+//   - "go build -o ./bin/" returns ""
+func GetBuildTarget(goBuildCmd []string) string {
```
y1yang0: The go build command supports numerous patterns, which might make exhaustively handling all of them a complex and difficult-to-implement task?

For example, it accepts patterns like:
```
go build cmd
go build std
go build main
go build ./abc/...
go build ./a/. ./b/.
```
Instead of trying to parse these patterns ourselves, perhaps we could leverage golang.org/x/tools/go/packages.Load(). We could pass the patterns (e.g., "cmd", "std", "./abc/...") directly to this function. The packages.Load() function can resolve these patterns and identify the directory D where the go.mod file resides. We could then use this directory D as the working directory for our compilation process?

txabman42: Thats great idea!!!

Indeed, I looked for something similar but I was only aware of the internal https://pkg.go.dev/cmd/go/internal/load. I didn't know there was that public Load function, although it has sense to exist one.

I'll work on that. TY

txabman42: @y1yang0 Added packages.Load for resolving patterns. I want to keep this PR simple.

My idea is to create another PR with: 
- Replace the hardcoded data like compiling dir, go.mod, package name, etc. with the package info.
- https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/210 as we already will be depending on the package info

tool/internal/setup/setup.go
```diff
@@ -51,6 +52,14 @@ func Setup(ctx context.Context, args []string) error {
 	sp := &SetupPhase{
 		logger: logger,
 	}
+
+	// Introduce additional hook code by generating otel.instrumentation.go
+	// Use GetPackage to determine the build target directory
+	pkgs, err := util.GetBuildPackages(args)
```
y1yang0: I guess the information returned by GetBuildPackages can be used throughout the entire setup phase. Maybe we can do something like this?

```
func (sp *SetupPhase) getBuildPackages(args []string) error {
    ...
    sp.buildPkgs = buildPkgs
}
```

tool/internal/setup/setup.go
```diff
@@ -61,10 +70,11 @@ func Setup(ctx context.Context, args []string) error {
...
 	}
-	// Introduce additional hook code by generating otel.instrumentation.go
-	err = sp.addDeps(matched)
-	if err != nil {
-		return err
+	for _, pkg := range pkgs {
+		// Introduce additional hook code by generating otel.instrumentation.go
+		if err = sp.addDeps(matched, pkg.Dir); err != nil {
+			return err
+		}
```
y1yang0, line 162: Using the above approach, no changes are needed here either.

tool/internal/setup/setup.go
```diff
@@ -77,11 +87,7 @@ func Setup(ctx context.Context, args []string) error {
 		return err
 	}
 	// Write the matched hook to matched.txt for further instrument phase
-	err = sp.store(matched)
-	if err != nil {
-		return err
-	}
-	return nil
+	return sp.store(matched)
```
y1yang0, line 175: Good catch!

tool/internal/setup/setup.go
txabman42: The problem of adding `pkgs` to `sp` and reuse them inside methods like `addDeps` is that we would need to loop over pkgs in only some of the methods, like `addDeps` or `syncDeps`. However, there are other steps independent to the number of pkgs.

My idea is to just use pkgs inside `func Setup(ctx context.Context, args []string) error` so we make setup steps agnostic to the number of pkgs + make more visible which are the steps done per pkg and which not.

Just thinking about the next implementation for supporting multi pkg build.

tool/internal/setup/testdata/multiple_rule_sets.otel.runtime.go.golden
```diff
@@ -0,0 +1,21 @@
...
+var _getstatck0 = _otel_debug.Stack
+
+//go:linkname _printstack0 github.com/example/pkg1.OtelPrintStackImpl
+var _printstack0 = func(bt []byte) { _otel_log.Printf(string(bt)) }
+
+//go:linkname _getstatck1 github.com/example/pkg2.OtelGetStackImpl
+var _getstatck1 = _otel_debug.Stack
+
+//go:linkname _printstack1 github.com/example/pkg2.OtelPrintStackImpl
+var _printstack1 = func(bt []byte) { _otel_log.Printf(string(bt)) }
```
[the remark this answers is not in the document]
txabman42: otel.runtime.go OOS. This is the autogenerated golden file

tool/internal/setup/testdata/single_func_rule.otel.runtime.go.golden
```diff
@@ -0,0 +1,13 @@
...
+import _ "github.com/example/pkg"
+import _otel_log "log"
+import _otel_debug "runtime/debug"
+import _ "unsafe"
+
+//go:linkname _getstatck0 github.com/example/pkg.OtelGetStackImpl
+var _getstatck0 = _otel_debug.Stack
+
+//go:linkname _printstack0 github.com/example/pkg.OtelPrintStackImpl
+var _printstack0 = func(bt []byte) { _otel_log.Printf(string(bt)) }
```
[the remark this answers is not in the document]
txabman42: otel.runtime.go OOS. This is the autogenerated golden file

tool/internal/setup/setup.go
```diff
@@ -51,6 +52,14 @@ func Setup(ctx context.Context, args []string) error {
 	sp := &SetupPhase{
 		logger: logger,
 	}
+
+	// Introduce additional hook code by generating otel.instrumentation.go
+	// Use GetPackage to determine the build target directory
```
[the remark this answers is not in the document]
txabman42: Lets keep GetPackage for now as its only supported one package. I'll update this in a next PR supporting multi target builds

tool/util/go.go
```diff
@@ -89,3 +91,70 @@ func SplitCompileCmds(input string) []string {
...
+}
+
+// GetBuildPackages loads all packages from the go build command arguments.
+// Returns a list of loaded packages. If no package patterns are found in args,
+// defaults to loading the current directory package.
+// The args parameter should be the go build command arguments (e.g., ["build", "-a", "./cmd"]).
+// Returns an error if package loading fails or if invalid patterns are provided.
+// For example:
+//   - args ["build", "-a", "cmd/"] returns packages for "./cmd/"
+//   - args ["build", "-a", "./app/vmctl"] returns packages for "./app/vmctl"
```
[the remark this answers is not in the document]
txabman42: 08c44c33484383e7676542002006ce4a87f4f2be

tool/internal/setup/setup.go
```diff
@@ -61,10 +70,11 @@ func Setup(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
-	// Introduce additional hook code by generating otel.instrumentation.go
-	err = sp.addDeps(matched)
-	if err != nil {
-		return err
+	for _, pkg := range pkgs {
+		// Introduce additional hook code by generating otel.instrumentation.go
```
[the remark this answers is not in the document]
txabman42: 08c44c33484383e7676542002006ce4a87f4f2be

tool/util/go_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+func extractPackageIDs(pkgs []*packages.Package) []string {
+	ids := make([]string, len(pkgs))
+	for i, pkg := range pkgs {
+		ids[i] = pkg.ID
+	}
+	return ids
+}
+
+// checkPackages returns a function that verifies all expected strings are found in the packages.
```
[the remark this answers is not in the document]
txabman42: 08c44c33484383e7676542002006ce4a87f4f2be

tool/internal/setup/setup.go
```diff
@@ -128,16 +134,20 @@ func GoBuild(ctx context.Context, args []string) error {
 		logger.DebugContext(ctx, "failed to back up files", "error", err)
 	}
 	defer func() {
-		err = os.RemoveAll(OtelRuntimeFile)
+		var pkgs []*packages.Package
+		pkgs, err = util.GetBuildPackages(os.Args[1:])
```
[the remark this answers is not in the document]
txabman42: 08c44c33484383e7676542002006ce4a87f4f2be

tool/internal/setup/setup.go
```diff
@@ -128,16 +134,20 @@ func GoBuild(ctx context.Context, args []string) error {
...
 		if err != nil {
-			logger.DebugContext(ctx, "failed to remove otel runtime file", "error", err)
+			logger.DebugContext(ctx, "failed to get build packages", "error", err)
 		}
-		err = os.RemoveAll(unzippedPkgDir)
-		if err != nil {
+		for _, pkg := range pkgs {
+			if err = os.RemoveAll(filepath.Join(pkg.Dir, OtelRuntimeFile)); err != nil {
+				logger.DebugContext(ctx, "failed to remove package", "path", pkg.PkgPath, "error", err)
```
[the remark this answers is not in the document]
txabman42: 08c44c33484383e7676542002006ce4a87f4f2be

tool/internal/setup/setup.go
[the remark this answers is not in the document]
txabman42: 08c44c33484383e7676542002006ce4a87f4f2be

tool/util/go.go
```diff
@@ -89,3 +91,70 @@ func SplitCompileCmds(input string) []string {
...
+			break
+		}
+
+		// If we hit a flag, stop. Packages come after all flags
+		// go build [-o output] [build flags] [packages]
+		if strings.HasPrefix(arg, "-") || arg == "go" || arg == "build" || arg == "install" {
+			break
+		}
+
+		pkgs, err := packages.Load(cfg, arg)
```
y1yang0: When running `go build -tags pkg main`, we might incorrectly load numerous packages. For instance, we could accidentally load the 'pkg' package.

tool/util/go.go
```diff
@@ -89,3 +91,70 @@ func SplitCompileCmds(input string) []string {
...
+		if err != nil {
+			return nil, ex.Wrapf(err, "failed to load packages for pattern %s", arg)
+		}
+		for _, pkg := range pkgs {
+			if pkg.Errors != nil || pkg.Module == nil {
+				continue
```
y1yang0: sp.DebugContext(pkg.errors)

tool/util/go.go
```diff
@@ -89,3 +91,70 @@ func SplitCompileCmds(input string) []string {
...
+// From: go help build
+//
+//nolint:gochecknoglobals // private lookup table
+var flagsWithPathValues = map[string]bool{
+	"-o":       true,
+	"-modfile": true,
+	"-overlay": true,
+	"-pgo":     true,
+	"-pkgdir":  true,
+}
```
y1yang0: The packages.Load function doesn't strictly require its arguments to be paths; they can also be patterns like "std" or "main". Therefore, we need to list all flags that take a value, not just those that take a path value.

tool/util/go.go
```diff
@@ -89,3 +91,70 @@ func SplitCompileCmds(input string) []string {
...
+// Returns a list of loaded packages. If no package patterns are found in args,
+// defaults to loading the current directory package.
+// The args parameter should be the go build command arguments (e.g., ["build", "-a", "./cmd"]).
+// Returns an error if package loading fails or if invalid patterns are provided.
+// For example:
+//   - args ["build", "-a", "./cmd"] returns packages for "./cmd"
+//   - args ["build", "-a", "cmd"] returns packages for the "cmd" package in the module
+//   - args ["build", "-a", ".", "./cmd"] returns packages for both "." and "./cmd"
+//   - args ["build"] returns packages for "."
+func GetBuildPackages(args []string) ([]*packages.Package, error) {
```
y1yang0: Move to setup package

y1yang0 (approved):
Otherwise LGTM

Issue #211, opened 2025-11-28
Unable to inject package with CGO code


y1yang0:
We have observed that our code injection process fails when applied to Go packages that utilize cgo. For example, consider the following compile command from a build-plan.log:

```
/-/go1.21/pkg/tool/darwin_arm64/compile -o $WORK/b043/_pkg_.a -trimpath "$WORK/b043=>" -p github.com/confluentinc/confluent-kafka-go/v2/kafka -lang=go1.14 -buildid 1DpLgbq5p2L3yK3jL3tE/1DpLgbq5p2L3yK3jL3tE -goversion go1.21.0 -c=4 -shared -nolocalimports -importcfg $WORK/b043/importcfg -pack /-/github.com/confluentinc/confluent-kafka-go/v2@v2.0.2/kafka/context.go $WORK/b043/_cgo_gotypes.go $WORK/b043/00version.cgo1.go $WORK/b043/adminapi.cgo1.go $WORK/b043/adminoptions.cgo1.go $WORK/b043/build_darwin_arm64.cgo1.go $WORK/b043/config.cgo1.go $WORK/b043/consumer.cgo1.go $WORK/b043/error.cgo1.go $WORK/b043/error_gen.cgo1.go $WORK/b043/event.cgo1.go $WORK/b043/generated_errors.cgo1.go $WORK/b043/handle.cgo1.go $WORK/b043/header.cgo1.go $WORK/b043/kafka.cgo1.go $WORK/b043/log.cgo1.go $WORK/b043/message.cgo1.go $WORK/b043/metadata.cgo1.go $WORK/b043/misc.cgo1.go $WORK/b043/mockcluster.cgo1.go $WORK/b043/offset.cgo1.go $WORK/b043/producer.cgo1.go $WORK/b043/testhelpers.cgo1.go $WORK/b043/time.cgo1.go $WORK/b043/_cgo_import.go
```
Because the source file consumer.go from the confluent-kafka-go package contains cgo code, the compile command does not target consumer.go directly. Instead, it targets consumer.cgo1.go, a file that is dynamically generated at compile time. This prevents our tooling from finding the correct source file during the "setup" phase.

A potential solution involves parsing the build-plan.log to identify the cgo tool invocation command, as shown below:
```
#
# github.com/confluentinc/confluent-kafka-go/v2/kafka
#

mkdir -p $WORK/b043/
[...]

PR #216, opened 2025-12-04
feat: support cgo

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/211

### DIFF
--- a/.github/workflows/test-unit.yaml
+++ b/.github/workflows/test-unit.yaml
@@ -18,6 +18,8 @@ jobs:
       pull-requests: read
     outputs:
       code: ${{ steps.filter.outputs.code }}
+      tool: ${{ steps.filter.outputs.tool }}
+      pkg: ${{ steps.filter.outputs.pkg }}
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
       - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36  # v3.0.2
@@ -33,6 +35,10 @@ jobs:
               - 'pkg/**'
               - 'test/**'
               - '.github/workflows/test-unit.yaml'
+            tool:
+              - 'tool/**'
+            pkg:
+              - 'pkg/**'
 
   test-unit-coverage:
     name: Coverage Unit Tests
@@ -48,8 +54,18 @@ jobs:
           go-version-file: go.mod
       - name: Run unit test coverage
         run: make test-unit/coverage
-      - name: Upload coverage report
+      - name: Upload tool coverage report
+        if: needs.changes.outputs.tool == 'true'
         uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7  # v5.5.1
+        with:
+          files: coverage-tool.txt
+          flags: tool
+      - name: Upload pkg coverage report
+        if: needs.changes.outputs.pkg == 'true'
+        uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7  # v5.5.1
+        with:
+          files: coverage-pkg.txt
+          flags: pkg
 
   test-unit:
     name: Unit Tests (go ${{ matrix.platform.os }} ${{ matrix.platform.arch }})
--- a/Makefile
+++ b/Makefile
@@ -261,14 +261,20 @@ test-unit/tool/coverage: package gotestfmt ## Run unit tests with coverage for t
 test-unit/pkg/coverage: package gotestfmt ## Run unit tests with coverage for pkg modules only
 	@echo "Running pkg unit tests with coverage..."
 	set -euo pipefail
-	@PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | grep -v "pkg/instrumentation/runtime" | grep -v "pkg/instrumentation/nethttp/semconv"); \
+	ROOT_DIR=$$(pwd); \
+	PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | grep -v "pkg/instrumentation/runtime" | grep -v "pkg/instrumentation/nethttp/semconv"); \
 	for moddir in $$PKG_MODULES; do \
 		echo "Testing $$moddir..."; \
-		(cd $$moddir && go mod tidy && go test -json -v -shuffle=on -timeout=5m -count=1 ./... -coverprofile=coverage.txt -covermode=atomic 2>&1 | tee -a ../../gotest-unit-pkg.log | gotestfmt); \
-	done
-	@echo "Running pkg unit tests with coverage report (semconv only - hook tests require full instrumentation)..."
-	set -euo pipefail
-	cd pkg/instrumentation/nethttp/semconv && go test -json -v -shuffle=on -timeout=5m -count=1 ./... -coverprofile=coverage.txt -covermode=atomic 2>&1 | tee ../../../gotest-unit-pkg.log | gotestfmt
+		(cd $$moddir && go mod tidy && go test -json -v -shuffle=on -timeout=5m -count=1 ./... -coverprofile=coverage-pkg.txt -covermode=atomic 2>&1 | tee -a ../../gotest-unit-pkg.log | gotestfmt); \
+	done; \
+	echo "Running pkg unit tests with coverage report (semconv only - hook tests require full instrumentation)..."; \
+	cd pkg/instrumentation/nethttp/semconv && go test -json -v -shuffle=on -timeout=5m -count=1 ./... -coverprofile=coverage-pkg.txt -covermode=atomic 2>&1 | tee ../../../gotest-unit-pkg.log | gotestfmt; \
+	cd "$$ROOT_DIR"; \
+	echo "Merging coverage files into coverage-pkg.txt..."; \
+	echo "mode: atomic" > coverage-pkg.txt; \
+	find pkg -name "coverage-pkg.txt" -exec grep -h -v "^mode:" {} \; >> coverage-pkg.txt; \
+	find pkg -name "coverage-pkg.txt" -delete; \
+	echo "Coverage merged into coverage-pkg.txt"
 
 .ONESHELL:
 test-integration: go-protobuf-plugins ## Run integration tests
@@ -282,7 +288,7 @@ test-integration/coverage: ## Run integration tests with coverage report
 test-integration/coverage: build build-demo gotestfmt
 	@echo "Running integration tests with coverage report..."
 	set -euo pipefail
-	go test -json -v -shuffle=on -timeout=10m -count=1 -tags integration ./test/integration/... -coverprofile=coverage.txt -covermode=atomic 2>&1 | tee ./gotest-integration.log | gotestfmt
+	go test -json -v -shuffle=on -timeout=10m -count=1 -tags integration ./test/integration/... -coverprofile=coverage-integration.txt -covermode=atomic 2>&1 | tee ./gotest-integration.log | gotestfmt
 
 .ONESHELL:
 test-e2e: ## Run e2e tests
@@ -296,7 +302,7 @@ test-e2e/coverage: ## Run e2e tests with coverage report
 test-e2e/coverage: build gotestfmt
 	@echo "Running e2e tests with coverage report..."
 	set -euo pipefail
-	go test -json -v -shuffle=on -timeout=10m -count=1 -tags e2e ./test/e2e/... -coverprofile=coverage.txt -covermode=atomic 2>&1 | tee ./gotest-e2e.log | gotestfmt
+	go test -json -v -shuffle=on -timeout=10m -count=1 -tags e2e ./test/e2e/... -coverprofile=coverage-e2e.txt -covermode=atomic 2>&1 | tee ./gotest-e2e.log | gotestfmt
 
 ##@ Utilities
 


tool/internal/rule/base.go
```diff
@@ -107,6 +109,11 @@ func (irs *InstRuleSet) SetPackageName(name string) {
 	irs.PackageName = name
 }
 
+// SetCgoFileMap sets the CGO file mapping for this rule set.
+func (irs *InstRuleSet) SetCgoFileMap(cgoFiles map[string]string) {
```
kakkoyun, line 113: nit: ```suggestion
func (irs *InstRuleSet) SetCGoFileMap(cgoFiles map[string]string) {
```

Any maybe sync all the usages accordingly. That being said, I wonder how it is used in the standard library, it would be nice to check. Naming things is hard.

File: tool/internal/setup/find.go [hunk 2]
```diff
@@ -30,51 +32,83 @@ func (d *Dependency) String() string {
...
+}
 
-	// Filter compile commands from build plan log
-	compileCmds := make([]string, 0)
-	scanner := bufio.NewScanner(buildPlanLog)
-	// Seek to the beginning of the file before reading
-	_, err := buildPlanLog.Seek(0, 0)
+// parseBuildPlan scans the build plan log once and extracts both compile commands
+// and CGO object directory mappings.
+func parseBuildPlan(buildPlanLog *os.File) ([]string, map[string]string, error) {
```
kakkoyun: I think it would be better to introduce a struct type for this.
```suggestion
type buildPlan struct {
   TBD []string
   TBD map[string]string
}
func parseBuildPlan(buildPlanLog *os.File) (buildPlan, error) {
```

tool/internal/setup/find.go
```diff
@@ -30,51 +32,83 @@ func (d *Dependency) String() string {
...
-		return nil, ex.Wrapf(err, "failed to seek to beginning of build plan log")
+		return nil, nil, err
 	}
-	// 10MB should be enough to accommodate most long line
-	buffer := make([]byte, 0, buildPlanBufSize)
-	scanner.Buffer(buffer, cap(buffer))
+
+	var compileCmds []string
+	cgoDirsMap := make(map[string]string)
+	var currentDir string
```
kakkoyun: nitpick: ```suggestion
	var (
		compileCmds []string
		cgoDirsMap = make(map[string]string)
		currentDir string
	)
```

File: tool/internal/rule/base.go (hunk 1 above)
Comment (txabman42) [reply tool/internal/rule/base.go:113]: In this case, seems that standard library uses Cgo instead of CGo ([example](https://cs.opensource.google/go/go/+/master:src/cmd/cgo/gcc.go;drc=a593ca9d657efb1ea021c9fc51cb528c398bbf4e;l=82)). But yes, naming is always the hardest thing.

File: tool/internal/setup/find.go (hunk 2 above)
Comment (txabman42) [reply tool/internal/setup/find.go]: Good idea f877c2528421d416ed840573db4cdee0a05f132e

```go
type parsedBuildPlan struct {
	CompileCmds []string
	CgoObjDirs  map[string]string
}
```

tool/internal/setup/find.go
txabman42: f877c2528421d416ed840573db4cdee0a05f132e

tool/internal/setup/find.go
```diff
@@ -30,51 +37,88 @@ func (d *Dependency) String() string {
 	return fmt.Sprintf("{%s@%s: %v}", d.ImportPath, d.Version, d.Sources)
 }
 
-// findCompileCommands finds the compile commands from the build plan log.
-func findCompileCommands(buildPlanLog *os.File) ([]string, error) {
-	const buildPlanBufSize = 10 * 1024 * 1024 // 10MB buffer size
+// parseCdDir extracts the directory path from a "cd" command line (case-insensitive).
+func parseCdDir(line string) (string, bool) {
+	if !strings.HasPrefix(strings.ToLower(line), "cd ") {
```
y1yang0: "cd" instead of "cd "

tool/util/sys.go
```diff
@@ -122,3 +122,7 @@ func PathExists(path string) bool {
 	_, err := os.Stat(path)
 	return err == nil
 }
+
+func NormalizePath(path string) string {
+	return strings.TrimSuffix(filepath.ToSlash(path), "/")
```
y1yang0: filepath.ToSlash(filepath.Clean(path))

tool/util/go.go
```diff
@@ -89,3 +92,37 @@ func SplitCompileCmds(input string) []string {
...
+}
+
+const (
+	cgoSuffix = ".cgo1.go"
+	goSuffix  = ".go"
+)
+
+// ResolveCgoFile maps a CGO-generated file back to its original source
+// in the specified source directory. Both cgoFile and sourceDir must be non-empty.
+func ResolveCgoFile(cgoFile, sourceDir string) (string, error) {
```
y1yang0: Please move to find.go as it is a non-reusable implementation detail of the 'find' functionality.

tool/internal/setup/find.go
```diff
@@ -152,9 +198,24 @@ func (sp *SetupPhase) findDeps(ctx context.Context, goBuildCmd []string) ([]*Dep
...
+				}
+				originalAbsFile, err1 := util.ResolveCgoFile(arg, sourceDir)
+				if err1 != nil {
+					// Skip non-CGO generated files (_cgo_gotypes.go, _cgo_import.go, ...)
+					sp.Debug("Skip generated file", "file", arg, "error", err1)
+					continue
+				}
+				dep.CgoFiles[originalAbsFile] = filepath.Base(arg)
+				dep.Sources = append(dep.Sources, originalAbsFile)
+				sp.Info("Resolved CGO source", "cgo", arg, "original", originalAbsFile)
```
y1yang0: The current implementation feels a bit odd. We identify the compile command list and objDir in parseBuildPlan, but the source files aren't linked. We could just identify the objDir and associate the source files immediately within findDeps, couldn't we? This would be cleaner, findCompileCommands could then focus only on finding the compile command list, and the changes would be limited to findDeps.

tool/util/sys.go
txabman42: 6572a55fa315bec907a1b70ab9ae82cf4b89c91a

tool/util/go.go
txabman42: 6572a55fa315bec907a1b70ab9ae82cf4b89c91a

tool/internal/setup/find.go
txabman42: I dont get why it feels odd. We just parseBuildPlan to get the required data (CompileCmds and CgoObjDirs) and use it to decide if an arg should be linked or not.

> This would be cleaner, findCompileCommands could then focus only on finding the compile command list

The problem is that findCompileCommands is scanning the file, so we would be scanning twice the file. That was the main reason for the small refactor.

y1yang0: I'm proposing a change to how we handle parsing. Currently, the parsing logic is fragmented: parseBuildPlan extracts cgodir, while findDeps is responsible for extracting the full cgodir:cgofiles+gofiles.

A better approach would be to let findCompileCommand remain as is—its only job should be to produce the raw compile command. We can then consolidate the parsing logic fatores cgodir:cgofiles+gofiles entirely within the findDeps function, like in the example below:

```
func (sp *SetupPhase) findDeps(ctx context.Context, goBuildCmd []string) ([]*Dependency, error) {
	...
	deps := make([]*Dependency, 0)
	for _, plan := range buildPlan {
		args := util.SplitCompileCmds(plan)
		if util.IsCompileCommand(plan) {
			dep, err := findGoSources(args)
			if err != nil {
				return nil, err
			}
			deps = append(deps, dep)
			sp.Info("Found go dependency", "dep", dep)
		} else if util.IsCgoCommand(plan) {
			dep, err := findGoSources(args)
			...
		} else {
			// Skip other commands like asm as we are not interested in them
			continue
		}
	}
	return deps, nil
}

```
This approach isolates the changes to just the findDeps function. Simple and intuitive.

tool/internal/instrument/instrument.go
```diff
@@ -71,14 +79,12 @@ func (ip *InstrumentPhase) instrument(rset *rule.InstRuleSet) error {
 		}
 		// Since trampoline-jump-if is performance-critical, perform AST level
 		// optimization for them before writing to file
-		err = ip.optimizeTJumps()
-		if err != nil {
+		if err = ip.optimizeTJumps(); err != nil {
```
y1yang0, line 130: I don't have a strong preference, but given that the project currently uses the ["happy path"](https://go.dev/wiki/CodeReviewComments#indent-error-flow) idiom, I'd suggest we follow that here for consistency. (The alternative would be to have a single patch dedicated to changing the style, but we should avoid mixing styles)

txabman42: Is not this following happy path idiom?

```go
		if err = ip.optimizeTJumps(); err != nil {
			return err
		}
```

tool/internal/setup/find.go
txabman42: ty, now I get your point. Good one tbh.

I've updated it, now:

```go
func (sp *SetupPhase) findDeps(ctx context.Context, goBuildCmd []string) ([]*Dependency, error) {
	buildPlan, err := sp.listBuildPlan(ctx, goBuildCmd)
	...
	for _, cmd := range buildPlan {
		currentDir = parseCdDir(cmd)
		
		if util.IsCompileCommand(cmd) {
			args := util.SplitCompileCmds(cmd)
			dep := findGoSources(sp, args, cgoObjDirs)
			deps = append(deps, dep)
		} else if util.IsCgoCommand(cmd) && currentDir != "" {
			args := util.SplitCompileCmds(cmd)
			if util.FindFlagValue(args, "-objdir") {
				cgoObjDirs[util.NormalizePath(objDir)] = currentDir
			}
		}
	}
	return deps, nil
}
```

tool/internal/instrument/instrument.go
kakkoyun: It is!

y1yang0: > Try to keep the normal code path at a minimal indentation
> If the if statement has an initialization statement, such as:
> 
> if x, err := f(); err != nil {
>     // error handling
>     return
> } else {
>     // use x
> }
> then this may require moving the short variable declaration to its own line:
> 
> x, err := f()
> if err != nil {
>     // error handling
>     return
> }
> // use x

tool/util/go.go
```diff
@@ -89,3 +99,12 @@ func SplitCompileCmds(input string) []string {
 func IsGoFile(path string) bool {
 	return strings.HasSuffix(strings.ToLower(path), ".go")
 }
+
+func NewFileScanner(file *os.File, size int) (*bufio.Scanner, error) {
+	if _, err := file.Seek(0, 0); err != nil {
+		return nil, ex.Wrapf(err, "failed to seek to beginning of build plan log")
```
y1yang0: "faild to seek file balah balah" instead of "build plan log"

tool/internal/setup/find.go
```diff
@@ -30,37 +32,44 @@ func (d *Dependency) String() string {
 	return fmt.Sprintf("{%s@%s: %v}", d.ImportPath, d.Version, d.Sources)
 }
 
-// findCompileCommands finds the compile commands from the build plan log.
-func findCompileCommands(buildPlanLog *os.File) ([]string, error) {
-	const buildPlanBufSize = 10 * 1024 * 1024 // 10MB buffer size
+// parseCdDir extracts the directory path from a "cd" command line.
+func parseCdDir(line string) (string, bool) {
+	if !strings.HasPrefix(strings.ToLower(line), "cd ") {
```
y1yang0: "cd" instead of "cd "

tool/internal/setup/find.go
```diff
@@ -105,71 +114,122 @@ func (sp *SetupPhase) listBuildPlan(ctx context.Context, goBuildCmd []string) ([
...
+			dep.Sources = append(dep.Sources, originalAbsFile)
+			sp.Info("Resolved CGO source", "cgo", arg, "original", originalAbsFile)
+			continue
 		}
-		// Extract the version from the source file path if available
-		if len(dep.Sources) > 0 {
-			dep.Version = findModVersion(dep.Sources[0])
+
+		// This is a Go source file, add it to the dependency sources
+		abs, _ := filepath.Abs(arg)
```
y1yang0: Why we ignore error

tool/internal/setup/find.go
```diff
@@ -105,71 +114,122 @@ func (sp *SetupPhase) listBuildPlan(ctx context.Context, goBuildCmd []string) ([
...
+			args := util.SplitCompileCmds(cmd)
+			dep := findGoSources(sp, args, cgoObjDirs)
+			deps = append(deps, dep)
+			sp.Info("Found dependency", "dep", dep)
+		} else if util.IsCgoCommand(cmd) && currentDir != "" {
+			args := util.SplitCompileCmds(cmd)
+			if objDir := util.FindFlagValue(args, "-objdir"); objDir != "" {
+				cgoObjDirs[util.NormalizePath(objDir)] = currentDir
+				sp.Debug("Found CGO objdir mapping", "objDir", objDir, "sourceDir", currentDir)
+			}
```
y1yang0: Perhaps we can use an assertion here instead of a conditional check, since -objdir $X is expected to be a fixed pattern in the command.
```
args := util.SplitCompileCmds(cmd)
objDir := util.FindFlagValue(args, "-objdir")
util.Assert(objDir != "", "sanity check")
cgoObjDirs[util.NormalizePath(objDir)] = currentDir
sp.Debug("Found CGO objdir mapping", "objDir", objDir, "sourceDir", currentDir)
```

tool/internal/instrument/instrument.go
```diff
@@ -11,23 +11,11 @@ import (
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
 )
 
-func groupRules(rset *rule.InstRuleSet) map[string][]rule.InstRule {
+func (ip *InstrumentPhase) groupRules(rset *rule.InstRuleSet) map[string][]rule.InstRule {
```
y1yang0: Maybe groupRules(workdir, reset), then we dont have to new InstrumentPhase instance in test

tool/util/go.go
txabman42: 09f43fe25e49cc153fcf23ce5d922ea5589960f8

tool/internal/setup/find.go
txabman42: we shouldn't, good catch, updated 09f43fe25e49cc153fcf23ce5d922ea5589960f8

tool/internal/setup/find.go
txabman42: agree 09f43fe25e49cc153fcf23ce5d922ea5589960f8

tool/internal/instrument/instrument.go
txabman42: IMO that refers to the case when you initialize `x` and `err`. In that case, don't do if + else, but just move the short variable declaration to its own line.

However, if you only have err := ... it should be fine as you dont have if + else pattern to avoid.

Issue #133, opened 2025-11-06
Create top-level `contrib` directory to aggregate the `pkg/instrumentation/*`


y1yang0:

y1yang0:
Thanks for your interest in this issue! It has now been assigned to you. Please be aware that the task can be quite challenging. We're looking forward to your contribution/pull request, and feel free to ask any questions in the discussion.

tushar1977:
@y1yang0 could you describe this issue in more detailed way?

kakkoyun:
> you mean keeping pkg/instrumentation/contrib also?

No, having `/instrumentation` and `/instrumentation/contrib`. For example, net/http and grpc instrumentation we try to add can go to `instrumentation/http` and `instrumentation/grpc` but in the feature we can have `instrumentation/contrib/gin` that is added by a community member.

kakkoyun:
> Before start moving things around, should we define what an instrumentation package should look like?
> 
> We have this issue for moving them to `contrib`, but we also have this other [#225](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/225) for moving hook definitions to instrumentation package.
> 
> I'd start doing a POC with the helloworld before moving all, so we focus the discussion on how the package should look like, how we are going to lint that, etc.
> 
> Sth like?
> 
> ```
> /
> ├── pkg/
> │   ├── inst/
> │   ├── shared/ 
> │   └── go.mod
> ├── instrumentation/
> │   ├── runtime/
> │   ├── http/
> │   ├── grpc/
> │   │   ├── rules.yaml
> │   │   ├── client/
> │   │   ├── server/
> │   │   ├── semconv/
> │   │   ├── go.mod
> │   │   ├── Makefile
> │   │   └── README.md
> │   └── contrib/
> │       └── helloworld/
> │           ├── rules.yaml
> │           ├── helloworld_hook.go
> │           ├── go.mod
> │           └── Makefile
> ```
> 
> For each package:
> 
> * Verify rules.yaml exist
> * make generate --> make test --> make build
> 
> Another point is the e2e tests. Right now, we have them under /test/e2e. It would be good if each contrib package also provides an e2e test to verify that it works correctly. We should provide some helpers for this.

💯 Agreed!

We can move the tests within the integrations. To give this flexibility we added the build tag so e2e tests and integration tests could be anywhere.

txabman42:
If thats blocking this, I would say so.

[...]

PR #226, opened 2025-12-08
feat(instrumentation): Add gRPC instrumentation

kakkoyun:


tool/internal/instrument/trampoline.go
```diff
@@ -871,6 +871,9 @@ func replaceTypeParamsWithAny(t dst.Expr, typeParams *dst.FieldList) dst.Expr {
 	case *dst.IndexListExpr:
 		// GenStruct[T, U] -> interface{} (for generic receiver methods with multiple type params)
 		return ast.InterfaceType()
+	case *dst.Ellipsis:
```
kakkoyun, line 874: Better fix introduced by https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/231

Thanks @y1yang0

demo/grpc/client/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Multi-stage Dockerfile for gRPC client with OpenTelemetry compile-time instrumentation
 # Stage 1: Build the otel tool
-FROM golang:1.25-alpine@sha256:26111811bc967321e7b6f852e914d14bede324cd1accb7f81811929a6a57fea9 AS otel-builder
+FROM golang:1.24-alpine@sha256:06545cc1ff10ddf04aebe20db1352ec7c96d1e43135767931c473557d0378202 AS otel-builder
```
txabman42: Why this version change? Before it was matching the http one. Also making sure its working with latest go versions is always good.

demo/grpc/client/main.go
```diff
@@ -41,29 +74,33 @@ func main() {
...
 		// Streaming RPC
 		streamClient, err := c.SayHelloStream(ctx)
 		if err != nil {
-			log.Fatalf("could not create stream: %v", err)
+			logger.Error("could not create stream", "error", err)
+			os.Exit(1)
 		}
 
 		// Send multiple requests
 		for i := 0; i < 5; i++ {
```
txabman42:
```suggestion
		for i := 0; i < *count; i++ {
```

demo/grpc/client/main.go
```diff
@@ -41,29 +74,33 @@ func main() {
...
 			req := &pb.HelloRequest{Name: *name + " " + time.Now().Format("15:04:05")}
 			if err := streamClient.Send(req); err != nil {
-				log.Fatalf("could not send: %v", err)
+				logger.Error("could not send", "error", err)
+				os.Exit(1)
 			}
-			log.Printf("Sent: %v", req.GetName())
+			logger.Info("sent request", "name", req.GetName())
```
txabman42: nitpick: stream logs `sent request` while unary `greeting + sending request`. Could be better to standarize logs messages and add an extra label with call type.

demo/grpc/client/main.go
txabman42, line 147: nitpick: this could be added inside a goroutine before sending the requests, so the demo app send requests and get them, instead of waiting to send all requests for receiving them.

demo/grpc/README.md
```diff
@@ -82,13 +92,20 @@ cd client
...
 # Combine options
 ./client -addr=localhost:50052 -name="Testing" -stream=true
 
 # Send a shutdown request to the server, this will exit the server process gracefully.
 ./client -shutdown
 ```
 
+#### Telemetry Export Note
+
+The client includes a 6-second sleep after completing requests to ensure the batch span processor (default 5-second timeout) has time to flush pending telemetry before the process exits. This is particularly important for single or low-count requests. For continuous monitoring, use a higher `-count` value which naturally provides enough runtime for span export.
```
txabman42: I would add a log in the client sth like `waiting to flush telemetry...`

demo/infrastructure/docker-compose/docker-compose.yml
```diff
@@ -128,9 +128,10 @@ services:
       context: ../../..
       dockerfile: demo/grpc/client/Dockerfile
     container_name: grpc-client
+    command: ["-addr", "grpc-server:50051", "-name", "demo-client", "-count", "10000"]
     environment:
-      - OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4317
-      - OTEL_EXPORTER_OTLP_PROTOCOL=grpc
+      - OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4318
+      - OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
```
txabman42, line 134: why not grpc? 🤔

demo/infrastructure/docker-compose/Makefile
```diff
@@ -382,3 +387,73 @@ quickstart-http: prereqs ports start-http health urls ## HTTP-only quickstart
 	@echo "$(GREEN)✓ HTTP observability stack is ready!$(NC)"
 	@echo "$(YELLOW)Note: gRPC demo excluded due to known build issues$(NC)"
 	@echo ""
+
+# =============================================================================
+# gRPC Targets (for testing gRPC instrumentation)
+# =============================================================================
+
+start-grpc: ## Start gRPC demo stack
```
txabman42, line 395: These actions are not very discoverable. I'd document the existence of them (http ones included) in the README. Also that building the project is a pre-requisite for running the demo.

Makefile
```diff
@@ -262,14 +271,17 @@ test-unit/pkg/coverage: package gotestfmt ## Run unit tests with coverage for pk
...
+		grep -v "pkg/instrumentation/nethttp/semconv" | \
+		grep -v "pkg/instrumentation/helloworld" | \
+		grep -v "/server$$" | \
+		grep -v "/client$$" | \
+		grep -v "/shared$$" | \
+		grep -v "^pkg$$"); \
 	for moddir in $$PKG_MODULES; do \
 		echo "Testing $$moddir..."; \
-		(cd $$moddir && go mod tidy && go test -json -v -shuffle=on -timeout=5m -count=1 ./... -coverprofile=coverage-pkg.txt -covermode=atomic 2>&1 | tee -a ../../gotest-unit-pkg.log | gotestfmt); \
+		(cd $$moddir && go mod tidy > /dev/null 2>&1 && go test -json -v -shuffle=on -timeout=5m -count=1 ./... -coverprofile=coverage-pkg.txt -covermode=atomic 2>&1 | tee -a ../../gotest-unit-pkg.log | gotestfmt); \
```
txabman42:
```suggestion
		(cd $$moddir && go mod tidy && go test -json -v -shuffle=on -timeout=5m -count=1 ./... -coverprofile=coverage-pkg.txt -covermode=atomic 2>&1 | tee -a ../../gotest-unit-pkg.log | gotestfmt); \
```

Makefile
```diff
@@ -237,7 +239,14 @@ test-unit/tool: build package gotestfmt ## Run unit tests for tool modules only
 test-unit/pkg: package gotestfmt ## Run unit tests for pkg modules only
 	@echo "Running pkg unit tests..."
 	set -euo pipefail
-	@PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | grep -v "pkg/instrumentation/runtime"); \
+	@PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | \
```
txabman42: why not just run all tests under pkg directly?

Makefile
```diff
@@ -262,14 +271,17 @@ test-unit/pkg/coverage: package gotestfmt ## Run unit tests with coverage for pk
...
-	PKG_MODULES=$$(find pkg -name "go.mod" -type f \
-							-not -path "pkg/instrumentation/nethttp/semconv/*" \
-							-not -path "pkg/instrumentation/runtime/*" \
-							-not -path "pkg/instrumentation/helloworld/*" \
-	 						-exec dirname {} \; ); \
+	PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | \
+		grep -v "pkg/instrumentation/runtime" | \
+		grep -v "pkg/instrumentation/nethttp/semconv" | \
+		grep -v "pkg/instrumentation/helloworld" | \
+		grep -v "/server$$" | \
```
txabman42: If we remove server, client, etc, which are the modules we are targeting with the pkg coverage testing? I though those were the modules to be tested

test/integration/grpc_server_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/test/app"
+)
+
+// TestGRPCServerIntegration tests gRPC server instrumentation
+func TestGRPCServerIntegration(t *testing.T) {
+	serverDir := filepath.Join("..", "..", "demo", "grpc", "server")
```
txabman42, line 19: Could be extracted all identical code (server startup, readiness detection, build, cleanup, ...) to some private funcs. That way less duplicate code, and even more important, easier to detect the key lines.

pkg/instrumentation/grpc/client/client_hook_test.go
```diff
@@ -0,0 +1,416 @@
...
+				assert.Greater(t, len(newOpts), len(tt.opts), "Expected stats handler to be added")
+			} else {
+				// Should not modify options when disabled
+				assert.Equal(t, len(tt.opts), len(newOpts))
+			}
+		})
+	}
+}
+
+func TestAfterNewClient(t *testing.T) {
```
txabman42, line 143: What is being tested in this test? that is not panicking? 🤔

test/integration/grpc_client_test.go
```diff
@@ -0,0 +1,169 @@
...
+	serverApp, outputPipe := app.Start(t, serverDir)
+	defer func() {
+		if serverApp.Process != nil {
+			_ = serverApp.Process.Kill()
+		}
+	}()
+
+	readyChan := make(chan struct{})
+	go func() {
+		serverOutput := strings.Builder{}
```
pdelewski: This is probably also related to `http`, but I would suggest to move this line above, to get better error message

test/integration/grpc_client_test.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+		}
+	}()
+
+	// Wait for server to be ready
+	select {
+	case <-readyChan:
+		t.Log("Server is ready!")
+	case <-time.After(5 * time.Second):
+		t.Fatal("timeout waiting for server to be ready")
```
pdelewski: It would be good to have better error message about root cause, for instance
``` listen tcp :50051: bind: address already in use```
```suggestion
		output := serverOutput.String()
+               if output == "" {
+                       t.Fatal("timeout waiting for server to be ready - no server output received")
+               } else {
+                       t.Fatalf("timeout waiting for server to be ready - server output:\n%s", output)
+               }
```

test/integration/grpc_server_test.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+		}
+	}()
+
+	// Wait for server to be ready
+	select {
+	case <-readyChan:
+		t.Log("Server is ready!")
+	case <-time.After(5 * time.Second):
+		t.Fatal("timeout waiting for server to be ready")
```
pdelewski: Same as above

demo/grpc/client/Dockerfile
kakkoyun: There was a hick-up. I forgot now. I'll try to upgrade.

demo/grpc/client/main.go
kakkoyun: Good catch!

demo/infrastructure/docker-compose/docker-compose.yml
kakkoyun: It created a deadlock with the gRPC instrumentation and I cut some corners. I'm planning to send a subsequent PR for this. I will create an issue.

demo/infrastructure/docker-compose/Makefile
kakkoyun: I'll add them to README.

There is always the option of using `make help`

Makefile
kakkoyun: Not all the packages actually have tests or testable (e.g. instrumentation/runtime) and running it like that fails. Hence the selection logic.

Makefile
kakkoyun: I think this is a merge mistake, it should be include not exclude. I'll investigate.

test/integration/grpc_server_test.go
kakkoyun: It can be but I like redundancy in the tests. They are same right now, but they will probably diverge a lot.

demo/infrastructure/docker-compose/Makefile
txabman42: That should be enough too. If you want to just run http or grpc, or know how to use some extra options --> run `make help`

test/integration/grpc_server_test.go
kakkoyun: I'll address this.

pkg/instrumentation/grpc/client/client_hook_test.go
kakkoyun: I'll add more test cases to check panics.

test/integration/grpc_client_test.go
kakkoyun: I narrowed the scope because the output is not used outside scope.

demo/infrastructure/docker-compose/docker-compose.yml
kakkoyun: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/233

txabman42 (approved):

demo/grpc/client/main.go
```diff
@@ -37,52 +70,117 @@ func main() {
...
+	// Wait for receiver to finish
+	<-done
+}
+
+func finish() {
+	logger.Info("client finished")
+	// Allow time for batch span processor to flush (default timeout is 5s)
+	// This ensures traces are exported before the process exits
+	logger.Info("waiting to flush telemetry...")
+	time.Sleep(6 * time.Second)
```
y1yang0: This hardcoded 6-second sleep seems fragile and slows down our CI/CD pipeline. Is there a more elegant solution, like TracerProvider.Shutdown?

test/integration/grpc_client_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Build the server and client applications with the instrumentation tool
+	app.Build(t, serverDir, "go", "build", "-a")
+	app.Build(t, clientDir, "go", "build", "-a")
+
+	t.Log("Starting gRPC server...")
+
+	// Start the server and wait for it to be ready
+	serverApp, outputPipe := app.Start(t, serverDir)
+	defer func() {
```
y1yang0: Already exists in app.Start

test/integration/helpers_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// waitForServerReady waits for a server to be ready by monitoring its output for "server started".
+// It returns:
+//   - func() string: a cleanup function that waits for the server to exit and returns its complete output
+//   - error: non-nil if the server failed to start within the timeout
+//
+// This helper provides better error messages on timeout by including the server's output.
+// Callers should check the error and use require.NoError to fail the test with proper context.
+func waitForServerReady(t *testing.T, serverCmd *exec.Cmd, output io.ReadCloser) (func() string, error) {
```
y1yang0: We can move this to app/infra.go to make it shared code.

Makefile
```diff
@@ -257,28 +285,27 @@ test-unit/tool/coverage: package gotestfmt ## Run unit tests with coverage for t
...
 	@echo "Running pkg unit tests with coverage..."
 	set -euo pipefail
-	ROOT_DIR=$$(pwd); \
-	PKG_MODULES=$$(find pkg -name "go.mod" -type f \
-							-not -path "pkg/instrumentation/nethttp/semconv/*" \
-							-not -path "pkg/instrumentation/runtime/*" \
-							-not -path "pkg/instrumentation/helloworld/*" \
-	 						-exec dirname {} \; ); \
+	rm -f ./gotest-unit-pkg.log
+	PKG_MODULES=$$(find pkg -maxdepth 3 -name "go.mod" -type f -exec dirname {} \; | grep -v "runtime" | grep -v "^pkg$$"); \
```
y1yang0, line 295: `-maxdepth 3` is fraigle

kakkoyun: Yes, it is but adding case by case filtering is more cumbersome. My proposed solution is moving the instrumentation out of `pkg` to top-level and adding `Makefile`s for each instrumentation and calling `make test` in each directory. It would make it more flexible and scalable.

kakkoyun: Update the https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/133 to encapsulate this idea.

y1yang0 (approved):

PR #231, opened 2025-12-09
fix: unexpected generic type: *dst.Ellipsis

y1yang0:
```
 # google.golang.org/grpc
Error:
[0] Unimplemented: unexpected generic type: *dst.Ellipsis
```


kakkoyun (approved):
LGTM!

Thanks for introducing a comprehensive fix for this. I have came across to this and just introduce a primitive patch!

Issue #224, opened 2025-12-08
chore: Validate the context propagation is supported


kakkoyun:
- [x] In-process (goroutine boundary?) https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/238
- [ ] HTTP https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/238#discussion_r2626650139
- [ ] gRPC https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/238#discussion_r2626650139

txabman42:
@kakkoyun  Could this issue be closed? We have a test for verifying the context propagation + tests for ensuring trace is propagated via HTTP/gRPC.

PR #238, opened 2025-12-17
chore(test): add context propagation test

txabman42:
Add in-process context propagation test https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/224

### DIFF
--- a/docs/ux-design.md
+++ b/docs/ux-design.md
@@ -208,6 +208,18 @@ used directly to build, run, and test go applications directly:
    go run github.com/open-telemetry/opentelemetry-go-compile-instrumentation/cmd/otel go test -shuffle=on ./...
    ```
 
+#### Building Multiple Packages
+
+The tool supports building multiple packages in a single command, which is useful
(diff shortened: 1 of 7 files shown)


pkg/instrumentation/helloworld/helloworld_hook.go
```diff
@@ -123,3 +123,15 @@ func BeforeUnderscore(ictx inst.HookContext, _ int, _ float32) {}
 func MyHookEllipsisBefore(ictx inst.HookContext, p1 []string) {
 	println("Ellipsis")
 }
+
+func FunctionABefore(ictx inst.HookContext, ctx context.Context) {
+	ctx, span := tracer.Start(ctx, "FunctionA")
+	ictx.SetParam(0, ctx)
```
kakkoyun, line 129: We should double check if we have the right calls in the `gRPC` and `HTTP` instrumentation.

And maybe add similar test to those.

kakkoyun (approved):
LGTM 💯

PR #242, opened 2025-12-18
fix: correctly pass variadic arguments

y1yang0:
a complete fix for "..." argument

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,9 @@ Highlights:
 ### 1. Build the Tool
 
 ```bash
-$ git clone https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation.git
-$ cd opentelemetry-go-compile-instrumentation
-$ make build
+git clone https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation.git
+cd opentelemetry-go-compile-instrumentation
+make build
 ```
 
 The `otel` binary will be built in the root directory.
@@ -40,16 +40,16 @@ The `otel` binary will be built in the root directory.
 Just prefix the original `go build` command with `otel`.
 
 ```bash
-$ cd demo/basic
-$ ../../otel go build
-$ ./basic
+cd demo/basic
+../../otel go build
+./basic
 [... output ...]
 ```
 
 ### 3. Run the Tests
 
 ```bash
-$ make test
+make test
 ```
 
 ## Community
--- a/tool/internal/instrument/trampoline.go
+++ b/tool/internal/instrument/trampoline.go
@@ -234,10 +234,10 @@ func countParameters(fields *dst.FieldList) int {
 // checkHookDecl checks if the hook function declaration is correct, i.e. if they
 // have correct signature
 func (ip *InstrumentPhase) checkHookDecl(hookFunc *dst.FuncDecl, before bool) error {
+	// TargetFunc:  func A(a int, b string) (ret string)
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -28,9 +28,9 @@ Highlights:
 ### 1. Build the Tool
 
 ```bash
-git clone https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation.git
-cd opentelemetry-go-compile-instrumentation
-make build
+$ git clone https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation.git
```
kakkoyun: nit: I think we shouldn't put `$`, it makes harder to just copy them and run them

tool/internal/instrument/trampoline.go
```diff
@@ -227,97 +221,105 @@ func getHookFunc(t *rule.InstFuncRule, before bool) (*dst.FuncDecl, error) {
...
+		beforeHookParams := ast.SplitMultiNameFields(beforeHook.Type.Params)
+		beforeTrampParamsCount := countParameters(beforeTrampParams)
+		beforeHookParamsCount := countParameters(beforeHookParams)
+		if (beforeTrampParamsCount + 1) != beforeHookParamsCount {
+			return ex.Newf("hook func signature mismatch, expected %d, got %d",
+				beforeTrampParamsCount+1, beforeHookParamsCount)
 		}
-		if ast.IsEllipsis(field.Type) {
-			attr.IsVariadic = true
+	} else {
```
kakkoyun: nit: Instead of `else` we can do an early return now.
```suggestion
	return nil
	} 
```

tool/internal/instrument/trampoline.go
```diff
@@ -227,97 +221,105 @@ func getHookFunc(t *rule.InstFuncRule, before bool) (*dst.FuncDecl, error) {
...
-	for i, field := range splitParams.List {
-		attr := ParamTrait{Index: i}
-		if ast.IsInterfaceType(field.Type) {
-			attr.IsInterfaceAny = true
+	return count
+}
+
+// checkHookDecl checks if the hook function declaration is correct, i.e. if they
+// have correct signature
+func (ip *InstrumentPhase) checkHookDecl(hookFunc *dst.FuncDecl, before bool) error {
```
txabman42, line 236: checkHookDecl just checks trampParams are hookParams+1. WDYT about also checking they are the expected type?

tool/internal/instrument/trampoline.go
```diff
@@ -872,8 +891,8 @@ func replaceTypeParamsWithAny(t dst.Expr, typeParams *dst.FieldList) dst.Expr {
 		// GenStruct[T, U] -> interface{} (for generic receiver methods with multiple type params)
 		return ast.InterfaceType()
 	case *dst.Ellipsis:
-		// ...T -> []T with type parameter replacement
-		return ast.ArrayType(replaceTypeParamsWithAny(tType.Elt, typeParams))
+		// ...T -> ...interface{} with type parameter replacement
```
txabman42:
```suggestion
		// Preserve variadic syntax: ...T -> ...interface{}
		// This maintains variadic semantics in the generated hook signatures
```

tool/internal/instrument/trampoline.go
y1yang0: Good idea, we can check here if the first argument is a HookContext and if the other arguments have the same type

Tracked in #245

PR #246, opened 2025-12-22
fix: ensure consistency between setup and instrumentation phase properly

### LINKED ISSUE #218: chore: Ensure consistency between setup and instrumentation phase properly
If the file versions used in the setup and instrument phases are inconsistent, it should actually be considered an error.

This is because the setup phase's matching process precisely locates a specific function in a specific version. If the instrument phase subsequently proceeds with a different version, it indicates that something went wrong in a process after the initial match was completed.

I will fix this.

_Originally posted by @y1yang0 in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/156#discussion_r2564137529_

y1yang0:
fix #218

### DIFF
--- a/docs/instrument-guide.md
+++ b/docs/instrument-guide.md
@@ -27,7 +27,7 @@ inject_to_grpc_newserver:
 ```
 
 * `target`: Import path of the package to instrument.
-* `version`: Version range to match.
+* `version`: Version range to match. The left bound is inclusive, the right bound is exclusive. If version is not specified, the rule is applicable to all versions.
 * `func`: Name of the function to hook.
 * `before` / `after`: Names of the hook functions.
 * `path`: Import path where the hook functions are defined.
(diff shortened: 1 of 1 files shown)


docs/instrument-guide.md
```diff
@@ -0,0 +1,111 @@
...
+  target: google.golang.org/grpc
+  version: v1.63.0,v1.70.0
+  func: NewServer
+  before: BeforeNewServer
+  after: AfterNewServer
+  path: github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/grpc/server
+```
+
+* `target`: Import path of the package to instrument.
+* `version`: Version range to match.
```
ralf0131: It is better to declare that the version is inclusive or exclusive.

docs/instrument-guide.md
```diff
@@ -0,0 +1,111 @@
...
+* `func`: Name of the function to hook.
```
ralf0131, line 31: What is the behavior if func does not exist?

docs/instrument-guide.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+If we cannot import a specific type (e.g., it is unexported), we can use `interface{}` in the hook signature.
+
+### Limitations
+
+When implementing hooks, we must adhere to certain limitations:
+
+1. **Restricted Imports**: If we are instrumenting a library (e.g., `github.com/foo/bar`), our hook code can only import {}
```
ralf0131: Is the last {} in this line unnecessary?

y1yang0: Yes, let me remove it

docs/instrument-guide.md
y1yang0: Changed to "Name of the function to hook. (Required)"

PR #248, opened 2025-12-24
chore(test): add unit test coverage for ex and util packages

### LINKED ISSUE #47: Add unit test for tool/{util, ast, ex, etc}
Add comprehensive unit tests for common libraries such as tool/{util, ast, ex}, etc.  

For some standalone utility functions within these packages, we hope to add more unit tests to improve coverage.

nithinputhenveettil:
Fixes #47 (Partially)

I have not added unit tests for all the missing packages mentioned in the issue yet. This PR currently includes tests for the files listed below.

- ex
  - error.go
- util
  - shared.go
  - assert.go
  - log.go

I am planning to continue working on this over the Christmas weekend to improve test coverage for the remaining missing packages as well.

Meanwhile, please have a look at this PR and let me know if you have any comments or suggestions.

We can also split the remaining tests into a separate PR if the reviewers feel that adding them here would make the PR too large. Alternatively, I can include more package tests in this PR as well.


tool/testutil/exec.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func RunSelfTest(t *testing.T, testName, env string) (int, string) {
```
y1yang0: We can skip testing the fatal functions (e.g., Fatal, util.Assert), so we don't need to add RunSelfTest. Otherwise LGTM

nithinputhenveettil: @y1yang0 :

 I will remove the fatal functions tests.

Also, Sorry for the lint failures, I tried running the linter locally, but `make lint` was not working yesterday due to some weird go version error. I will fix those issues as well.

Do you think it’s okay to add more files to this PR, or should I raise a separate PR for the additional unit test cases?

nithinputhenveettil:
@y1yang0 : Removed fatal tests and also addressed lint issues.

tool/testutil/exec.go
y1yang0: > Ooh.. do you think we can skip all the Fatal tests?

Yes, I think we can skip it.

> Do you think it’s okay to add more files to this PR, or should I raise a separate PR for the additional unit test cases?

I'm fine with either approach. You can complete it in one patch or split it into smaller patches, depending on your preference.

nithinputhenveettil: @y1yang0: In that case, I would prefer to keep this PR open until the end of this weekend. I’m mostly free this weekend, so I can improve the tests further and add them to the same PR.

Also, do you see any issue with the lint checks? Locally, I’m getting zero issues when running `make lint/go`, but it is failing in the pipeline.

y1yang0: Maybe you need to use golangci-lint-2.1.6

nithinputhenveettil: Ah, okay. I was using the latest version, 2.7.2. Now I can see the same failures locally as well. I have addressed those issues in the latest commit.

tool/util/log_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+	"log/slog"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLoggerFromContext_Default(t *testing.T) {
+	//nolint:usetesting // explicitly testing behaviour with context.Background()
```
nithinputhenveettil, line 15: This test intentionally uses `context.Background()` to validate the exact behaviour with a base context. Using `t.Context()` (as suggested by the linter) does not make sense here.

tool/ex/error_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func TestUnwrap(t *testing.T) {
+	baseErr := errors.New("root cause")
+	w1 := Wrap(baseErr)
+	w2 := Wrap(w1)
+
+	assert.ErrorIs(t, w2, baseErr)
+}
+
+func TestStackTracePresent(t *testing.T) {
```
txabman42, line 47: The main idea for this stackfulError is to carry stracktrace info. Basically, log stracktrace during Fatal/Fatalf.

I would add a test for Fatal to ensure the expected stracktrace is logged (not just the format). This would ensure that captureStack is working as expected and that Fatal logs correctly the info in frame.

It could also be interesting to verify the panic thrown by Fatal.

tool/util/assert_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAssert_Pass(t *testing.T) {
+	Assert(true, "should not fail")
+}
+
+func TestAssertType_Pass(t *testing.T) {
```
txabman42, line 17: I would test also non-happy path cases (assert and assertType). They are quite interesting in this case, as they call ex.Fatalf which it makes the app panic.

tool/util/log_test.go
```diff
@@ -0,0 +1,41 @@
...
+	ctx := context.Background()
+	customLogger := slog.New(slog.NewTextHandler(nil, nil))
+
+	ctx = ContextWithLogger(ctx, customLogger)
+
+	logger := LoggerFromContext(ctx)
+	assert.Equal(t, customLogger, logger)
+}
+
+func TestLoggerFromContext_IgnoresWrongType(t *testing.T) {
```
txabman42, line 34: Keep all tests for the same function together. Right now its:
1. TestLoggerFromContext_...
2. TestContextWithLogger_...
3. TestLoggerFromContext_...

tool/util/shared_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+func TestBackupAndRestoreFile(t *testing.T) {
+	dir := t.TempDir()
+	t.Setenv(EnvOtelWorkDir, dir)
+
+	// Setup dummy file to backup
+	err := os.MkdirAll(filepath.Join(dir, BuildTempDir, "backup"), 0o755)
+	require.NoError(t, err)
+
+	fn := filepath.Join(dir, "file1.txt")
```
txabman42, line 84: Create `file1.txt` and `backup` as constants and `[]byte("some content")` as variable.

tool/util/shared_test.go
```diff
@@ -0,0 +1,118 @@
...
+	t.Chdir(dir)
+
+	// backup
+	err = BackupFile([]string{"file1.txt"})
+	require.NoError(t, err)
+
+	backupLoc := filepath.Join(dir, BuildTempDir, "backup", "file1.txt")
+	backedData, err := os.ReadFile(backupLoc)
+	require.NoError(t, err)
+	assert.Equal(t, []byte("some content"), backedData)
```
txabman42, line 98: It could be interesting to also assert that the original file content hasn't changed after backup.

txabman42 (changes requested):
Nice tests! Just a few comments with some improvement ideas

tool/ex/error_test.go
nithinputhenveettil: @txabman42:

I had actually added the Fatal test initially, it is present in commit 4c411c4b8. As suggested by @y1yang0, I removed it. I also feel it would be better to test the Fatal cases with the full panic/exit behaviour.

I will bring back those changes (hope @y1yang0 is okay with this) and will also double check and address all your suggestions.

y1yang0: I think this is just a helper for exit-time diagnostics, so covering it isn't a high priority. Rather than doing that, I suggest we add tests for wrapOrCreate directly.

txabman42: Not high prioritary but its an interesting use case to have tested.

I usually prefer to test directly public functions which are the ones consumed by other packages, but its ok for me to test direclty `wrapOrCreate`. In the end the main logic to be tested is `wrapOrCreate` and verify the content of frame which implicitly tests `captureStack`.

kakkoyun:
@nithinputhenveettil Do you want to revive this?

PR #251, opened 2025-12-26
fix: debug files in .otel-build use incorrect absolute paths

### LINKED ISSUE #247: Debug files in .otel-build use incorrect absolute paths
The build process generates instrumented files in the .otel-build/debug directory. Currently, these files are saved using their absolute file system path, which is incorrect.

## Current Behavior

A modified file is placed in the debug directory with its full absolute path. For example:
```
demo/basic/.otel-build/debug/main/home/-/opentelemetry-go-compile-instrumentation/demo/basic/go.mod
```
## Expected Behavior

The files should be placed in the debug directory while preserving their relative path within the project. The correct path for the example above should be:
```
demo/basic/.otel-build/debug/main/go.mod
```

y1yang0:
fixes #247

### DIFF
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -27,12 +27,12 @@ func (sp *SetupPhase) Debug(msg string, args ...any) { sp.logger.Debug(msg, args
 
 // keepForDebug copies the file to the build temp directory for debugging
 // Error is tolerated as it's not critical.
-func (sp *SetupPhase) keepForDebug(name string) {
-	base := filepath.Base(name)
-	dstFile := filepath.Join(util.GetBuildTemp("debug"), "main", base)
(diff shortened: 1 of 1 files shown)


tool/internal/setup/setup.go
```diff
@@ -28,7 +28,8 @@ func (sp *SetupPhase) Debug(msg string, args ...any) { sp.logger.Debug(msg, args
 // keepForDebug copies the file to the build temp directory for debugging
 // Error is tolerated as it's not critical.
 func (sp *SetupPhase) keepForDebug(name string) {
```
txabman42: Right now keepForDebug expects name. But this change treats it as it was a path.

I would do one of these 2:
- Send directly the filename to keepForDebug
- Rename `name` attribute to `file` or `filepath`

Issue #234, opened 2025-12-11
test: Find better ways to assert for integration and e2e tests rather than checking logs


kakkoyun:
Some ideas

For integration tests:

- https://pkg.go.dev/go.opentelemetry.io/otel/sdk/trace/tracetest
- https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest

For end-to-end tests:

spawn actual collector and use https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/

---

e2e should verify those spans, now that we support http and aligned with otel semconv. Worth an issue?

_Originally posted by @txabman42 in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/156#discussion_r2565335282_

txabman42:
@kakkoyun is this duplicated from https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/214

We can keep this and close the other one. WDYT?

> Find better ways to assert spans and add them to e2e and integration tests

txabman42:
In this case, our e2e and integration tests build and start the services (http client/server, etc.) so it'll be hard to use things like tracetest there. I think the best approach is to do sth similar to what `opentelemetry-go-instrumentation` has done. Create a small custom server which receives the spans and parse them ([code](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/872e6935f6793817594b9e229824ae89d05f3246/internal/test/e2e/integration.go#L84)).

I'll give it a try to this approach.

PR #254, opened 2026-01-02
chore(tests): check e2e spans

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/234

Check spans during e2e tests.

Inspired by [opentelemetry-go-instrumentation](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/internal/test/e2e/integration.go)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: Release
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+jobs:
+  release:
+    name: Build and Release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c  # v6.1.0
+        with:
+          go-version-file: 'go.mod'
+      - name: Build binary
+        run: make build-all
+      - name: Create Release and Upload Assets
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -330,28 +330,28 @@ test-integration: go-protobuf-plugins ## Run integration tests
 test-integration: build build-demo gotestfmt
 	@echo "Running integration tests..."
 	set -euo pipefail
-	go test -json -v -shuffle=on -timeout=10m -count=1 -tags integration ./test/integration/... 2>&1 | tee ./gotest-integration.log | gotestfmt
+	cd test && go test -json -v -shuffle=on -timeout=10m -count=1 -tags integration ./integration/... 2>&1 | tee ../gotest-integration.log | gotestfmt
```
y1yang0, line 333: Why does this need to be changed?

txabman42: Because now tests have their own module, so tests must run within it. It would fail if not.

Note that I've added a go module to test folder to avoid adding extra dependencies to root go.mod. However, I'd prefer to have e2e tests directly inside inst packages https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/133#issuecomment-3660097896, but out of scope from this PR.

test/app/collector.go
```diff
@@ -0,0 +1,54 @@
...
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+// Collector represents an in-memory OTLP collector for testing
+type Collector struct {
```
kakkoyun, line 16: I love it ❤️

kakkoyun (approved):
LGTM 🚀 

Thanks for tackling this!

Issue #245, opened 2025-12-22
Stricter hook definition checks


y1yang0:
checkHookDecl just checks trampParams are hookParams+1. WDYT about also checking they are the expected type?

_Originally posted by @txabman42 in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/242#discussion_r2637727637_

PR #255, opened 2026-01-02
fix: make hook def check stricter

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/245

It checks the arguments base types from hook and trampoline are the same.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,14 +110,14 @@ build-demo: build-demo-grpc build-demo-http
 build-demo-grpc: go-protobuf-plugins ## Build gRPC demo server and client
 	@echo "Building gRPC demo..."
 	@rm -f demo/grpc/server/otel.runtime.go demo/grpc/client/otel.runtime.go
-	@(cd demo/grpc/server && go generate && go build -o server .)
-	@(cd demo/grpc/client && go build -o client .)
+	@(cd demo/grpc/server && go generate && go build -a -o server .)
+	@(cd demo/grpc/client && go build -a -o client .)
 
 build-demo-http: ## Build HTTP demo server and client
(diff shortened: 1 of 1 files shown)


tool/internal/instrument/trampoline.go
```diff
@@ -221,48 +222,79 @@ func getHookFunc(t *rule.InstFuncRule, before bool) (*dst.FuncDecl, error) {
...
-		beforeTramp := ip.beforeTrampFunc
-		beforeHook := hookFunc
-		beforeTrampParams := ast.SplitMultiNameFields(beforeTramp.Type.Params)
-		beforeHookParams := ast.SplitMultiNameFields(beforeHook.Type.Params)
-		beforeTrampParamsCount := countParameters(beforeTrampParams)
-		beforeHookParamsCount := countParameters(beforeHookParams)
-		if (beforeTrampParamsCount + 1) != beforeHookParamsCount {
-			return ex.Newf("hook func signature mismatch, expected %d, got %d",
-				beforeTrampParamsCount+1, beforeHookParamsCount)
+		beforeTrampParams := ast.SplitMultiNameFields(ip.beforeTrampFunc.Type.Params)
```
y1yang0, line 255: maybe we can check `hook vs target`? we dont need to extract base type then

txabman42: I haven't found an easy way to compare hook.Type with target.Type, being both dst.Expr.

It works parsing them to string, but it ends up being same solution as current one.

y1yang0: Okay then.

y1yang0 (approved):

PR #266, opened 2026-01-14
feat(tool): support custom rule configuration

y1yang0:
Allow specifying the rule configuration at compile time, i.e.
```
$ ./otel -rules=custom.yml go build -o app ./cmd/app
```
After this, we can move tool/data/*.yaml into their respective directories.

### DIFF
--- a/tool/internal/setup/match_test.go
+++ b/tool/internal/setup/match_test.go
@@ -303,9 +303,9 @@ func validateCreatedRule(t *testing.T, createdRule rule.InstRule, ruleName strin
 	}
 }
 
-func writeCustomRules(t *testing.T, name string, content string) string {
+func writeCustomRules(t *testing.T, name, content string) string {
 	path := filepath.Join(t.TempDir(), name)
-	err := os.WriteFile(path, []byte(content), 0644)
+	err := os.WriteFile(path, []byte(content), 0o644)
 	require.NoError(t, err)
 	return path
 }
@@ -322,7 +322,7 @@ func TestLoadDefaultRules(t *testing.T) {
   raw: "_ = 1"`
 	p1 := writeCustomRules(t, "r1.yaml", content1)
 	p2 := writeCustomRules(t, "r2.yaml", content2)
-	os.Setenv(util.EnvOtelRules, p1)
+	t.Setenv(util.EnvOtelRules, p1)
 
 	// Prepare setup phase and set custom rules via environment variable and flag
 	sp := newTestSetupPhase()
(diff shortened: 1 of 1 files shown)


tool/internal/setup/match.go
```diff
@@ -212,9 +215,24 @@ func (sp *SetupPhase) preciseMatching(
...
+func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
+	// Load custom rules from config file if specified
+	if sp.ruleConfig != "" {
+		content, err := os.ReadFile(sp.ruleConfig)
+		if err != nil {
+			return nil, ex.Wrap(err)
+		}
+		return parseRuleFromYaml(content)
+	}
+	// TODO: Load rules from environment variable if specified, e.g. OTEL_RULES
```
kakkoyun: Let's create an issue for this (If we haven't already).

tool/internal/setup/match.go
```diff
@@ -212,9 +215,24 @@ func (sp *SetupPhase) preciseMatching(
 	return set, nil
 }
 
+func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
+	// Load custom rules from config file if specified
+	if sp.ruleConfig != "" {
```
kakkoyun, line 230: I believe we should support loading several files. What I mean the value can be a comma separated list or if it is a directory we can load all the files. Probably, we might need to change the suffix to something like `*xxx*.otelc.yaml` or `*.tool.yaml`.

We don't have to do it as part of this PR, we can create an issue for it.

kakkoyun (approved):
LGTM.

I have some suggestions to improve things.

tool/internal/setup/match.go
y1yang0: Great suggestion. Tracked in #268

tool/internal/setup/match.go
y1yang0: It doesn’t seem very complicated—let me implement it in this patch. (I’ll add the tests tomorrow, thanks.)

tool/internal/setup/match_test.go
```diff
@@ -299,9 +303,36 @@ func validateCreatedRule(t *testing.T, createdRule rule.InstRule, ruleName strin
 }
 
 func TestMaterializeRules(t *testing.T) {
```
txabman42: I miss some tests related to loadRules. We need to ensure priority order (EnvOtelRules > Config > Default) and that each case is working as expected.

y1yang0: Done

Issue #271, opened 2026-01-15
Fresh Go cache for each build


txabman42:
Use a fresh Go cache for each build to prevent polluting the global Go cache (this is important as we may modify core packages like runtime or net/http)

Related to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/257#issuecomment-3721615866

PR #272, opened 2026-01-15
fix(go): use independent go cache for each build

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/271


tool/internal/setup/setup.go
```diff
@@ -208,6 +208,31 @@ func Setup(ctx context.Context, cmd *cli.Command) error {
 	return sp.store(matched)
 }
 
+// setupFreshCache creates a temporary GOCACHE if one isn't already set.
+// This prevents cache pollution when modifying core packages via //go:linkname.
+func setupFreshCache(ctx context.Context, env []string) ([]string, func(), error) {
```
y1yang0:
```suggestion
func setupGoCache(ctx context.Context, env []string) ([]string, func(), error) {
```

y1yang0 (approved):

PR #273, opened 2026-01-19
feat(tool): support multiple custom rules

### LINKED ISSUE #268: Support loading several YAML rule files
I believe we should support loading several files. What I mean the value can be a comma separated list or if it is a directory we can load all the files. Probably, we might need to change the suffix to something like `*xxx*.otelc.yaml` or `*.rule.yaml`.

We don't have to do it as part of this PR, we can create an issue for it.

_Originally posted by @kakkoyun in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/266#discussion_r2690126873_

e.g.
```
$ otel -rules=a.yaml,b.yaml go build
```

tushar1977:
## Description

Allow specifying the multiple rule configuration at compile time, i.e.

```bash
$ ./otel --rules=a.yml,b.yml go build -o app ./cmd/app
```

Fixes #268

### DIFF
--- a/tool/internal/setup/match_test.go
+++ b/tool/internal/setup/match_test.go
@@ -311,7 +311,6 @@ func writeCustomRules(t *testing.T, name, content string) string {
 }
 
 func TestMultipleRuleFiles(t *testing.T) {
-	// Write custom rules to temporary files
 	content1 := `h1:
   target: main
   func: Example
@@ -320,23 +319,35 @@ func TestMultipleRuleFiles(t *testing.T) {
   target: main
   func: Example
   raw: "_ = 1"`
+
 	p1 := writeCustomRules(t, "r1.yaml", content1)
 	p2 := writeCustomRules(t, "r2.yaml", content2)
+
 	t.Setenv(util.EnvOtelRules, "")
 
-	// Prepare setup phase and set custom rules via environment variable and flag
 	sp := newTestSetupPhase()
 	err := sp.extract()
 	require.NoError(t, err)
+
 	sp.ruleConfig = p1 + "," + p2
 
-	// Get all rules from rule Files
 	rules, err := sp.loadRules()
 	require.NoError(t, err)
-	require.NotEmpty(t, rules)
 	require.Len(t, rules, 2)
 	require.Equal(t, "h1", rules[0].GetName())
 	require.Equal(t, "h2", rules[1].GetName())
+
+	// Check for duplicate rule by name
+	sp = newTestSetupPhase()
+	err = sp.extract()
+	require.NoError(t, err)
+
+	sp.ruleConfig = p1 + "," + p1
+
+	rules, err = sp.loadRules()
+	require.NoError(t, err)
+	require.Len(t, rules, 1)
+	require.Equal(t, "h1", rules[0].GetName())
 }
 
 func TestLoadDefaultRules(t *testing.T) {


tushar1977:
any test cases that should be created?

tool/internal/setup/match.go
```diff
@@ -226,13 +226,33 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
+		ruleFiles := strings.SplitSeq(sp.ruleConfig, ",")
+
+		for file := range ruleFiles {
+			file = strings.TrimSpace(file)
+			// Starting Point for each rule file
+			content, err := os.ReadFile(file)
+			if err != nil {
+				return nil, ex.Wrapf(err, "failed to read %s from -rules flag", file)
+			}
+			rules, _ := parseRuleFromYaml(content)
```
kakkoyun: We shouldn't just ignore the error. I would say we should stop the operation and propagate the error to the user.

tool/internal/setup/match.go
```diff
@@ -226,13 +226,33 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
-			return nil, ex.Wrapf(err, "failed to read %s from -rules flag", sp.ruleConfig)
+		// Custom map to store deduplicate rules
+		ruleSet := make(map[string]rule.InstRule)
+		var allRules []rule.InstRule
+		ruleFiles := strings.SplitSeq(sp.ruleConfig, ",")
+
+		for file := range ruleFiles {
+			file = strings.TrimSpace(file)
+			// Starting Point for each rule file
+			content, err := os.ReadFile(file)
```
kakkoyun, line 240: There was a second part of the issue where we test for the directories and look for a special suffix to identify any rule files in the directory. If we want to tackle this in a subsequent PR, let's make sure we have an issue for it.

tool/internal/setup/match.go
```diff
@@ -226,13 +226,33 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
+			rules, _ := parseRuleFromYaml(content)
+
+			for _, rule := range rules {
+				key := rule.GetName()
+				ruleSet[key] = rule
+			}
 		}
-		return parseRuleFromYaml(content)
+
+		allRules = make([]rule.InstRule, 0, len(ruleSet))
```
kakkoyun: nit: I'm sure there is a modern way to do this now in either `slice` or `maps` package.

tushar1977: so you mean to implement ```allRules``` using maps or slice?
EDIT - i checked up slice package and found this
```go
		allRules := make([]rule.InstRule, 0, len(ruleSet))
		for _, r := range ruleSet {
			allRules = append(allRules, r)
		}
		return slices.Clone(allRules), nil
```
could you review this?

tool/internal/setup/match.go
```diff
@@ -226,13 +226,33 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
+		for _, r := range ruleSet {
+			allRules = append(allRules, r)
+		}
+		return allRules, nil
```
kakkoyun:
```suggestion
		return slices.Collect(maps.Values(ruleSet)), nil
```

tool/internal/setup/match.go
kakkoyun: See my suggestion https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/273#discussion_r2704158753

(I haven't tested it)

tool/internal/setup/match.go
tushar1977: sure, i'll create a seperate pr and issue for this

tool/internal/setup/match.go
tushar1977: Got it! Thanks for the suggestion. Although, the meaning is same but its modern way to write.

tool/internal/setup/match.go
kakkoyun: #275

PR #276, opened 2026-01-21
feat(tool): Add support for directories as rule source

### LINKED ISSUE #275: Add support for directories as rule source
There was a second part of the issue where we test for the directories and look for a special suffix to identify any rule files in the directory. If we want to tackle this in a subsequent PR, let's make sure we have an issue for it.

_Originally posted by @kakkoyun in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/273#discussion_r2703985499_

tushar1977:
## Description
Added support for directories as rule source. Right now, it supports this pattern ```*.otelc.yaml```

```bash
$ ./otel --rules=a/,b.yml go build -o app ./cmd/app
```
Fixes #275

### DIFF
--- a/tool/cmd/main.go
+++ b/tool/cmd/main.go
@@ -33,7 +33,7 @@ func main() {
 				Value:     util.GetBuildTempDir(),
 			},
 			&cli.BoolFlag{
-				Name:    "-debug",
+				Name:    "debug",
 				Aliases: []string{"d"},
 				Usage:   "Enable debug mode",
 				Value:   false,


tool/util/sys.go
```diff
@@ -144,3 +144,29 @@ func PathExists(path string) bool {
 func NormalizePath(path string) string {
 	return filepath.ToSlash(filepath.Clean(path))
 }
+
+func GetRuleFromDir(path string) ([]string, error) {
```
y1yang0: Move to match.go as it's not shared function

tool/internal/setup/match.go
```diff
@@ -234,21 +234,30 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
 
-		for file := range ruleFiles {
-			file = strings.TrimSpace(file)
-			// Starting Point for each rule file
-			content, err := os.ReadFile(file)
-			if err != nil {
-				return nil, ex.Wrapf(err, "failed to read %s from -rules flag", file)
-			}
-			rules, err := parseRuleFromYaml(content)
+		for path := range ruleFiles {
```
y1yang0: Since it is becoming more complex, we could consider extracting it into a separate function, for example,` if sp.ruleConfig != "" { return loadCustomRules(sp.ruleConfig) }`.

tool/internal/setup/match.go
```diff
@@ -234,21 +234,30 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
-				ruleSet[key] = rule
+			for _, file := range files {
+				// Starting Point for each rule file
+				content, Staterr := os.ReadFile(file)
+				if Staterr != nil {
+					return nil, ex.Wrapf(err, "failed to read %s from -rules flag", file)
+				}
+				rules, err := parseRuleFromYaml(content)
+				if err != nil {
+					return nil, ex.Wrapf(err, "failed to parse rules from file %q", file)
```
y1yang0: If an error does not originate from a third-party or the standard library, it is already a "stackful error". There's no need to wrap it; just return err directly.

tool/util/sys.go
```diff
@@ -144,3 +144,29 @@ func PathExists(path string) bool {
...
+	ruleFilePatterns := []string{"*.otelc.yaml"}
```
kakkoyun: We can support nested directories.

kakkoyun (commented):
Looking good. We need good tests and I agree with y1yang0's comment.

tool/internal/setup/match.go
tushar1977: sure

tool/util/sys.go
tushar1977: For nested directories, we can use WalkDir (https://pkg.go.dev/path/filepath#WalkDir)  and then use glob.

kakkoyun: Sorry, that I missed it.

tushar1977: > We need good tests 

Can you describe this like what test cases are needed with this?

tool/internal/setup/match.go
```diff
@@ -216,43 +218,100 @@ func (sp *SetupPhase) preciseMatching(
 	return set, nil
 }
 
-func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
-	// Load rules from environment variable OTEL_RULES if specified. It has the
-	// highest priority.
-	rulePath := os.Getenv(util.EnvOtelRules)
-	if rulePath != "" {
-		content, err := os.ReadFile(rulePath)
+func GetRuleFromDir(path string) ([]string, error) {
```
kakkoyun:
```suggestion
func ruleFromDir(path string) ([]string, error) {
```
No need to export.

tool/internal/setup/match.go
```diff
@@ -216,43 +218,100 @@ func (sp *SetupPhase) preciseMatching(
...
+	ruleFilePatterns := []string{"*.otelc.yaml"}
```
kakkoyun:
```suggestion
	ruleFilePatterns := []string{"*.otelc.yaml", "*.otelc.yml"}
```

tool/internal/setup/match.go
```diff
@@ -216,43 +218,100 @@ func (sp *SetupPhase) preciseMatching(
...
+			for _, r := range rules {
+				ruleSet[r.GetName()] = r
 			}
 		}
+	}
+
+	return slices.Collect(maps.Values(ruleSet)), nil
+}
+
+func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
```
kakkoyun, line 301: Let's put some debug log lines for the selection here. It would be helpful to know whether the rules come from the flag or environment variable while troubleshooting.

tool/internal/setup/match_test.go
```diff
@@ -310,6 +310,38 @@ func writeCustomRules(t *testing.T, name, content string) string {
...
+	err = os.WriteFile(filepath.Join(dir, "r2.otelc.yaml"), []byte(content2), 0o644)
+	require.NoError(t, err)
+
+	t.Setenv(util.EnvOtelRules, "")
+
+	sp := newTestSetupPhase()
+	err = sp.extract()
+	require.NoError(t, err)
+
+	sp.ruleConfig = dir
```
kakkoyun, line 342: Could we add test cases for nested directories?

kakkoyun (commented):
We are close. I have added a couple of comments.

tool/internal/setup/match_test.go
tushar1977: yes

tushar1977:
@kakkoyun @y1yang0 

I noticed a bug in otel flag
```bash
➜ otel --help
NAME:
   otel - OpenTelemetry Go Compile-Time Instrumentation Tool

USAGE:
   otel [global options] [command [command options]]

COMMANDS:
   setup    
   go       
   version  
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --work-dir string, -w string    The path to a directory where working files will be written (default: "/home/tushar/.otel-build")
   ---debug, -d                    Enable debug mode           <-----------------------
   --rules string, --rules string  The path to the rules configuration file
   --help, -h                      show help
```

The ``debug`` flag has an extra ```-```. I rechecked by cloning and building otel again. Still same flag issue

PR #278, opened 2026-01-21
feat(tool): function call join point support

darccio:
## Description

Implemented call‑site wrapping rules with template‑based AST substitution:

- Adds function‑call rules, template compilation, and basic import handling.
- Adds extensive unit tests for call wrapping, alias resolution, placeholder behavior, and template errors.
- Updates rule documentation to describe call wrapping, qualified matching, and current template output constraints.

## Motivation

Implements #277's function call join point.

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format`
- [x] Linters pass: `make lint`
- [x] Tests pass: `make test`
- [x] Tests added for new functionality
- [x] Documentation updated (if applicable)

Issue #277, opened 2026-01-21
OP3: Join point support


darccio:
- [x] OP3.1: Function call
- [x] OP3.2: Named declaration (function, type, global variable or constant)
- [x] OP3.3: Value declaration of a given type
- [x] OP3.4: Metaprogramming annotations (magical comments)
- [x] OP3.5: Function or method with specific signature

darccio:
All done.

PR #279, opened 2026-01-22
feat(tool): function call join point support

darccio:
## Description

Adds support for a new **Call Wrapping Rule** — a join point that matches function *call sites* and wraps them with a user-supplied template expression.

This complements the existing Function Hook Rule (which instruments function *definitions*) for cases where the caller, not the callee, must be modified — e.g. when the target function belongs to a third-party package with no exported hook points.

### What's included

- **New rule type: `function-call`** — matches calls to a qualified function (`package/path.FunctionName`) anywhere in the instrumented package and rewrites each call site using a Go `text/template` expression. The `{{ . }}` placeholder is substituted with the original call expression.

- **Template system** — uses Go's `text/template` for code generation, with compile-time validation of the template and the resulting AST node type.

- **Import injection for call rules** — the `imports` field is fully supported: any packages referenced in the wrapper template are automatically added to the instrumented file (same mechanism as other rule types).

- **Setup integration** — `preciseMatching` now registers `InstCallRule` entries so they reach the instrumentation phase without panicking.

- **End-to-end and golden tests** — new test cases cover basic wrapping (`call-rule-only`), template features (`call-rule-template-features`), and import injection (`call-rule-with-import`).

### Rule format

```yaml
wrap_my_call:
  target: main
  function-call: some/pkg.MyFunc        # qualified: import-path.FuncName
  template: "wrapper.Wrap({{ . }})"     # {{ . }} is the original call
  imports:
    wrapper: "my/instrumentation/wrapper"
```

### Rule fields

| Field           | Type              | Required | Description |
|-----------------|-------------------|----------|-------------|
| `function-call` | string            | yes      | `import/path.FuncName` to match |
| `template`      | string            | yes      | `text/template` expression; `{{ . }}` → original call |
| `imports`       | map[string]string | no       | Extra imports needed by the template |

### Limitations

- The template must produce a single **call expression** (not a statement
  block). Templates that produce other expression kinds are rejected at
  instrumentation time.
- Only top-level, non-generic function calls are matched in this initial
  implementation.

## Motivation

Implements #277's function call join point. Duplicates #278 as it was created from a branch pushed to the repo instead of doing it in a fork.

---

## Checklist

- [x] Tests pass: `make test`


docs/rules.md
```diff
@@ -13,7 +13,7 @@ All rules share a set of common fields that define the target of the instrumenta
 
 ## Rule Types
 
-There are four types of rules, each designed for a specific kind of code modification.
+There are five types of rules, each designed for a specific kind of code modification.
```
kakkoyun: I think we should omit the number here. It's hard to keep it consistent.

tool/internal/rule/base.go
```diff
@@ -59,16 +60,18 @@ func NewInstRuleSet(importPath string) *InstRuleSet {
...
 		FuncRules:   make(map[string][]*InstFuncRule),
 		StructRules: make(map[string][]*InstStructRule),
+		CallRules:   make(map[string][]*InstCallRule),
 		FileRules:   make([]*InstFileRule, 0),
 	}
 }
 
 func (irs *InstRuleSet) String() string {
-	return fmt.Sprintf("{%s: %v, %v, %v, %v}",
+	return fmt.Sprintf("{%s: %v, %v, %v, %v, %v}",
```
kakkoyun: It's better to convert this to a string builder or similar where we pass a slice and it would print.

Also maybe a way to auto-discover the rules.

tool/internal/instrument/apply_call_helpers.go
```diff
@@ -0,0 +1,178 @@
...
+	if pkgPath != "" {
+		return pkgPath == importPath
+	}
+
+	resolvedPath, ok := importAliases[ident.Name]
+	return ok && resolvedPath == importPath
+}
+
+// addImportsFromRule adds any imports specified in the rule to the file.
+func addImportsFromRule(root *dst.File, r *rule.InstCallRule) {
```
kakkoyun: I imagine importing facilities could be shared among other possible future rules. Nothing the change, just something to keep in mind.

kakkoyun (commented):
Thank you so much!

I just did a first sweep. I will take another look at it today or tomorrow.

I'm impressed this was really fast! Kudos!


One thing, I realized we can have classification now to divide the rules. Call-site vs Trampoline. I image there can be rules can be clustered into two and maybe we can carve out some interfaces and collect some shared functionality around these interfaces. However, let's not rush it, just something to keep in mind, when we start adding similar rules.

txabman42:
I agree with keeping both approaches. The library-side approach should be the default and most common one (this should be reflected in the documentation). It's like having 2 flavours for function instrumentation. However, the call-side approach adds extra flexibility, which some users may find useful. It adds some overlapping, but at the same time, it also adds new instrumentation possibilities.

Could you share some examples of real-world use cases for this feature coming from Orchestrion?

Additionally, this direction aligns well with making the solution compatible with Orchestrion.

Regarding the implementation, I still think we should try to unify how we instrument a function, regardless of whether it’s targeted from the call side or the library side. I would expect both share same `insertTJump` logic.

y1yang0:
I’ve addressed all of my concerns! Please feel free to leave comments so we can start the review.

darccio:
@y1yang0 @kakkoyun Ready for your review!

tool/internal/instrument/apply_call.go
```diff
@@ -0,0 +1,89 @@
...
+		return true
+	})
+
+	// Now wrap each matching call
+	for _, call := range matchingCalls {
+		if err := wrapCall(call, r); err != nil {
+			// Log but continue processing other calls
+			ip.Warn("Failed to wrap call", "error", err)
+			continue
+		}
```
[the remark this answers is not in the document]
darccio: This is on purpose. Ignoring.

tool/internal/rule/base.go
```diff
@@ -132,3 +147,12 @@ func (irs *InstRuleSet) GetStructRules() []*InstStructRule {
 	}
 	return rules
 }
+
+// GetCallRules returns all call rules from the rule set.
+func (irs *InstRuleSet) GetCallRules() []*InstCallRule {
```
kakkoyun: This function is not used. Considering my previous comment. Maybe this was meant to be used in `String()` method?

tool/internal/setup/match.go
```diff
@@ -207,6 +209,9 @@ func (sp *SetupPhase) preciseMatching(
 					set.AddRawRule(source, rt)
 					sp.Info("Match raw rule", "rule", rt, "dep", dep)
 				}
+			case *rule.InstCallRule:
+				set.AddCallRule(source, rt)
+				sp.Info("Match call rule", "rule", rt, "dep", dep)
```
kakkoyun, line 219: Unlike every other rule type in `preciseMatching` — which verifies the AST before adding (`FindFuncDecl` for func rules, `FindStructDecl` for struct rules) — call rules are added unconditionally to *all* source files in the target package.

This is functionally correct (no matching calls = no-op), but it performs unnecessary AST traversals on every file in the package and breaks the "precise matching" contract the other cases establish. For a package with 50 files where only 3 contain calls to the target function, all 50 will be visited by `applyCallRule`.

Consider at minimum a comment explaining why the unconditional match is intentional, or scan each file for the import alias + function name string before adding the rule.

tool/internal/rule/call_rule.go
```diff
@@ -0,0 +1,166 @@
...
+
+package rule
+
+import (
+	"encoding/json"
+	"regexp"
+	"strings"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/ex"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/instrument/template"
```
kakkoyun: medium: This import creates a dependency from the `rule` layer into `instrument/template`, inverting the normal dependency direction (where `instrument` depends on `rule`). All other rule types (`InstFuncRule`, `InstRawRule`, `InstStructRule`) are pure data objects with no dependency on the instrumentation layer.

Today it works because `template` is a leaf package, but if `template` ever needs to import from `rule` you'll hit a circular import.

Consider moving `tool/internal/instrument/template/` → `tool/internal/template/` as a sibling to both `rule` and `instrument`, so both layers can import it cleanly.

tool/internal/rule/call_rule.go
```diff
@@ -0,0 +1,166 @@
...
+func (r *InstCallRule) validate() error {
+	// FunctionCall format already validated in NewInstCallRule
+	if strings.TrimSpace(r.FunctionCall) == "" {
+		return ex.Newf("function-call cannot be empty")
+	}
+
+	if strings.TrimSpace(r.Template) == "" {
+		return ex.Newf("template cannot be empty")
+	}
+	if !strings.Contains(r.Template, "{{ . }}") {
```
kakkoyun: medium: This exact-string check rejects valid Go `text/template` placeholder variants:
- `{{.}}` (no spaces — idiomatic and common)
- `{{- . -}}` (trimmed whitespace)
- `{{ .  }}` (extra spaces)

Consider using a regex like `\{\{-?\s*\.\s*-?\}\}`, or execute the template with a sentinel value and verify the sentinel appears in the rendered output.

tool/internal/instrument/apply_call.go
```diff
@@ -0,0 +1,89 @@
...
+	// Use the template to compile the wrapped expression
+	wrappedExpr, err := tmpl.CompileExpression(call)
+	if err != nil {
+		return ex.Wrapf(err, "failed to compile template")
+	}
+
+	// Verify we got a call expression back
+	wrappedCall, ok := wrappedExpr.(*dst.CallExpr)
+	if !ok {
+		return ex.Newf("template did not produce a call expression, got %T", wrappedExpr)
```
kakkoyun: medium: `CompileExpression` successfully handles any Go expression (binary, selector, etc.), but this layer silently rejects anything that isn't a `*dst.CallExpr`. A user writing `"{{ . }} + 1"` or `"{{ . }}.Field"` gets this error with no guidance on the restriction.

The limitation is documented in the PR description, but consider a more actionable error message:
```go
return ex.Newf("template output must be a call expression (e.g. \"wrapper({{ . }})\") but got %T; see docs/rules.md for supported template patterns", wrappedExpr)
```

tool/internal/instrument/apply_call_helpers.go
```diff
@@ -0,0 +1,133 @@
...
+			if alias == "" || alias == "_" || alias == "." {
+				continue
+			}
+			aliases[alias] = importPath
+		}
+	}
+	return aliases
+}
+
+func defaultImportAlias(importPath string) string {
```
kakkoyun, line 94: low: This function reimplements Go's import alias inference (versioned `v2`/`v9` stripping, `gopkg.in` convention). Edge cases like package names containing `-`, `_test` packages, or build-tag-conditional imports may behave differently from the Go compiler's own resolution. A comment acknowledging this would help future maintainers understand where mismatches can arise.

tool/internal/instrument/template/template.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
+)
+
+// Template represents a code template that can be used to wrap or transform
+// Go expressions. It uses Go's text/template package for template execution
+// and supports placeholder substitution for AST nodes.
+type Template struct {
+	template *template.Template
+	Source   string
```
kakkoyun: low: `Source` is exported but never read outside of this struct (only assigned during `NewTemplate`). If it's for debugging, consider making it unexported or adding a `String()` method.

tool/internal/template/template.go
```diff
@@ -0,0 +1,130 @@
...
+`))
+
+// NewTemplate creates a new Template from the provided template text.
+// The template text should contain {{ . }} as a placeholder for the expression
+// being wrapped.
+//
+// Example:
+//
+//	NewTemplate("wrapper({{ . }})")
+func NewTemplate(text string) (*Template, error) {
```
kakkoyun: nit: Consider [`fasttemplate`](https://github.com/valyala/fasttemplate) as a replacement for `text/template` here. For single-variable substitution (`{{ . }}` → a string), `fasttemplate` is up to 10x faster because it avoids the full template engine parse/execute overhead. Since `CompileExpression` is called once per matched call site during compilation, the impact is minor today, but it would simplify the implementation — `fasttemplate` substitution is a single `ExecuteString` call with no need for the `_.PLACEHOLDER_0` sentinel pattern.

darccio: I don't think our bottleneck comes from `text/template` but I'd open to do it in a separate PR.

tool/internal/instrument/template/template.go
darccio: Done in [f543f31](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/f543f31a2190731355b4abe0dde10304c56f88dd)

tool/internal/instrument/apply_call_helpers.go
darccio: Done in [f543f31](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/f543f31a2190731355b4abe0dde10304c56f88dd)

tool/internal/instrument/apply_call.go
darccio: Done in [f543f31](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/f543f31a2190731355b4abe0dde10304c56f88dd)

tool/internal/rule/call_rule.go
darccio: Done in [f543f31](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/f543f31a2190731355b4abe0dde10304c56f88dd)

tool/internal/setup/match.go
darccio: Done in [f543f31](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/f543f31a2190731355b4abe0dde10304c56f88dd)

tool/internal/rule/call_rule.go
darccio: Done in [f543f31](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/f543f31a2190731355b4abe0dde10304c56f88dd)

tool/internal/instrument/testdata/golden/call-rule-template-features/rules.yml
```diff
@@ -0,0 +1,4 @@
+wrap_nested_call:
+  target: "main"
+  function-call: "unsafe.Sizeof"
```
y1yang0: Exiting code uses `function_call` style

tool/internal/template/placeholder.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"github.com/dave/dst"
+	"github.com/dave/dst/dstutil"
+)
+
+// replacePlaceholder replaces all occurrences of _.PLACEHOLDER_0 in the AST
+// with the given node. This is used to inject the original call expression
+// into the template-generated code.
+func replacePlaceholder(ast, node dst.Node) (dst.Node, bool) {
```
y1yang0: Since template.go and placeholder.go are only used in one place right now, I suggest we move them to tool/internal/instrument. We might even be able to merge them into one file?

tool/internal/template/placeholder.go
```diff
@@ -0,0 +1,41 @@
...
+		ast,
+		func(cursor *dstutil.Cursor) bool {
+			selectorExpr, ok := cursor.Node().(*dst.SelectorExpr)
+			if !ok {
+				return true
+			}
+
+			// Check if this is _.PLACEHOLDER_0
+			ident, ok := selectorExpr.X.(*dst.Ident)
+			if !ok || ident.Name != "_" {
```
y1yang0: util.IdentIgnore

tool/internal/instrument/testdata/golden/call-rule-template-features/rules.yml
darccio: Done in [a45c13e](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/a45c13e9b5d89584be8ee4ad4788b969b0ed806e)

tool/internal/template/placeholder.go
darccio: Done in [0bf38a5](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/279/commits/0bf38a57b7d9c892235067956c1dd7a7d6ba504a)

PR #283, opened 2026-01-29
feat(tooll): `importcfg` manipulation support

### LINKED ISSUE #282: Support 3rd party libraries in hooks
From our [docs](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#limitations):

> Limitations
> When implementing hooks, we must adhere to certain limitations:
> 
> Restricted Imports: If we are instrumenting a library (e.g., github.com/foo/bar), our hook code can only import from:
>   - The Target Library (github.com/foo/bar)
>   - OpenTelemetry packages
>   - Standard Library packages
>
> Importing other third-party libraries is not allowed.

Support importing other third-party libraries in hooks.

darccio:
## Description

This PR adds support for declaring import dependencies in instrumentation rules. Rules can now specify an `imports` field that automatically adds the required packages to instrumented files, enabling injected code to reference external packages without manual import management.

Previously, instrumentation rules that injected code referencing external packages would fail to compile. Users had to ensure target files already imported the needed packages, or the build would fail with "could not import X" errors. This severely limited the flexibility of raw code injection and other rule types.

## Solution

Implement end-to-end import handling across the instrumentation toolchain:

### AST Import Management
- New `imports` package to parse existing imports from DST files with bidirectional alias/path mapping
- Handle edge cases: blank imports (`_`), dot imports (`.`), and explicit aliases
- Detect conflicts when the same alias points to different packages

### Alias Validation
- Validate that rule import aliases match existing file aliases before injection
- Catch cases where injected code uses `ctx.Context` but the file imports `context.Context`
- Detect dot-import conflicts where rules expect qualified identifiers

### Go Toolchain Integration
- Parse and update `importcfg` files that the Go compiler uses to locate package archives
- New imports must be registered in importcfg before compilation can succeed
- Preserve unrecognized directives and ensure deterministic output for reproducible builds

### Build Coordination
- Per-process import tracking files to avoid race conditions during parallel compilation
- Intercept the link phase to merge all tracked imports into the linker's importcfg
- Forward build context flags (`-race`, `-tags`, `-mod`, etc.) to resolve archives for the correct build configuration

### Rule Integration
- Add `imports` field to `InstBaseRule` allowing any rule type to declare dependencies
- Wire import handling to file, func, raw, and struct rule application

## Usage

```yaml
raw_with_logging:
  target: main
  func: Example
  raw: |
    fmt.Printf("Hash: %x\n", sha256.Sum256([]byte("data")))
  imports:
    fmt: "fmt"
    sha256: "crypto/sha256"
```

## Testing

- Golden test case `raw-rule-with-imports` verifying imports are properly injected
- Updated demo/basic rules demonstrating real-world usage with `context`, `fmt`, and `crypto/sha256`

## Verification

It's been verified using `go-size-analyzer` against binaries compiled from `demo/basic`:

`go build` (requires manual edition):
```
➜ gsa basic | grep crypto           
time=12.378958ms level=INFO msg="Parsing binary..."
time=15.67375ms level=INFO msg="Parsed binary"
time=15.689041ms level=INFO msg="Finding build info..."
time=15.694416ms level=INFO msg="Found build info"
time=15.700958ms level=INFO msg="Loading sections..."
time=15.712583ms level=INFO msg="Loaded sections"
time=15.71675ms level=INFO msg="Loading version flag"
time=49.856875ms level=INFO msg="Loaded version flag"
time=49.880666ms level=INFO msg="Loading packages..."
time=61.046583ms level=INFO msg="Loaded packages done"
time=61.0705ms level=INFO msg="Loading meta info..."
time=61.073125ms level=INFO msg="Loaded meta info"
time=61.076083ms level=INFO msg="Parsing DWARF..."
time=89.892875ms level=INFO msg="Parsed DWARF"
time=91.231166ms level=INFO msg="Analyzing symbols..."
time=91.488458ms level=INFO msg="Analyzing symbols done"
time=91.502041ms level=INFO msg="Disassemble functions..."
time=91.589916ms level=WARN msg="Disassembler not supported on this architecture"
time=92.692958ms level=INFO msg="Printing text report"
time=92.944666ms level=INFO msg="Report rendered"
time=92.959208ms level=INFO msg="Report written"
time=92.961375ms level=INFO msg="Printing done"
time=92.963583ms level=INFO msg="Ready to exit"
```

`otel go build`:
```
➜ gsa basic | grep crypto           
time=14.278917ms level=INFO msg="Parsing binary..."
time=20.423833ms level=INFO msg="Parsed binary"
time=20.440708ms level=INFO msg="Finding build info..."
time=20.449167ms level=INFO msg="Found build info"
time=20.4555ms level=INFO msg="Loading sections..."
time=20.474083ms level=INFO msg="Loaded sections"
time=20.479042ms level=INFO msg="Loading version flag"
time=61.914667ms level=INFO msg="Loaded version flag"
time=61.93475ms level=INFO msg="Loading packages..."
time=81.875833ms level=INFO msg="Loaded packages done"
time=81.895667ms level=INFO msg="Loading meta info..."
time=81.897833ms level=INFO msg="Loaded meta info"
time=81.900292ms level=INFO msg="Parsing DWARF..."
time=137.369958ms level=INFO msg="Parsed DWARF"
time=141.748167ms level=INFO msg="Analyzing symbols..."
time=142.490625ms level=INFO msg="Analyzing symbols done"
time=142.540333ms level=INFO msg="Disassemble functions..."
time=142.73125ms level=WARN msg="Disassembler not supported on this architecture"
time=146.007625ms level=INFO msg="Printing text report"
time=146.351542ms level=INFO msg="Report rendered"
time=146.363292ms level=INFO msg="Report written"
time=146.36575ms level=INFO msg="Printing done"
time=146.367792ms level=INFO msg="Ready to exit"
│ 4.91%   │ crypto                                                                        │ 379 kB │ std       │
│ 0.00%   │ vendor/golang.org/x/crypto/chacha20poly1305                                   │ 376 B  │ std       │
│ 0.00%   │ vendor/golang.org/x/crypto/cryptobyte                                         │ 340 B  │ std       │
```

## Motivation

Rules can now safely reference packages not originally imported by the target file. The tool dynamically resolves and registers these packages with the Go compiler.

Closes #282

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -343,10 +343,6 @@ func BuildWithToolexec(ctx context.Context, cmd *cli.Command) error {
 	newArgs = append(newArgs, "-work")
 	// Add "-toolexec=..."
 	newArgs = append(newArgs, insert)
-	// TODO: We should support incremental build in the future, so we don't need
-	// to force rebuild here.
-	// Add "-a" to force rebuild
-	newArgs = append(newArgs, "-a")
 	// Add the rest
 	newArgs = append(newArgs, args[1:]...)
 	logger.InfoContext(ctx, "Running go build with toolexec", "args", newArgs)
--- a/Makefile
+++ b/Makefile
@@ -91,17 +91,17 @@ install: package ## Install otel to $$GOPATH/bin (auto-packages instrumentation)
 package: ## Package the instrumentation code into binary
 	@echo "Packaging instrumentation code into binary..."
 	@set -euo pipefail
-	rm -rf $(INST_PKG_TMP)
-	if [ ! -d pkg ]; then \
+	@rm -rf $(INST_PKG_TMP)
+	@if [ ! -d pkg ]; then \
 		echo "Error: pkg directory does not exist"; \
 		exit 1; \
 	fi
-	cp -r pkg $(INST_PKG_TMP)
-	(cd $(INST_PKG_TMP) && go mod tidy)
-	tar -czf $(INST_PKG_GZIP) --exclude='*.log' $(INST_PKG_TMP)
-	mkdir -p tool/data/
-	mv $(INST_PKG_GZIP) tool/data/
-	rm -rf $(INST_PKG_TMP)
+	@cp -r pkg $(INST_PKG_TMP)
+	@(cd $(INST_PKG_TMP) && go mod tidy)
+	@tar -czf $(INST_PKG_GZIP) --exclude='*.log' $(INST_PKG_TMP)
+	@mkdir -p tool/data/
+	@mv $(INST_PKG_GZIP) tool/data/
+	@rm -rf $(INST_PKG_TMP)
 	@echo "Package created successfully at tool/data/$(INST_PKG_GZIP)"
 
 build-demo: ## Build all demos
--- a/test/testutil/fixture.go
+++ b/test/testutil/fixture.go
@@ -58,6 +58,10 @@ func NewTestFixture(t *testing.T, opts ...TestFixtureOption) *TestFixture {
 		f.collector = StartCollector(t)
 
 		// Configure OTEL env vars (can be overridden by test after this)
+		// Clear signal-specific endpoints to ensure OTEL_EXPORTER_OTLP_ENDPOINT takes precedence
+		t.Setenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", "")
+		t.Setenv("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT", "")
+		t.Setenv("OTEL_EXPORTER_OTLP_LOGS_ENDPOINT", "")
 		t.Setenv("OTEL_SERVICE_NAME", f.serviceName)
 		t.Setenv("OTEL_TRACES_EXPORTER", "otlp")
 		t.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", f.collector.URL)
--- a/tool/ex/error.go
+++ b/tool/ex/error.go
@@ -141,7 +141,7 @@ func Fatalf(format string, args ...any) {
 
 func Fatal(err error) {
 	if err == nil {
-		fmt.Fprintln(os.Stderr, "Fatal error: unknown")
+		_, _ = fmt.Fprintln(os.Stderr, "Fatal error: unknown")
 		os.Exit(1)
 	}
 	e := &stackfulError{}
@@ -149,7 +149,7 @@ func Fatal(err error) {
 		em := ""
 		var emSb149 strings.Builder
 		for i, m := range e.message {
-			emSb149.WriteString(fmt.Sprintf("[%d] %s\n", i, m))
+			_, _ = fmt.Fprintf(&emSb149, "[%d] %s\n", i, m)
 		}
 		em += emSb149.String()
 		_, _ = fmt.Fprintf(os.Stderr, "Error:\n%s\nStack:\n%s\n",


tool/internal/imports/imports.go
```diff
@@ -0,0 +1,307 @@
...
+	ImportPath string   `json:"ImportPath"`
+	Export     string   `json:"Export"`
+	Deps       []string `json:"Deps"`
+}
+
+// ResolvePackageInfo attempts to retrieve the archive for the designated import path
+// and its dependencies using `go list -export -json`.
+// The buildFlags parameter allows passing build flags (like -tags) to ensure the resolved
+// archives match the current build context.
+func ResolvePackageInfo(ctx context.Context, importPath string, buildFlags ...string) (map[string]string, error) {
```
y1yang0: Can setup.getBuildPackages be used as a replacement for this?

tool/util/shared.go
```diff
@@ -59,3 +75,32 @@ func BackupFile(names []string) error {
...
+	if err := json.Unmarshal([]byte(encoded), &flags); err != nil {
+		// Malformed JSON, return nil
+		return nil
+	}
+	return flags
+}
+
+// EncodeBuildFlags encodes build flags as a JSON string for storage in an environment variable.
+// This preserves arguments that contain spaces (e.g., -tags "foo bar").
+func EncodeBuildFlags(flags []string) string {
```
y1yang0: Move to toolexec.go as they are not shared utilities

tool/internal/instrument/toolexec.go
```diff
@@ -74,10 +84,25 @@ func interceptCompile(ctx context.Context, args []string) ([]string, error) {
...
+		workDir:          filepath.Dir(target),
+		compileArgs:      args,
+		importConfigPath: importCfgPath,
+	}
+
+	// Parse existing importcfg if present
+	if importCfgPath != "" {
+		imports, err := imports.ParseImportCfg(importCfgPath)
+		if err != nil {
+			return nil, fmt.Errorf("parsing importcfg: %w", err)
```
y1yang0: A global comment:

We dont use fmt.Errorf. If an error originates from a standard library API or a third-party library function, please wrap it with ex.Wrap(err) before returning. For errors generated within our framework, a direct return err is sufficient.

(Please see the comments in ex/error.go).

tool/internal/imports/imports.go
```diff
@@ -0,0 +1,307 @@
...
+		root.Decls = append([]dst.Decl{newImportDecl}, root.Decls...)
+	}
+
+	return nil
+}
+
+// CollectPaths returns a map of all unique import paths in the file.
+// The map uses import path as both key and value, ensuring that multiple blank (_) or
+// dot (.) imports don't collapse to a single entry.
+func CollectPaths(ctx context.Context, root *dst.File, buildFlags ...string) map[string]string {
```
y1yang0, line 214: A global comment:

Since many functions are only used within this package, they can be unexported (e.g., CollectPaths => collectPaths).

tool/internal/instrument/imports.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if err := ip.updateImportConfig(paths); err != nil {
+		return ex.Wrapf(err, "updating import config for file imports in %s", ruleName)
+	}
+
+	return nil
+}
+
+// handleRuleImports processes imports for a rule and updates the import config.
+// ruleType is used for error messages (e.g., "file rule", "func rule").
```
y1yang0: ruleType is used for error messages —— It's redundant. Errors wrapped with `ex.Newf` will generate a full stack trace, which already includes the origin file (e.g., `apply_func.go`, `apply_file.go`). This stack trace effectively tells us the 'rule type'

tool/internal/instrument/apply_file.go
```diff
@@ -53,6 +53,17 @@ func (ip *InstrumentPhase) applyFileRule(rule *rule.InstFileRule, pkgName string
...
+	// Without this, the compiler will fail with "could not import X" errors.
+	if err = ip.updateImportConfigForFile(root, rule.Name); err != nil {
+		return err
+	}
+
+	// Add any rule-specified additional imports if defined.
+	if err = ip.handleRuleImports(root, rule.Imports, rule.Name, "file rule"); err != nil {
+		return err
+	}
+
```
y1yang0: Allowing imports in file.go makes more sense to me. I'd suggest we just keep updateImportConfigForFile and remove `handleRuleImports`.

tool/internal/instrument/imports.go
```diff
@@ -0,0 +1,99 @@
...
+//
+// This function validates that if a rule expects to use an import with a specific alias,
+// and the file already imports the same package with a different explicit alias, an error is returned.
+// This prevents silent failures where injected code uses an alias that doesn't exist in the file.
+func (ip *InstrumentPhase) handleRuleImports(
```
y1yang0: Optional suggestion: rename handleRuleImports to addRuleImports for clarity.

tool/internal/instrument/toolexec.go
```diff
@@ -105,19 +130,298 @@ func interceptCompile(ctx context.Context, args []string) ([]string, error) {
...
+			}
+		}
+	}
+
+	if !updated {
+		return nil
+	}
+
+	// Atomic write: write to temp file first
+	tempPath := ip.importConfigPath + ".tmp"
```
y1yang0: I don't think the backup and copy is needed here. It's single-threaded, and each package gets a dedicated importcfg, so there's nothing to protect against.

docs/rules.md
```diff
@@ -75,6 +102,23 @@ add_new_field:
...
+If your new struct fields use types from external packages, specify those imports:
+
+Example:
+
+```yaml
+add_context_field:
+  target: main
+  struct: MyStruct
+  new_field:
+    - name: Ctx
```
kakkoyun:
```suggestion
    - name: ctx
```

pkg/instrumentation/basic/basic.yaml
```diff
@@ -10,12 +10,23 @@ add_new_field:
   new_field:
     - name: NewField
       type: string
+    - name: Ctx
```
kakkoyun:
```suggestion
    - name: ctx
```

tool/util/go.go
```diff
@@ -11,28 +11,76 @@ import (
...
+			return true
+		}
+	}
+	return false
+}
+
+// IsCompileArgs checks if the args slice represents a compile command.
+// This is preferred over IsCompileCommand when you have the args as a slice,
+// as it correctly handles tool paths with spaces (common on Windows).
+func IsCompileArgs(args []string) bool {
```
kakkoyun: Being verbose wouldn't hurt here.
```suggestion
func IsCompileCommandWithArgs(args []string) bool {
```

tool/util/go.go
```diff
@@ -11,28 +11,76 @@ import (
...
+	// PGO compile command is different, skip it
+	if hasFlag(args, "-pgoprofile") {
+		return false
+	}
+
+	return true
+}
+
+// IsLinkArgs checks if the args slice represents a link command.
+func IsLinkArgs(args []string) bool {
```
kakkoyun:
```suggestion
func IsLinkCommandWithArgs(args []string) bool {
```

tool/util/go.go
```diff
@@ -87,6 +142,11 @@ func SplitCompileCmds(input string) []string {
 		args = append(args, arg.String())
 	}
 
+	// Handle unquoted Windows paths with spaces (e.g., from go build -x -n output)
+	// These look like: C:/Program Files/Go/pkg/tool/windows_amd64/compile.exe
+	// which gets incorrectly split into ["C:/Program", "Files/Go/pkg/tool/windows_amd64/compile.exe", ...]
+	args = mergeWindowsPathsWithSpaces(args)
+
 	// Fix the escaped backslashes on Windows
 	if IsWindows() {
```
kakkoyun, line 151: Not for this PR, but we should rather use build tags and move windows related things into a dedicated file.

tool/internal/setup/setup.go
```diff
@@ -228,6 +231,98 @@ func setupGoCache(ctx context.Context, env []string) ([]string, error) {
...
+			// Handle boolean flags in =value format (e.g., -race=true, -race=false)
+			// strconv.ParseBool accepts: 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False
+			if buildContextBoolFlags[flagName] {
+				if enabled, err := strconv.ParseBool(flagValue); err == nil {
+					boolFlagState[flagName] = boolFlagValue{set: true, value: enabled} // Last value wins
+				}
+				// Parse error: ignore invalid value
+				continue
+			}
+			continue
```
kakkoyun, line 292: Let's document this `continue`

tool/internal/instrument/toolexec.go
```diff
@@ -105,19 +130,298 @@ func interceptCompile(ctx context.Context, args []string) ([]string, error) {
...
+
+	var updated bool
+	for _, importPath := range newImports {
+		if importPath == "unsafe" || importPath == "C" {
+			// unsafe is built-in, C is the cgo pseudo-package; neither has an archive file
+			continue
+		}
+
+		if _, exists := ip.importConfig.PackageFile[importPath]; exists {
+			// Already have this import
```
kakkoyun, line 154: What happens if use adds an import with different alias and use it in the injected code? We need to point this case out for the user.

kakkoyun: This case might be handled in `imports.go`? 🤔

tool/internal/imports/imports.go
darccio: Let me check that.

tool/internal/instrument/toolexec.go
darccio: There is concurrency at `go build` level: https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies

```
-p n
	the number of programs, such as build commands or
	test binaries, that can be run in parallel.
	The default is GOMAXPROCS, normally the number of CPUs available.
```

I think this is needed to handle the `go build` concurrency.

y1yang0: The -p flag is for compiling multiple packages using multiple processes. For a single command invocation, such as the one below:

`/usr/local/go/pkg/tool/linux_amd64/compile ... -importcfg $WORK/b005/importcfg ...`

Since only a single process is handling it, the importcfg file can be safely modified in-place, without needing the new_temp + backup + rename sequence.

y1yang0:
> I think we should avoid adding dynamically dependencies that aren't in the go.mod.

Agreed

> We could introduce a pin feature that generates a gotelc.tool.go file for introducing a default set of to-be-injected dependencies - using blank imports - and allow users to customize it. It's how Orchestrion does. WDYT?

We already have a similar mechanism for this: otel.runtime.go

tool/internal/imports/imports.go
darccio: I think we should have a linter for getting this guideline consistently applied. I'll take care of that in another PR.

tool/internal/instrument/toolexec.go
darccio: Ok, I didn't realize that. Thanks!

darccio:
> We already have a similar mechanism for this: otel.runtime.go

Thanks @y1yang0, I'm still getting familiar with the codebase.

tool/internal/imports/imports.go
darccio: @y1yang0 They serve different purposes. `ResolvePackageInfo` allows to get the export archive file path (.a files), something that `setup.getBuildPackages` doesn't.

It's feasible to make `setup.getBuildPackages` return that information, but as it uses `packages.Load` that means to walk the graph instead of relying in the simpler output of `go list`.

tool/internal/instrument/toolexec.go
darccio: Done in 93a74219ae3b7def72c26f1a19f066b82c2a397d

tool/internal/imports/imports.go
darccio: I did a scan and I confirmed that most of them are used by another package:

  - ResolvePackageInfo, FindNew, AddToFile, CollectPaths, Resolution — used by instrument package
  - ParseImportCfg, ImportConfig, WriteFile — used by instrument package
  - CleanupImportTrackingFiles, Toolexec — used by setup and cmd packages
  - GetAddedImportsFileForProcess, GetAddedImportsPattern, GetBuildFlags, EncodeBuildFlags — used across instrument and setup
  - IsCompileArgs, IsLinkArgs — used across multiple packages

darccio: I did at least `ResolvePackageName` in a7f6a729b900e94c62c21a09e62cfed56182ca7a

tool/internal/instrument/imports.go
darccio: Done in 98d37031be73adb31faac709221e89fa6974b6b1

tool/internal/instrument/imports.go
darccio: Done in 7171508d03b5bc9ff2e7c344ad14abb00e881c03

tool/internal/instrument/apply_file.go
darccio: Done in ed07d5dc60b39708f452b02bfd3491aec58748d3

tool/internal/instrument/toolexec.go
darccio: Done in 60947fc50b4b59d35920f72d3dd3d236c67c7d80

pkg/instrumentation/basic/basic.yaml
darccio: Done in 8909293ac040e35ebfdbde0988daccdd4dddd1b2

tool/util/go.go
darccio: Done in [0b8b738](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/commit/0b8b738ff0c1721b2c83b9747dc5e6c5bfeb0d1c)

tool/util/go.go
darccio: Done in 0b8b738ff0c1721b2c83b9747dc5e6c5bfeb0d1c

docs/rules.md
darccio: Done in 8909293ac040e35ebfdbde0988daccdd4dddd1b2

tool/internal/setup/setup.go
darccio: Done in 4d4c099c41c901262d5e4368aabcf79c8f378d2a

tool/internal/instrument/toolexec.go
darccio: Let's do a quick example:

1. `FindNew` sees `crypto/sha256` that already exists; path not added to `NewImports`.
2. Validation checks `resolution.ExplicitAliases["crypto/sha256"]` is `false`.
3. This skips validation: `if !resolution.ExplicitAliases[importPath] { continue }`.
4. No import is added, no error is raised.
5. Injected code uses hasher.New() but only `sha256` is available, so... silent compile failure.

I've decided to implement a fix: when the alias is implicit, still validate that the rule alias matches the resolved package name.

Done in 6726e3ecfaab10457113cb323a2363646d7b663c.

tool/internal/imports/imports.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// ResolvePackageInfo attempts to retrieve the archive for the designated import path
+// and its dependencies using `go list -export -json`.
+// The buildFlags parameter allows passing build flags (like -tags) to ensure the resolved
+// archives match the current build context.
+func ResolvePackageInfo(ctx context.Context, importPath string, buildFlags ...string) (map[string]string, error) {
+	// Build the go list command with any build flags
+	args := make([]string, 0, 4+len(buildFlags)+1)
+	args = append(args, "list", "-export", "-json", "-deps")
```
kakkoyun: I wonder if we can use https://pkg.go.dev/golang.org/x/tools/go/packages instead of `go list ...`?

darccio: Done in 313b84bf770e18c04ecd568eeb147b103f9491f7

tool/internal/imports/imports.go
darccio: After getting questions about this, I decided to implement a `package.Load` solution that both functions use in 313b84bf770e18c04ecd568eeb147b103f9491f7.

tool/internal/setup/setup.go
```diff
@@ -258,6 +353,14 @@ func BuildWithToolexec(ctx context.Context, cmd *cli.Command) error {
 	util.Assert(pwd != "", "invalid working directory")
 	env = append(env, fmt.Sprintf("%s=%s", util.EnvOtelWorkDir, pwd))
 
+	// Extract and forward build flags that affect the build context
+	// This ensures `go list` resolves archives matching the current build
+	if buildFlags := extractBuildFlags(args); len(buildFlags) > 0 {
```
y1yang0, line 358: Thanks for this awesome contribution. Since we are no longer using go list (we're using package.Load instead), is this code still needed?

PR #292, opened 2026-02-11
feat(instrumentation): Add Redis instrumentation

NameHaibinZhang:
add redis(github.com/redis/go-redis/v9) support

## Motivation

Fixes #

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/demo/basic/main.go
+++ b/demo/basic/main.go
@@ -95,6 +95,7 @@ func main() {
 	m := &MyStruct{}
 	// Add a new field to the struct
 	m.NewField = "abc"
+	m.ctx = context.Background()
 	m.Example()
 	MyStruct{}.Example2()
 
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -8,6 +8,16 @@ All rules share a set of common fields that define the target of the instrumenta
 
 - `target` (string, required): The import path of the Go package to be instrumented. For example, `golang.org/x/time/rate` or `main` for the main package.
 - `version` (string, optional): Specifies a version range for the target package. The rule will only be applied if the package's version falls within this range. The format is `start_inclusive,end_exclusive`. For example, `v0.11.0,v0.12.0` means the rule applies to versions greater than or equal to `v0.11.0` and less than `v0.12.0`. If omitted, the rule applies to all versions.
+- `imports` (map[string]string, optional): A map of imports to inject into the instrumented file. The key is the import alias and the value is the import path. For standard imports without an alias, use the package name as both key and value. For blank imports, use "_" as the key.
+
+  Examples:
+
+  ```yaml
+  imports:
+    fmt: "fmt"                                    # Standard import: import "fmt"
+    ctx: "context"                                # Aliased import: import ctx "context"
+    _: "unsafe"                                   # Blank import: import _ "unsafe"
+  ```
 
 ---
 
@@ -46,6 +56,23 @@ hook_helloworld:
 
 This rule will inject `MyHookBefore` at the start of the `Example` function in the `main` package, and `MyHookAfter` at the end. The hook functions are located in the specified `path`.
 
+**Import Handling:**
+
+The function hook rule automatically imports the package specified in `path` for the hook functions. If your hook functions need additional imports beyond the hook package, you can specify them using the `imports` field.
+
+Example with additional imports:
+
+```yaml
+hook_with_imports:
+  target: main
+  func: Example
+  before: MyHookBefore
+  path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld"
+  imports:
+    log: "log"        # Additional import needed by the hook
+    _: "unsafe"       # Blank import if needed
+```
+
 ### 2. Struct Field Injection Rule
 
 This rule adds one or more new fields to a specified struct type.
@@ -75,6 +102,23 @@ add_new_field:
 
 This rule adds a new field named `NewField` of type `string` to the `MyStruct` struct in the `main` package.
 
+**Import Handling:**
+
+If your new struct fields use types from external packages, specify those imports:
+
+Example:
+
+```yaml
+add_context_field:
+  target: main
+  struct: MyStruct
+  new_field:
+    - name: ctx
+      type: context.Context
+  imports:
+    context: "context"
+```
+
 ### 3. Raw Code Injection Rule
 
 This rule injects a string of raw Go code at the beginning of a target function. This offers great flexibility but should be used with caution as the injected code is not checked for correctness at definition time.
@@ -103,6 +147,21 @@ raw_helloworld:
 
 This rule injects a new goroutine that prints "RawCode" at the start of the `Example` function in the `main` package.
 
+**Import Handling:**
+
+Raw code rules often need imports for the code being injected. Use the `imports` field to specify any packages your raw code references.
+
+Example:
+
+```yaml
+raw_with_logging:
+  target: main
+  func: Example
+  raw: 'log.Printf("Executing Example with arg: %v", someVar)'
+  imports:
+    log: "log"
+```
+
 ### 4. File Addition Rule
 
 This rule adds a new Go source file to the target package.
@@ -127,3 +186,18 @@ add_new_file:
 ```
 
 This rule would take the file `new_helpers.go` from the `github.com/my-org/my-repo/instrumentation/helpers` package and add it to the `main` package during compilation.
+
+**Import Handling:**
+
+File rules typically don't need the `imports` field because the added file already contains its own import declarations. However, you can use it to add additional imports to the file being added:
+
+Example:
+
+```yaml
+add_file_with_extra_imports:
+  target: main
+  file: "helpers.go"
+  path: "github.com/my-org/my-repo/instrumentation/helpers"
+  imports:
+    log: "log"  # Add extra import to the copied file
+```
--- a/pkg/instrumentation/basic/basic.yaml
+++ b/pkg/instrumentation/basic/basic.yaml
@@ -10,12 +10,23 @@ add_new_field:
   new_field:
     - name: NewField
       type: string
+    - name: ctx
+      type: context.Context
+  imports:
+    context: "context"
 
 raw_helloworld:
   target: main
   func: Example
   raw: |
-    go func(){ println("RawCode") }()
+    go func(){
+      h := sha256.New()
+      h.Write([]byte("RawCode"))
+      fmt.Printf("RawCode: %x\n", h.Sum(nil))
+    }()
+  imports:
+    fmt: "fmt"
+    sha256: "crypto/sha256"
 
 hook_recv:
   target: main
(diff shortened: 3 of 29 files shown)


pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+	return &otelRedisHook{
+		Addr: addr,
+	}
+}
+
+func (o *otelRedisHook) ProcessHook(next redis.ProcessHook) redis.ProcessHook {
+	return func(ctx context.Context, cmd redis.Cmder) error {
+		if !redisEnabler.Enable() {
+			logger.Debug("Redis Client instrumentation disabled")
+			return nil
```
y1yang0: Maybe you should next(ctx, cmd) even if hook is disabled?

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+
+		return nil
+	}
+}
+
+func (o *otelRedisHook) ProcessPipelineHook(next redis.ProcessPipelineHook) redis.ProcessPipelineHook {
+	return func(ctx context.Context, cmds []redis.Cmder) error {
+		if !redisEnabler.Enable() {
+			logger.Debug("Redis Client instrumentation disabled")
+			return nil
```
y1yang0: Maybe you should next(ctx, cmd) even if hook is disabled?

y1yang0 (approved):
Otherwise LGTM. Please resolve all the comments from the Copilot bot

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,259 @@
...
+			trace.WithAttributes(attrs...),
+		)
+		defer span.End()
+
+		if err := next(ctx, cmd); err != nil && err != redis.Nil {
+			span.SetStatus(codes.Error, err.Error())
+			return err
+		}
+
+		return nil
```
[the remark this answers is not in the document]
NameHaibinZhang: redis nil  user not think this is err,so not set status,but return always err, not change the origin return

tushar1977:
Hi @NameHaibinZhang Could you tell me the purpose of ```client_hook.go``` and ```hook.go```?

kakkoyun (commented):
Sorry for the late review — catching up after merge. See inline comments.

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
```
kakkoyun, line 106: This swallows `redis.Nil` instead of propagating it. The condition correctly skips marking the span as error, but then falls through to `return nil` — so the caller never sees `redis.Nil` and can't distinguish a cache miss from success.

```go
// fix
err := next(ctx, cmd)
if err != nil && err != redis.Nil {
}
return err
```

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+			trace.WithAttributes(attrs...),
+		)
+		defer span.End()
+
+		if err := next(ctx, cmds); err != nil && err != redis.Nil {
+			span.SetStatus(codes.Error, err.Error())
+			return err
+		}
+
+		return nil
```
kakkoyun, line 154: Same `redis.Nil` propagation bug as `ProcessHook` — returns `nil` instead of `redis.Nil` on a pipeline cache miss.

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+		b = append(b, ": "...)
+		b = redisV9AppendArg(b, cmd.Name())
+	}
+
+	return redisV9String(b)
+}
+
+func redisV9String(b []byte) string {
+	return *(*string)(unsafe.Pointer(&b))
+}
```
kakkoyun, line 193: The old pre-1.17 `*(*string)(unsafe.Pointer(&b))` trick. Since Go 1.20 use `unsafe.String(unsafe.SliceData(b), len(b))`, or just `string(b)` — the allocation is negligible compared to the Redis call.

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+
+func redisV9Bytes(s string) []byte {
+	return *(*[]byte)(unsafe.Pointer(
+		&struct {
+			string
+			Cap int
+		}{s, len(s)},
+	))
+}
```
kakkoyun, line 207: Same issue — the struct-based `*(*[]byte)(unsafe.Pointer(...))` hack. Replace with `unsafe.Slice(unsafe.StringData(s), len(s))` or just `[]byte(s)`.

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+		b = append(b, ": "...)
+		b = append(b, err.Error()...)
+	}
+
+	if cmd, ok := cmd.(*redis.Cmd); ok {
+		b = append(b, ": "...)
+		b = redisV9AppendArg(b, cmd.Name())
+	}
+
+	return redisV9String(b)
```
kakkoyun, line 188: This records every argument verbatim, so `SET session:abc supersecrettoken` ends up in the trace as `db.query.text`. Redis is commonly used for sessions and tokens — needs at minimum a prominent doc warning, ideally a sanitizer option that's opt-out for full args.

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+
+	if err := cmd.Err(); err != nil && err != redis.Nil {
+		b = append(b, ": "...)
+		b = append(b, err.Error()...)
+	}
+
+	if cmd, ok := cmd.(*redis.Cmd); ok {
+		b = append(b, ": "...)
+		b = redisV9AppendArg(b, cmd.Name())
+	}
```
kakkoyun, line 186: Two issues here. The error string is already on the span status — putting it in `db.query.text` is noise. And `cmd.Name()` is already the first element of `cmd.Args()`, so it's already in the statement. Both blocks should go.

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (o *otelRedisHook) DialHook(next redis.DialHook) redis.DialHook {
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+		conn, err := next(ctx, network, addr)
+		if err != nil {
+			return nil, err
+		}
+		return conn, err
+	}
+}
```
kakkoyun, line 166: This is just `return next` with extra steps — the wrapper does nothing. Either instrument dial (connect-time span) or `return next` directly.

pkg/instrumentation/redis/v9/hook.go
```diff
@@ -0,0 +1,260 @@
...
+
+func redisV9String(b []byte) string {
+	return *(*string)(unsafe.Pointer(&b))
+}
+
+func redisV9AppendUTF8String(dst, src []byte) []byte {
+	dst = append(dst, src...)
+	return dst
+}
```
kakkoyun, line 198: This is just `append(dst, src...)`. The name implies UTF-8 validation that isn't happening. Inline it or rename it.

pkg/instrumentation/redis/v9/client_hook.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func afterNewClusterClientV9(call inst.HookContext, client *redis.ClusterClient) {
+	client.OnNewNode(func(rdb *redis.Client) {
+		rdb.AddHook(newOtelRedisHook(rdb.Options().Addr))
+	})
+}
+
+func afterNewSentinelClientV9(call inst.HookContext, client *redis.SentinelClient) {
+	client.AddHook(newOtelRedisHook(client.String()))
```
kakkoyun, line 47: `SentinelClient.String()` returns something like `"SentinelClient(...)"`, not `host:port`. `net.SplitHostPort` will fail on it and `server.address` ends up with the full string.

pkg/instrumentation/redis/v9/client_hook.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func afterClientConnV9(call inst.HookContext, client *redis.Conn) {
+	client.AddHook(newOtelRedisHook(client.String()))
```
kakkoyun, line 51: Same as `SentinelClient` above — `Conn.String()` returns `"Redis<addr db:N>"`, not a parseable address.

pkg/instrumentation/redis/v9/client_hook.go
```diff
@@ -0,0 +1,52 @@
...
+// redisClientEnabler controls whether client instrumentation is enabled
+type redisClientEnabler struct{}
+
+func (g redisClientEnabler) Enable() bool {
+	return shared.Instrumented(instrumentationKey)
+}
+
+var redisEnabler = redisClientEnabler{}
+
+func afterNewRedisClientV9(ictx inst.HookContext, client *redis.Client) {
```
kakkoyun, line 26: `ictx` here, `call` everywhere else. Let's pick one of them.

pkg/instrumentation/redis/semconv/client.go
```diff
@@ -0,0 +1,42 @@
...
+		host = req.Endpoint
+	}
+
+	attrs := []attribute.KeyValue{
+		semconv.DBSystemNameRedis,
+		semconv.DBOperationName(req.FullName),
+		semconv.ServerAddress(host),
+		semconv.NetworkTransportTCP,
+		semconv.DBQueryText(req.Statement),
+	}
```
kakkoyun, line 33: Missing `db.namespace` — the Redis DB index is part of the OTel DB semconv and useful for grouping queries by database. It's available via `client.Options().DB` on `*redis.Client`. Would need to add a `DB int` field to `RedisRequest`.

pkg/instrumentation/redis/v9/hook_test.go
```diff
@@ -0,0 +1,408 @@
...
+	assert.Equal(t, "get", span.Name())
+
+	// Verify attributes
+	attrMap := make(map[string]interface{})
+	for _, attr := range span.Attributes() {
+		attrMap[string(attr.Key)] = attr.Value.AsInterface()
+	}
+	assert.Equal(t, "redis", attrMap["db.system.name"])
+	assert.Equal(t, "get", attrMap["db.operation.name"])
+	assert.Equal(t, "localhost:6379", attrMap["network.peer.address"])
```
kakkoyun, line 201: Two bugs in this assertion. The attribute key is `server.address` (set by `semconv.ServerAddress`), not `network.peer.address`. And the expected value should be `"localhost"` — `net.SplitHostPort` strips the port before passing it to `ServerAddress`.

pkg/instrumentation/redis/v9/hook_test.go
```diff
@@ -0,0 +1,408 @@
...
+	defer tp.Shutdown(context.Background())
+
+	hook := newOtelRedisHook("localhost:6379")
+	processHook := hook.ProcessHook(func(ctx context.Context, cmd redis.Cmder) error {
+		return redis.Nil
+	})
+
+	cmd := redis.NewCmd(context.Background(), "get", "nonexistent")
+	err := processHook(context.Background(), cmd)
+	assert.NoError(t, err)
```
kakkoyun, line 243: This asserts `NoError`, which passes because the hook incorrectly swallows `redis.Nil`. Should be `assert.ErrorIs(t, err, redis.Nil)` to validate that the sentinel is properly propagated.

PR #297, opened 2026-02-12
feat(instrumentation): Add support for slog and zap logging instrumentation

kestrelcjx:
## Description

Add support for slog and zap logging instrumentation
## Motivation

---

## Checklist

- [x] Tests pass: `make test`


pkg/instrumentation/shared/setup.go
```diff
@@ -252,6 +260,27 @@ func setupMeterProvider(ctx context.Context, res *resource.Resource) error {
 	return nil
 }
 
+// setupLoggerProvider creates and configures the logger provider
+func setupLoggerProvider(ctx context.Context, res *resource.Resource) error {
```
kakkoyun, line 264: Missing endpoint guard

`setupTraceProvider` and `setupMeterProvider` both check for `OTEL_EXPORTER_OTLP_ENDPOINT` before creating an exporter, returning early (with a nil provider) when no endpoint is configured. This function creates a log exporter unconditionally.

This means any app instrumented with slog/zap will attempt to export logs even when there's no collector running, which either fails noisily or silently drops logs.

Suggestion:
```go
endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
if endpoint == "" {
    return nil // No collector configured, skip log exporter
}
```

pkg/instrumentation/log/slog/slog.yaml
```diff
@@ -0,0 +1,25 @@
+handler_wrapper:
+  target: "log/slog"
+  file: "handler_wrapper.go"
+  path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/log/slog/internal"
+
+slog_handler_hook:
+  target: "log/slog"
+  func: Handler
```
kakkoyun, line 8: Hooking `Handler()`, a getter with side effects

`Handler()` is semantically a read-only accessor. Users call it to inspect the logger's current handler (e.g. for testing, wrapping, or passing to child loggers). Hooking it here means any such call will permanently mutate the logger by wrapping its handler — which is a significant surprise.

Is there a way to inject the wrapper at a point that corresponds to actual log emission (e.g. hooking `log`, `logAttrs`, or `output`) or at logger construction time instead? The current approach is observable from user code in ways that could cause subtle bugs.

pkg/instrumentation/log/slog/internal/handler_wrapper.go
```diff
@@ -0,0 +1,8 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package slog
+
+func (l *Logger) WrapHandler(wrapperFunc func(Handler) Handler) {
```
kakkoyun, line 6: Permanent mutation inside a getter

Every call to `logger.Handler()` invokes this, which permanently replaces `l.handler` with the wrapped version. Since `wrapperFunc` checks for double-wrapping (`if _, ok := h.(*HandlerWrapper); ok { return h }`), repeated calls are idempotent — but the mutation on every read is still surprising.

More critically: if user code calls `logger.Handler()` for inspection (a totally reasonable thing to do), this modifies the logger as a side effect. That violates the principle of least surprise.

pkg/instrumentation/log/slog/internal/types.go
```diff
@@ -0,0 +1,14 @@
...
+
+package slog
+
+import (
+	logslog "log/slog"
+)
+
+type Handler logslog.Handler
+
+type Logger struct {
```
kakkoyun, line 12: Coupling to private `slog.Logger` struct layout

This mirrors the private `handler` field of `log/slog.Logger`. If the Go stdlib renames or reorders this field, the instrumentation will break — not at development time, but at instrumentation time (when the tool injects the file), and potentially silently.

Two things would help:
1. Add a comment documenting the dependency: `// Mirrors log/slog.Logger as of Go 1.21+. Field name and type must stay in sync.`
2. Add a `version` constraint in `slog.yaml` so the rule only applies to known-good Go versions
3. Ideally, a test that fails at CI time if the struct layout changes

pkg/instrumentation/log/slog/logger_hook.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	instrumentationName = "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/log"
+	instrumentationKey  = "LOG"
+)
+
+var (
+	logger           = shared.Logger()
+	initializerStart = new(atomic.Bool)
+	handlerProvier   = new(atomic.Value)
```
kakkoyun, line 24: typo: `handlerProvier` → `handlerProvider`

This appears multiple times in this file. Please fix — it affects grep-ability and readability.

pkg/instrumentation/log/slog/logger_hook.go
```diff
@@ -0,0 +1,121 @@
...
+		// Start runtime metrics (respects OTEL_GO_ENABLED/DISABLED_INSTRUMENTATIONS)
+		if err := shared.StartRuntimeMetrics(); err != nil {
+			logger.Error("failed to start runtime metrics", "error", err)
+		}
+
+		logger.Info("slog logger instrumentation initialized")
+	}
+}
+
+// slogLoggerEnabler controls whether server instrumentation is enabled
```
kakkoyun, line 64: Copy-paste artifact

```go
// slogLoggerEnabler controls whether server instrumentation is enabled
```

This should say "slog logger instrumentation", not "server instrumentation". Same issue exists in `zap/logger_hook.go`.

pkg/instrumentation/log/slog/types.go
```diff
@@ -0,0 +1,38 @@
...
+	logslog "log/slog"
+)
+
+type HandlerWrapper struct {
+	logslog.Handler
+	otelHandler logslog.Handler
+}
+
+func (w *HandlerWrapper) Enabled(ctx context.Context, level logslog.Level) bool {
+	_ = w.otelHandler.Enabled(ctx, level)
```
kakkoyun, line 17: OTel handler's `Enabled()` result is silently discarded

```go
_ = w.otelHandler.Enabled(ctx, level)  // result ignored
return w.Handler.Enabled(ctx, level)
```

This means that if the OTel handler is configured with a higher log level than the original handler, records below the OTel threshold will still be passed to `w.otelHandler.Handle()` — wasting resources and producing incorrect behavior.

Suggestion: only call `Handle()` on the OTel handler when it reports itself as enabled:
```go
func (w *HandlerWrapper) Handle(ctx context.Context, r logslog.Record) error {
    if w.otelHandler.Enabled(ctx, r.Level) {
        _ = w.otelHandler.Handle(ctx, r)
    }
    return w.Handler.Handle(ctx, r)
}
```

pkg/instrumentation/log/slog/types.go
```diff
@@ -0,0 +1,38 @@
...
+	return w.Handler.Enabled(ctx, level)
+}
+
+func (w *HandlerWrapper) Handle(ctx context.Context, r logslog.Record) error {
+	_ = w.otelHandler.Handle(ctx, r)
```
kakkoyun, line 22: OTel handler error silently dropped

```go
_ = w.otelHandler.Handle(ctx, r)  // error swallowed
```

If the OTel handler fails (e.g. exporter misconfigured, context cancelled), the error is discarded with no feedback. At minimum, consider logging it at debug level:
```go
if err := w.otelHandler.Handle(ctx, r); err != nil {
    shared.Logger().Debug("otel handler error", "error", err)
}
```

pkg/instrumentation/log/zap/logger_hook.go
```diff
@@ -0,0 +1,80 @@
...
+		// Start runtime metrics (respects OTEL_GO_ENABLED/DISABLED_INSTRUMENTATIONS)
+		if err := shared.StartRuntimeMetrics(); err != nil {
+			logger.Error("failed to start runtime metrics", "error", err)
+		}
+
+		logger.Info("zap logger instrumentation initialized")
+	})
+}
+
+// zapLoggerEnabler controls whether server instrumentation is enabled
```
kakkoyun, line 62: Copy-paste artifact

```go
// zapLoggerEnabler controls whether server instrumentation is enabled
```

Should say "zap logger instrumentation", not "server instrumentation".

demo/log/zap/main.go
```diff
@@ -0,0 +1,29 @@
...
+	}()
+
+	ctx := context.Background()
+
+	logger, err := zap.NewProductionConfig().Build()
+	if err != nil {
+		panic(err)
+	}
+
+	logger.With(zap.Any("ctx", ctx)).
```
kakkoyun, line 26: Context passed as a field value, not for trace correlation

```go
logger.With(zap.Any("ctx", ctx))
```

`zap.Any("ctx", ctx)` serializes the context object as a log field — it doesn't propagate trace context for OTel log-trace correlation. Since this is a compile-time instrumentation demo (where correlation should happen automatically), consider removing the manual ctx field entirely and just showing:
```go
logger.With(zap.Int32("id", 12345)).Info("hello world!")
```
This better demonstrates the "no-code-change" value of compile-time instrumentation.

Makefile
```diff
@@ -52,7 +52,7 @@ all: build format lint test
 build/pkg: ## Build all pkg modules to verify compilation
 	@echo "Building pkg modules..."
 	@set -euo pipefail
-	@PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | grep -v "pkg/instrumentation/runtime"); \
+	@PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | grep -v -e "pkg/instrumentation/runtime" -e "pkg/instrumentation/log/slog"); \
```
kakkoyun, line 55: slog module excluded from build target without explanation

The `runtime` package exclusion is understandable (it uses unsafe patterns). But the slog exclusion is new and undocumented. The zap module is **not** excluded, which makes the inconsistency more visible.

If this is because `pkg/instrumentation/log/slog` uses file injection that only compiles correctly with the instrumentation tool applied (i.e. `internal/handler_wrapper.go` references `slog.Logger.handler` which doesn't exist in the stdlib package without the tool), please add a comment:
```makefile
# slog module excluded: uses file injection that requires the instrumentation tool to compile
```

kakkoyun (changes requested):
Thanks for the PR! 

The **zap implementation** is clean and follows our existing hook patterns really well — `BeforeWrite` hooking `CheckedEntry.Write` and adding an OTel core via `AddCore` is a natural fit with zap's extensibility model. Nice.

The **slog implementation** gives me more pause. It introduces two novel patterns for this project: injecting a method into `log/slog` via file rules to access private fields, and wrapping the handler on every call to `Handler()` rather than at span/log emission time. These are interesting ideas, but there are some correctness and robustness issues I'd like to resolve before merging.

**Must fix before merge:**
- Missing endpoint guard in `setupLoggerProvider` — inconsistent with how `setupTraceProvider` and `setupMeterProvider` behave
- `BeforeHandler` hooks the `Handler()` getter, which is semantically read-only — any call to `logger.Handler()` for inspection or testing permanently mutates the logger, which will surprise users
- `internal/types.go` mirrors `slog.Logger`'s private struct layout without any Go version constraint or comment — a stdlib rename silently breaks the instrumentation
- No tests for `HandlerWrapper` or integration tests — every other instrumentation in this repo has them

**Quick fixes:**
- Typo: `handlerProvier` → `handlerProvider` (appears ~6 times)
- Copy-paste artifact: enabler struct comments still say "server instrumentation" in both slog and zap
- `Enabled()` discards the OTel handler's result, so records below OTel's threshold still get passed to `Handle()`
- `Handle()` swallows the OTel handler's error silently

**Nit:** The slog module is silently excluded from the `build/pkg` Makefile target. If that's intentional (because the module uses struct field injection that only works with the instrumentation tool), please add a comment so the next reader isn't left wondering.

PR #301, opened 2026-02-25
feat(instrumentation): Add MySQL instrumentation

NameHaibinZhang:
Change-Id: I26a4712058094769c327bbaf0da430fe0f3e83ab

Add Db Sql Instrumentation

## Motivation

Fixes #

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/test/e2e/db_test.go
+++ b/test/e2e/db_test.go
@@ -35,9 +35,9 @@ func TestDB(t *testing.T) {
 		t.Fatalf("Expected at least 7 spans, got %d", len(spans))
 	}
 
-	// For "testdb" driver, parseDSN falls back to the full DSN as the endpoint.
-	// net.SplitHostPort cannot parse a raw DSN, so server.address = full DSN, no port.
-	const serverAddr = "user:pass@tcp(127.0.0.1:3306)/testdb?charset=utf8"
+	// For "testdb" driver, parseDSN returns an error (unknown driver),
+	// so beforeOpenInstrumentation falls back to "unknown" as the endpoint.
+	const serverAddr = "unknown"
 
 	// Verify ping span
 	pingSpan := testutil.RequireSpan(t, f.Traces(),
--- a/test/integration/db_client_test.go
+++ b/test/integration/db_client_test.go
@@ -23,7 +23,7 @@ func TestDBClientPing(t *testing.T) {
 	testutil.RequireDBClientSemconv(t, span,
 		"PING",
 		"ping",
-		"user:pass@tcp(127.0.0.1:3306)/testdb?charset=utf8", 0,
+		"unknown", 0,
 		"testdb",
 	)
 }
@@ -38,7 +38,7 @@ func TestDBClientExec(t *testing.T) {
 	testutil.RequireDBClientSemconv(t, span,
 		"INSERT",
 		"INSERT INTO users (name, email) VALUES (?, ?)",
-		"user:pass@tcp(127.0.0.1:3306)/testdb?charset=utf8", 0,
+		"unknown", 0,
 		"testdb",
 	)
 }
@@ -53,7 +53,7 @@ func TestDBClientQuery(t *testing.T) {
 	testutil.RequireDBClientSemconv(t, span,
 		"SELECT",
 		"SELECT id, name FROM users WHERE name = ?",
-		"user:pass@tcp(127.0.0.1:3306)/testdb?charset=utf8", 0,
+		"unknown", 0,
 		"testdb",
 	)
 }


pkg/instrumentation/databasesql/semconv/db.go
```diff
@@ -0,0 +1,32 @@
...
+	Endpoint   string
+	DriverName string
+	Dsn        string
+	Params     []any
+	DbName     string
+}
+
+func DbClientRequestTraceAttrs(req DatabaseSqlRequest) []attribute.KeyValue {
+	attrs := []attribute.KeyValue{
+		semconv.DBSystemNameMongoDB,
```
y1yang0: Should not semconv.DBSystemNameMongoDB

pkg/instrumentation/databasesql/parse.go
```diff
@@ -0,0 +1,232 @@
...
+	u, err := nurl.Parse(url)
+	if err != nil {
+		return "", err
+	}
+
+	if u.Scheme != "postgres" && u.Scheme != "postgresql" {
+		return "", fmt.Errorf("invalid connection protocol: %s", u.Scheme)
+	}
+
+	return u.Host + ":" + u.Port(), nil
```
y1yang0: What if `u.Port()` is empty? It returns "hostname:" ? which seems wrong.

NameHaibinZhang: fix

pkg/instrumentation/databasesql/semconv/db.go
NameHaibinZhang: fix

Makefile
```diff
@@ -52,7 +52,7 @@ all: build format lint test
 build/pkg: ## Build all pkg modules to verify compilation
 	@echo "Building pkg modules..."
 	@set -euo pipefail
-	@PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | grep -v "pkg/instrumentation/runtime"); \
+	@PKG_MODULES=$$(find pkg -name "go.mod" -type f -exec dirname {} \; | grep -v "pkg/instrumentation/runtime" | grep -v "pkg/instrumentation/databasesql"); \
 	for moddir in $$PKG_MODULES; do \
 		echo "Building $$moddir..."; \
 		(cd $$moddir && go mod tidy && go build ./...); \
```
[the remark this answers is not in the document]
NameHaibinZhang: db instrument is add new obj to db struct, so exclude

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,677 @@
...
+func (n dbClientEnabler) Enable() bool {
+	return shared.Instrumented(instrumentationKey)
+}
+
+var clientEnabler = dbClientEnabler{}
+
+func beforeOpenInstrumentation(ictx inst.HookContext, driverName, dataSourceName string) {
+	addr, err := parseDSN(driverName, dataSourceName)
+	if err != nil {
+		addr = dataSourceName
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

demo/db/server/main.go
```diff
@@ -0,0 +1,392 @@
...
+		level = slog.LevelWarn
+	case "error":
+		level = slog.LevelError
+	default:
+		level = slog.LevelInfo
+	}
+	logger = slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: level}))
+
+	logger.Info("connecting to mysql", "dsn", *dsn)
+
```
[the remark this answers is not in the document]
NameHaibinZhang: demo

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,677 @@
...
+	if conn == nil {
+		return
+	}
+	instrumentStart(ictx, ctx, "start", "START TRANSACTION", conn.Endpoint, conn.DriverName, conn.DSN, conn.DbName)
+}
+
+func afterConnTxInstrumentation(ictx inst.HookContext, tx *sql.Tx, err error) {
+	if !clientEnabler.Enable() {
+		return
+	}
```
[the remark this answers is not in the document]
NameHaibinZhang: not use change tx

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,677 @@
...
+	dsn, ok := data["dsn"]
+	if ok {
+		stmt.Data["dsn"] = dsn
+	}
+	dbName, ok := data["dbName"]
+	if ok {
+		stmt.Data["dbName"] = dbName
+	}
+}
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,677 @@
...
+
+func calOp(sql string) string {
+	sqls := strings.Split(sql, " ")
+	var op string
+	if len(sqls) > 0 {
+		op = sqls[0]
+	}
+	return op
```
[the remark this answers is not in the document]
NameHaibinZhang: ok

demo/infrastructure/docker-compose/docker-compose.yml
```diff
@@ -182,6 +182,41 @@ services:
...
+      context: ../../..
+      dockerfile: demo/db/server/Dockerfile
+    container_name: db-server
+    ports:
+      - "8081:8081"
+    command: ["-port", "8081", "-dsn", "demo:demo@tcp(mysql:3306)/demodb?parseTime=true"]
+    environment:
+      - OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4318
+      - OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
+      - OTEL_SERVICE_NAME=db-server
```
kakkoyun, line 219: **`db-server` service is incomplete**

The `db-server` block ends here without `networks`, `depends_on`, or `restart` that every other service in this file has. Compare with `redis-client` or `grpc-server` — they all declare:

```yaml
    networks:
      - observability
    depends_on:
      mysql:
        condition: service_healthy
    restart: unless-stopped
```

Without `networks: observability`, the container cannot resolve `mysql` by hostname, so the DSN `demo:demo@tcp(mysql:3306)/demodb` will fail immediately on startup. Without `depends_on`, the server may start before MySQL is healthy.

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,681 @@
...
+// dbClientEnabler controls whether client instrumentation is enabled
+type dbClientEnabler struct{}
+
+func (n dbClientEnabler) Enable() bool {
+	return shared.Instrumented(instrumentationKey)
+}
+
+var clientEnabler = dbClientEnabler{}
+
+func beforeOpenInstrumentation(ictx inst.HookContext, driverName, dataSourceName string) {
```
kakkoyun, line 42: **Missing `clientEnabler.Enable()` guard**

Every other `before*Instrumentation` function in this file starts with:
```go
if !clientEnabler.Enable() {
    return
}
```
but `beforeOpenInstrumentation` skips this check. This means DSN parsing and `ictx.SetData` run even when instrumentation is disabled. The data gets consumed by `afterOpenInstrumentation`, so it is not a correctness bug today — but it breaks the consistent pattern and will silently do work when disabled.

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,681 @@
...
+	addr, err := parseDSN(driverName, dataSourceName)
+	if err != nil {
+		addr = "unknown"
+	}
+	dbName := ParseDbName(dataSourceName)
+	ictx.SetData(map[string]string{
```
kakkoyun, line 48: **Raw DSN (with credentials) stored in hook context**

```go
"dsn": dataSourceName,
```

The DSN typically contains the username and password (e.g. `demo:demo@tcp(mysql:3306)/demodb`). Storing the raw DSN propagates credentials into span attributes via `semconv.DatabaseSqlRequest.Dsn`. This is a sensitive data leak.

Consider scrubbing credentials from the DSN before storing it, or not storing the DSN at all — the address, driver name, and db name can be parsed once during `Open` and stored separately (as is already done for `endpoint` and `dbName`).

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,681 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package db
```
kakkoyun, line 4: **Package name `db` does not match directory `databasesql`**

Go convention is that the package name matches (or is an obvious abbreviation of) the directory name. Here the directory is `databasesql` but `package db` is declared. Compare with the project's other instrumentations: `nethttp` → `package nethttp`, `grpc` → `package grpc`.

This breaks `go doc` navigation and surprises users importing the package. Suggest renaming to `package databasesql` (or `package dbsql` as an abbreviation), updating the `parse.go` package declaration to match.

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,681 @@
...
+}
+
+func beforePrepareContextInstrumentation(ictx inst.HookContext, db *sql.DB, ctx context.Context, query string) {
+	if !clientEnabler.Enable() {
+		return
+	}
+	if db == nil {
+		return
+	}
+	ictx.SetData(map[string]string{
```
kakkoyun, line 106: **Inconsistent data map types between hook pairs**

`beforePrepareContextInstrumentation` stores `map[string]string{...}` here, but `instrumentStart` (called by most other before-hooks) stores `map[string]interface{}{...}`.

This inconsistency means any generic helper that tries to read hook data (e.g. via `ictx.GetKeyData`) must handle two different type assertions. It also means `instrumentEnd`'s `ictx.GetKeyData("span")` call will silently return nil for the prepare path since the span is never stored there.

Suggest either having prepare also call `instrumentStart`, or standardising on one map type throughout.

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,681 @@
...
+}
+
+func beforeTxCommitInstrumentation(ictx inst.HookContext, tx *sql.Tx) {
+	if !clientEnabler.Enable() {
+		return
+	}
+	if tx == nil {
+		return
+	}
+	instrumentStart(ictx, context.Background(), "commit", "COMMIT", tx.Endpoint, tx.DriverName, tx.DSN, tx.DbName)
```
kakkoyun, line 508: **`context.Background()` creates orphaned spans for Commit/Rollback**

```go
instrumentStart(ictx, context.Background(), "commit", "COMMIT", tx.Endpoint, ...)
```

Using `context.Background()` means the Commit/Rollback span has no parent — it will appear as a root span in the trace, disconnected from the transaction that initiated it. The same issue affects `beforeTxRollbackInstrumentation` at line 525.

The `Tx` struct should carry (or be derivable from) the context from when `BeginTx` was called. Look at how `grpc` or `nethttp` propagates context through the transaction — the context from `beforeBeginTxInstrumentation` should be stored and retrieved here.

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,681 @@
...
+		Dsn:        dsn,
+		Params:     args,
+		DbName:     dbName,
+	}
+	// Get trace attributes from semconv
+	attrs := semconv.DbClientRequestTraceAttrs(req)
+
+	// Start span
+	ctx, span := tracer.Start(ctx,
+		req.OpType,
```
kakkoyun, line 602: **Span name should follow OTel semantic conventions**

```go
ctx, span := tracer.Start(ctx,
    req.OpType,  // e.g. "SELECT", "INSERT"
```

The [OTel DB span naming spec](https://opentelemetry.io/docs/specs/semconv/database/database-spans/) says the span name should be `{db.operation.name} {db.namespace}` when both are available, e.g. `"SELECT demodb"`. Using only the SQL verb makes traces harder to filter by database in observability tools.

Suggest:
```go
spanName := req.OpType
if req.DbName != "" {
    spanName = req.OpType + " " + req.DbName
}
ctx, span := tracer.Start(ctx, spanName, ...)
```

pkg/instrumentation/databasesql/client.go
```diff
@@ -0,0 +1,681 @@
...
+		logger.Debug("instrumentEnd: no span from before hook")
+		return
+	}
+	defer span.End()
+	if err != nil {
+		span.SetStatus(codes.Error, err.Error())
+	}
+}
+
+func calOp(sql string) string {
```
kakkoyun, line 635: **`calOp` naming and redundant guard**

Two issues:
1. The name `calOp` is cryptic — `extractSQLVerb` or `sqlOperation` would be clearer.
2. The `if trimmed == ""` check is redundant: `strings.Fields("")` returns an empty slice, so the subsequent `if len(fields) == 0` already handles this. The two guards collapse to one.

Simplified:
```go
func extractSQLVerb(sql string) string {
    fields := strings.Fields(sql)
    if len(fields) == 0 {
        return ""
    }
    return strings.ToUpper(fields[0])
}
```

pkg/instrumentation/databasesql/semconv/db.go
```diff
@@ -0,0 +1,56 @@
...
+		semconv.DBQueryText(req.Sql),
+	}
+
+	if err == nil {
+		if port, convErr := strconv.Atoi(portStr); convErr == nil && port > 0 {
+			attrs = append(attrs, semconv.ServerPort(port))
+		}
+	}
+
+	switch req.DriverName {
```
kakkoyun, line 44: **Missing `db.system.name` mappings for ClickHouse, Oracle, SQL Server**

The `parseDSN` in `parse.go` handles `clickhouse`, `godror`/`oracle`/`oci8`, and `mssql`/`sqlserver` drivers, but the switch here only maps `mysql`, `postgres`, and `sqlite3` — everything else falls through to `DBSystemNameOtherSQL`.

This means ClickHouse, Oracle, and SQL Server spans will have `db.system.name = "other_sql"` instead of the correct OTel value. Suggest adding:
```go
case "clickhouse":
    attrs = append(attrs, semconv.DBSystemNameClickHouse)
case "godror", "oracle", "oci8", "go-oci8":
    attrs = append(attrs, semconv.DBSystemNameOracle)
case "mssql", "sqlserver":
    attrs = append(attrs, semconv.DBSystemNameMicrosoftSQLServer)
```

pkg/instrumentation/databasesql/parse.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+
+			break
+		}
+	}
+	return name
+}
+
+func parseDSN(driverName, dsn string) (addr string, err error) {
+	// TODO: need a more delegate DFA
```
kakkoyun, line 34: **Typo in comment**

```go
// TODO: need a more delegate DFA
```

"delegate" should be "dedicated" (or "elaborate" / "complete"). Suggest: `// TODO: need a more complete DSN parser`.

pkg/instrumentation/databasesql/client.go
y1yang0: FYI: SetKeyData/GetKeyData maybe a more better choice?

PR #302, opened 2026-02-25
feat(instrumentation): Add Kafka instrumentation

### LINKED ISSUE #296: feat(instrumentation): Support Kafka instrumentation

### Package

I'll start planning and finalize the code structure in next couple of days. will post here for review

### Version

_No response_

### Environment

Kafka

### Use-case

_No response_

### Telemetry

_No response_

### Anything else?

_No response_

### Tip

tushar1977:
## Description

Adds compile-time instrumentation for [`segmentio/kafka-go`](https://github.com/segmentio/kafka-go), covering both the producer and consumer paths.

**Producer** — instruments `(*kafka.Writer).WriteMessages`:
- Starts a `SpanKindProducer` span per message (`<topic> send`).
- Injects W3C trace context into each message's headers so downstream consumers can continue the trace.
- Records error status when the write fails.

**Consumer** — instruments `(*kafka.Reader).ReadMessage`:
- Extracts the propagated trace context from message headers and starts a `SpanKindConsumer` span (`<topic> receive`) linked to the producer.
- Records error status on read failures.

**Semantic conventions** — emits OTel messaging attributes (semconv v1.37.0): `messaging.system=kafka`, `messaging.operation.name`/`type`, `messaging.destination.name`, `server.address`/`server.port`, `messaging.kafka.message.key`, `messaging.destination.partition.id`, `messaging.kafka.offset`, etc. Optional attributes are omitted when empty.

Instrumentation is gated by the standard enable/disable env vars (`OTEL_GO_ENABLED_INSTRUMENTATIONS` / `OTEL_GO_DISABLED_INSTRUMENTATIONS`, key `kafka`).

### Files

- `instrumentation/github.com/segmentio/kafka-go/` — hooks (`hook.go`), enabler + header carrier (`kafka.go`), semconv builder (`semconv/client.go`), `otelc.yaml` rules, and unit tests.
- `test/apps/kafkaclient/` — test app. The producer path uses an in-process fake `RoundTripper` (mirroring the `database/sql` test app's in-memory driver) so no real broker is required; the consumer path targets a closed port so `ReadMessage` fails fast, still exercising the injected hook end-to-end.
- `test/integration/kafka_client_test.go` — integration test asserting produced/consumed spans and attributes.
- `docs/getting-started.md` — adds `segmentio/kafka-go` to the supported libraries table.

## Motivation

Adds Kafka support to the set of available instrumentations, letting `segmentio/kafka-go` users get producer/consumer traces with context propagation via compile-time instrumentation and no code changes.

Fixes #296

---

## Checklist

- [x] Tests pass: `make test`


kakkoyun:
@tushar1977 I will ask for your patience here. We will get to this. There are a lot of other PRs that touches to the core functionality and comments needs to be address. Thanks a lot for the contributions.

This is still on my list, don't worry 😊

tushar1977:
@kakkoyun thanks for mentioning this pr in #446

kakkoyun:
@tushar1977 We need this for v1 so thank you.

Could you please make sure everything is green on the CI?

tushar1977:
> @tushar1977 We need this for v1 so thank you.
> 
> Could you please make sure everything is green on the CI?

sure i'll do that

tushar1977:
> > @tushar1977 We need this for v1 so thank you.
> > Could you please make sure everything is green on the CI?
> 
sure i'll do that please give me this week time

tushar1977:
@kakkoyun please check. I noticed that the instrumentation folder is now seperated from the pkg folder thats great but my code still references some part of it. I tried to fix majority of things but if you want i can fix some more errors and make it equivalent to newer codebase

amazingakai (changes requested):
@tushar1977 Yes, that would be great.

The `instrumentation/` directory follows the same convention as `opentelemetry-go-contrib/instrumentation/`. For `kafka-go`, the expected layout would be something like:

```
instrumentation/github.com/segmentio/kafka-go/
    go.mod
    hook.go
    semconv/
    ...
```

I skimmed through the PR but haven't done a full review yet. A few things I noticed:

* There doesn't seem to be an `otelc.yaml` rules file. Could you add one?
* The current implementation appears to only instrument consumers (`Read`). We'll also need producer (`Write`) instrumentation.
* There's a `test/apps/kafkaclient`, but there's no integration test for it. Please add one as well.

I'd be happy to do a full review once these are addressed and CI is green.

tushar1977:
> @tushar1977 Yes, that would be great.
> 
> The `instrumentation/` directory follows the same convention as `opentelemetry-go-contrib/instrumentation/`. For `kafka-go`, the expected layout would be something like:
> 
> ```
> instrumentation/github.com/segmentio/kafka-go/
>     go.mod
>     hook.go
>     semconv/
>     ...
> ```
> 
> I skimmed through the PR but haven't done a full review yet. A few things I noticed:
> 
> * There doesn't seem to be an `otelc.yaml` rules file. Could you add one?
> * The current implementation appears to only instrument consumers (`Read`). We'll also need producer (`Write`) instrumentation.
> * There's a `test/apps/kafkaclient`, but there's no integration test for it. Please add one as well.
> 
> I'd be happy to do a full review once these are addressed and CI is green.
> 

give me this week time

amazingakai:
@tushar1977 I'd be happy to pick this up if you're busy, since we're already past the release deadline and this is on the release roadmap.

NameHaibinZhang:
I'll handle this issue; I'll submit the PR today. @tushar1977 @kakkoyun @amazingakai

tushar1977:
> @tushar1977 I'd be happy to pick this up if you're busy, since we're already past the release deadline and this is on the release roadmap.

sorry for no reply, i was busy for few days
you can go ahead

kakkoyun:
@NameHaibinZhang Thank you so much!

Let's directly contribute to this branch. So that we can have @tushar1977's contributions in.

NameHaibinZhang:
> @NameHaibinZhang Thank you so much!
> 
> Let's directly contribute to this branch. So that we can have @tushar1977's contributions in.

ok

instrumentation/github.com/segmentio/kafka-go/hook.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	spans, ok := ictx.GetData().([]trace.Span)
+	if !ok {
+		return
+	}
+	for _, span := range spans {
+		if span == nil {
+			continue
+		}
+		if err != nil {
```
kakkoyun: `WriteMessages` returns `kafka.WriteErrors` (a `[]error`, index-aligned with `msgs`) when some messages succeed and others fail. This code applies the single returned `err` to every span in the batch, so on a partial failure, successfully-written messages get marked `codes.Error` too.

Check `errors.As(err, &writeErrs)` and apply status per index, falling back to the current behavior for a plain `error`.

instrumentation/github.com/segmentio/kafka-go/hook.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	// Propagate the header-injected messages to the real WriteMessages call.
+	ictx.SetParam(2, msgs)
+	ictx.SetData(spans)
+}
+
+// AfterWriteMessages finalizes the producer spans created by BeforeWriteMessages.
+func AfterWriteMessages(ictx hook.HookContext, err error) {
+	if !kafkaEnabler.Enable() {
```
kakkoyun: `kafkaEnabler.Enable()` reads env vars fresh in both `BeforeWriteMessages` (line 80) and here. If the enabled state changes between the two calls, this returns before `span.End()` runs on the spans `Before` created. For the producer path, the trace headers are already written to the outgoing message by then, so the consumer links to a span that will never exist in the backend.

Snapshot the decision once in `Before` (e.g. via `SetData`), and have `After` always end whatever spans it finds instead of re-checking `Enable()`.

instrumentation/github.com/segmentio/kafka-go/hook.go
```diff
@@ -0,0 +1,223 @@
...
+		Operation:       semconv.KafkaOperationReceive,
+		ConsumerGroupID: data.groupID,
+		MessageKey:      string(msg.Key),
+		MessageBodySize: len(msg.Value),
+		Partition:       msg.Partition,
+		Offset:          msg.Offset,
+		HasPartition:    err == nil,
+		HasOffset:       err == nil,
+	}
+	_, span := tracer.Start(parent, topic+" receive",
```
kakkoyun, line 202: The extracted producer context (`parent`, line 199) is used to start this span and then discarded (`_, span :=`). `ReadMessage` doesn't return a context, so any span the application creates in its own message-processing code afterward is not a child of the traced call. The trace breaks exactly where a caller would expect it to continue.

Expose something like `ExtractContext(msg kafka.Message) context.Context` built on the existing `headerCarrier`, so downstream processing spans can parent correctly.

kakkoyun (changes requested):

amazingakai (approved):
LGTM once the other review comments are addressed.

A couple of non-blocking comments from my side:

1. Should we also have an e2e test for this?
2. grpc and net/http instrumentations use `client/` and `server/` subdirectories. Could we follow the same convention here and use `producer/` and `consumer/`? It's not super important right now, but with `otel.instrumentation.go` support users could choose to enable only one of them.

Also, once this is merged, we'll need to add it to the OpenTelemetry registry (same for OpenAI instrumentation), similar to https://github.com/open-telemetry/opentelemetry.io/pull/10609. Would you like to take care of that, or should I?

kakkoyun:
> A couple of non-blocking comments from my side:

@amazingakai I think these are great points and they should be blocking.

tushar1977:
thanks @NameHaibinZhang @kakkoyun @amazingakai

instrumentation/github.com/segmentio/kafka-go/hook.go
NameHaibinZhang: Done. Added a public `ExtractContext(msg kafka.Message) context.Context` function that extracts the trace context from message headers using the existing `headerCarrier`. Users can call it with the message returned by `ReadMessage` to continue the trace in downstream processing code. Added a test (`TestExtractContext`) to verify.

instrumentation/github.com/segmentio/kafka-go/hook.go
NameHaibinZhang: Done. Used `errors.As(err, &writeErrs)` to handle `kafka.WriteErrors` — when the returned error is a `WriteErrors`, each span now gets its own per-index error status instead of all being marked as error. Falls back to the original behavior for a plain `error`. Added a test (`TestAfterWriteMessages_PartialFailure`) to verify.

instrumentation/github.com/segmentio/kafka-go/hook.go
NameHaibinZhang: Done. Removed the `Enable()` re-check from both `AfterWriteMessages` and `AfterReadMessage`. The `After` hooks now rely solely on the presence of data set by `Before`. If `Before` was disabled, no data is stored, so `After` returns early via the nil-data check. This prevents span leaks when env vars change between the two calls. Added a test (`TestAfterWriteMessages_AlwaysEndsSpans`) to verify.

kakkoyun (approved):
LGTM

Issue #217, opened 2025-12-04
chore: Introduce locally reproducible mechanism for keeping submodules in sync


kakkoyun:
Nit: as we have multiple nested modules, shouldn't we check in these files? I'm ok both ways, but from experience these seem to be useful in the repository so everybody has the same versions around.

_Originally posted by @darccio in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/156#discussion_r2585841420_

- [ ] Better to use [crosslink](https://pkg.go.dev/go.opentelemetry.io/build-tools/crosslink)

darccio:
@txabman42 Doesn't using `crosslink` for development cause noise from the `replace` edits done to multiple `go.mod` files?

We decided to keep `go.work` and `go.work.sum` in `dd-trace-go` when we migrated it to be a root module with nested modules for each contrib (inst pkg) while keeping all the `replace` directives in all the `go.mod` files for development because they don't have impact when using them outside of the repository.

PR #326, opened 2026-03-06
chore: add crosslink

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/217

Add [crosslink](https://pkg.go.dev/go.opentelemetry.io/build-tools/crosslink) for local development:
- `make crosslink`: verify that all replace directives exist in go.mod or add them if not
- `make go-work`: generate go.work with all submodules (for local development)
- `make go-mod-tidy`: execute `go mod tidy` in all submodules

### DIFF
--- a/.config/golangci.yml
+++ b/.config/golangci.yml
@@ -1,7 +1,7 @@
 # This file is licensed under the terms of the MIT license https://opensource.org/license/mit
 # Copyright (c) 2021-2025 Marat Reymers
 
-## Golden config for golangci-lint v2.1.6
+## Golden config for golangci-lint v2.11.3
 #
 # This is the best config for golangci-lint based on my experience and opinion.
 # It is very strict, but not extremely strict.
--- a/.github/tools/go.mod
+++ b/.github/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/open-telemetry/opentelemetry-go-compile-instrumentation/.github/tools
 
-go 1.24.0
+go 1.25.0
 
 require github.com/google/go-github/v79 v79.0.0
 
--- a/.github/workflows/lint-golang.yaml
+++ b/.github/workflows/lint-golang.yaml
@@ -9,7 +9,7 @@ permissions:
   contents: read
 env:
   # TODO: Configure dependabot or renovate to update this version.
-  GOLANGCI_LINT_VERSION: v2.1.6
+  GOLANGCI_LINT_VERSION: v2.11.3
 
 jobs:
   # This is a required workflow, so it must always run and report a status.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ for a summary description of past meetings. You can also get in touch on slack c
 
 This project uses several tools for development. Most tools will be automatically installed when you first run the corresponding `make` target. However, you need to have:
 
-- [Go](https://golang.org/dl/) 1.21 or later
+- [Go](https://golang.org/dl/) 1.25 or later
 - [Git](https://git-scm.com/)
 - Make (usually pre-installed on macOS and Linux)
 
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <div align="center">
   <img src="./docs/assets/otel-logo.png" alt="OpenTelemetry Logo" width="500" />
   <br />
-  <img src="https://img.shields.io/badge/Go-1.21%2B-4A90E2?style=flat&logo=go" alt="Go" />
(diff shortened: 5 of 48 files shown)


.tools/tools.go
```diff
@@ -0,0 +1,12 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build tools
+// +build tools
+
+package tools // import "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tools"
```
y1yang0, line 7: Can we avoid creating this `.tools` at project root?

Maybe somethink like `go install crosslink && crosslink ....`

test/apps/grpcclient/go.sum
```diff
@@ -14,14 +14,14 @@ go.opentelemetry.io/auto/sdk v1.2.1 h1:jXsnJ4Lmnqd11kwkBV2LgLoFMZKizbCi5fNZ/ipaZ
 go.opentelemetry.io/auto/sdk v1.2.1/go.mod h1:KRTj+aOaElaLi+wW1kO/DZRXwkF4C5xPbEe3ZiIhN7Y=
 go.opentelemetry.io/otel v1.39.0 h1:8yPrr/S0ND9QEfTfdP9V+SiwT4E0G7Y5MO7p85nis48=
 go.opentelemetry.io/otel v1.39.0/go.mod h1:kLlFTywNWrFyEdH0oj2xK0bFYZtHRYUdv1NklR/tgc8=
-go.opentelemetry.io/otel/metric v1.39.0 h1:d1UzonvEZriVfpNKEVmHXbdf909uGTOQjA0HF0Ls5Q0=
-go.opentelemetry.io/otel/metric v1.39.0/go.mod h1:jrZSWL33sD7bBxg1xjrqyDjnuzTUB0x1nBERXd7Ftcs=
-go.opentelemetry.io/otel/sdk v1.39.0 h1:nMLYcjVsvdui1B/4FRkwjzoRVsMK8uL/cj0OyhKzt18=
-go.opentelemetry.io/otel/sdk v1.39.0/go.mod h1:vDojkC4/jsTJsE+kh+LXYQlbL8CgrEcwmt1ENZszdJE=
+go.opentelemetry.io/otel/metric v1.38.0 h1:Kl6lzIYGAh5M159u9NgiRkmoMKjvbsKtYRwgfrA6WpA=
```
kakkoyun, line 17: hmm, why the downgrade?

Makefile
```diff
@@ -21,14 +21,16 @@ INST_PKG_GZIP = otel-pkg.gz
 INST_PKG_TMP = pkg_temp
 API_SYNC_SOURCE = pkg/inst/context.go
 API_SYNC_TARGET = tool/internal/instrument/api.tmpl
-
+TOOLS_DIR = .tools
+GO_VERSION = 1.24
```
kakkoyun, line 25: This should be solved in another PR, but since now 1.26 is out. The minimum supported version should be 1.25

.tools/tools.go
txabman42: In this case `go install` doesn't work as crosslink has a replace directive in it's [go mod](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/crosslink/go.mod).

Usually, other opentelemtry projects use .tools pattern or adds the tool folder inside internal [example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3df953e353d2f1c8ff1e8bd551835cbc94363975/internal/tools/go.mod#L16)

WDYT about refactoring other `go install` and move to this approach? That way, we dont force users to install all this tools globally and avoid unwanted tools, version mismatch, etc.

Makefile
txabman42: Good point, https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/331

test/apps/grpcclient/go.sum
txabman42: not sure how the 1.39 was added. I think its related to how the replace directives are working as it has one pointing to grpcserver which has the v1.39.0 version.

Right now, it using v1.38.0 coming from google.golang.org/grpc

.tools/tools.go
y1yang0: I have indeed run into version mismatch issues several times. Since the purpose of `root/.config` is also to support tooling, could we rename `root/.config` to `root/.tools` and centralize all helper tools there?
(This could be done in another PR)

txabman42: I like the idea https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/348

PR #350, opened 2026-03-12
chore: add AGENTS.md

txabman42:
First iteration of AGENTS.md

blocked by https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/326

### DIFF
--- a/pkg/inst/insttest/mock_hook_context.go
+++ b/pkg/inst/insttest/mock_hook_context.go
@@ -1,0 +1,84 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package insttest provides test helpers for instrumentation hook testing.
+package insttest
+
+// MockHookContext is a shared test implementation of inst.HookContext.
+// The data field mirrors HookContextImpl
+type MockHookContext struct {
+	Params      []interface{}
+	ReturnVals  []interface{}
+	SkipCall    bool
+	FuncName    string
+	PackageName string
+	data        interface{}
+}
+
+// NewMockHookContext creates a new MockHookContext with optional initial params.
+func NewMockHookContext(params ...interface{}) *MockHookContext {
+	return &MockHookContext{
+		Params:      params,
+		FuncName:    "mockFunc",
+		PackageName: "mock",
+	}
+}
+
+func (m *MockHookContext) SetSkipCall(skip bool) { m.SkipCall = skip }
+func (m *MockHookContext) IsSkipCall() bool      { return m.SkipCall }
+
+func (m *MockHookContext) SetData(data interface{}) { m.data = data }
+func (m *MockHookContext) GetData() interface{}     { return m.data }
+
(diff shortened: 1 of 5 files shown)


AGENTS.md
```diff
@@ -0,0 +1,62 @@
+# AGENTS.md
+
+This file provides guidelines for AI-assisted contributions to
+`opentelemetry-go-compile-instrumentation`. For the project's AI usage policy, see
+[AI_POLICY.md](AI_POLICY.md).
+
+## General Rules and Guidelines
+
+The most important rule is **never post AI-generated comments on issues or pull requests**.
```
ralf0131, line 9: We use Github copilot for code review, which means AI will generate comments on pull request. I actually do not oppose to AI generated comments as long as it indicates that it is created by AI.

AI_POLICY.md
```diff
@@ -0,0 +1,84 @@
...
+Assisted-by: GitHub Copilot
+Assisted-by: Claude Opus 4
+Assisted-by: Cursor with Claude Sonnet 4
+```
+
+This helps maintainers calibrate their review and set appropriate expectations. Note that for
+AI-assisted code that has been substantially reviewed, reworked, and understood by the author,
+disclosure is appreciated but not required — the human author owns the result either way.
+
+Do not use `Co-Authored-By` lines for AI tools. As the Go project notes, AI tools cannot be
```
ralf0131: Actually we use `Co-Authored-By` pretty intensive in our internal repo. Could you explain more about why this is not a good practice?

y1yang0 (approved):
Looks good to me. Just wondering if OpenTelemetry as a whole has any agent policy requirements, or if each repo is free to do its own thing?

AI_POLICY.md
txabman42: I took that idea from [here](https://groups.google.com/g/golang-dev/c/4Li4Ovd_ehE/m/8L9s_jq4BAAJ). Basically, `Co-Authored-By` could be a bit misleading when we talk about legal issues.

The idea is to use `Assisted-by` assited by instead.

AGENTS.md
txabman42: WDYT to avoid AI comments except the ones generated by AI code review tools configured by the repo maintainers?

The idea is to keep comment sections for human discussions. If we start filling all our PRs with tons of AI generated comments, it could be hard to follow them, and it could bring too much noise.

AI_POLICY.md
kakkoyun: I don't see any value on any of this. It's just a tool. I don't "add this is created by vim" to my commits.

We can add labels to PRs to let the review know, but other than that I don't see any value. Things are changing and most of the commits will be written by the assistance of these tools.

txabman42: I don't have a strong position on this yet. Removed. We can always change this in the future.

AGENTS.md
txabman42: Updated c2e0697ad4fb24a04199688d7386b630a08b6600

PR #358, opened 2026-03-12
fix(tool): handle blank identifier in function return parameters

### LINKED ISSUE #152: Unable to inject functions with a "_" in return parameters
Unable to inject functions with a blank identifier (_) in their return parameters, for example:

```go
func foo() (_ error)
```

kakkoyun:
## Description

Adds a test case and golden file for functions that use `_` (blank identifier) as a named return parameter, e.g.:

```go
func UnderscoreReturnFunc() (_ int, _ error) { return 0, nil }
```

The existing `collectReturnValues` logic in `apply_func.go` already renames blank return identifiers to `_ignoredParam0`, `_ignoredParam1`, etc. before generating the trampoline. This PR adds the missing test coverage that verifies this behaviour and locks it in via golden files.

## Motivation

Issue #152 reported that blank identifiers in return parameter lists were not handled correctly. Investigation revealed the code path was correct but completely untested — the fix is to add the missing test coverage. Without it, a future refactor could silently break this edge case.

Fixes #152

---

## Checklist

- [x] Tests pass: `make test`


tool/internal/instrument/testdata/hook.go
```diff
@@ -48,3 +48,5 @@ func GenericFuncAfter(ctx inst.HookContext, r1 interface{}, r2 error) {}
 func GenericMethodBefore(ctx inst.HookContext, recv interface{}, p1 interface{}, p2 string) {}
 
 func GenericMethodAfter(ctx inst.HookContext, r1 interface{}, r2 error) {}
+
+func HUnderscoreReturnAfter(ctx inst.HookContext, r1 int, r2 error) {}
```
kakkoyun: This needs to be `H12UnderscoreReturnAfter`
```suggestion
func H12UnderscoreReturnAfter(ctx inst.HookContext, r1 int, r2 error) {}
```

kakkoyun: Fixed — renamed to `H12UnderscoreReturnAfter` in hook.go, rules.yml, and the golden file. f2884b7

Issue #332, opened 2026-03-11
Automatically detect imports in hook code


y1yang0:
PR #283 allows us to import arbitrary packages in our hook code using the imports directive, like so:

```yaml
hook_with_imports:
  target: main
  func: Example
  before: MyHookBefore
  path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld"
  imports:
    log: "log"        # Additional import needed by the hook
    _: "unsafe"       # Blank import if needed
```

I propose that we add a new capability to automatically analyze the dependencies of the MyHookBefore file. This would eliminate the need to manually add the imports field, which is a rather brittle and tedious approach.

PR #361, opened 2026-03-14
chore(hooks): verify auto detect import hooks

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/332

> I propose that we add a new capability to automatically analyze the dependencies of the MyHookBefore file.

Checking the code, this should already be solved. I've added a test to verify this and updated the doc.

Basically, setup `add.go` just adds as blank import the matched hook package, e.g. `_ "pkg/instrumentation/basic" `. This blank import forces go build to compile the hook package and its entire transitive dependency graph, which means, all of them will end up in the target's importcfg.


docs/rules.md
```diff
@@ -56,22 +56,7 @@ hook_helloworld:
...
-hook_with_imports:
-  target: main
-  func: Example
-  before: MyHookBefore
-  path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/helloworld"
-  imports:
-    log: "log"        # Additional import needed by the hook
-    _: "unsafe"       # Blank import if needed
-```
+The tool automatically reads the hook source file and ensures all of its imports are present in the build. No `imports:` field is needed for function hook rules.
```
y1yang0, line 59:
```suggestion
The tool automatically reads the hook source file and ensures all of its imports are present in the build. No `imports` field is needed for function hook rules.
```

y1yang0 (approved):

PR #364, opened 2026-03-18
feat(tool): value declaration join point

darccio:
## Description

Adds the **`value_declaration` join point** (OP3.3 from #277), which targets package-level `var` and `const` declarations by their *declared type* rather than by name.

This complements the `declaration` join point from the parent branch: where `declaration_of` pins a specific named declaration, `value_declaration` targets all declarations of a given type across the package — useful for overriding feature flags, replacing interface implementations, or injecting defaults wherever a particular type is declared.

Assisted-By: `claude-sonnet-4-6`

### What's included

- **New rule type: `value_declaration`** (`rule/value_decl_rule.go`) — matches all package-level `var`/`const` declarations whose annotated type matches a given type expression and replaces their initializer with `assign_value`. Supports built-in types (`bool`, `string`), fully-qualified types (`net/http.Client`), and pointer types (`*net/http.Request`). The type string is parsed into components (pointer flag, import path, identifier) once at construction time.

- **Type matching** (`instrument/apply_value_decl.go`) — `matchesType` handles pointer wrapping, qualified selectors, and plain identifiers. Import alias resolution reuses `collectImportAliases` and is skipped entirely for built-in types (no import path to resolve).

- **`matchesQualifiedSelector`** (`apply_call_helpers.go`) — the `pkg.Name` selector-resolution logic (`ident.Path` first, then alias map fallback) was duplicated between `matchesCallRule` and the new `matchesType`. It is extracted into a shared helper used by both.

- **Retroactive test coverage for `declaration_of`** — `decl_rule_test.go` and `apply_decl_test.go` were missing from the parent branch and are added here, covering rule construction/validation and the `applyDeclRule` apply function.

- **Setup integration** — `preciseMatching` registers `InstValueDeclRule` unconditionally for all source files in a matched package (type resolution requires import aliases, which are only available at instrumentation time); files with no matching declarations are a silent no-op.

- **Improved `FindNamedDecl` documentation** — the docstring now spells out the non-uniform concrete return type per `kind` value (`*dst.FuncDecl` / `*dst.ValueSpec` / `*dst.GenDecl`), avoiding future type-assertion surprises.

- **Golden test** (`testdata/golden/value-decl-rule/`) and unit tests (`value_decl_rule_test.go`, `apply_value_decl_test.go`).

### Rule format

```yaml
override_feature_flags:
  target: mypackage
  value_declaration: "bool"        # built-in type
  assign_value: "true"

replace_transport:
  target: net/http
  value_declaration: "*net/http.Transport"   # pointer to qualified type
  assign_value: "&http.Transport{MaxIdleConns: 100}"
  imports:
    http: "net/http"
```

### Rule fields

| Field | Type | Required | Description |
|---|---|---|---|
| `value_declaration` | string | yes | Type expression: `bool`, `pkg/path.Type`, `*pkg/path.Type` |
| `assign_value` | string | yes | Go expression assigned to every matched declaration |
| `imports` | map[string]string | no | Extra imports needed by the injected expression |

### Limitations

- Only declarations with an **explicit type annotation** are matched — untyped declarations such as `const x = true` are skipped by design.
- `assign_value` for multi-name specs (`var a, b T = ...`) assigns the same expression (cloned) to every name in the spec.

## Motivation

Implements OP3.3 from #277. Builds on the `declaration_of` join point from the parent branch.

---

## Checklist

- [x] Tests pass: `make test`


y1yang0:
Since InstDeclRule (kind + identifier + value) already gives us a precise top-level declaration modification, I’m not yet convinced that a separate type-based bulk matcher is necessary, unless we can point to concrete cases where enumerating the named declarations is impractical.

I have a few questions and would appreciate some clarification on the design intent:

- What concrete scenarios cannot be expressed reasonably with the existing InstDeclRule?
In those scenarios, why is "matching by declared type" the right way, rather than explicitly enumerating the named declarations(or regex-based decl if these names do have some patterns, e.g. "Flag.*"/"Experimental.*" )?
- How do we prevent overly broad matches (blast radius), especially in packages that contain multiple declarations of the same type where we might only want to override some of them?
- Is the expected usage really common enough to justify a completely separate rule family, rather than treating it as a more specialized extension (e.g., regex-based identifier in InstDeclRule) later?

Since InstDeclRule is already available to handle declarations, I think the bar (justification) for adding a second decl-oriented rule should be fairly high.

darccio:
Closing it. We'll reopen if we have more pressing need to use it in multiple integrations.

PR #375, opened 2026-03-19
feat(tool): metaprogramming annotations (magical comments; directives)

darccio:
## Description

Adds a new **Directive Rule** — a join point that matches AST nodes annotated with magic comments (e.g., `//dd:span span.name:"my operation"`). This is a **pure filter** with no advice: it matches during the setup phase but performs no AST transformation during instrumentation. It becomes useful when `all-of` combinators land, allowing directive annotations to be combined with other join points.

### What's included

- **New rule type: `directive`** — matches source files containing a `//directive` comment. The directive name is specified without the `//` prefix. Validation rejects empty directives, directives with spaces, and directives that start with `//`.

- **Directive matching** (`matchDirective`) — validates that the comment is a line comment (not block), has no space after `//`, matches the full directive name (not a prefix), and requires whitespace after the name if arguments follow. Ports the matching semantics from Orchestrion's `directive.matches`.

- **Argument parsing** (`ParseDirectiveArgs`, `scanArgs`, `tokenize`) — parses `key:value` arguments from directive comments. Values may be Go double-quoted strings (handled via `strconv.Unquote`). Single quotes and malformed input are rejected. This infrastructure is ready for use when combinators wire directive args into the instrument phase.

- **Setup integration** — `createRuleFromFields` dispatches on the `directive` YAML key. `preciseMatching` uses a lazy-loaded `os.ReadFile` + `bytes.Contains` pre-filter (since `ParseFileFast` skips comments, the raw bytes are needed).

- **Instrument phase** — directive rules are intentionally **excluded from `groupRules`** to avoid a wasted parse+write cycle for files that only have directive rules (no AST transformation to apply).

- **Tests** — unit tests for matching (10 cases), arg parsing (9 cases), directive args (3 cases), rule validation (5 cases), rule factory (2 cases), `TestGroupRules` exclusion, and a golden integration test.

### Rule format

```yaml
my_directive_rule:
  target: main
  directive: "otelc:span"
```

### Rule fields

| Field       | Type   | Required | Description |
|-------------|--------|----------|-------------|
| `directive` | string | yes      | Directive name to match (without `//`) |

### Design decisions

- **Pure filter, no advice** — no `before`/`after`/`assign_value` fields. The rule only participates in setup-phase matching.
- **`bytes.Contains` pre-filter** — a loose but fast check on raw file bytes. False positives (e.g., the directive appearing inside a string literal) are acceptable because the precise `matchDirective` function will be used for real matching when combinators land.
- **Excluded from `groupRules`** — since `applyDirectiveRule` would be a no-op, directive rules don't enter the instrument phase's file-processing loop, avoiding unnecessary AST parse+write cycles.

## Motivation

Implements the Orchestrion-equivalent `directive` join point for compile-time instrumentation. This is a prerequisite for the `all-of` combinator feature, which will allow rules like "instrument functions annotated with `//dd:span`" by combining directive matching with function join points.

Design follows feedback on Orchestrion PR https://github.com/DataDog/orchestrion/pull/665 regarding Go double-quoted string support for directive argument values.

---

## Checklist

- [x] Tests pass: `make test`


tool/internal/ast/directive_test.go
```diff
@@ -0,0 +1,209 @@
...
+func TestMatchDirective(t *testing.T) {
+	tests := []struct {
+		name      string
+		dec       string
+		directive string
+		expected  bool
+	}{
+		{
+			name:      "exact match",
+			dec:       "//dd:span",
```
kakkoyun: Let's use `otelc:span` even in tests.

tool/internal/setup/match.go
```diff
@@ -225,6 +232,20 @@ func (sp *SetupPhase) preciseMatching(
 				// Files without matching calls are a no-op in applyCallRule.
 				set.AddCallRule(source, rt)
 				sp.Info("Match call rule", "rule", rt, "dep", dep)
+			case *rule.InstDirectiveRule:
```
y1yang0, line 230: Can we identify these directives from `tree`?

tool/internal/ast/directive_test.go
darccio: Done in [785e4c](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/375/commits/785e4c0aa19e8761e671a009648dc353ccbba477)

tool/internal/instrument/testdata/golden/directive-rule/rules.yml
```diff
@@ -0,0 +1,3 @@
+span_directive:
```
kakkoyun, line 1: I think we have to configure some modifications to demonstrate the directive here. This does nothing, IMHO this should fail.

tool/internal/ast/directive.go
```diff
@@ -0,0 +1,146 @@
...
+	r, _ = utf8.DecodeRuneInString(rest)
+	if !unicode.IsSpace(r) {
+		return "", false
+	}
+	return rest, true
+}
+
+// ParseDirectiveArgs finds the directive in the decoration string, extracts
+// the text after the directive name, and parses it into key:value arguments.
+func ParseDirectiveArgs(dec, directive string) ([]DirectiveArg, error) {
```
txabman42, line 65: When is this going to be called?

tool/internal/setup/match.go
```diff
@@ -225,6 +232,20 @@ func (sp *SetupPhase) preciseMatching(
...
+				// Lazy-read raw bytes on first directive rule (read once per source file).
+				// ParseFileFast skips comments, so we need the raw bytes to check
+				// for directives.
+				if rawBytes == nil {
+					rawBytes, err = os.ReadFile(source)
```
txabman42: If multiple directive rules appear in the same file, the file will be loaded into memory multiple times.

tool/internal/setup/match.go
darccio: It should be possible, let me give it a try.

tool/internal/ast/directive.go
darccio: This is intentional as stated in the comment in `instrument.go:21-24`:

> DirectiveRules are intentionally excluded: they are a pure filter with no advice, so there is no AST transformation to apply.

The function is infrastructure waiting for the advice layer. When `InstDirectiveRule` gains an advice field (or when combinators like "all-of" land that compose a directive filter with a func rule), the instrument phase will need to:

1. Walk function declarations, check `funcDecl.Decs.Start` for the directive with `MatchDirective`
2. Call `ParseDirectiveArgs` on the matched decoration to extract the arguments (e.g., `span.name:"my op"`)
3. Use those args to configure the injected instrumentation

tool/internal/setup/match.go
darccio: You are right. I wanted to optimize to detect comments by pattern instead of parsing. Per @y1yang0 request, I explored and implemented the same feature on AST tree in 641115c62ee2be176e5a682052c0e2a7582c67f5.

tool/internal/setup/match.go
darccio: Done in 641115c62ee2be176e5a682052c0e2a7582c67f5

tool/internal/instrument/testdata/golden/directive-rule/rules.yml
darccio: Done in 1e7d06c2c973e06f965df69f016fa545d706b1ab.

tool/internal/ast/directive.go
```diff
@@ -0,0 +1,196 @@
...
+				funcs = append(funcs, funcDecl)
+				break
+			}
+		}
+	}
+	return funcs
+}
+
+// tokenize splits input on whitespace, respecting double-quoted strings.
+func tokenize(input string) ([]string, error) {
```
y1yang0, line 158: Quite similiar to `SplitCompileCmds`? They could be consolidate in separate PR

tool/internal/instrument/apply_raw.go
```diff
@@ -49,6 +49,31 @@ func insertRaw(r *rule.InstRawRule, decl *dst.FuncDecl) error {
 	return nil
 }
 
+// applyDirectiveRule finds all functions annotated with the directive, renders
+// the template for each, and prepends the resulting Go statements into the
+// function body — mirroring Orchestrion's prepend-statements approach.
+func (ip *InstrumentPhase) applyDirectiveRule(ctx context.Context, r *rule.InstDirectiveRule, root *dst.File) error {
```
y1yang0: We can create `apply_directive.go` to accommodate this

tool/internal/rule/directive_rule.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	if _, err := template.New("").Parse(r.Template); err != nil {
+		return ex.Wrapf(err, "invalid template syntax")
+	}
+	return nil
+}
+
+// Render executes the template with the given function context and returns
+// the resulting Go source snippet.
+func (r *InstDirectiveRule) Render(data DirectiveTemplateData) (string, error) {
```
y1yang0: This could be moved to `apply_directive.go`(for consistency)

tool/internal/instrument/apply_raw.go
darccio: Done in b26b134f062cff9618af2f074dc10285eb6f6f8d

tool/internal/rule/directive_rule.go
darccio: Done in ee4e7ca3f883adc6c9c558b0be1f30f21e0abdd5

PR #377, opened 2026-03-19
feat(tool): add join point filter interface

### LINKED ISSUE #346: OP4: Join point composition & filtering
- [ ] OP4.1: Intersection combinator (all-of)
- [ ] OP4.2: Union combinator (one-of)
- [ ] OP4.3: Negation combinator (not)
- [ ] OP4.4: Import path filtering
- [ ] OP4.5: Package name filtering (possibly limited to "main")
- [ ] OP4.6: Test package filtering ("test main" need

kakkoyun:
## Description

Adopts the 2-tier \`where\`/\`do\` rule schema agreed in the design discussion on this PR. This is branch 0 of the OP4 stack (#381–#386).

Top-level \`target\` + \`version\` scope the package. \`where\` carries all non-package selectors (point selectors + \`where.file\` for file-level predicates; implicit \`all-of\` at the top, with \`one-of\`/\`not\` as named sub-groups). \`do\` carries modifiers, and the modifier name in \`do\` declares the rule type.

**What is in this branch:**

- \`func\`, \`recv\` (file-level via \`where.file\`), \`struct\` leaf filters implemented
- \`all-of\`, \`one-of\`, \`not\` combinators and \`has_directive\` are schema-present but return descriptive errors at runtime (wired up in follow-up PRs #381–#386)
- \`normalize\` layer translating structured \`where\`/\`do\` form to the internal flat fields that \`createRuleFromFields\` consumes
- Golden fixtures migrated to canonical 2-tier form
- [ADR-0003](docs/adr/0003-structured-rule-schema.md) records the decision in Nygard form (Status/Context/Decision/Consequences)
- [docs/rules.md](docs/rules.md) contains the full field reference, valid/invalid shape examples, and flat→structured migration guide

**Deferred items (not in this PR):**

- Target globbing — \`rulesByTarget\` remains exact-match; glob on \`target\` tracked as a follow-up
- \`is_test\` / \`test_main\` filters — reserved for a follow-up PR
- \`defaults:\` section sugar — deferred to a future iteration
- \`import_path\` / \`package_name\` filters — dropped from the stack; single \`*\` on \`target\` deferred

Fixes #346 (partial — branch 0 of the stack)

---

## Checklist

- [x] Code formatted: \`make format\`
- [x] Linters pass: \`make lint\`
- [x] Tests pass: \`make test\`


tool/internal/filter/filter.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// MatchContext carries per-file contextual information for filter evaluation.
+// It is constructed once per source file in preciseMatching and passed to all
+// filters associated with the rules being evaluated for that file.
+type MatchContext struct {
+	// ImportPath is the Go import path of the package containing the source file.
+	ImportPath string
+
+	// SourceFile is the absolute path to the source file being evaluated.
```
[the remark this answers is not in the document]
kakkoyun: Fixed in b9baa34 — changed test helper to use `"/tmp/source.go"` (absolute path) matching the `SourceFile` doc contract. Production code in `preciseMatching` always provides absolute paths from `dep.Sources`.

tool/internal/setup/match.go
```diff
@@ -173,14 +167,46 @@ func (sp *SetupPhase) runMatch(
...
+	for _, r := range rules {
+		var f filter.Filter
+		if where := r.GetWhere(); where != nil {
+			var err error
+			f, err = filter.Build(where)
+			if err != nil {
+				return nil, ex.Wrapf(err, "build filter for rule %q", r.GetName())
+			}
+		}
+		ruleFilters = append(ruleFilters, ruleFilter{rule: r, filter: f})
```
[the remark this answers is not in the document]
kakkoyun: Fixed in d262604 — updated ADR-0003 to say "compiled once per rule before source-file iteration in `preciseMatching`". Also added a comment in `preciseMatching` explaining the design choice (each rule targets one import path, so filters are effectively built once).

tool/internal/setup/match.go
```diff
@@ -191,51 +220,75 @@ func (sp *SetupPhase) preciseMatching(
 		if err != nil {
 			return nil, err
 		}
-		if tree == nil {
-			return nil, ex.Newf("failed to parse file %s", source)
-		}
```
darccio, line 190: Removing this clause might cause a potential nil dereference.

darccio (approved):
Approving with a minor comment that might need a small change.

docs/adr/0003-join-point-filter-interface.md
```diff
@@ -0,0 +1,57 @@
+# 3. Join Point Filter Interface
```
txabman42: I’m used to creating ADRs when documenting why a particular option was chosen over others or documenting the pros and cons of a chosen decision.

I would keep documentation at the code level and include a guide under /doc explaining how to use it.

WDYT?

kakkoyun: I don't have strong opinion on this. I'd be happy to do so.

tool/internal/setup/match.go
kakkoyun: Good catch. I traced the contract:

`ParseFileFast` → `Parse` → `parser.ParseFile` (Go stdlib) → `DecorateFile` (dave/dst)

The Go stdlib `parser.ParseFile` always returns non-nil `*ast.File` on success — `nil` only accompanies a non-nil error. `DecorateFile` follows the same contract. No code path produces `(nil, nil)`.

No other caller of `ParseFileFast` in the codebase guards against `tree == nil` after an error check (including `findHookFile` in `trampoline.go`).

Rather than adding dead code for an unreachable path, I'll add a comment at the call site documenting this contract so the reasoning is visible.

docs/adr/0003-join-point-filter-interface.md
kakkoyun: Fair point. ADRs work best for recording rejected alternatives and trade-offs, and the current ones are more "design overview" than "decision record." Happy to convert them to a usage guide under `/docs` in a follow-up once the design discussion settles — the rationale will stay in code-level comments either way.

docs/adr/0003-join-point-filter-interface.md
```diff
@@ -0,0 +1,125 @@
...
+two phases allow cheap early-exit before source files are parsed.
+
+## Decision
+
+### 3-Tier Granularity Model
+
+Rules are evaluated in three tiers of increasing specificity:
+
+```
+Tier 1 — Package Scope:   target (exact or glob) + version
```
pdelewski: Do we need to distinguish all these tiers?
Why not to generalize it to something like 
```
instrument:
  function: Open
  where:
    target: example.com/svc
    before:
      name: BeforeOpen
      path: example.com/hooks
    or:
      - has_function: init
      - has_method: Setup
```
Both from usage perspective as well as implementation one.

pdelewski: I've seen that this was discussed already

y1yang0:
@txabman42 @kakkoyun @pdelewski  My thinking is, since we've reached an agreement on the 3-tier model, let's focus our efforts on advancing the v1.0.0 roadmap. After that, we can merge PR #377 and the other stacked PRs that align with the 3-tier model.

We can save the discussion on topics like the structured-schema/action-blocks/distinguish-tier3 for v1.1.0.

kakkoyun:
> @txabman42 @kakkoyun @pdelewski My thinking is, since we've reached an agreement on the 3-tier model, let's focus our efforts on advancing the v1.0.0 roadmap. After that, we can merge PR #377 and the other stacked PRs that align with the 3-tier model.
> 
> We can save the discussion on topics like the structured-schema/action-blocks/distinguish-tier3 for v1.1.0.

From my side this is a blocker for v1.0.0 No need to rush things. Let's make sure we have a consensus on the initial design here and also the user facing configuration and semantics. We still have time to hit our goals for June deadline.

I will check the comments further and try to respond.

kakkoyun:
@txabman42 @y1yang0 @pdelewski I think there is a compromise here that keeps the semantics explicit and still preserves the selector/modifier distinction.

I am increasingly convinced this should be framed as a **2-tier model**:

1. **Top-level package scope**: `target` (exact or wildcard) + `version`
2. **`where`**: the second-tier selector block

In other words, `target` / `version` stay special because they answer the first question for the rule writer: **which package(s) can this rule apply to?**

Then `where` answers the second question: **within those packages, what exactly are we selecting?** That can include the point selector for the rule kind and, when needed, additional file-level predicates.

Finally, `do` answers: **what modification do we apply once the selector matches?**

From a UX point of view, this is the part I care about most: config writers should be able to tell at a glance which fields are selectors and which are modifiers.

A few examples in that shape:

```yaml
hook_open:
  target: database/sql
  version: "v1.0.0,v2.0.0"
  where:
    func: Open
  do:
    inject_hooks:
      before: beforeOpenInstrumentation
      after: afterOpenInstrumentation
      path: "github.com/.../databasesql"
```

```yaml
hook_open_only_in_files_with_init:
  target: database/sql
  version: "v1.0.0,v2.0.0"
  where:
    func: Open
    file:
      has_func: init
  do:
    inject_hooks:
      before: beforeOpenInstrumentation
      after: afterOpenInstrumentation
      path: "github.com/.../databasesql"
```

```yaml
gin_bind_monitoring:
  target: github.com/gin-gonic/gin
  version: "v1.9.0,v2.0.0"
  where:
    recv: "*Context"
    one-of:
      - func: Bind
      - func: ShouldBind
      - func: BindJSON
  do:
    inject_code:
      raw: |-
        defer appsec.MonitorParsedHTTPBody(c.Request.Context(), obj)
```

And if we agree wildcard package selection is needed, I would still prefer that to live on `target` directly rather than introducing a second package selector under `where`:

```yaml
wrap_http_calls:
  target: "github.com/myorg/**"
  version: "v1.0.0"
  where:
    function_call: net/http.Get
  do:
    wrap_call:
      template: "tracedGet({{ . }})"
```

So my current preference is:
- keep `target` / `version` at top level as package-scope selectors
- let `where` dictate the second tier
- keep `do` explicitly for modifiers/actions

If we can agree on that semantic split here, I can update #395 accordingly.

txabman42:
- [Option 1](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/377#issuecomment-4184002406) is better for per-rule locality: one block to read, one doc page per rule type.
```yaml
hook_open:
  target: database/sql
  where:              # (or where_file) file-level predicates only
    has_func: init
  inject_hooks:       # rule type + its selectors + its modifiers
    func: Open
    before: beforeOpen
    after: afterOpen
```
- [Option 2](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/377#issuecomment-4303461332) is better for global orthogonality: selectors and modifiers never mix, combinators live in one place.
```yaml
hook_open:
  target: database/sql
  where:              # all selectors (point + file)
    func: Open
    file:
      has_func: init
  do:
    inject_hooks:     # modifiers only
      before: beforeOpen
      after: afterOpen
```

My main concern with `Option 2` is that `where` filters depend on the rule type defined under `do`. IMO, that could be a bit confusing to document.

We are focusing on how the hook definition looks and reads, but we should also consider how easy it is to understand how to configure it.

kakkoyun:
> My main concern with `Option 2` is that `where` filters depend on the rule type defined under `do`. IMO, that could be a bit confusing to document.
> 
> We are focusing on how the hook definition looks and reads, but we should also consider how easy it is to understand how to configure it.

I do believe the number one priority should be user experience and ease of understanding the rule at a glance (thus my comment in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/377#issuecomment-4141960500). The best way to solve this to have a linter for configuration files (i.e. `otelc validate` or `otelc lint`), we can even provide json schema and enable editor LSPs so that linters would work on the fly.

I do think option 2 is more flexible and you separate selectors and modifiers cleanly and it is easier mental model. 

Let's see what others think as well.

kakkoyun:
option 3
```yaml
hook_open:
  target: database/sql
  version: v
  inject_hooks:     # modifiers only
	  where:              # all selectors (point + file)
	    func: Open
	    file:
	      has_func: init
	  do:
	    before: beforeOpen
	    after: afterOpen
   import_package:
	  where:              # all selectors (point + file)
	    file:
	      prefix: **something**
	  do:
		XXX
```

kakkoyun:
After the extended discussion on this PR (March–May 2026), the schema shape is now settled and implemented. This comment summarizes what was decided and links to the artifacts so reviewers can verify the loop is closed.

**Agreed model: 2-tier `where`/`do`**

```yaml
rule_name:
  target: net/http           # package scope — exact import path
  version: "v1.0.0,v2.0.0"  # optional version range
  where:                     # selectors (read intent)
    func: ServeHTTP
    recv: serverHandler
    file:
      has_func: init         # file-level predicate
  do:                        # modifiers (write intent)
    - inject_hooks:
        before: BeforeServeHTTP
        after: AfterServeHTTP
        path: github.com/example/nethttp/server
```

**What was dropped from earlier proposals:**

- `import_path` / `package_name` filters — agreed to drop (redundant with `target`; glob deferred to `target` field in a follow-up per txabman42's suggestion)
- 3-tier Package→File→Point model — merged into 2-tier: `target`/`version` remain top-level, everything else lives under `where` (point selectors + `where.file` for file predicates)
- ADR-0004 (import path glob matching) — deleted; strategy folded into `target` globbing follow-up

**What is now in the branch:**

- Code: `rule.Normalize` (`tool/internal/rule/normalize.go`) translates the structured `where`/`do` form to the internal flat fields. `filter.Build` (`tool/internal/filter/build.go`) compiles `FilterDef` to runtime `Filter`. Both are tested.
- Docs: ADR-0003 trimmed to Nygard form (decision record only). Schema reference, field tables, invalid-shape examples, and flat→structured migration guide live in `docs/rules.md`.
- Golden fixtures under `tool/internal/instrument/testdata/golden/**/rules.yml` are all in canonical 2-tier form.

Requesting re-review from @darccio, @y1yang0, and @txabman42. The open inline threads have existing replies; please mark them resolved if satisfied.

docs/adr/0003-join-point-filter-interface.md
kakkoyun: Addressed in commit `1328f9bd`: ADR-0003 is now a pure Nygard decision record (Status/Context/Decision/Consequences, no options tables). The schema reference, field tables, valid/invalid shape examples, and flat→structured migration guide have been moved to [`docs/rules.md`](https://github.com/kakkoyun/opentelemetry-go-compile-instrumentation/blob/kakkoyun/op4/base/docs/rules.md). The user guide now lives in `docs/` as you suggested.

tool/internal/setup/match.go
kakkoyun: Confirming no code change needed. The contract is: `ParseFileFast` calls `parser.ParseFile` (Go stdlib), whose return guarantee is `(f *ast.File, err error)` where `f` is non-nil whenever err is nil — this is enforced by `go/parser` internals. The dave/dst `DecorateFile` call that follows has the same guarantee. So `ParseFileFast` cannot return `(nil, nil)`. The full trace-down is in the earlier reply at this thread. Happy to add a contract comment to `ParseFileFast` if that would give you more confidence.

darccio: It's good enough as it is.

docs/rules.md
```diff
@@ -2,28 +2,243 @@
```
txabman42, line 2: Worth a TOC in this document. Too many sections

docs/rules.md
```diff
@@ -2,28 +2,243 @@
...
+| `target`  | yes      | Package import path. Exact match against compile-time `-p` flag.   |
+| `version` | no       | Version range `start_inclusive,end_exclusive`. Omit to match all.  |
+| `where`   | no       | Non-package selectors and file-level predicates.                   |
+| `do`      | yes      | Ordered modifier list. Modifier name declares the rule type.       |
+| `imports` | no       | `alias: path` map merged into instrumented files.                  |
+| `name`    | no       | Explicit rule name; defaults to the YAML map key.                  |
+
+### `where` semantics
+
+- Flat selector keys inside `where` are an implicit `all-of`.
```
txabman42, line 105: Could we add a demo example covering where + where.file scenarios? It should be easier to understand for someone without context. We can collapse these sections for deeper insights.

Sth like:

```yaml
foo_hook:
  ...
  where:
    func: Open # or: struct, function_call, directive, kind, identifier
    recv: "*DB" # optional; method receiver (usually with func)
    file:
      has_func: init # or: has_struct, has_directive
      recv: "*DB" # optional; narrows has_func
      all-of: # or: one-of, not
        - has_func: init # leaves: has_func | has_struct | has_directive
        - has_struct: connector
  do:
    ...
```

It could also be interesting to specify the combination of multiple all-of, one-of, etc.

docs/rules.md
```diff
@@ -2,28 +2,243 @@
...
+  target: net/http
+  where:
+    file: init       # ERROR: where.file must be a map
+  do:
+    - add_file:
+        file: helpers.go
+        path: github.com/example/helpers
+```
+
+### Migrating from the flat schema
```
txabman42: Do we need this section?

docs/rules.md
```diff
@@ -2,28 +2,243 @@
...
+| `raw` | `do: - inject_code: raw:` |
+| `new_field` | `do: - add_struct_fields: new_field:` |
+| `file` (add_file) | `do: - add_file: file:` |
+| `replace`, `append_args` (wrap_call) | `do: - wrap_call:` |
+| `replace`, `wrap` | `do: - assign_value: replace:` or `do: - assign_value: wrap:` |
+| `imports` | top-level (unchanged) |
+
+---
+
 ## Common Fields
```
txabman42: This section is quite related to `### Top-level fields`. Could just merge them into one?

docs/rules.md
```diff
@@ -35,10 +250,13 @@ This is the most common rule type. It injects function calls at the beginning (`
...
 
-**Fields:**
+**Selectors (under `where`):**
 
 - `func` (string, required): The name of the target function to be instrumented.
 - `recv` (string, optional): The receiver type for a method. For a standalone function, this field should be omitted. For a pointer receiver, it should be prefixed with `*`, e.g., `*MyStruct`.
+
+**Modifier (`do: - inject_hooks:`):**
+
 - `before` (string, optional): The name of the function to be called at the entry of the target function.
```
txabman42, line 256: Out of PR scope

Adding other modifiers would silently be dropped. What about creating an issue for linting rules? So we ensure we don't add invalid modifiers/selectors.

docs/rules.md
```diff
@@ -391,25 +650,34 @@ This rule instruments functions annotated with a magic comment (a "directive") b
 
 **Fields:**
 
+**Selectors (under `where`):**
+
 - `directive` (string, required): The directive name to match, without the leading `//`. Must not contain spaces. For example, `otelc:span` matches the comment `//otelc:span`. Note that a space after `//` (e.g., `// otelc:span`) does **not** match — the directive must immediately follow `//`.
+
+**Modifier (`do: - expand_directive:`):**
+
 - `template` (string, required): Go statements to prepend to each matching function body. Rendered with [fasttemplate](https://github.com/valyala/fasttemplate) using `{{` / `}}` delimiters. The only supported placeholder is `{{FuncName}}`, which is replaced with the name of the annotated function.
```
txabman42, line 725: In the example we are using `replace` instead of `template`. Mismatch in naming

tool/internal/setup/testdata/where/func-filter/expected.json
```diff
@@ -0,0 +1,5 @@
+{
```
txabman42: Are these tests under `setup/testdata/where` even executed?

tool/internal/rule/normalize.go
```diff
@@ -0,0 +1,196 @@
...
+//	version: "v1.0.0,v2.0.0"
+//	func: Open
+//	before: BeforeOpen
+//	path: "github.com/example/sql"
+//	where:
+//	  file:
+//	    has_func: init
+//	imports:
+//	  fmt: fmt
+func Normalize(fields map[string]any) ([]map[string]any, error) {
```
txabman42, line 90: I miss some unit tests for all this logic. Right now there are some tests related in the matcher, but I wouldn't like to just keep the unit testing of this logic in another package.

docs/rules.md
```diff
@@ -2,28 +2,243 @@
...
+  `identifier`.
+- File-level predicates live under `where.file`.
+- `target` and `version` **must not** appear inside `where`. They are
+  package-scope selectors and stay top-level.
+
+### `where.file` semantics
+
+- Predicate keys: `has_func`, `recv`, `has_struct`, `has_directive`.
+  Combinator keys: `all-of`, `one-of`, `not`.
+- `recv` inside `where.file` narrows `has_func` to a specific receiver type.
```
txabman42: `has_recv` in order to continue with the naming format? is_ or has_ under where.file

tool/internal/rule/normalize.go
```diff
@@ -0,0 +1,196 @@
...
+
+package rule
+
+import (
+	"maps"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/ex"
+)
+
+// Normalize detects the structured target/version + where + do format defined
```
txabman42, line 51: Do we need to end up with a flat version of the rule? I see ok current approach under the scope of this PR, but, which are the benefits of keeping a flat version? Could rule constructors be refactored to work directly with the new rule YAML schemas?

tool/internal/rule/normalize.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	for _, modifierRaw := range modifier {
+		modifierFields, ok := modifierRaw.(map[string]any)
+		if !ok {
+			return nil, ex.Newf("do modifier payload must be a map")
+		}
+		return []map[string]any{maps.Clone(modifierFields)}, nil
+	}
+	// Unreachable: len-1 map iteration guaranteed above.
+	return nil, ex.Newf("do reached an unsupported shape")
```
amazingakai: `util.ShouldNotReachHere`?

docs/rules.md
```diff
@@ -163,24 +401,30 @@ This rule wraps function calls at call sites with instrumentation code. Unlike t
...
 
-**Replacement Template System:**
+**Template System:**
 
-The `replace` field uses Go's standard `text/template` package for code generation. This provides:
+The `template` field uses Go's standard `text/template` package for code generation. This provides:
 
 - **Placeholder Substitution**: `{{ . }}` is replaced with the original function call's AST node
-- **Type Safety**: The replacement template is compiled at rule creation time and validated
+- **Type Safety**: The template is compiled at rule creation time and validated
```
amazingakai: `template` was renamed to `replace`

docs/rules.md
```diff
@@ -2,28 +2,243 @@
...
+| ------------------- | ------------------- |
+| `inject_hooks`      | `InstFuncRule`      |
+| `inject_code`       | `InstRawRule`       |
+| `add_struct_fields` | `InstStructRule`    |
+| `add_file`          | `InstFileRule`      |
+| `wrap_call`         | `InstCallRule`      |
+| `expand_directive`  | `InstDirectiveRule` |
+| `assign_value`      | `InstDeclRule`      |
+
+Rule type comes from the modifier key in `do`, not from field-presence priority.
```
amazingakai: This does not seem to be true yet. The implementation still appears to ignore modifier names and infer rule types from field presence like before. Do you plan to do a followup for this?

amazingakai (commented):
I tried reviewing this a bit, though I probably lack some context since many of the design decisions here were likely discussed previously in meetings.

* `instrument/testdata/golden/all-rule-external-wrapper`, `tool/internal/instrument/testdata/golden/decl-rule-wrap-expression` still uses the old rule style.

* `setup/testdata/where` appears unused.

## Questions About Future Direction

* Since `target` is not part of `where`, how would we support call-site instrumentation use cases where we want to exclude specific packages? For example, allowing all targets except `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc` for gRPC instrumentation.

* `WhereDef` contains `func`, `recv`, and other rule-specific fields, but right now we flatten the `where`/`do` schema back into flat rule structs where both selectors and modifiers live together. I wonder if it would make sense long-term to model selectors and modifiers separately (like FuncSelector -> FuncHookModifier (inject_hooks), etc.) and validate whether a given join point supports the requested modifiers. This would probably require a much larger refactor though. Maybe worth tracking in a followup issue?

## Needs Followup (Not directly related to this PR but noticed while reviewing)

* `do` supports sequences, but sequence support for `func` rules seems broken right now because it generates multiple rules that all try to emit trampoline functions with the same name. Probably worth addressing in a followup issue if we want to support multiple hooks/advice chains for a single function.

tool/internal/filter/build.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
+)
+
+// Build constructs a runtime Filter from a structured where clause.
+//
+// A nil result is valid and means the rule has no executable where.file
+// predicate.
+//
+//nolint:nilnil // nil Filter means "no executable file predicate"
+func Build(where *rule.WhereDef) (Filter, error) {
```
y1yang0, line 93: `filter` is an attribute of rule. I suggest getting rid of the tool/instrument/filter folder and using tool/internal/rule/filter.go instead.

tool/internal/filter/leaf.go
```diff
@@ -0,0 +1,30 @@
...
+package filter
+
+import "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/ast"
+
+var (
+	_ Filter = (*FuncFilter)(nil)
+	_ Filter = (*StructFilter)(nil)
+)
+
+// FuncFilter matches source files that declare the named function or method.
```
y1yang0: These files are quite simple and are essentially all part of the 'filter'. We can just use a single filter.go instead of having build.go + leaf.go + filter.go

tool/internal/setup/match.go
```diff
@@ -32,9 +33,10 @@ const (
 	matchDepsConcurrencyMultiplier = 2
 )
 
-// createRuleFromFields creates a rule instance based on the field type present in the YAML
+// createRuleFromFields creates a rule instance based on the field type present
+// in the (already-normalized) flat YAML fields map produced by [rule.Normalize].
 //
-//nolint:nilnil // factory function
+//nolint:nilnil // factory function; default branch is unreachable per normalizeRule
```
y1yang0: Maybe we can remove nilnil check, I've seen many usages in my mind

y1yang0 (commented):
First batch of review comments.

Also, an open question: is there a better name for the `Do` keyword? Do you think the current `Do` is appropriate?

cc @other_reviewers.

tool/internal/rule/normalize.go
```diff
@@ -0,0 +1,196 @@
...
+//	func: Open
+//	before: BeforeOpen
+//	path: "github.com/example/sql"
+//	where:
+//	  file:
+//	    has_func: init
+//	imports:
+//	  fmt: fmt
+func Normalize(fields map[string]any) ([]map[string]any, error) {
+	_, hasWhere := fields["where"]
```
txabman42: Lets use constants for strings like "where", "do", etc.

tool/internal/rule/normalize.go
```diff
@@ -0,0 +1,196 @@
...
+		return nil, ex.Newf("target must be top-level, not inside where")
+	}
+	if _, exists := where["version"]; exists {
+		return nil, ex.Newf("version must be top-level, not inside where")
+	}
+
+	normalized := make(map[string]any)
+	for key, value := range where {
+		switch key {
+		case "func", "recv", "struct", "function_call", "directive", "kind", "identifier":
```
txabman42: I can buy this for now (we can work in another PR) but it will end up causing problems in the future. It's really easy to forget about updating code like this, referencing selectors manually.

I would have a specific place for registering all possible selectors/modifiers. You can have public methods like IsValidSelector(), etc.

You encapsulate all selector/modifier logic in one place. This would also open up the idea of creating a direct dependence between rules and selectors/modifiers, making them first-class citizens.

tool/internal/rule/base.go
```diff
@@ -19,10 +19,50 @@ import (
...
+	HasDirective string `json:"has_directive,omitempty" yaml:"has_directive,omitempty"`
+}
+
+// WhereDef carries the structured where clause after package selectors have
+// been split back out to top-level target/version fields.
+//
+// Today the setup phase only executes where.file predicates. The remaining
+// selector and qualifier fields are preserved here so the agreed syntax surface
+// can be normalized now without forcing the broader internal refactor yet.
+type WhereDef struct {
```
txabman42, line 52: `normalizeWhere` is not aligned with this. Some are dead code right now.

tool/internal/setup/filter.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, ex.Newf("where all-of selector composition is not yet supported")
+	}
+	if len(where.OneOf) > 0 {
+		return nil, ex.Newf("where one-of selector composition is not yet supported")
+	}
+	if where.Not != nil {
+		return nil, ex.Newf("where not selector composition is not yet supported")
+	}
+
+	if where.Func != "" || where.Recv != "" || where.Struct != "" ||
```
txabman42, line 108: `WhereDef` could expose this logic

txabman42 (changes requested):
Really nice PR in the right direction! ❤️ 

My main concern:

I ended up accepting the tier-2 approach based on that selectors/modifiers would become first-class citizens. Right now, tasks like OP4.1-6 will touch the schema, `WhereDef`, the flat struct, and the docs. Where/Do are not independent with the tier-2 approach, they are tightly coupled. So basically, the same concern @amazingakai commented:

> WhereDef contains func, recv, and other rule-specific fields, but right now we flatten the where/do schema back into flat rule structs where both selectors and modifiers live together. I wonder if it would make sense long-term to model selectors and modifiers separately (like FuncSelector -> FuncHookModifier (inject_hooks), etc.) and validate whether a given join point supports the requested modifiers. ...

This shouldn't be a blocker right now, but WDYT about a follow-up issue for pushing this?

Extra:
- Normalization required in the long term?
- Linting for rules as a future improvement?

docs/rules.md
kakkoyun: Good catch!

docs/rules.md
kakkoyun: Yes, this PR lays out a plan. I would be happy to keep this section in the ADR.

kakkoyun:
> Also, an open question: is there a better name for the `Do` keyword? Do you think the current `Do` is appropriate?

@y1yang0 How about `apply`?

tool/internal/setup/match.go
kakkoyun: I think so yes. We have too many exceptions.

kakkoyun:
Pushed 4 commits addressing the review feedback. Quick summary:

**In this PR:**

- `6a103fa3`: removed orphan `setup/testdata/where/` fixtures (no test referenced them).
- `1a2a7557`: consolidated `tool/internal/filter/` into `tool/internal/rule/filter.go`, renamed `where.file.recv` → `has_recv` for symmetry with `has_func`/`has_struct`/`has_directive`, extracted schema-key constants reused across `normalize.go` and `setup/match.go`, added `normalize_test.go` for direct unit coverage, dropped the `//nolint:nilnil` exception (`createRuleFromFields` now returns an explicit error for the unknown-shape branch), and switched the unreachable case in `normalize.go` to `util.ShouldNotReachHere` to match the existing convention.
- `e6b09b32`: added a TOC and a `Quick demo` section to `docs/rules.md`, aligned the call-wrap prose to `replace` where the YAML key is `replace`, folded the duplicate `Common Fields` section into `Top-level fields`, removed the (unreleased-schema) migration section, and softened the modifier-name-dispatch claim to mark it as planned.
- `2d38194e`: migrated `all-rule-external-wrapper` and `decl-rule-wrap-expression` golden fixtures to the canonical `where`/`do` form. The generated `*.go.golden` outputs are byte-identical because `normalize.go` re-emits the same flat fields.

**Follow-ups filed (each issue has full context):**

- #541 — Consume structured `where`/`do` in rule constructors (drop `normalize.go`).
- #542 — Central selector/modifier registry with `IsValidSelector` / `IsValidModifier`.
- #543 — Rule schema linter (depends on #542).
- #544 — `do` sequence emits duplicate trampoline names for `func` rules.
- #545 — Target exclusion / `not target:` for call-site instrumentation.
- #546 — Derive rule type from `do` modifier name (likely subsumed by #541).

---

@y1yang0 on the `do` keyword name — kept `do` for this PR; it pairs naturally with `where` and reads cleanly as "set of things to do" in line with AOP advice terminology. Happy to revisit if a clearer pair emerges.

@txabman42 the architectural concern about the flat-rule normalization layer is tracked in #541 with full context; the constants added in this PR are a stepping stone toward the typed registry in #542. Re-requesting your review.

@amazingakai old-style fixtures migrated; follow-ups filed for the `do` sequence trampoline collision (#544) and the target exclusion question (#545).

docs/rules.md
kakkoyun: Added TOC in `e6b09b32`.

docs/rules.md
kakkoyun: Added a worked `where` + `where.file` Quick demo in `e6b09b32`.

docs/rules.md
kakkoyun: Removed in `e6b09b32`. The schema is unreleased so the migration narrative adds noise.

docs/rules.md
kakkoyun: Folded into `Top-level fields` in `e6b09b32`.

docs/rules.md
kakkoyun: Filed #543 for the rule-schema linter (depends on the central registry in #542).

docs/rules.md
kakkoyun: Aligned the call-wrap prose to `replace` in `e6b09b32`. The directive section's `template` is a separate fasttemplate field and stays as-is.

tool/internal/setup/testdata/where/func-filter/expected.json
kakkoyun: They were never wired into any test. Deleted in `6a103fa3`; equivalent coverage lives in `tool/internal/rule/testdata/where/`.

tool/internal/rule/normalize.go
kakkoyun: Added `tool/internal/rule/normalize_test.go` with passthrough / hoisting / error-path coverage in `1a2a7557`.

docs/rules.md
kakkoyun: Renamed to `has_recv` throughout — YAML tag, docs, and testdata fixtures all updated (`1a2a7557` + `e6b09b32`).

tool/internal/rule/normalize.go
kakkoyun: Agreed. This will be tracked in #541 with full context. Keeping branch 0 scoped so the structural shift lands in a dedicated PR.

tool/internal/rule/normalize.go
kakkoyun: Switched to `util.ShouldNotReachHere()` in `1a2a7557` to match the convention used in `match.go` / `filter.go`.

docs/rules.md
kakkoyun: Aligned to `replace` in `e6b09b32`.

docs/rules.md
kakkoyun: Softened the doc wording to mark modifier-name-driven dispatch as planned (`e6b09b32`). Follow-up tracked in #546 (likely subsumed by #541).

tool/internal/filter/build.go
kakkoyun: Moved to `tool/internal/rule/filter.go` in `1a2a7557`.

tool/internal/filter/leaf.go
kakkoyun: Consolidated into a single `tool/internal/rule/filter.go` in `1a2a7557`.

tool/internal/setup/match.go
kakkoyun: Removed in `1a2a7557` — `createRuleFromFields` now returns an explicit error for the unknown-shape branch.

tool/internal/rule/normalize.go
kakkoyun: Replaced inline literals with exported `KeyWhere` / `KeyDo` / `SelXxx` / `WhereXxx` constants in `1a2a7557`; reused them in `setup/match.go`.

tool/internal/rule/normalize.go
kakkoyun: Agreed, and filed #542 for the typed `IsValidSelector` / `IsValidModifier` registry. The constants added in this PR are a stepping stone.

tool/internal/rule/base.go
kakkoyun: The unset `WhereDef` selector fields are unreachable after normalize hoists them out. Left them in place because the structured-form constructor refactor (#541) will consume them directly.

tool/internal/setup/filter.go
kakkoyun: Considered exposing as `(*WhereDef).Build()`; kept it a free `rule.Build` to limit churn in this PR. Easy to convert when #541 lands.

txabman42 (approved):
GJ! TY for pushing this! More than happy to approve it!

Nice PR and nice follow-up issues ❤️

tool/internal/setup/match.go
```diff
@@ -66,16 +64,22 @@ func parseRuleFromYaml(content []byte) ([]rule.InstRule, error) {
...
+		for _, flatFields := range flatRules {
+			raw, err1 := yaml.Marshal(flatFields)
+			if err1 != nil {
+				return nil, ex.Wrap(err1)
+			}
 
-		r, err2 := createRuleFromFields(raw, name, fields)
-		if err2 != nil {
-			return nil, err2
+			r, err2 := createRuleFromFields(raw, name, flatFields)
```
darccio, line 76: Is it possible that `name` will be the same for each custom rule? Wouldn't that cause that `loadCustomRules`, which uses `r.GetName()` to deduplicate, doesn't load all the rules?

A minimal repro shape would be:

```yaml
combo:
  target: main
  where:
    func: Example
  do:
    - inject_hooks:
        before: BeforeExample
        path: example.com/hooks
    - inject_code:
        raw: "_ = 1"
```

Expected after `parseRuleFromYaml`: 2 rules, both named `combo`.

Expected after `loadCustomRules`: currently 1 rule, because `map["combo"] = r` is assigned twice.

If we want to preserve the existing “same rule file passed twice should dedupe” behavior, the fix should dedupe at the YAML-entry level while retaining all expanded rules for that entry, e.g. `map[string][]rule.InstRule` rather than `map[string]rule.InstRule`.

darccio (changes requested):
I think we need to migrate these match-narrowing fields from my PRs:

 ```go                                                                           
   Signature         *FuncSignature `yaml:"signature"`                           
   SignatureContains *FuncSignature `yaml:"signature_contains"`                  
   Result            string         `yaml:"result"`                              
   LastResult        string         `yaml:"last_result"`                         
   Param             string         `yaml:"param"`                               
 ```

Agent suggest something like this.

Add constants:

```go
SelSignature = "signature"
SelSignatureContains = "signature_contains"
SelResult = "result"
SelLastResult = "last_result"
SelParam = "param"
```

Then hoist them in `normalizeWhere`:

```go
case SelFunc,
    SelRecv,
    SelStruct,
    SelFunctionCall,
    SelDirective,
    SelKind,
    SelIdentifier,
    SelSignature,
    SelSignatureContains,
    SelResult,
    SelLastResult,
    SelParam:
    common[key] = value
```

That keeps current internal flat constructors unchanged.

Then add a normalize test showing:

```yaml
where:
  func: Open
  signature:
    args: [context.Context]
do:
  - inject_hooks:
      before: BeforeOpen
      path: example.com/hooks
```

normalizes into a flat func rule with both `func` and `signature`.

tool/internal/setup/match.go
kakkoyun: Confirmed and fixed in `df4f6e97`. Switched `loadCustomRules` to `map[string][]rule.InstRule` as you suggested. Each entry name keys the full slice of rules it expands into, and re-reading the same entry replaces the group as a unit, so the "same file passed twice dedupes" behavior is preserved. Added `TestDoSequenceLoadsAllExpandedRules` using your repro; it fails on the old code (1 rule) and passes with the fix (2 rules).

kakkoyun:
> I think we need to migrate these match-narrowing fields from my PRs:
> 
> ```go
>   Signature         *FuncSignature `yaml:"signature"`                           
>   SignatureContains *FuncSignature `yaml:"signature_contains"`                  
>   Result            string         `yaml:"result"`                              
>   LastResult        string         `yaml:"last_result"`                         
>   Param             string         `yaml:"param"`                               
> ```
> 
> Agent suggest something like this.
> 
> Add constants:
> 
> ```go
> SelSignature = "signature"
> SelSignatureContains = "signature_contains"
> SelResult = "result"
> SelLastResult = "last_result"
> SelParam = "param"
> ```
> 
> Then hoist them in `normalizeWhere`:
> 
> ```go
> case SelFunc,
>     SelRecv,
>     SelStruct,
>     SelFunctionCall,
>     SelDirective,
>     SelKind,
>     SelIdentifier,
>     SelSignature,
>     SelSignatureContains,
>     SelResult,
>     SelLastResult,
>     SelParam:
>     common[key] = value
> ```
> 
> That keeps current internal flat constructors unchanged.
> 
> Then add a normalize test showing:
> 
> ```yaml
> where:
>   func: Open
>   signature:
>     args: [context.Context]
> do:
>   - inject_hooks:
>       before: BeforeOpen
>       path: example.com/hooks
> ```
> 
> normalizes into a flat func rule with both `func` and `signature`.

@darccio Done in `63ca9799`. Registered the five signature selectors (`signature`, `signature_contains`, `result`, `last_result`, `param`) and added them to the hoist case in `normalizeWhere`, so a `where` block can carry them next to `func`. The flat constructors are unchanged, as you suggested. Added a normalize test asserting all five hoist into the flat func rule, and migrated the `docs/rules.md` signature examples to the `where`/`do` form.

darccio (approved):
GJ!

kakkoyun:
@y1yang0 Are you okay to proceed?

amazingakai (approved):
LGTM. Thanks for creating the followup issues ❤️

y1yang0 (approved):
Sorry Kemal, I only approved it in my head and forgot to leave a physical record.

LGTM!

PR #380, opened 2026-03-20
feat(instrumentation): Add Otel SDK instrumentation and use gls for span

NameHaibinZhang:
## Description

This PR is mainly used to instrument the OTel SDK, implementing context propagation of spanContext. It also supports compatibility with OTel SDK version 1.39 and above. Currently, the saving of baggage has not yet been implemented, and requires the tool to support a file replace approach.

## Motivation

Fixes #

---

## Checklist

- [x] Tests pass: `make test`


pkg/instrumentation/otel/sdk/trace/otel_trace_context.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func traceContextAddSpan(span trace.Span) {
+	tc := getOrInitTraceContext()
+	if tc.add(span) {
+		setTraceContext(tc)
+	}
+}
+
+func getTraceAndSpanId() (string, string) {
```
txabman42: getTraceContext(), getTraceAndSpanId(), clearTraceContext() and localRootSpanFromGLS() are not used? Are they planned to be used in a future PR?

pkg/instrumentation/otel/sdk/trace/otel_trace_test_func_holder.go
```diff
@@ -0,0 +1,11 @@
+//go:build ignore
+
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package trace
+
+var (
```
txabman42: These func holders are not used.

pkg/instrumentation/otel/sdk/trace/otel_trace_context.go
```diff
@@ -0,0 +1,160 @@
...
+	prev *spanWrapper
+}
+
+func (tc *traceContext) size() int {
+	return tc.n
+}
+
+func (tc *traceContext) add(span trace.Span) bool {
+	if tc.n > 0 {
+		if tc.n >= maxSpans {
```
txabman42: In this case, I'd follow the other rules approach (like grpc) and add a README.md for this rule. I think it's important to explain how this rule does.

Also its important to have documented values like maxSpans, as dropping spans silently if a goroutine creates >300 spans is not ideal. It should be configurable.

pkg/instrumentation/otel/sdk/baggage/otel_baggage_util.go
```diff
@@ -0,0 +1,49 @@
...
+
+func SetBaggageToGLS(baggage *Baggage) {
+	runtime.SetBaggageContainerToGLS(&BaggageContainer{baggage})
+}
+
+func ClearBaggageInGLS() {
+	SetBaggageToGLS(nil)
+}
+
+func DeleteBaggageMemberInGLS(key string) bool {
```
txabman42: Not used for now AFAIU. Could be moved to a new PR?

Same for `otel_metric_test_func_holder.go`

pkg/instrumentation/otel/hook/span_setup.go
```diff
@@ -0,0 +1,30 @@
...
+	_ "unsafe"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst"
+	"go.opentelemetry.io/otel/trace"
+)
+
+//go:linkname traceContextDelSpan go.opentelemetry.io/otel/sdk/trace.traceContextDelSpan
+func traceContextDelSpan(span trace.Span)
+
+func nonRecordingSpanEndOnEnter(ictx inst.HookContext, span interface{}, options ...interface{}) {
```
txabman42: `nonRecordingSpanEndOnEnter` and `recordingSpanEndOnEnter`. I suggest moving the code to a private func, so its clearer that both funcs do the same.

pkg/instrumentation/otel/hook/tracer_setup.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst"
+	"go.opentelemetry.io/otel/trace"
+)
+
+//go:linkname traceContextAddSpan go.opentelemetry.io/otel/sdk/trace.traceContextAddSpan
+func traceContextAddSpan(span trace.Span)
+
+func newRecordingSpanOnExit(ictx inst.HookContext, span interface{}) {
+	if span != nil {
```
txabman42: Same as https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/380/changes#r3018434400

txabman42 (changes requested):
Nice PR and nice fixes in the tool!

Documentation request for this rule + some nitpicks

pkg/instrumentation/otel/sdk/trace/otel_trace_context.go
NameHaibinZhang: yes, i will use in log instrument trace and span id, and other sdk.

pkg/instrumentation/otel/sdk/trace/otel_trace_test_func_holder.go
NameHaibinZhang: delete

pkg/instrumentation/otel/sdk/trace/otel_trace_context.go
NameHaibinZhang: use a env to get maxSpans

pkg/instrumentation/otel/sdk/baggage/otel_baggage_util.go
NameHaibinZhang: delete

pkg/instrumentation/otel/hook/span_setup.go
NameHaibinZhang: ok

pkg/instrumentation/otel/hook/tracer_setup.go
NameHaibinZhang: ok

pkg/instrumentation/otel/hook/tracer_setup.go
```diff
@@ -0,0 +1,30 @@
...
+	_ "unsafe"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst"
+	"go.opentelemetry.io/otel/trace"
+)
+
+//go:linkname traceContextAddSpan go.opentelemetry.io/otel/sdk/trace.traceContextAddSpan
+func traceContextAddSpan(span trace.Span)
+
+func newRecordingSpanOnExit(ictx inst.HookContext, span interface{}) {
```
y1yang0: We prefer to use `Before` and `After` suffix.

pkg/instrumentation/otel/sdk/trace/otel_trace_context.go
```diff
@@ -0,0 +1,165 @@
...
+
+	trace "go.opentelemetry.io/otel/trace"
+)
+
+const defaultGLSMaxSpans = 1000
+
+var otelGLSMaxSpans = defaultGLSMaxSpans
+
+func init() {
+	ms := os.Getenv("OTEL_GLS_MAX_SPANS")
```
y1yang0, line 21: Document this env var

test/apps/otelsdk/main.go
```diff
@@ -0,0 +1,69 @@
...
+	http.HandleFunc("/otel", otelHandler)
+
+	go func() {
+		if err := http.ListenAndServe(addr, nil); err != nil {
+			log.Fatalf("failed to serve: %v", err)
+		}
+	}()
+
+	// Wait for the server to start
+	time.Sleep(2 * time.Second)
```
y1yang0: Instead of this fixed-duration wait, could we use a more robust mechanism? A reference implementation for this pattern can be found in demo/app/http/server/main.go.

tool/internal/instrument/apply_file.go
```diff
@@ -46,8 +51,13 @@ func (ip *InstrumentPhase) applyFileRule(ctx context.Context, rule *rule.InstFil
...
 
-	// Parse the new file into AST nodes and modify it as needed
-	root, err := ip.parseFile(file)
+	// Parse the new file into AST nodes and modify it as needed.
+	// Keep processing in-memory to avoid mutating shared temp rule files.
+	data, err := os.ReadFile(file)
+	if err != nil {
+		return ex.Wrapf(err, "reading rule source file %s", file)
+	}
+	root, err := ast.NewAstParser().ParseSource(stripBuildIgnoreTag(string(data)))
```
y1yang0, line 60: Maybe you can parseFile and replace `buildignore` comment with "" by inspecting the entire ast root?

tool/internal/instrument/apply_func.go
```diff
@@ -96,12 +96,19 @@ func collectArguments(funcDecl *dst.FuncDecl) []string {
 		}
 	}
 	for _, field := range funcDecl.Type.Params.List {
-		for _, name := range field.Names {
-			if name.Name == ast.IdentIgnore {
-				name.Name = fmt.Sprintf("%s%d", ignoredParam, idx)
-				idx++
+		if len(field.Names) == 0 {
```
y1yang0: When `len(field.Names) == 0`, why do we need to explicitly add an ignoredParam? This seems incorrect.

tool/internal/instrument/trampoline.go
```diff
@@ -431,19 +431,28 @@ func findTargetParamType(targetFunc *dst.FuncDecl) *dst.FieldList {
 	if ast.HasReceiver(targetFunc) {
 		splitRecv := ast.SplitMultiNameFields(targetFunc.Recv)
 		recvField := util.AssertType[*dst.Field](dst.Clone(splitRecv.List[0]))
-		for _, names := range recvField.Names {
-			names.Name = fmt.Sprintf("%s%d", "recv", idx)
-			idx++
+		if len(recvField.Names) == 0 {
```
y1yang0: Ditto. When `len(field.Names) == 0`, why do we need to explicitly add an ignoredParam? This seems incorrect.

tool/internal/instrument/trampoline.go
```diff
@@ -431,19 +431,28 @@ func findTargetParamType(targetFunc *dst.FuncDecl) *dst.FieldList {
...
 	idx = 0
 	splitParams := ast.SplitMultiNameFields(targetFunc.Type.Params)
 	for _, field := range splitParams.List {
 		paramField := util.AssertType[*dst.Field](dst.Clone(field))
-		for _, names := range paramField.Names {
-			names.Name = fmt.Sprintf("%s%d", "param", idx)
+		if len(paramField.Names) == 0 {
+			paramField.Names = []*dst.Ident{{Name: fmt.Sprintf("param%d", idx)}}
 			idx++
+		} else {
```
y1yang0: Ditto

tool/internal/instrument/trampoline.go
```diff
@@ -462,9 +471,14 @@ func findTargetResultType(targetFunc *dst.FuncDecl) *dst.FieldList {
 		idx := 0
 		for _, field := range splitResults.List {
 			retField := util.AssertType[*dst.Field](dst.Clone(field))
-			for _, names := range retField.Names {
-				names.Name = fmt.Sprintf("%s%d", "arg", idx)
+			if len(retField.Names) == 0 {
+				retField.Names = []*dst.Ident{{Name: fmt.Sprintf("arg%d", idx)}}
 				idx++
+			} else {
```
y1yang0: Ditto

tool/internal/setup/match.go
```diff
@@ -172,6 +172,13 @@ func (sp *SetupPhase) runMatch(
 	}
 
 	if len(preciseRules) == 0 {
+		if !set.IsEmpty() && len(dep.Sources) > 0 {
```
y1yang0, line 175: This is expensive, but I don't have a better solution for now. Maybe we can leave a TODO for it, e.g., // TODO: Optimize this.

tool/internal/instrument/apply_func.go
NameHaibinZhang: // Unnamed parameters (e.g. func F(int)) are valid in declarations but
			// cannot be referenced in generated trampoline calls. Assign a synthetic
			// identifier so later instrumentation code can take its address.

docs/gls-operation.md
```diff
@@ -0,0 +1,80 @@
+# GLS Operation for OTel SDK Instrumentation
```
txabman42, line 1: Let's put hook-related docs inside the hook's folder. Similar to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/pkg/instrumentation/grpc/README.md

txabman42 (approved):

docs/gls-operation.md
y1yang0: Since GLS is shared across all hook code, I suggest keeping it as a top-level document.

txabman42: We can have a small doc explaining how GLS operations works with otelc, but I would keep also the hook readme with a more detailed explanation of how it works. I like the idea of having each hook with each own README.md file. They are self-contained. In the future, if we add capabilities for deciding which hooks to enable or not, it'd be easier to understand what each of them does.

PR #381, opened 2026-03-20
feat(filter): implement AllOf combinator (OP4.1)

kakkoyun:
## Summary

Implements the `AllOf` filter combinator that matches when **all** child filters match. An empty `AllOf` returns true (vacuous truth). Short-circuits on the first non-matching child.

Part of [OP4: Join point composition & filtering](#346), stacked on top of the filter interface foundation.

### Changes
- `tool/internal/filter/combinator.go`: `AllOf []Filter` type + `Match` method
- `tool/internal/filter/build.go`: recursive construction from `FilterDef.AllOf`
- `tool/internal/filter/combinator_test.go`: vacuous truth, short-circuit, nesting tests
- `tool/internal/instrument/testdata/golden/all-of-filter-match/`: filter passes → instrumented
- `tool/internal/instrument/testdata/golden/all-of-filter-no-match/`: filter blocks → original source

### Test plan
- [ ] Golden no-match file is identical to original source (no instrumentation applied)


tool/internal/filter/build.go
```diff
@@ -27,9 +27,17 @@ func Build(def *rule.FilterDef) (Filter, error) {
 }
 
 func buildDef(def *rule.FilterDef) (Filter, error) {
-	// Combinators are not yet implemented; return a clear error.
+	// AllOf combinator: match when all children match.
 	if len(def.AllOf) > 0 {
```
txabman42: This could be tricky. If AllOf is defined, all other clauses like Func, Struct, OneOf, etc. are ignored. We should document this at least

txabman42 (commented):
LGTM! No major comments except for the missing doc 😄

tool/internal/instrument/instrument_test.go
```diff
@@ -147,6 +153,38 @@ func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet
...
+		return true
+	}
+	f, err := filter.Build(&filterDef)
+	if err != nil || f == nil {
+		return true
+	}
+	tree, err := pkgast.ParseFileFast(sourceFile)
+	if err != nil {
+		return true
+	}
```
darccio: nit: would it be possible to avoid reparsing the file per rule?

tool/internal/instrument/testdata/golden/all-of-filter-match/rules.yml
```diff
@@ -0,0 +1,10 @@
+hook_all_of:
+  target: main
+  func: Func1
+  before: H1Before
+  after: H1After
+  path: testdata
+  where:
+    all-of:
+      - func: Func1
+      - struct: T
```
darccio: nit: would be good to have an empty `all-of` golden test?

darccio (commented):
Once set as open I'll approve. Right now the approval would be lost.

tool/internal/setup/filter.go
```diff
@@ -121,7 +139,13 @@ func Build(where *rule.WhereDef) (Filter, error) {
 //nolint:nilnil // unreachable default branch is guarded by util.ShouldNotReachHere
 func buildFile(def *rule.FilterDef) (Filter, error) {
 	if len(def.AllOf) > 0 {
-		return nil, ex.Newf("where.file all-of predicate composition is not yet supported")
+		// all-of owns the composition for this node; sibling predicates would be
+		// silently ignored, so reject the ambiguous combination explicitly.
+		if def.HasFunc != "" || def.HasRecv != "" || def.HasStruct != "" ||
+			def.HasDirective != "" || len(def.OneOf) > 0 || def.Not != nil {
```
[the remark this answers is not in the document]
kakkoyun: Fixed in `1cfd282`. `buildFile` now detects combinator presence via a non-nil slice (`def.AllOf != nil`) instead of `len > 0`, so an explicit `all-of: []` compiles to an empty `AllOf{}` that matches vacuously rather than erroring with "no active predicate". The sibling-predicate guard still runs for the empty case.

tool/internal/filter/build.go
kakkoyun: Documented in `1cfd282`: `buildFile` carries a comment stating that when `all-of` is present it owns composition for that node — sibling leaf predicates and other combinators are rejected at build time, not silently ignored (enforced by the guard, which also runs for `all-of: []`). Note this code is now in `tool/internal/setup/filter.go` after the package relocation.

tool/internal/instrument/instrument_test.go
```diff
@@ -163,6 +174,59 @@ func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet
...
+	case def.HasFunc != "":
+		return ast.FindFuncDecl(tree, def.HasFunc, def.HasRecv) != nil
+	case def.HasStruct != "":
+		return ast.FindStructDecl(tree, def.HasStruct) != nil
+	default:
+		t.Fatalf("golden fixture uses a where.file predicate the harness cannot "+
+			"evaluate (%+v); extend fileFilterMatches to mirror setup.buildFile", def)
+		return false
+	}
+}
```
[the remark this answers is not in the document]
kakkoyun: Tightened in `1cfd282`: `fileFilterMatches` now uses `def.AllOf != nil` to mirror `setup.buildFile`'s presence semantics, and the `t.Fatalf` default still rejects any predicate the evaluator does not model so the golden harness cannot certify shapes production would reject.

tool/internal/instrument/instrument_test.go
```diff
@@ -147,6 +153,38 @@ func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet
...
+		return true
+	}
+	var filterDef rule.FilterDef
+	if err = yaml.Unmarshal(whereBytes, &filterDef); err != nil {
+		return true
+	}
+	f, err := filter.Build(&filterDef)
+	if err != nil || f == nil {
+		return true
+	}
```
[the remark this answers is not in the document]
kakkoyun: Already safe (confirmed in `1cfd282`): `whereFileMatches` uses `require.NoError` on the YAML unmarshal, and the source parse now happens once via `require.NoError(t, parseErr)` at the parse-once site. No path returns `true` on a parse/unmarshal error.

tool/internal/instrument/instrument_test.go
kakkoyun: Fixed in `1cfd282`: the source AST is now parsed once via `ast.ParseFileFast` before the rule loop in `loadRulesYAML`, and the shared `*dst.File` is passed into the gating helper — N rules no longer trigger N reparses. Per-rule gating is unchanged.

tool/internal/instrument/testdata/golden/all-of-filter-match/rules.yml
kakkoyun: Added in `1cfd282`: golden fixture `tool/internal/instrument/testdata/golden/all-of-filter-empty/` demonstrates that an empty `all-of: []` is vacuously true and leaves the target instrumented, plus unit test `TestBuild_AllOf_Empty` and snapshot `ok_all_of_empty`.

tool/internal/filter/build.go
```diff
@@ -27,9 +27,17 @@ func Build(def *rule.FilterDef) (Filter, error) {
...
-		return nil, ex.Newf("all-of combinator is not yet supported")
+		children := make(AllOf, 0, len(def.AllOf))
+		for i := range def.AllOf {
+			child, err := buildDef(&def.AllOf[i])
```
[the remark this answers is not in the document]
kakkoyun: Addressed in `1cfd282` — see the reply on the current-code thread. The combinator now compiles `all-of: []` to a vacuously-true `AllOf{}`. (This file moved to `tool/internal/setup/filter.go` in the rebase onto the merged base.)

tool/internal/filter/build_test.go
```diff
@@ -119,15 +119,55 @@ func TestBuild_Error_MultipleActiveLeaves(t *testing.T) {
 	}
 }
 
+func TestBuild_AllOf(t *testing.T) {
+	t.Run("single func child", func(t *testing.T) {
+		def := &rule.FilterDef{AllOf: []rule.FilterDef{{Func: "Foo"}}}
+		f, err := filter.Build(def)
+		if err != nil {
+			t.Fatalf("Build(%+v) error = %v, want nil", def, err)
+		}
```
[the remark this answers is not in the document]
kakkoyun: Added empty-`AllOf` coverage in `1cfd282`: `TestBuild_AllOf_Empty` plus the `ok_all_of_empty` snapshot. (Unit tests are now in `tool/internal/setup/filter_test.go` after the package relocation.)

darccio (approved):
LGTM!

txabman42 (approved):
LGTM! ❤️ 

Same as before, PR looks great, I just miss some docs. We are going to add new combinators, and users would need to understand what they could do and how.

PR #382, opened 2026-03-20
feat(filter): implement import path glob filter (OP4.4)

kakkoyun:
## Summary

Implements `ImportPathFilter` with glob matching (`*` = one segment, `**` = zero-or-more segments). Also refactors `matchDeps` to split rules into `exactRules` map + `globRules` slice — rules with an `ImportPath` filter are evaluated against every dependency, letting the filter select the matching packages.

Part of [OP4: Join point composition & filtering](#346), stacked on `kakkoyun/op4/base` (parallel to OP4.1–4.3).

### Changes
- `tool/internal/filter/import_path.go`: `ImportPathFilter`, `matchGlob`, `matchSegments`, `ContainsImportPath`
- `tool/internal/filter/build.go`: build-time pattern validation via `path.Match`
- `tool/internal/setup/match.go`: `exactRules`/`globRules` split; `runMatch` signature changed to `[]InstRule`
- Golden tests: `import-path-filter-match/` + `import-path-filter-no-match/`

### Test plan

### DIFF
--- a/tool/internal/instrument/instrument_test.go
+++ b/tool/internal/instrument/instrument_test.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"testing"
 
+	pkgast "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/ast"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/filter"
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/rule"
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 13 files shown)


tool/internal/instrument/instrument_test.go
```diff
@@ -147,6 +153,38 @@ func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet
...
+// On any parse or build error the rule is included (fail-open).
+func whereFilterMatches(props map[string]any, sourceFile string) bool {
+	whereRaw, ok := props["where"]
+	if !ok {
+		return true
+	}
+	whereBytes, err := yaml.Marshal(whereRaw)
+	if err != nil {
+		return true
+	}
```
darccio, line 167: Should we log any YAML-related error?

kakkoyun:
Closing per the OP4 schema-narrowing decision recorded in #377: `import_path` was dropped from the stack in favor of folding glob support into the existing `target` selector, so package selection has one surface instead of two overlapping ones. The glob engine and exact/glob split here are a useful reference — tracked in #549 (target globbing), which links this PR as the reference implementation. Note the code targets the pre-merge `tool/internal/filter` layout (now `tool/internal/setup/filter.go`), so it needs porting rather than a direct cherry-pick.

PR #383, opened 2026-03-20
feat(filter): implement package name filter (OP4.5)

kakkoyun:
## Summary

Implements `PackageNameFilter` that matches source files whose declared package name equals the configured value. Reads from `ctx.AST.Name.Name` directly — no redundant field on `MatchContext`.

Note: non-test files in a package share the same declared name; external test files may declare a different name (e.g. `foo_test`).

Part of [OP4: Join point composition & filtering](#346), stacked on `kakkoyun/op4/base`.

### Changes
- `tool/internal/filter/package_name.go`: `PackageNameFilter` struct + `Match`
- `tool/internal/filter/build.go`: wired in `buildDef`
- Golden tests: `package-name-filter-match/` + `package-name-filter-no-match/`

### Test plan

### DIFF
--- a/tool/internal/instrument/instrument_test.go
+++ b/tool/internal/instrument/instrument_test.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"testing"
 
+	pkgast "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/ast"
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/filter"
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/rule"
 	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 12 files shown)


tool/internal/instrument/instrument_test.go
```diff
@@ -147,6 +153,38 @@ func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet
...
+	if err != nil {
+		return true
+	}
+	ctx := &filter.MatchContext{
+		ImportPath: mainPackage,
+		SourceFile: sourceFile,
+		AST:        tree,
+	}
+	return f.Match(ctx)
+}
```
darccio, line 186: Is this duplicated across PRs?

kakkoyun: Yeah, stacking has issues right now.

kakkoyun:
Closing per the OP4 schema-narrowing decision in #377: `package_name` was dropped from the stack. There is no current substitute (matching is by import path, not declared package clause), so the intent is preserved in #551 as a low-priority follow-up that links this PR as the reference implementation. The `_test` use case overlaps the test-package filter (#550); #551 notes that target globbing + test filter may cover every real case, in which case it can be closed as "won't do".

PR #384, opened 2026-03-20
feat(filter): implement test package filter (OP4.6)

kakkoyun:
## Summary

Implements `TestMainFilter` that selects or excludes test packages based on whether the import path carries the `.test` suffix appended by the Go toolchain. `ShouldMatch: true` → match only test packages; `ShouldMatch: false` → match only non-test packages.

`FilterDef.TestMain` is `*bool` (not `bool`) to distinguish "not specified" (`nil`) from "explicitly false" — since `false` is a meaningful value meaning "non-test packages only".

Part of [OP4: Join point composition & filtering](#346), stacked on `kakkoyun/op4/base`.

### Changes
- `tool/internal/filter/test_main.go`: `TestMainFilter` struct + `Match`
- `tool/internal/filter/build.go`: wired in `buildDef`
- Golden tests: `test-main-filter-match/` + `test-main-filter-no-match/`

### Test plan

### DIFF
--- a/tool/internal/filter/build.go
+++ b/tool/internal/filter/build.go
@@ -84,7 +84,7 @@ func buildDef(def *rule.FilterDef) (Filter, error) {
 	case def.PackageName != "":
 		return nil, ex.Newf("package_name filter is not yet supported")
 	case def.TestMain != nil:
-		return nil, ex.Newf("test_main filter is not yet supported")
+		return &TestMainFilter{ShouldMatch: *def.TestMain}, nil
 	default:
 		// Unreachable: active == 1 guarantees one of the cases above matched.
 		// If this fires, a new FilterDef field was added without a matching case.
--- a/tool/internal/filter/test_main.go
+++ b/tool/internal/filter/test_main.go
@@ -1,0 +1,32 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package filter
+
+import "strings"
+
+// Compile-time check that TestMainFilter implements Filter.
+var _ Filter = (*TestMainFilter)(nil)
+
+// TestMainFilter matches or excludes test packages based on whether the
+// package's import path carries the ".test" suffix that Go appends when
+// compiling a test binary.
+//
+// ShouldMatch == true  → match only test packages (import path ends in ".test")
+// ShouldMatch == false → match only non-test packages
+//
+// This corresponds directly to the YAML field:
+//
+//	where:
+//	  test_main: true   # instrument test packages only
+//	  test_main: false  # instrument non-test packages only
+type TestMainFilter struct {
+	ShouldMatch bool
+}
+
+// Match reports whether the test-ness of the source file's package matches
+// f.ShouldMatch.
+func (f *TestMainFilter) Match(ctx *MatchContext) bool {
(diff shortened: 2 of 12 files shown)


tool/internal/filter/build.go
txabman42, line 18: I'd remove not-yet-implemented cases, as they are already being implemented.

tool/internal/filter/test_main.go
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package filter
+
+import "strings"
+
+// Compile-time check that TestMainFilter implements Filter.
+var _ Filter = (*TestMainFilter)(nil)
+
+// TestMainFilter matches or excludes test packages based on whether the
```
txabman42, line 11: This needs to be documented also under `/docs`.

tool/internal/instrument/instrument_test.go
```diff
@@ -147,6 +153,38 @@ func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet
 	return ruleSet
 }
 
+// whereFilterMatches evaluates the optional 'where' filter for a rule.
```
txabman42, line 156: This comment doesn't strictly target this PR, but nevertheless, I see important to document where the filter doesn't apply for file rules. [setup/match.go](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/76aacbf6e3406b0c9713695d4a30c21c4db363db/tool/internal/setup/match.go#L159) just adds them directly. So this kind of filters wont affect that type of rule.

tool/internal/instrument/testdata/golden/test-main-filter-no-match/rules.yml
```diff
@@ -0,0 +1,8 @@
+hook_test_main:
+  target: main
+  func: Func1
+  before: H1Before
+  after: H1After
+  path: testdata
+  where:
+    test_main: true
```
txabman42, line 8: It's missing the test case where `test_main: true` matches `main.test`

txabman42 (commented):
Clean MR, just few comments about doc and a missing test case.

tool/internal/instrument/instrument_test.go
```diff
@@ -147,6 +153,38 @@ func loadRulesYAML(t *testing.T, testName, sourceFile string) *rule.InstRuleSet
...
+	if err != nil {
+		return true
+	}
+	ctx := &filter.MatchContext{
+		ImportPath: mainPackage,
+		SourceFile: sourceFile,
+		AST:        tree,
+	}
+	return f.Match(ctx)
+}
```
darccio, line 186: I confirm this seems to be duplicated. Is it because the stacking issue?

kakkoyun: Yes, it is a stacking issue.

kakkoyun:
Closing and converting to a tracked follow-up per #377, which reserved `is_test` / `test_main` for a later PR. Captured in #550, which links this PR as the reference implementation. The `.test`-suffix detection and tri-state `*bool` modeling here are the design to carry forward; it needs porting to `tool/internal/setup/filter.go` and the 2-tier schema.

PR #388, opened 2026-03-23
feat(tool): function or method with specific signature filter for `InstFuncRule`

darccio:
## Description

Extends `InstFuncRule` with five optional signature sub-filters that narrow function matching beyond name and receiver type:

| YAML field | Semantics |
|---|---|
| `signature` | Exact match on parameter and result types (field-list order) |
| `signature_contains` | OR-match: any listed arg type found in params **or** any listed return type found in results |
| `result_implements` | Any return type exactly matches the named type |
| `final_result_implements` | Only the last return type matches the named type |
| `argument_implements` | Any parameter type exactly matches the named type |

All filters are optional; when none are set the rule behaves exactly as before. All specified filters must match (AND logic).

Example rule:

```yaml
my_rule:
  target: example.com/pkg
  func: Open
  signature:
    args: [string]
    returns: ["*os.File", error]
  before: OnOpen
  path: example.com/hooks
```

Type-name matching is structural (DST-level, no type checker). For the `*_implements` filters this means an exact type-name match rather than full interface-satisfaction checking — matching Orchestrion's fast-path behaviour.

**New files:**
- `tool/internal/ast/typename.go` — private type-name parser + `parsedTypeName.matches` + `fieldListContainsType`
- `tool/internal/ast/typename_test.go` — unit tests for type-name parsing and field matching
- `tool/internal/ast/funcmatch_test.go` — unit tests for `FuncDeclMatchesFilters` (all five filters, combined AND logic, edge cases)
- `tool/internal/rule/func_rule_test.go` — YAML unmarshal tests for the new fields
- `tool/internal/instrument/testdata/golden/func-signature/` — golden integration test matching `Func1(string, int) (float32, error)` by exact signature

## Motivation

Instrumentation rules targeting overloaded-by-signature patterns (common in middleware, handler registration, and interface adapter code) currently require unique function names. The new sub-filters allow a rule to target a specific function variant without relying solely on its name.

Implements OP3.5 from #277.

---

## Checklist

- [x] Tests pass: `make test`


tool/internal/rule/func_rule.go
```diff
@@ -29,6 +50,13 @@ type InstFuncRule struct {
 	Before string `json:"before" yaml:"before"` // The function we inject at the target function entry
 	After  string `json:"after"  yaml:"after"`  // The function we inject at the target function exit
 	Path   string `json:"path"   yaml:"path"`   // The module path where hook code is located
+
+	// Optional signature sub-filters (all non-nil/non-empty filters must match).
+	Signature             *FuncSignature `json:"signature,omitempty"               yaml:"signature"`
+	SignatureContains     *FuncSignature `json:"signature_contains,omitempty"      yaml:"signature_contains"`
+	ResultImplements      string         `json:"result_implements,omitempty"       yaml:"result_implements"`
```
kakkoyun: Maybe thesse should also be a pointer (*string) to indicate these are optional?

tool/internal/instrument/testdata/golden/func-signature/rules.yml
```diff
@@ -0,0 +1,9 @@
+func_sig_hook:
```
kakkoyun, line 1: I think we should have test cases for each of the optional matchers separately.

tool/internal/ast/shared.go
```diff
@@ -190,6 +191,98 @@ func AddStructField(decl dst.Decl, name, t string) {
...
+// FuncDeclMatchesFilters reports whether funcDecl satisfies all signature
+// sub-filters in r.  Returns true when no sub-filters are set.
+//
+// Matching uses structural comparison of dst.Expr nodes (no type checker).
+// For the *_implements filters this means an exact type-name match rather than
+// full interface-satisfaction checking.
+func FuncDeclMatchesFilters(funcDecl *dst.FuncDecl, r *rule.InstFuncRule) bool {
+	ft := funcDecl.Type
+
+	if r.Signature != nil {
```
kakkoyun, line 281: This indicate a precedence order and also only one of them can be defined at a time. So two points: 1) Is this what we want? 2) This should be documented.

kakkoyun (commented):
Looking good. A couple of questions and suggestions, I have.

tool/internal/ast/shared.go
darccio: I don't see a better way to support all the cases. I added docs in 5dbee7a93cbe72cb97c2201120dd1a0b351c3022

tool/internal/instrument/testdata/golden/func-signature/rules.yml
darccio: Done in 5dbee7a93cbe72cb97c2201120dd1a0b351c3022

docs/rules.md
```diff
@@ -58,6 +58,65 @@ This rule will inject `MyHookBefore` at the start of the `Example` function in t
...
+| `argument_implements` | string | Any parameter type matches the named type |
+
+`signature` and `signature_contains` each take an object with two optional lists:
+
+- `args` — type names for parameters
+- `returns` — type names for results
+
+Type names follow the form `[*][pkg.]Name`, for example `error`, `context.Context`, `*http.Request`. Matching is structural (AST-level) rather than type-checker-based, so the package qualifier must match the local identifier used in the source file (typically the last path component, e.g. `http` for `"net/http"`).
+
+> **Note on `*_implements` semantics:** Because there is no type checker at instrumentation time, `result_implements`, `final_result_implements`, and `argument_implements` perform an exact type-name match rather than full interface-satisfaction checking. For example, `result_implements: error` matches functions that literally return the `error` type, but not functions that return a concrete type (e.g. `*MyError`) that happens to implement `error`.
```
txabman42: `_implements` could be misleading, its not satisfacing any interface. It just do a type-name checking. WDYT sth like `_type`? e.g. `argument_type`

tool/internal/ast/typename.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/dave/dst"
+)
+
+// typeNameRe parses type-name strings of the form [*][pkg.]Name.
+// It handles identifiers, qualified identifiers, and pointers to those.
+// Limitations: does not handle chan, func, map, slice, or interface literals.
```
txabman42, line 18: Not matching these not-supported types returns directly false. We need to be sure that the user is aware of these. I propose documenting this well in rule.md + logging at least a warning.

tool/internal/ast/shared.go
```diff
@@ -190,6 +191,102 @@ func AddStructField(decl dst.Decl, name, t string) {
...
+// matchesExactSignature returns true when funcType has exactly the parameter
+// and result types listed in sig, compared field-by-field in order.
+func matchesExactSignature(ft *dst.FuncType, sig *rule.FuncSignature) bool {
+	return matchesFieldList(sig.Args, ft.Params) &&
+		matchesFieldList(sig.Returns, ft.Results)
+}
+
+// matchesFieldList returns true when expected type strings match the types in
+// fields exactly (same count, same order).
+func matchesFieldList(expected []string, fields *dst.FieldList) bool {
```
txabman42: `fields` should be split via `SplitMultiNameFields` before comparison.

kakkoyun (approved):

txabman42 (approved):
LGTM ❤️ 

Just a small nitpick: It's missing the refactor of golden file names to `_type` e.g. `.../func_sig_last_result_implements.otelc.globals.go.golden`

darccio:
@y1yang0 Any feedback on this PR? Thanks!

y1yang0:
Sorry for the delay. We just got back from a long holiday (through May 5). I'll finish my review later this week.

kakkoyun (approved):

kakkoyun:
@darccio CI seems red. If you can fix those we can merge it.

y1yang0:
> We shouldn't be only thinking in the trampoline approach. This approach helped us immensely with call-site instrumentation.

However, it is clear that the changes in this PR only target `InstFuncRule`, so my concerns still remain, just like the two cases I mentioned earlier.

Additionally, I double-checked the origins of these features. From what I can find:

- `signature`: 3 occurrences, 2 of which are `signature: {}`
- `signature-contains`: 1 occurrence
- `result_implements`: 0 occurrences
- `final_result_implements`: 0 occurrences
- `argument_implements`: 0 occurrences

I always prefer simplicity. We should only introduce complexity when there is a pressing and justifiable need

darccio:
@y1yang0 Let's look at this the other way around: if these filters are not available, how can we implement the equivalent instrumentation already present in `dd-trace-go`? We'd be impaired and it'd be difficult for us to deprecate Orchestrion in favour of `otelc`.

Let me show you with two examples.

## How to trace errors

Imagine that we want to target functions returning an `error` value:

 ```yaml
   join-point:
     all-of:
       - package-filter:
           root: true
       - function-body:
           function:
             - signature-contains:
                 returns: [error]
   advice:
     template: |-
       {{ $ret := .Function.LastResultThatImplements "error" }}
       {{ with $ret }}
         defer func() {
           if {{ $ret }} != nil {
             {{ $ret }} = errortrace.Wrap({{ $ret }})
           }
         }()
       {{- end -}}
 ```

Right now we can't express this in `otelc`. The existing func rule requires `func: <name>`: a single, concrete function name. There is no wildcard, no glob, no iteration over all functions in a package. We could use a directive rule so it iterates all functions that carry a specific comment annotation, but it requires to opt-in by annotating. The raw rule is also name-driven. There is simply no existing rule primitive that can say "apply this to every function whose return list contains error". Without `signature_contains`, the only conceivable workaround would be to enumerate every error-returning function by name in the rule file but this is impossible to do statically for a user's application code, whose functions are not known to the instrumentation author.

The same goes for `final_result_implements`. Even if some workaround existed to match the right functions, the advice itself depends on knowing the variable name of the last error-typed return value. 

## When a type overrides an embedded type's function

Imagine this:

```go
type Base struct{}

func (b *Base) Run(name string, _ http.Handler) bool {
	fmt.Printf("Base.Run: %q\n", name)
	return true
}

type Suite struct{ Base }

type MyRunner struct {
	Suite
	custom string
}

func (mr *MyRunner) Run(name string, _ *http.ServeMux) bool {
	fmt.Printf("Custom[%q]: %q\n", mr.custom, name)
	return true
}

func main() {
	mr := new(MyRunner)
	mr.Run("ohai", http.DefaultServeMux)
}
```

If we want to instrument both `Run` methods, we need two rules for that. If we have multiple cases, the list of rules keeps growing. `argument_implements` would reduce this to a single rule.

NameHaibinZhang:
I think that in certain situations, it is necessary.

NameHaibinZhang (approved):

y1yang0:
@darccio Thanks for the examples. They make sense to me. Let's reivew the code itself.

tool/internal/setup/match.go
```diff
@@ -229,7 +229,7 @@ func (sp *SetupPhase) matchOneRule(
 	switch rt := r.(type) {
 	case *rule.InstFuncRule:
 		funcDecl := ast.FindFuncDecl(tree, rt.Func, rt.Recv)
-		if funcDecl != nil {
+		if funcDecl != nil && ast.FuncDeclMatchesFilters(funcDecl, rt) {
```
y1yang0: FuncDeclMatchesFilters can be inlined into FindFuncDecl, since it's essentially responsible for finding a FuncDecl based on the rule.

tool/internal/rule/func_rule.go
```diff
@@ -29,6 +52,14 @@ type InstFuncRule struct {
...
 	After  string `json:"after"  yaml:"after"`  // The function we inject at the target function exit
 	Path   string `json:"path"   yaml:"path"`   // The module path where hook code is located
+
+	// Optional signature sub-filters (all non-nil filters must match; combined
+	// with AND logic so any combination is allowed).
+	Signature         *FuncSignature `json:"signature,omitempty"          yaml:"signature"`
+	SignatureContains *FuncSignature `json:"signature_contains,omitempty" yaml:"signature_contains"`
+	ResultType        *string        `json:"result_type,omitempty"        yaml:"result_type"`
+	LastResultType    *string        `json:"last_result_type,omitempty"   yaml:"last_result_type"`
+	ArgumentType      *string        `json:"argument_type,omitempty"      yaml:"argument_type"`
```
y1yang0: Use `string` for consistency

tool/internal/instrument/apply_func.go
```diff
@@ -352,6 +352,12 @@ func (ip *InstrumentPhase) applyFuncRule(ctx context.Context, rule *rule.InstFun
 	if funcDecl == nil {
 		return ex.Newf("can not find function %s", rule.Func)
 	}
+	// Setup phase matched by name; skip if signature filters no longer match.
+	if !ast.FuncDeclMatchesFilters(funcDecl, rule) {
```
y1yang0: After FuncDeclMatchesFilters inlined into FindFuncDecl, this can be removed

tool/internal/instrument/testdata/golden/func-sig-argument-type/source.go
```diff
@@ -0,0 +1,45 @@
+// Copyright The OpenTelemetry Authors
```
y1yang0, line 1: After #411, every golden test now has its own dedicated source.go file, could you please clean it up so it only contains the target function that is waiting to be injected? Thanks.

tool/internal/ast/typename.go
```diff
@@ -0,0 +1,99 @@
...
+// fieldListContainsType reports whether any field in fields has a type that
+// matches typeStr.  Returns false when typeStr is invalid or fields is empty.
+func fieldListContainsType(fields *dst.FieldList, typeStr string) bool {
+	if fields == nil || len(fields.List) == 0 {
+		return false
+	}
+	tn, err := parseTypeName(typeStr)
+	if err != nil {
+		//nolint:sloglint // no context available
+		slog.Warn("signature filter: invalid type string; filter will never match", "type", typeStr, "err", err)
```
y1yang0: It seems this error should not be ignored, as a parse error typically indicates an issue with the corresponding field in the rule.

tool/internal/ast/typename.go
```diff
@@ -0,0 +1,99 @@
...
+
+	case *dst.InterfaceType:
+		// Only the empty interface matches "any".
+		return len(n.Methods.List) == 0 && t.importPath == "" && t.name == "any"
+
+	default:
+		// Unsupported AST node types (chan, func, map, slice, array, interface
+		// literals) cannot be matched by type-name filters.
+		//nolint:sloglint // no context available
+		slog.Debug("signature filter: unsupported type node; filter will not match",
```
y1yang0: Maybe util.Unimplemented() instead of slog.Debug?

tool/internal/ast/shared.go
```diff
@@ -270,6 +272,119 @@ func AddStructField(decl dst.Decl, name, t string) {
...
+		}
+	}
+	if len(expected) != len(types) {
+		return false
+	}
+	for i, typeStr := range expected {
+		tn, err := parseTypeName(typeStr)
+		if err != nil {
+			//nolint:sloglint // no context available
+			slog.Warn("signature filter: invalid type string; filter will never match", "type", typeStr, "err", err)
```
y1yang0: Ditto. https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/388/changes#r3226398376

tool/internal/rule/func_rule.go
```diff
@@ -29,6 +52,14 @@ type InstFuncRule struct {
 	Before string `json:"before" yaml:"before"` // The function we inject at the target function entry
 	After  string `json:"after"  yaml:"after"`  // The function we inject at the target function exit
 	Path   string `json:"path"   yaml:"path"`   // The module path where hook code is located
+
```
y1yang0, line 55:
```
type InstFuncRule struct {
	InstBaseRule 
	Func   string 
	Recv   string 
	Before string 
	After  string 
	Path   string 

	Signature         *FuncSignature 
	SignatureContains *FuncSignature 
	ResultType        *string        
	LastResultType    *string        
	ArgumentType      *string        
}
```
From an aesthetic standpoint, the naming feels a bit inconsistent. I'd suggest something like:
```
Signature
SignatureContains
Result
LastResult
Param
```
I admit this still seems somewhat inconsistent, but I don't have any better ideas for now.

darccio: I tried but I'm in the same place regarding better naming. I applied your suggestion.

tool/internal/instrument/apply_func.go
kakkoyun: @darccio is this resolved?

tool/internal/setup/match.go
kakkoyun: @darccio is this resolved?

darccio: @kakkoyun Yes!

tool/internal/instrument/apply_func.go
darccio: @kakkoyun Yes, I was waiting for @y1yang0 approval before resolving these.

y1yang0: How about we just inline the logic from FuncDeclMatchesFilters into FindFuncDecl? Then we wouldn't have to change all the call sites, and it looks like their logic can be combined pretty easily. Something like this:
```
func FindFuncDecl(....) {
    ....
    if rule.Signature != "" {
        ...
    }
    if rule.SignatureContains != "" {
        ...
    }
    return ...
}
```

y1yang0 (approved):
Otherwise LGTM. Thanks for the contribution

tool/internal/instrument/apply_func.go
darccio: @y1yang0 I'm going to do it in a separate PR. Thanks!

PR #395, opened 2026-03-26
refactor(tool): structured rule schema with where/do grouping

kakkoyun:
## Description

Introduce explicit `where` (selectors) and `do` (modifiers) blocks in rule YAML files, replacing the previous flat field layout where matcher and modifier fields were mixed at the same level.

**Before:**
```yaml
server_hook:
  target: net/http
  func: ServeHTTP
  recv: serverHandler
  before: BeforeServeHTTP
  after: AfterServeHTTP
  path: "github.com/.../server"
```

**After:**
```yaml
server_hook:
  where:
    target: net/http
    func: ServeHTTP
    recv: serverHandler
  do:
    inject_hooks:
      before: BeforeServeHTTP
      after: AfterServeHTTP
      path: "github.com/.../server"
```

The modifier name inside `do` determines the rule type: `inject_hooks`, `inject_code`, `add_struct_fields`, `add_file`, `wrap_call`, `expand_directive`, `assign_value`.

A `normalizeRule()` function at the YAML parsing boundary translates the new format to the internal flat representation — rule structs, constructors, validation, and JSON serialization are unchanged.

## Motivation

Rules have two concerns — **what to match** and **what to do** — but they were expressed as a flat bag of fields. Scanability suffers: it takes effort to tell selectors from modifiers without knowing the rule type. The new schema makes intent visible at a glance and gives future modifier types (`prepend_statements`, `modify_body`, etc.) a natural place to live.

ADR: `docs/adr/0003-structured-rule-schema.md`

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/docs/adr/0003-structured-rule-schema.md
+++ b/docs/adr/0003-structured-rule-schema.md
@@ -1,0 +1,108 @@
+# 3. Structured Rule Schema (where/do)
+
+Date: 2026-03-26
+
+## Status
+
+Accepted
+
+## Context
+
+Rule YAML files define compile-time instrumentation rules. Each rule has two concerns:
+
+- **Matching**: which code to target (`target`, `func`, `recv`, `struct`, `version`, …)
+- **Modifying**: what transformation to apply (`before`, `after`, `path`, `raw`, `new_field`, …)
+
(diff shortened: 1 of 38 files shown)


kakkoyun:
This will be handling as part of #377

PR #400, opened 2026-03-26
feat(tool): add --stats flag for toolexec timing

### LINKED ISSUE #389: chore: add built-in pprof support for otelc
> One of the features in `orchestrion` helped us a lot is built-in profiling support, so I will create another issue for it https://datadoghq.dev/orchestrion/contributing/performance/#built-in-profiling 

 _Originally posted by @kakkoyun in [#351](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/351#issuecomment-4110387023)_

kakkoyun:
## Description

Adds a hidden `--stats` flag (and `OTELC_STATS` env var) that logs wall-clock duration for every toolexec command invocation. When enabled, each compile/link/asm call emits a structured log entry:

```
tool=compile package=github.com/foo/bar duration=123.4ms duration_ms=123
```

**Changes:**
- `tool/util/shared.go` — `EnvOtelcStats = "OTELC_STATS"` constant, shared between `cmd` and `internal/instrument` to avoid import cycles
- `tool/cmd/stats.go` — new: `initStats` Before-hook reads `--stats` flag and sets the env var
- `tool/cmd/main.go` — hidden `--stats` flag with `Sources: cli.EnvVars(util.EnvOtelcStats)`; `initStats` chained after `initProfiling`
- `tool/internal/instrument/toolexec.go` — wraps `util.RunCmd` with `time.Now()`/`time.Since()` when `OTELC_STATS` is set; logs tool, package, duration

## Motivation

Needed for benchmarking otelc against Orchestrion. Per-tool timing identifies which compilation phases are slowest and where instrumentation overhead concentrates.

`OTELC_STATS` propagates to child toolexec processes automatically via `os.Environ()` in `BuildWithToolexec` — the same pattern used by `OTELC_PROFILE_PATH`.

Fixes #389

---

## Checklist

- [x] Tests pass: `make test`


tool/internal/instrument/toolexec.go
```diff
@@ -366,5 +367,19 @@ func Toolexec(ctx context.Context, args []string) error {
...
+	}
+	tool := filepath.Base(args[0])
+	pkg := util.FindFlagValue(args, "-p")
+	start := time.Now()
+	err := util.RunCmd(ctx, args...)
+	elapsed := time.Since(start)
+	util.LoggerFromContext(ctx).InfoContext(ctx, "toolexec stats",
+		"tool", tool,
+		"package", pkg,
+		"duration", elapsed,
```
txabman42, line 381: `duration` and `duration_ms` are redundant. Do we need both?

kakkoyun: Probably we don't. It depends on the precision you want.

tool/internal/instrument/toolexec.go
```diff
@@ -366,5 +367,19 @@ func Toolexec(ctx context.Context, args []string) error {
 	}
 
 	// Run the command
-	return util.RunCmd(ctx, args...)
+	if os.Getenv(util.EnvOtelcStats) == "" {
```
y1yang0, line 370: Maybe stat in Go build? there, we can get the flags directly from cmd.cli.Command

```
func GoBuild(){
	...
	start := time.Now()
	err := Setup(ctx, cmd)
	if err != nil {
		return err
	}
	logger.InfoContext(ctx, "Setup completed successfully")
	end := time.Now()
	logger.DebugContext("setup stats",...)

	start = time.Now()
	err = BuildWithToolexec(ctx, cmd)
	if err != nil {
		return err
	}
	logger.InfoContext(ctx, "Instrumentation completed successfully")
	end = time.Now()
	logger toolexec stat....
	return nil
}

}
```

PR #405, opened 2026-03-30
fix(tool): mixed std error and stackful error

y1yang0:
mixed std error and stackful error


.tools/golangci.yml
```diff
@@ -212,6 +212,12 @@ linters:
         # Forbid fmt.Print* in non-main packages (use structured logging instead)
         - pattern: '^(fmt\.Print(|f|ln)|print|println)$'
           msg: use structured logging instead of fmt.Print*
+        # Forbid fmt.Errorf - use ex.Newf or ex.Wrapf instead for stack traces
+        - pattern: '^fmt\.Errorf$'
```
kakkoyun, line 216: 👍

tool/internal/profile/profile.go
```diff
@@ -124,32 +125,30 @@ func (s *Session) Stop() error {
 		return nil
 	}
 
-	var errs []error
-
 	if s.cpuFile != nil {
 		pprof.StopCPUProfile()
 		if err := s.cpuFile.Close(); err != nil {
-			errs = append(errs, fmt.Errorf("close CPU profile: %w", err))
+			return ex.Wrapf(err, "close CPU profile %q", s.cpuFile.Name())
```
kakkoyun: We need joins here, these early returns breaks the intended functionality. We want to clean things up properly.

kakkoyun (changes requested):
We need some sort of `Join` to support multiple mode failures at once without early returns.

tool/internal/profile/profile.go
y1yang0: You are right, I removed `ex.Join` initially to expedite the fix, let me implement it completely

kakkoyun (approved):

Issue #351, opened 2026-03-12
On-going performance benchmarking (overhead vs. vanilla Go)


txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/261

CI/CD: On-going performance benchmarking (overhead vs. vanilla Go)

---

Proposal: add a CI benchmark that tracks compile times for our demo apps over time.
	•	Run in a clean/isolated environment (pinned Go version, consistent machine/runner).
	•	Measure compile time for the same program using:
	1.	plain go toolchain (baseline)
	2.	otelc (instrumented)
	•	Execute each benchmark N times (we can use Go’s native benchmarking harness to reduce noise).
	•	Store results somewhere persistent (details TBD).
	•	Analyze trends over time and define a regression threshold.

Automation idea: a GitHub Action evaluates the latest run vs. historical baseline:
	•	If there’s a statistically significant regression: automatically open (or update) an issue and block the next release until it’s triaged.
	•	If the regression is expected (known change): document it and close/annotate the alert.

y1yang0:
This roadmap item was proposed by @kakkoyun . @kakkoyun , could you please elaborate on it? Thanks

txabman42:
I'll try to push this issue

txabman42:
This first PR https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/425 allows us to control the overhead limit.

We can complicate this a bit more by adding the following features using [continuous-benchmark](https://github.com/marketplace/actions/continuous-benchmark):
- Store historical bench results
- Compare the PR bench results against the last ones and block the PR if it goes over the defined threshold

WDYT about pushing any of them right now?

y1yang0:
+1 for `Store historical bench results`

kakkoyun:
Let's start with storing and start gathering the trends. However, I'm a little bit reluctant to block anything at this stage. However, after v1.0.0 we should definitely put the gates in place.

txabman42:
Agree on just storing the bench results for now (https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/430).

Currently, we compare go vs otelc under the same runner. However, in case we want to compare the bench results with the previous runs, we need to be very careful, as the current runners could apply ~20-30% variation in bench results. In case we go in the future in this direction, we'd need have more control over our runners.

txabman42:
[...]

PR #425, opened 2026-04-09
feat(test): add bench tests

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/351

Current bench tests compare `go vs otelc` compile times under 3 different scenarios:
- `baseline`: just fmt
- `multi`: currently all available rules instrumented
- `largeidle`: a good amount of dependencies not instrumented

I'm open to discussing other scenarios.

The good thing about `go vs otelc` is that we don't care too much about historical perf evolution, which means we also don't care about the runners (which usually you end up needing dedicated or baremetal runners, etc.). So no need to push the adoption of tools like https://github.com/marketplace/actions/continuous-benchmark. With just some basic tricks (documented in `benchmarking.md`), we can obtain good enough results for `go vs otelc` comparison. I've experienced around ~<5% deviation between runs.

I let a 50% of extra overhead as an initial value, although we should define and document clearly our "SLOs".

Current results example:

```
2026/04/09 14:54:46 scenario=baseline      plain=3.443s (±0.007)  otelc=4.366s (±0.006)  overhead=+26.8%
2026/04/09 14:57:22 scenario=largeidle     plain=12.205s (±0.027)  otelc=13.625s (±0.002)  overhead=+11.6%
2026/04/09 15:01:13 scenario=multi         plain=14.900s (±0.076)  otelc=23.471s (±0.089)  overhead=+57.5%
```


txabman42:
Extra bench features to be discussed in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/351#issuecomment-4215766975, I want to keep this issue scoped to scenario definitions and bench workflow

y1yang0 (approved):

test/bench/scenarios/largeidle/main.go
```diff
@@ -0,0 +1,71 @@
...
+	_ "go.uber.org/multierr"
+	_ "golang.org/x/crypto/bcrypt"
+	_ "golang.org/x/crypto/sha3"
+	_ "golang.org/x/exp/maps"
+	_ "golang.org/x/text/language"
+	_ "golang.org/x/text/unicode/norm"
+	_ "gopkg.in/yaml.v3"
+)
+
+func main() {}
```
kakkoyun: How about making this more realistic? With the help of some agentic coding, I think we can make use of most of those packages. I want make sure we are not prone to any future compiler optimizations for deadcode elimination.

Exactly something like `scenarious/multi/main.go`

test/bench/cmd/bench/main.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+	sort.Strings(names)
+	return names, nil
+}
+
+func buildArgs(prefix []string) []string {
+	args := make([]string, 0, len(prefix)+5)
+	args = append(args, prefix...)
+	return append(args, "go", "build", "-a", "-o", "app", ".")
```
kakkoyun, line 158: nit: Let's add a comment that we deliberately ignore all the cache and rebuild everything each time from scratch.

kakkoyun (approved):
LGTM. I only have small suggestions. 

Thanks for doing this!

test/bench/scenarios/largeidle/main.go
txabman42: Sure, done [a37157e](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/commit/a37157e48ace2b0ec364741f4ef4b7fd9bcce46d)

test/bench/cmd/bench/main.go
txabman42: [a37157e](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/commit/a37157e48ace2b0ec364741f4ef4b7fd9bcce46d)

kakkoyun (approved):

Issue #406, opened 2026-03-31
Add test capability to check if current hook satisfies latest SDK version


y1yang0:
As discussed in the slack, maintaining fixed version ranges for instrumented libraries (e.g., `version: v1.2.3, v1.7.0)` is a manual and potentially unscalable process. To ensure our instrumentation remains compatible with new library releases without constant manual intervention, we propose implementing a new automated test.

#### **Proposed Solution**

We should introduce a new CI test, tentatively named **`FetchingLatestVersion`**, which will allow us to use open-ended version ranges (e.g., `version: v1.2.3` only) with confidence.

#### **Expected Behavior**

1.  The test should run as part of the CI pipeline.
2.  For a given instrumented library, it should automatically fetch its **latest stable version** from its registry
3.  It will then run the corresponding integration/instrumentation tests against this latest version.
4.  **If the tests pass:** The build is successful. This confirms our assumption of forward compatibility.
5.  **If the tests fail**: The build must fail. This serves as an immediate alert to maintainers about a breaking change. In this scenario, the maintainer will need to split the instrumentation rule
    1. The existing rule's version range will be capped to define an upper bound (e.g., changing [v1.2.3,) to [v1.2.3, v4.5.6)).
    2. A new rule will be created for the new and subsequent versions (e.g., [v4.5.6,)).
    3. This new rule will require its own instrumentation implementation to handle the breaking changes.

#### **Benefits**

*   **Reduces Maintenance Burden:** Automates the process of verifying compatibility with new releases.
*   **Proactive Failure Detection:** Allows us to discover breaking changes immediately, rather than waiting for user reports.
*   **Increases Confidence:** Provides continuous assurance that our instrumentation works with the latest dependencies.

This feature is a critical testing capability for the long-term health of the project. This issue is to track its implementation and is open for community contribution.

txabman42:
I can take this, but I'd like to clarify the behavior first.

Some concerns with current proposal:
- Coupling CI to "latest" would make PRs break, which in the end is not related to the PR author
- "latest stable" could be tricky or at least it would require a better definition (github releases vs module proxy mismatch, etc.) 
- A flaky upstream release or having too restrictive assertions could make the test fail, which in reality it was perfectly compatible
- Could be too slow

All this could generate noise in the PRs

I'm thinking about splitting this into 2 tests:
[...]

PR #426, opened 2026-04-10
feat(instrumentation): Add k8s-client-go informers instrumentation

### LINKED ISSUE #462: feat(instrumentation) Support Kubernetes instrumentation
Implements k8s-client-go informers auto instrumentation to automatically emit traces from informer handlers

amazingakai:
Implements k8s-client-go informers auto instrumentation to automatically emit traces from informer handlers.

Follows [K8S semconv](https://opentelemetry.io/docs/specs/semconv/resource/k8s/) with some additional keys:
- k8s.object.uid (for generic/unknown objects)
- k8s.object.name (for generic/unknown objects)
- k8s.object.api_version
- k8s.object.kind

- k8s.objects.count
- k8s.objects.is_in_initial_list

Inspired by: https://github.com/alibaba/loongsuite-go-agent/tree/main/pkg/rules/k8s-client-go

Fixes #462


pkg/instrumentation/k8s-client-go/semconv/k8s_test.go
```diff
@@ -0,0 +1,348 @@
...
+func TestK8SObjectInfoTraceAttrs(t *testing.T) {
+	tests := []struct {
+		name     string
+		objInfo  K8SObjectInfo
+		expected map[string]any
+	}{
+		{
+			name: "basic pod",
+			objInfo: K8SObjectInfo{
+				UID:        "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
```
y1yang0, line 22: Can we generate real UUID instead?

pkg/instrumentation/k8s-client-go/hook.go
```diff
@@ -0,0 +1,179 @@
...
+
+	spanName := "k8s.informer.objects.process"
+	ctx, span := tracer.Start(context.Background(),
+		spanName,
+		trace.WithSpanKind(trace.SpanKindInternal),
+		trace.WithAttributes(attrs...),
+	)
+
+	ictx.SetParam(0, newK8SOtelEventHandler(handler, ctx))
+	ictx.SetData(map[string]any{
```
y1yang0: I suggest using SetKeyData instead

y1yang0 (commented):
Not an expert on this, but I have few comments

pkg/instrumentation/k8s-client-go/hook.go
amazingakai: Done

pkg/instrumentation/k8s-client-go/semconv/k8s_test.go
amazingakai: I dont think generating a UUID here would change the test. I can use a valid hardcoded one if you'd prefer.

amazingakai:
@y1yang0 I'd appreciate it if you could run a Copilot review on this PR. I don't have permission to request reviews myself.

amazingakai:
This is still missing a demo, I will add that in a separate PR to keep things reviewable.

NameHaibinZhang (approved):

NameHaibinZhang:
LGTM!

kakkoyun:
@amazingakai Let's rebase this and resolve conflicts. This is closer to done.

amazingakai:
@kakkoyun Done.

`TestLatestLibBuild` is currently failing because the latest `k8s.io/client-go` now requires Go 1.26.0. Should we skip the k8s client case in `TestLatestLibBuild` for now?

kakkoyun:
> @kakkoyun Done.
> 
> `TestLatestLibBuild` is currently failing because the latest `k8s.io/client-go` now requires Go 1.26.0. Should we skip the k8s client case in `TestLatestLibBuild` for now?

@amazingakai Yeah, unfortunately we will face similar issues in the future as well. We have the promise of supporting the latest two major releases. 

But I don't want us to skip the tests. So, here is the idea, let's split `LatestLibBuild` by major go version, and somehow maintain a list of integrations to test on specific versions. `stable` and `oldstable` already an establish pattern for `setup-go` and we already have separation for arch and platform. What do you think? Could you handle this in another PR? so that we can merge that one first.

txabman42:
@kakkoyun are you proposing sth similar to what discussed in this issue? https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/406

### Idea 1
I've in mind to create a new issue for testing the current supported versions. My idea is sth like:
- [min_v, max_v] --> test both min_v and max_v
- min_v --> test min_v

For latest as upper limit --> just `LatestLibBuild` + `LatestLibRun` to catch them.

This wouldnt target the test for supporting latest 2 major releases, but boundaries of the supported versions per pkg

### Idea 2

Specify latest 2 major releases per pkg and test against them. Use renovate for upgrading those pkg versions

kakkoyun:
> @kakkoyun are you proposing sth similar to what discussed in this issue? #406
> 
> ### Idea 1
> I've in mind to create a new issue for testing the current supported versions. My idea is sth like:
> 
> * [min_v, max_v] --> test both min_v and max_v
> * min_v --> test min_v
> 
> For latest as upper limit --> just `LatestLibBuild` + `LatestLibRun` to catch them.
> 
> This wouldnt target the test for supporting latest 2 major releases, but boundaries of the supported versions per pkg
> 
> ### Idea 2
> Specify latest 2 major releases per pkg and test against them. Use renovate for upgrading those pkg versions

@txabman42 I liked both of the ideas. Ideally, our tool should be more resilient to minimum supported versions of the integrations. We can actually solve this by moving out the integrations from our workspace which we plan. 

I also like the ideas discussed in the issue. We can have a daily CI action to check whether the latest upstream changes are broken and maybe open an issue for those.

This definitely needs more thought. Let's discuss a plan in the next meeting. What do you think?

@amazingakai In the meanwhile to make this green, let's add a target version boundary for the supported client-go, and open an issue to track support for the most recent versions.

amazingakai:
@kakkoyun The CI would still fail even after adding a version boundary because `TestLatestLibBuild` tries to bump dependencies regardless of whether the latest version falls within our supported range, even though `otelc` doesn't actually instrument those dependencies. Should I fix this by skipping the bump when the latest version exceeds the max supported version in rules? I can do that in a separate PR :)

kakkoyun:
> Should I fix this by skipping the bump when the latest version exceeds the max supported version in rules? I can do that in a separate PR :)

@amazingakai I think this is the way to go. I'd appreciate the PR.

amazingakai:
Added some changes:
* Updated version bound in k8s instrumentation rules. The rules now support client-go versions >= v0.34.x and < v0.36.x.
* Created #556 for tracking client-go v0.36.x support.
* Updated the k8s integration test to run in parallel with the rest of the integration tests.
* Updated to use `where`/`do` schema.

kakkoyun (approved):

PR #431, opened 2026-04-20
fix(instrumentation): prevent reentrant OTel SDK initialization

amazingakai:
When a project depends on golang.google.org/grpc and OTEL_EXPORTER_OTLP_PROTOCOL is set to grpc, initializing the exporters (autoexport.NewSpanExporter/NewMetricReader) inside initInstrumentation/SetupOTelSDK calls grpc.NewClient.

This invokes BeforeNewClient gRPC instrumentation hook which calls initInstrumentation/SetupOTelSDK again. Since initInstrumentation/SetupOTelSDK use sync.Once, this recursive call causes a deadlock.

### DIFF
--- a/pkg/instrumentation/grpc/client/client_hook.go
+++ b/pkg/instrumentation/grpc/client/client_hook.go
@@ -40,11 +40,12 @@ type int64Hist interface {
 }
 
 var (
-	logger     = shared.Logger()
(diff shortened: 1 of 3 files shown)


kakkoyun (approved):
LGTM. Thanks for taking care of it.

PR #442, opened 2026-04-27
chore: bump dependency version at once

y1yang0:
bump dependency version at once


.tools/go.mod
```diff
@@ -6,7 +6,7 @@ module github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tools
 go 1.25.0
 
 require (
-	github.com/campoy/embedmd/v2 v2.0.0
+	github.com/campoy/embedmd v1.0.0
```
kakkoyun, line 9: This seems like a downgrade?

kakkoyun (commented):
I have seen only one issue. Otherwise, it looks good to me.

.tools/go.mod
y1yang0: because .tools/tools.go uses v1

kakkoyun (approved):

Issue #202, opened 2025-11-24
Add join-points to support opentelemetry-go-contrib/instrumentation/net/http/otelhttp


kakkoyun:
Add join-points to support https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/net/http/otelhttp

## Required Join-points

- [ ] TBD

ref: #164

amazingakai:
Hello, I am interested in picking this up and wanted to clarify a couple of points before proceeding.

1. We currently do not have a way to wrap `http.Client`/`http.Server` structs to modify fields such as `Transport` & `Handler` with the values `otelhttp` provides. This would require a `struct_literal` join point, similar to what orchestrion has. I had something like this in mind:
```yaml
wrap_struct_literal:
    target: "main"
    struct_literal: "net/http.Server"
    match: value-only # value-only|pointer-only|any
    template: |
        func(s http.Server) {
            s.Handler = /* wrap within otelhttp handler */
            return s
        }({{ . }})
```

2. ~~Additionally, we will require a way to apply the advices in user's code. At the moment, we use `target` field to only to modify library side code. We can use pattern-matching with `target` field so we can target any packages (and also exclude internal instrumentation packages to avoid span bombing).~~ 
EDIT: I had originally misunderstood `package-name` and `package-filter` join points in orchestrion. After reading through discussions in #382 and #377, I think the negation combinator along with `target` should be sufficient for this use case.

If this direction seems good, I can start by working on struct-literal join point. Also, if there have been any prior discussions on this, I would appreciate being pointed to them.

y1yang0:
> Hello, I am interested in picking this up and wanted to clarify a couple of points before proceeding.
> 
> 1. We currently do not have a way to wrap `http.Client`/`http.Server` structs to modify fields such as `Transport` & `Handler` with the values `otelhttp` provides. This would require a `struct_literal` join point, similar to what orchestrion has. I had something like this in mind:

I guess the InstRawRule with POS already covers this functionality?

amazingakai:
> > Hello, I am interested in picking this up and wanted to clarify a couple of points before proceeding.
> > 
> > 1. We currently do not have a way to wrap `http.Client`/`http.Server` structs to modify fields such as `Transport` & `Handler` with the values `otelhttp` provides. This would require a `struct_literal` join point, similar to what orchestrion has. I had something like this in mind:
> 
[...]

PR #450, opened 2026-04-30
feat(tool): add support for customizable injection in InstRawRule with pattern matching

### LINKED ISSUE #324: Add pos field to InstRawRule forcustomizable injection position
**1. Background**

Currently, an `InstRawRule` is defined with the following structure:

```yaml
raw_helloworld:
  target: main
  func: Example
  raw: |
    go func(){
      h := sha256.New()
      h.Write([]byte("RawCode"))
      fmt.Printf("RawCode: %x\n", h.Sum(nil))
    }()
  imports:
    fmt: "fmt"
    sha256: "crypto/sha256"
```

This configuration injects the code snippet from the `raw` field at the very beginning (the entry point) of the target function, such as `main.Example()`. The injection position is not customizable.

**2. Proposal**

I propose introducing a new optional field named `pos`. This field will allow users to specify a precise injection point within the function body using a regular expression.

When the `pos` field is present, the instrumenter will search for the first line in the function that matches the regex and inject the code snippet **before** that line. If `pos` is omitted, the behavior defaults to injecting at the function's entry point, preserving backward compatibility.

**3. Example Usage**

Consider the following target function:

```go
func Example(){
  a()
  b()
  c()
}
```

To inject code before the call to `b()`, the configuration would be:

```yaml
my_rule:
  target: main
  func: Example
  pos: "b\\(\\)" # Inject before the line matching b()
  raw: |
    // ... your code snippet ...
```

Similarly, `pos: "a\\(\\)"` would inject the code before `a()`.

amazingakai:
## Description

- `pattern` (string, optional): The position within the function where the raw code should be injected. By default, the code is injected at the start of the function body. If provided, the value must be a valid regular expression.

  The pattern is matched against the canonical gofmt representation for each statement in the function body (not always the exact original source formatting). The injected code is placed immediately before/after the first statement that matches the pattern.

  If no statement matches the pattern, an error is returned.

- `placement` (string, optional): Determines where to inject the raw code when a `pattern` is specified. Can be either `before` (default) or `after`.

## Motivation

It helps customize injection point for raw code inside functions.

Fixes #324 

---

## Checklist

- [x] Tests pass: `make test`


amazingakai:
@y1yang I was thinking about extending this a bit and wanted your input.

Instead of `pos` being a plain string, what do you think about making it a structured field like:
```yaml
pos:
  where: before | after | start | end
  pattern: "..."
```
- If `pattern` is omitted, `where: start | end` controls insertion at the beginning or end of the function body.
- If `pattern` is provided, `where: before | after` controls insertion relative to the first matched statement.

This would make the behavior more explicit and allow supporting "after" insertion as well. WDYT?

y1yang0:
@amazingakai 
- start/end: This seems like an uncommon use case. The proper way to insert at a function's start/end is usually with InstFuncRule.
- before/after: Although this is a valid scenario, the field feels redundant. We can achieve the same result by simply modifying the matching code to target the next line.
- Naming: I strongly agree that pattern is a much better name than pos.

amazingakai:
@y1yang0 

> start/end: This seems like an uncommon use case. The proper way to insert at a function's start/end is usually with InstFuncRule.

True. I did not think of that.

> before/after: Although this is a valid scenario, the field feels redundant. We can achieve the same result by simply modifying the matching code to target the next line.

There is one usecase for this. For example we are at the end of a block:
```go
if true {
  print("Hello")
  // we want to inject code here
}
print("World")
```

If we target `print("World")`, the code would be outside the if statement.
There is no way to inject code at the end of "if" block without an explicit placement concept.

> Naming: I strongly agree that pattern is a much better name than pos.

Also it may make sense to call the `where` param `placement` to avoid confusing it with the `where`/`do` schema.
So the rule would look like

```yaml
pos:
   placement: before|after # optional, default=before
   pattern: ...
```

y1yang0:
My point is, we can always default `placement: after` (and omit this field), in your example, we can use
```
rule:
  ...
  pattern: "print(\"Hello\")"
  raw: insertCode()
```
to inject code in the position you mentioned.
```
if true {
  print("Hello")
  insertCode()
}
print("World")
```
But your proposal makes me realize that neither `always_placement_after + pattern` nor `placement + pattern` can solve the problem below:
```
for range list {
	for range list {
		for range list {
			...
		}
		.... // insert there
	}
	.... // or there
}
```

----

I'm thinking we could keep InstRawRule as it is (convenient, but limited), while introducing a generic fallback InstCommonRule based on AST matching?

For example:
```
universal_rule:
  target: abc
  version: test
  match: MyInstrument()
  path: github.com/where/MyInstrument/located

//github.com/where/MyInstrument/located
package newpkg

func MyInstrument(node dst.Node) {
	if node.Type == ABC .... {
		// Matched node, modifiy/replace/it
		....
	}
}
```

I can see a few benefits to this approach:

- We always have a fallback matching mechanism. For anything existing rules can't handle, InstCommonRule can cover it.
- For vague or evolving requirements (e.g., [comment](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/202#issuecomment-4365559886)), we can initially implement them using InstCommonRule. If we later find that the number of InstCommonRule instances exceeds a certain threshold (e.g., >=7), we can then formalize them into a new dedicated rule. This saves a lot of back-and-forth discussion on whether to introduce a new rule type upfront.

amazingakai:
I may have misunderstood the nested `for range` example, but if the loops are structurally distinct, I think `placement + pattern` would still handle it correctly. For example:

```yaml
pos:
  pattern: "^for j := range l2"
  placement: after
```

would inject after the second loop block.

If the concern is multiple identical matches (e.g. the same `for _ = range list` pattern appearing several times), then I think that becomes more of a match disambiguation problem rather than a placement problem.

Also, if the preference is to avoid nesting in the rule schema, we could move `pattern` and `placement` to the top level of the raw rule and omit `pos` entirely. But I think defaulting `placement` to `before` would be more intuitive.

---

Regarding `InstCommonRule`, My concern is that if instrumentation authors start relying on AST-level matching directly, then, as you also [suggested earlier](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/324#issuecomment-4323703791), it may increase the complexity + maintenance burden of instrumentations over time.

y1yang0:
> If the concern is multiple identical matches (e.g. the same for _ = range list pattern appearing several times), then I think that becomes more of a match disambiguation problem rather than a placement problem.

Yes, this is my concern

> Regarding InstCommonRule, My concern is that if instrumentation authors start relying on AST-level matching directly, then, as you also https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/324#issuecomment-4323703791, it may increase the complexity + maintenance burden of instrumentations over time.

Regarding InstCommonRule (or InstUniversalRule?), it serves as a fallback solution. This means it will be used infrequently, so the complexity of AST matching is acceptable. This is different from imposing AST matching complexity on InstRawRule, which has a relatively high frequency of use (for quick debugging purposes, etc.)

@kakkoyun @darccio @txabman42  your input is also welcome. Thanks.

amazingakai:
> Yes, this is my concern

Having code like that is not very common, but perhaps we could also support matching the nth occurrence of a pattern (e.g. via something like `occurrence`) rather than always selecting the first match.

> Regarding InstCommonRule (or InstUniversalRule?), it serves as a fallback solution. This means it will be used infrequently, so the complexity of AST matching is acceptable. This is different from imposing AST matching complexity on InstRawRule, which has a relatively high frequency of use (for quick debugging purposes, etc.)

That sounds useful too, although I do not have much experience working with compile-time instrumentation, so I would appreciate input from other maintainers here as well.

Perhaps that could also be introduced later as a follow-up PR, while still having a simpler raw rule with `pattern`, `placement` (and maybe `occurrence`) fields, which i think is already quite powerful.

txabman42:
Lots of interesting ideas in this thread, throwing my vision...

From README.md:
> This project provides a tool to automatically instrument Go applications with [OpenTelemetry](https://opentelemetry.io/) at compile-time.

From docs/rules:
> Raw code injection rule ... This offers great flexibility but should be used with caution

This repo focuses on OTEL instrumentation, rather than just providing full code injection capabilities at compile time. This should help us with decision-making.

Some questions:
- Why do we need `pos` at raw rule? Which need is it solving?
- In case we end up needing it, could the simplest approach (a single regex) be enough for now? As documented, raw should be used in really special cases.

I think we should focus more on making clear when/how to use each rule type before adding more flexibility. Too much flexibility without guardrails could end up hiding underlying problems instead of surfacing them.

Regarding `InstCommonRule`: sounds interesting, IMO worth its own issue. My main concern is that this kind of flexibility translates into higher implementation speed, no blockers. However, is that extra implementation speed worth it?

amazingakai:
> Why do we need `pos` at raw rule? Which need is it solving?

`InstRawRule` currently injects at the start of the function body, which is already quite useful for certain instrumentation/debugging scenarios.

The `pos/pattern` matching part, from my understanding, is intended as an additional feature meant for lightweight debugging. Though, Yi Yang might have created the original issue with some other use-case in mind.

One example scenario would be a function that has a side-effect instead of returning values. In this case we cannot use `unnamedRetValueX` to modify that state. Being able to inject raw code at a certain position inside a function would be useful in this case:
```go
func doSmth() {
    // some code here

    // modifies some global state here
    x.Foo = 123 // injection point

    // some code here
}
```

With a simple pattern match, we could inject before/after `x.Foo = 123` and modify additional state around that operation.


> In case we end up needing it, could the simplest approach (a single regex) be enough for now? As documented, raw should be used in really special cases.

I'd say yes, the simplest form with just `pattern` would already be powerful enough. Since this uses `format.Node` for matching, I'd also say this isn't the most reliable option and should stay limited to debugging/special-cases and should be avoided in general instrumentation use-cases.

> Regarding InstCommonRule: sounds interesting, IMO worth its own issue. My main concern is that this kind of flexibility translates into higher implementation speed, no blockers. However, is that extra implementation speed worth it?

I agree this should be discussed in its own issue.

kakkoyun:
Thanks for looping me in. Quick context: in orchestrion (https://github.com/DataDog/orchestrion) (the other compile-time Go instrumentation tool we work on) we have essentially this feature as prepend-statements / append-statements against a matcher. It's been in real use for a while and has held up — so I'm coming at this from the "we already know roughly what shape works" angle rather than a greenfield design.

1. +1 to `pattern/placement` over `pos`. Strong agreement with @y1yang0 on the rename.

2. I'd keep both before and after placements. "Just target the next line" works for most cases, but it breaks down at block boundaries — exactly the `if true { print("hello") /* HERE */ }` example from earlier in the thread, where there is no "next line" inside the block to target. That's the same reason orchestrion ended up with both prepend and append. Default before matches the current behavior, so existing rules don't change.

3. Disambiguation (nested identical statements) is a non-goal for `InstRawRule`. Agree with the concern, but the moment we'd need occurrence: N or path-qualified matching we've outgrown regex-on-formatted-source and should reach for a different rule type, not pile more knobs on this one. Punt.

4. On `InstCommonRule` / `InstUniversalRule`: +1 to splitting it into its own issue (agreeing with @txabman42). Worth designing carefully — a "give users `dst.Node`" escape hatch tends to become the rule everyone reaches for, which is exactly the maintenance-burden trap flagged in #324. Let's not couple that decision to this PR.

One small ask regardless of which direction this lands: the current implementation matches against `format.Node` (gofmt'd) output of each statement, not the user's source text. Could we add a debug subcommand or `-v` flag that prints the formatted statements of a target function? That makes patterns writable without guessing, and would prove useful for the rare cases we even need raw.

amazingakai:
> One small ask regardless of which direction this lands: the current implementation matches against format.Node (gofmt'd) output of each statement, not the user's source text. Could we add a debug subcommand or -v flag that prints the formatted statements of a target function? That makes patterns writable without guessing, and would prove useful for the rare cases we even need raw.

Added `logger.Debug` for this, so the formatted statements are visible now.

Also switched to `pattern` + `placement`.
waiting for another look :)

kakkoyun (approved):

tool/internal/instrument/apply_raw.go
```diff
@@ -44,6 +112,28 @@ func insertRaw(r *rule.InstRawRule, decl *dst.FuncDecl) error {
...
 	}
+
+	// if specified, insert raw code at the position matched by the regex
+	if r.Pattern != "" {
+		restorer := decorator.NewRestorer()
+		if _, restoreErr := restorer.RestoreFile(root); restoreErr != nil {
+			return ex.Wrapf(restoreErr, "failed to restore the AST")
+		}
+
+		pattern := regexp.MustCompile(r.Pattern)
```
txabman42, line 123: Lets store this as variable so we dont need to recompile always, similar approach to https://github.com/txabman42/opentelemetry-go-compile-instrumentation/blob/b6c6c1eb44524a3f1713addf83e5fe92032064ba/tool/internal/rule/call_rule.go#L87

txabman42 (approved):
LGTM! Just a minor nitpick

tool/internal/instrument/apply_raw.go
amazingakai: The regex is only compiled once per rule application. Since a raw rule targets a single function declaration and `Pattern` can differ across rules, I wouldn't expect repeated compilation of the same pattern.

amazingakai:
Updated to support `where`/`do` schema.

PR #456, opened 2026-05-02
feat(instrumentation): use otelgrpc for gRPC instrumentation

### LINKED ISSUE #203: Add join-points to support opentelemetry-go-contrib/instrumentation/grpc/otelgrpc
Add join-points to support https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc

## Required Join-points

- [ ] TBD

ref: #164

yuluo-yx:
Fixes #203.

This adds compile-time join-point support for upstream otelgrpc by injecting
otelgrpc stats handlers into grpc.NewServer, grpc.NewClient, and grpc.DialContext.


pkg/instrumentation/grpc/client/client_hook.go
```diff
@@ -202,130 +128,23 @@ func AfterDialContext(ictx inst.HookContext, conn *grpc.ClientConn, err error) {
...
-}
 
-// TagConn is called when a new connection is established
-func (h *clientStatsHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo) context.Context {
-	return ctx
-}
+	endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+	if endpoint == "" {
+		endpoint = os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")
+	}
```
amazingakai: Should we also skip for `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` and `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` here? (Missed in #431)

amazingakai:
With this change, can we remove the code in `grpc/semconv` package? most of it seems to be unused now, except for `IsOTELExporterPath`. Maybe that could be moved out?

pkg/instrumentation/grpc/client/client_hook.go
yuluo-yx: done

yuluo-yx:
> With this change, can we remove the code in `grpc/semconv` package? most of it seems to be unused now, except for `IsOTELExporterPath`. Maybe that could be moved out?

done

kakkoyun:
@yuluo-yx Thanks for this. I will have a look soon.

NameHaibinZhang:
I think relying on contrib to support gRPC is not a good choice, as it will lead to compatibility issues such as gRPC version and SDK version mismatches. We not only need to support higher versions of gRPC, but also need to support lower versions in the future.

yuluo-yx:
so? @NameHaibinZhang 

What should I do? Should I withdraw the changes, or is there a better solution? Your review comments are leaving me unsure of what to do next.

pkg/instrumentation/grpc/client/client_hook.go
amazingakai: @yuluo-yx  Could we split this fix into a separate PR? IMO this shouldn't be blocked by the decision around switching to `otelgrpc`.

amazingakai: #621

darccio:
Closing it manually, just to clean up backlog a bit faster.

Issue #434, opened 2026-04-22
compile-time-benchmarks consistently fail


y1yang0:
For example, see this CI run: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/24773461213/job/72486815859

Log chunk:
```
/usr/bin/git -c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= pull ***github.com/open-telemetry/opentelemetry-go-compile-instrumentation.git gh-pages
From https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation
 * branch            gh-pages   -> FETCH_HEAD
Already up to date.
Could not find data.js at dev/bench/data.js. Using empty default: Error: ENOENT: no such file or directory, open 'dev/bench/data.js'
/usr/bin/git -c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= add dev/bench/data.js
/usr/bin/git -c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= add dev/bench/index.html
Created default index.html at dev/bench/index.html
/usr/bin/git -c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= commit -m add otelc Compile-Time Benchmarks (customSmallerIsBetter) benchmark result for 8bbedd79023381b71ede2a27d7bf7cf80a0e7ad7
[gh-pages 30c0f0a] add otelc Compile-Time Benchmarks (customSmallerIsBetter) benchmark result for 8bbedd79023381b71ede2a27d7bf7cf80a0e7ad7
 2 files changed, 1147 insertions(+)
 create mode 100644 dev/bench/data.js
 create mode 100644 dev/bench/index.html
/usr/bin/git -c user.name=github-action-benchmark -c user.email=github@users.noreply.github.com -c http.https://github.com/.extraheader= push ***github.com/open-telemetry/opentelemetry-go-compile-instrumentation.git gh-pages:gh-pages --no-verify
remote: error: GH013: Repository rule violations found for refs/heads/gh-pages.        
remote: Review all repository rules at https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/rules?ref=refs%2Fheads%2Fgh-pages        
remote: 
remote: - Required status check "EasyCLA" is expected.        
remote: 
[...]

PR #457, opened 2026-05-03
feat(test): add codspeed and refactor benchmark tests

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/434

Refactor how we approach benchmarking. I've requested codspeed access to OTEL so we can follow same approach as `opentelemetry-go` and `opentelemetry-collector`.

Bench tests have been refactored as codespeed requires go-bench tests.

Current proposed approach:
- PR --> Check if otelc compile time is less than 150% vs golang base
- master after merge --> Runs bench tests and stores results in codespeed

> NOTE: I've push result from this branch just to 👀 how things looks like

Wins:
- We keep the 150% check vs base (we can add extra checks in the future, e.g. % increase vs last run once we got the bare-metal runners)
- [profiling](https://codspeed.io/open-telemetry/opentelemetry-go-compile-instrumentation/branches/txabman42%3Axabier%2Fadd-codspeed?utm_source=github&utm_medium=comment-v2&utm_content=button&uri=test%2Fbench%2Fbench_test.go%3A%3ABenchmarkCompile%3A%3Amulti%2Fotelc&runnerMode=WallTime&page=1)
- Store the historical times in `main` per scenario (it should be enabled once merged in main), from codespeed:
> No baseline data is yet available
>Once your main branch has generated a first performance report, you will see the list of commits and their performance impact here.

Following steps:
- Use [OTEL bare-metal runners](https://github.com/open-telemetry/community/blob/main/assets.md#bare-metal-runners)


Makefile
```diff
@@ -349,36 +349,23 @@ check-golden-files: package
 
 ##@ Benchmarking
 
-BENCH_HARNESS_DIR := test/bench/cmd/bench
```
kakkoyun, line 352: Shall we keep this for local runs?

kakkoyun (commented):
Thanks for handling this.

I only have one concern. Shall we keep the existing facilities to run these on our local machines while experimenting?

Makefile
txabman42: It's not needed. Now, benchmark tests are normal go-bench tests which could be run via `go test -bench=...` or `benchmark/codspeed`

kakkoyun (approved):
Thanks for the clarification. LGTM!

.github/workflows/benchmark.yaml
```diff
@@ -5,13 +5,17 @@ on:
...
 
 jobs:
-  benchmark:
-    name: Compile-Time Benchmarks
-    runs-on: ubuntu-latest
+  # CodSpeed posts PR comments by comparing the PR head to the most recent main
+  # baseline.
+  codspeed:
+    name: CodSpeed Benchmarks
+    runs-on: ubuntu-latest  # TODO: oracle-bare-metal-64cpu-1024gb-x86-64-ubuntu-24
```
y1yang0, line 17: Just curious, is this TODO intended to run benchmark on the high-performance machine?

y1yang0 (approved):

.github/workflows/benchmark.yaml
txabman42: Yes, the idea is to use otel bare metal runner for running this. I think there are other repos using it. I've created this [issue](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/467) for checking it, so we can close other bench related issues

PR #459, opened 2026-05-04
feat(rule): add wrap_struct_literal join-point for struct literal instrumentation

gyanranjanpanda:
## Summary

Implements [#202](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/202) — adds a new `wrap_struct_literal` join-point that enables template-based wrapping of struct instantiations at compile time.

## Motivation

The current instrumentation framework supports wrapping function calls (`wrap_call`) and injecting hooks into function bodies (`func`), but has no mechanism to intercept **struct literal creation**. This is needed for instrumenting patterns like `&http.Server{}` or `&http.Transport{}` where the instrumentation point is the instantiation itself, not a function call.

### Rule Schema

```yaml
wrap_struct_literal:
  target: "main"
  struct_literal: "net/http.Server"
  match: "pointer-only"  # value-only | pointer-only | any
  template: |
    func(s *http.Server) *http.Server {
        s.Handler = otelhttp.NewHandler(s.Handler)
        return s
    }({{ . }})
```

### Match Modes

| Mode | Matches `&http.Server{}` | Matches `http.Server{}` |
|------|:---:|:---:|
| `pointer-only` | ✅ | ❌ |
| `value-only` | ❌ | ✅ |
| `any` (default) | ✅ | ✅ |

### Architecture

The implementation follows the established rule engine pattern:

1. **Rule definition** (`InstStructLiteralRule`) — mirrors `InstCallRule` with struct-specific fields
2. **Rule matching** — collected during the setup phase using glob-based target patterns
3. **AST transformation** — walks `dst.CompositeLit` nodes, resolves import aliases, and applies the template wrap

### New files
| File | Purpose |
|------|---------|
| `tool/internal/rule/struct_literal_rule.go` | Rule type definition with validation |
| `tool/internal/rule/struct_literal_rule_test.go` | Rule creation and validation tests |
| `tool/internal/instrument/apply_struct_literal.go` | AST transformation logic |
| `tool/internal/instrument/apply_struct_literal_test.go` | 5 test cases: pointer, value, mismatch, any, aliased imports |

### Modified files
| File | Change |
|------|--------|
| `tool/internal/rule/base.go` | Added `StructLiteralRules` to `InstRuleSet`, `Excludes` to `InstBaseRule` |
| `tool/internal/instrument/instrument.go` | Dispatch `InstStructLiteralRule` to `applyStructLiteralRule` |
| `tool/internal/setup/match.go` | Parse `wrap_struct_literal` rule type, glob-based target matching |
| `tool/internal/setup/match_test.go` | Updated tests for refactored `runMatch` signature |

## Testing

- All existing unit tests pass
- All integration tests pass
- All e2e tests pass
- `make all` completes successfully (build → format → lint → test)

## Checklist

- [x] `make all` passes
- [x] No new dependencies introduced
- [x] Apache 2.0 license headers present on all new files
- [x] Follows existing code patterns (`InstCallRule`, `applyCallRule`)


tool/internal/setup/match.go
```diff
@@ -142,8 +145,37 @@ func (sp *SetupPhase) runMatch(
...
+			if exclMatch {
+				excluded = true
+				break
+			}
+		}
+
+		if !excluded {
+			relevantRules = append(relevantRules, r)
+		}
+	}
```
amazingakai: I think this should be a separate PR. Also this is already being implemented in form of not combinator (#386) + import path glob filter (#382).

tool/internal/rule/struct_literal_rule.go
```diff
@@ -0,0 +1,64 @@
...
+//	  match: "value-only" # value-only|pointer-only|any
+//	  template: |
+//	    func(s http.Server) http.Server {
+//	        return s
+//	    }({{ . }})
+type InstStructLiteralRule struct {
+	InstBaseRule `yaml:",inline"`
+
+	StructLiteral string `json:"struct_literal" yaml:"struct_literal"` // The type name of the struct literal to be matched
+	Match         string `json:"match"          yaml:"match"`          // "value-only", "pointer-only", or "any"
```
amazingakai, line 28: Orchestrion also supported field level advising, mutually exclusive with match. Might be worth adding but it should be discussed first.
See: https://datadoghq.dev/orchestrion/contributing/aspects/join-points/struct-literal/

amazingakai (changes requested):
Thank you for looking into this. For context, [my comment](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/202#issuecomment-4365559886) was meant to start a discussion rather than serve as a ready proposal. I think it might be too soon to implement this without some maintainer input on the design first.

amazingakai:
This PR is also missing golden tests and documentation.

gyanranjanpanda:
could @amazingakai  check this if there any other isuue pls let me know

y1yang0:
Too early to review now, we should first discuss the necessity and design.

PR #466, opened 2026-05-06
chore(test): give each golden test its own source and hook files

vyagh:
## Description
  This PR updates the instrumentation golden test fixtures so each golden test case owns its input files.

  changes done:
  - moved `source.go` into the individual golden test case directories
  - added case-local `hook.go` files for hook-based golden tests
  - updated hook rule paths to point at the case-local fixture directories
  - regenerated affected golden files where the `//go:linkname` hook paths changed
  
  Fixes #411

  ---

  ## Checklist

  - [ ] Tests pass: `make test`


y1yang0 (approved):
LGTM. 

Currently, all tests share the same source.go, but they should be isolated. For instance, before-only/source.go should only contain Func1, while ellipsis-syntax/source.go should only contain EllipsisFunc. We can handle this in a follow-up PR.

vyagh:
happy to do a follow up PR for this :)

Issue #161, opened 2025-11-12
isSetup - Check if setup phase is already done(because we can either `otel setup + otel go build` or just `otel go build`)


kakkoyun:

amazingakai:
#468 is a start towards this. Currently, running `otelc setup` on its own is not very useful ~~and can even break the project state.~~

~~At the moment, `otelc setup` updates `go.mod` and generates `otelc.runtime.go` files. Later, when `otelc go build` runs, it unconditionally triggers setup again, which will:~~
 * ~~add duplicate dependencies to `go.mod`~~
 * ~~attempt to instrument the generated hook/runtime code itself~~

EDIT: On reading the code around `addDeps`/`syncDeps`, I see nothing that would cause this and also upon trying this again I'm unable to reproduce this problem. Something was probably wrong in my original setup when testing the `otelc setup` command. The issue still stands but not as serious as I originally had thought: `otelc setup` modifies user project state which we probably shouldn't be doing. Also, running `otelc setup` again backups the modified `go.mod` and the original `go.mod` file is lost.

This raises a broader question of whether `otelc setup` should remain a separate command at all, or whether it should become an internal step of `otelc go build`.

If we do want to keep a separate setup command, I had the following approach in mind:

1. During setup, update `go.mod` + generate `otelc.runtime.go`, run `go mod tidy`, and store the new `go.mod` + runtime files inside `.otelc-build/setup`.
2. If `Setup()` was invoked from the standalone `otelc setup` command, immediately restore the original `go.mod` and remove the generated runtime files afterwards. The cached copies inside `.otelc-build/setup` would still remain.
3. During `otelc go build`, we would check whether the setup state already exists (`matched.json`, `setup/go.mod`, runtime files). If present, restore them before instrumentation/build begins. Otherwise, invoke setup internally but the generated files should be kept.
4. Subsequent builds can then reuse the cached setup state. If dependencies or source code change, the user would need to either rerun `otelc setup` or clean the existing setup state before building again.

I would love to hear your thoughts on this approach. If it sounds reasonable, I could try working on it.

txabman42:
I like the direction of the proposal.

Just a small comment regarding point 3:
>  we would check whether the setup state already exists

We don't just need to check if it exist, but compare the hash to know if something has change. That would allow us to know if we need to reuse it or not. For hash, we can take into account rules, go.mod, envs, etc.

> otelc setup modifies user project state which we probably shouldn't be doing

[...]

PR #468, opened 2026-05-06
fix(tool): otelc setup command failing to run

amazingakai:
## Description

Fixes `otelc setup` command failing to run.
Also allows passing `build`, `install` or `go` as package names to `getBuildPackages` function.

## Motivation

Currently running otelc setup command fails because `setupPhase.listBuildPlan()` expects arguments as `go build ...`:
```
$ otelc setup   
Error:
[0] at least 2 arguments are required

Stack:
[0] github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/setup/find.go:78 tool/internal/setup.(*SetupPhase).listBuildPlan
[1] github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/setup/find.go:208 tool/internal/setup.(*SetupPhase).findDeps
[2] github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/setup/setup.go:175 tool/internal/setup.Setup
[3] github.com/urfave/cli/v3@v3.6.1/command_run.go:354 github.com/urfave/cli/v3.(*Command).run
[4] github.com/urfave/cli/v3@v3.6.1/command_run.go:288 github.com/urfave/cli/v3.(*Command).run
[5] github.com/urfave/cli/v3@v3.6.1/command_run.go:94 github.com/urfave/cli/v3.(*Command).Run
[6] github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/cmd/main.go:100 main.main
[7] runtime/proc.go:285 runtime.main
```

Also, passing "build", "install" or "go" as package names to `getBuildPackages` skips them as the loop breaks and it fallbacks to `.` as the default.

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/find.go
+++ b/tool/internal/setup/find.go
@@ -69,17 +69,10 @@ func findCommands(buildPlanLog *os.File) ([]string, error) {
 	return commands, nil
 }
 
-// listBuildPlan lists the build plan by running `go build/install -a -x -n`
+// listBuildPlan lists the build plan by running `go build -a -x -n`
 // and then filtering the commands (cd, cgo, compile) from the build plan log.
-func (sp *SetupPhase) listBuildPlan(ctx context.Context, goBuildCmd []string) ([]string, error) {
-	const goBuildMinArgs = 2 // go build
+func (sp *SetupPhase) listBuildPlan(ctx context.Context, cmdArgs []string) ([]string, error) {
 	const buildPlanLogName = "build-plan.log"
-	if len(goBuildCmd) < goBuildMinArgs {
-		return nil, ex.Newf("at least %d arguments are required", goBuildMinArgs)
-	}
-	if goBuildCmd[1] != "build" && goBuildCmd[1] != "install" {
-		return nil, ex.Newf("must be go build/install, got %s", goBuildCmd[1])
-	}
 
 	// Create a build plan log file in the temporary directory
 	buildPlanLog, err := os.Create(util.GetBuildTemp(buildPlanLogName))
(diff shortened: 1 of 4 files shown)


tool/internal/setup/find.go
```diff
@@ -88,16 +81,13 @@ func (sp *SetupPhase) listBuildPlan(ctx context.Context, goBuildCmd []string) ([
...
-	args = append(args, goBuildCmd[:goBuildMinArgs]...) // go build/install
-	args = append(args, []string{"-a", "-x", "-n"}...)  // -a -x -n
-	if len(goBuildCmd) > goBuildMinArgs {               // {...} remaining
-		args = append(args, goBuildCmd[goBuildMinArgs:]...)
-	}
-	sp.Info("New build command", "new", args, "old", goBuildCmd)
+	buildArgsLen := 4
+	args := make([]string, 0, buildArgsLen+len(cmdArgs))
+	args = append(args, "build", "-a", "-x", "-n") // build -a -x -n
+	args = append(args, cmdArgs...)                // args from original build/install or setup command
```
txabman42:
```suggestion
	prefix := []string{"build", "-a", "-x", "-n"}
	args := make([]string, 0, len(prefix)+len(cmdArgs))
	args = append(args, prefix...)
	args = append(args, cmdArgs...)                // args from original build/install or setup command
```

tool/internal/setup/setup.go
```diff
@@ -140,17 +140,20 @@ func getPackageDir(pkg *packages.Package) string {
...
 		logger.InfoContext(ctx, "Setup has already been completed, skipping setup.")
 		return nil
 	}
 
+	// Since Setup can be invoked in different contexts (i.e, via `otelc setup` or as part of `otelc go build`),
+	// we need to handle the arguments accordingly. If the command is `go build` or `go install`, we should trim the first argument
+	if cmd.Name == "go" {
+		args = args[1:] // trim build/install
+	}
+
```
txabman42: setup.go shouldn't be aware of who is calling it. The fix should be similar to what you added in the defer statement.

```suggestion
	args := cmd.Args().Tail()
```

tool/internal/setup/setup.go
```diff
@@ -386,9 +389,17 @@ func GoBuild(ctx context.Context, cmd *cli.Command) error {
 	// to prevent stale data from affecting this build.
 	instrument.CleanupImportTrackingFiles()
 
+	if !cmd.Args().Present() {
```
txabman42, line 392: All new logic must be unit tested.

txabman42 (changes requested):
Nice PR! TY for the fix ❤️ 

Just some minor nitpicks

tool/internal/setup/setup.go
amazingakai: Wouldn't this be incorrect if someone runs `otelc setup ./mypkg`? In that case it would skip `./mypkg`.

If we keep `setup` as a separate command, my plan was to move the actual `Setup()` implementation into a shared function so the command-specific logic stays in the command layer, while both `otelc setup` and `otelc go build` can reuse the same setup flow.

tool/internal/setup/setup.go
amazingakai: I have added unit tests for `listBuildPlan`. For `Setup()`, I think integration tests would be more valuable than unit tests.

I do plan to work on those, but I was holding off because of #161 so we can first decide how we want to move forward with `setup` before adding integration test coverage around the current behavior.

tool/internal/setup/setup.go
txabman42: Agree, this is ok now:

```go
	args := cmd.Args().Slice()
	if cmd.Name == "go" {
		args = cmd.Args().Tail() // trim build/install
	}
```

tool/internal/setup/setup.go
txabman42: Agree with this plan. I have #161 in my backlog pending for review.

txabman42 (approved):
LGTM! ❤️

kakkoyun (approved):
LGTM

PR #469, opened 2026-05-06
feat(tool): add //otelc:ignore and //otelc:instrument source directives

pradhyum6144:
## Description

Two source-level directives for opting out of (or back into) auto-instrumentation without a rebuild:

- `//otelc:ignore` on a function declaration skips all func rules for that function. The function body is left completely unmodified.
- `//otelc:ignore` before the `package` keyword skips all rules for the entire file.
- `//otelc:instrument` on a specific function within an ignored file overrides the file-level opt-out and forces instrumentation.

The directive check runs inside `applyFuncRule`, before any trampoline code is generated, so ignored functions are left completely unmodified.

This mirrors the `//orchestrion:ignore` mechanism in DataDog's Orchestrion and is motivated by the same production use-cases: hot paths already manually instrumented, third-party files vendored into a package, and performance-sensitive loops where the trampoline overhead is unacceptable.

## Motivation

Without a selective opt-out, every large codebase will have edge cases that block adoption wholesale. Teams currently have zero recourse other than a full rebuild without otelc.

---

## Checklist

- [x] Tests pass: `make test`


amazingakai:
InstFuncRule is used to target library side code, of which user has no control. I think these directives would make sense for something like InstFuncCallRule once we support instrumenting user-side code, see:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/382
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/377

kakkoyun:
> InstFuncRule is used to target library side code, of which user has no control. I think these directives would make sense for something like InstFuncCallRule once we support instrumenting user-side code, see: #382 #377

@amazingakai I will revive them this week.

kakkoyun:
@pradhyum6144 This can be revived. Now that #377 is merged by following the design decisions there.

tool/internal/ast/directive.go
```diff
@@ -135,6 +135,29 @@ func FileHasDirective(file *dst.File, directive string) bool {
...
+// This is distinct from FileHasDirective, which searches all nodes in the file.
+func FileLeadHasDirective(file *dst.File, directive string) bool {
+	for _, dec := range file.Decs.Start {
+		if MatchDirective(dec, directive) {
+			return true
+		}
+	}
+	return false
+}
+
```
darccio, line 160: `FileLeadHasDirective` assumes a pre-`package` `//otelc:ignore` always lands in `file.Decs.Start`. Both golden fixtures put a **blank line** between the directive and `package main` (see `ignore-file-directive/source.go`), which keeps it out of the package doc-comment slot. If a user instead writes the directive directly above `package` with no blank line, `dst` may attach it as the package doc comment instead — and it's untested whether that still shows up in `Decs.Start`. If it doesn't, the file-level opt-out silently no-ops (no build error, no log line the user would see by default). Worth either a unit test pinning both layouts, or an explicit note in user-facing docs that a blank line is required.

darccio, line 160: `FuncHasDirective` and `FileLeadHasDirective` are exported and load-bearing for this feature, but `directive_test.go` wasn't extended — the only coverage is indirect, via the two golden fixtures in `tool/internal/instrument/testdata/golden/`. A golden test proves the end-to-end pipeline works for one layout; it doesn't pin the helper's contract (leading-only vs. whole-file scan, blank-line vs. no-blank-line placement, directive on a method with a receiver, multiple leading comments where only one matches). Recommend adding table-driven unit tests here similar to the existing `TestFileHasDirective`/`TestMatchDirective` tests in this file.

tool/internal/ast/directive.go
```diff
@@ -135,6 +135,29 @@ func FileHasDirective(file *dst.File, directive string) bool {
 	return found
 }
 
+// FuncHasDirective reports whether the function declaration's leading
+// decorations contain the given directive.
+func FuncHasDirective(funcDecl *dst.FuncDecl, directive string) bool {
```
darccio, line 140: Minor naming nit: `FileHasDirective` (above) scans the *whole file* via `dst.Inspect`, `FileLeadHasDirective` (below) scans only the *leading* decorations, and `FuncHasDirective` also scans only leading decorations but has no "Lead" qualifier — so a reader skimming the exported API could reasonably expect `FuncHasDirective` to also check trailing/body decorations for symmetry with `FileHasDirective`. Since a `dst.FuncDecl` doesn't really have a meaningful "whole subtree" directive concept the way a file does (current callers only care about leading `//otelc:ignore` right above the func), the current behavior is correct — this is purely about naming clarity. Suggest either renaming to `FuncLeadHasDirective` for consistency, or adding a doc-comment sentence explicitly noting it deliberately excludes body/trailing comments (mirroring the "This is distinct from..." sentence already on `FileLeadHasDirective`).

darccio (commented):
Review notes on the directive helpers in `tool/internal/ast/directive.go` (Area A of a full walkthrough). Non-blocking observations, flagging for discussion.

tool/internal/instrument/apply_func.go
```diff
@@ -354,6 +357,10 @@ func (ip *InstrumentPhase) applyFuncRule(ctx context.Context, rule *rule.InstFun
 	if !ok {
 		return ex.Newf("can not find function %s", rule.Func)
 	}
+	if ast.FuncHasDirective(funcDecl, directiveIgnore) {
+		ip.Debug("Skip func rule due to //otelc:ignore", "func", rule.Func, "rule", rule.Name)
+		return nil
```
darccio, line 362: This early `return nil` is reached *after* the caller has already decided globals are needed. In `instrument.go`'s `applyOneRule`, the `*rule.InstFuncRule` case does `return true, ip.applyFuncRule(...)` — the `true` (isFuncRule) is returned unconditionally, regardless of whether this function actually inserts a trampoline or short-circuits here on `//otelc:ignore`. Back in `instrument()`, `hasFuncRule = hasFuncRule || funcRule` then drives `writeGlobals(...)`.

Both new golden fixtures always pair an ignored function with a legitimately-instrumented sibling in the same file, so `hasFuncRule` ends up `true` for the right reason and this is masked. But a file whose *only* func rules target `//otelc:ignore`'d functions would still get `otelc.globals.go` written with an unused `HookContext` interface and trampoline-support types — dead code with no test catching it. Suggest having this function signal "skipped, no globals needed" back to the caller (e.g. via a bool return), rather than `applyOneRule` assuming every `*rule.InstFuncRule` needs globals.

tool/internal/instrument/instrument.go
```diff
@@ -81,8 +82,28 @@ func (ip *InstrumentPhase) instrument(ctx context.Context, rset *rule.InstRuleSe
...
+
 		// Apply the rules to the target file
 		for _, r := range rules {
+			if fileIgnored {
+				fr, isFuncRule := r.(*rule.InstFuncRule)
+				if !isFuncRule {
+					ip.Debug("Skip non-func rule due to file-level //otelc:ignore", "rule", r.GetName())
+					continue
+				}
+				funcDecl, ok, err1 := ast.FindFuncDecl(root, fr)
```
darccio, line 98: This `FindFuncDecl(root, fr)` call resolves the same function that `applyFuncRule` (in `apply_func.go`) resolves again a few lines later via its own `ast.FindFuncDecl(root, rule)` call, once the rule passes this override check and reaches `applyOneRule`. Same lookup, same rule, run twice per iteration when the file is ignored. Not a correctness issue — both calls are idempotent AST walks — but it's a duplicated code path that has to be kept in sync if `FindFuncDecl`'s contract changes. Might be worth threading the already-resolved `*dst.FuncDecl` through instead of re-resolving, though that would touch `applyOneRule`'s generic `rule.InstRule` signature.

tool/internal/instrument/instrument.go
```diff
@@ -81,8 +82,28 @@ func (ip *InstrumentPhase) instrument(ctx context.Context, rset *rule.InstRuleSe
...
+		fileIgnored := ast.FileLeadHasDirective(root, directiveIgnore)
+		if fileIgnored {
+			ip.Debug("File-level //otelc:ignore found, only //otelc:instrument functions will be instrumented", "file", file)
+		}
+
 		// Apply the rules to the target file
 		for _, r := range rules {
+			if fileIgnored {
+				fr, isFuncRule := r.(*rule.InstFuncRule)
+				if !isFuncRule {
```
darccio, line 94: Non-func rules (struct/raw/call/decl/directive) have no override mechanism at all in an ignored file — this `continue` drops them unconditionally with no way to write an equivalent of `//otelc:instrument` to force one through. That may well be intentional (the override is inherently function-scoped, since only a func decl can carry the override comment), but it isn't called out anywhere — not in the PR description, not in a code comment here, and no golden fixture combines a file-level `//otelc:ignore` with a non-func rule to prove this path behaves as expected. Worth an explicit comment here (or in user docs) stating that only func rules are overridable, so it doesn't read as an oversight.

tool/internal/instrument/instrument.go
```diff
@@ -81,8 +82,28 @@ func (ip *InstrumentPhase) instrument(ctx context.Context, rset *rule.InstRuleSe
...
+					ip.Debug("Skip non-func rule due to file-level //otelc:ignore", "rule", r.GetName())
+					continue
+				}
+				funcDecl, ok, err1 := ast.FindFuncDecl(root, fr)
+				if err1 != nil {
+					return ex.Wrapf(err1, "finding function %s", fr.Func)
+				}
+				if !ok || !ast.FuncHasDirective(funcDecl, directiveInstrument) {
```
darccio, line 102: Precedence question: if a function carries **both** `//otelc:instrument` and `//otelc:ignore` in a file that also has the file-level `//otelc:ignore`, this check passes (override matches), the rule reaches `applyOneRule`/`applyFuncRule`, and `applyFuncRule`'s own directive check then skips it — so file-level-ignore + func-level-instrument + func-level-ignore nets out to **ignore wins**. That's a defensible resolution (most specific/closest-to-declaration directive wins), but it's untested and undocumented, and it's the same code path that produces the dead-globals-file side effect from the comment on line 51 above. Suggest a golden fixture (or unit test) pinning this precedence explicitly, plus a one-line doc note on the expected outcome.

darccio (commented):
Review notes on the file-level ignore/override gate in `tool/internal/instrument/instrument.go`, plus one finding that spans back into `apply_func.go`. Non-blocking, flagging for discussion.

tool/internal/instrument/testdata/golden/ignore-file-directive/rules.yml
```diff
@@ -0,0 +1,13 @@
+hook_skipped:
```
darccio, line 1: Both rules here are func rules (`InstFuncRule`). There's no fixture where a file-level `//otelc:ignore` coexists with a non-func rule (struct/raw/call/decl/directive). The `instrument.go` loop unconditionally `continue`s past any non-func rule when the file is ignored — that branch currently has no golden coverage proving it behaves as expected. Worth adding a third rule of a non-func kind to this fixture (or a new one) to pin that path.

tool/internal/instrument/testdata/golden/ignore-file-directive/source.go
```diff
@@ -0,0 +1,13 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//otelc:ignore
+
+package main
+
+func SkippedFunc(p1 string) {}
+
+//otelc:instrument
```
darccio, line 10: Would be good to also cover the case where a function carries **both** `//otelc:instrument` and `//otelc:ignore` at once (e.g. add this to `ForcedFunc` too, or a third function). Right now the precedence between the two directives on the same function is only inferable from reading the code — nothing pins whether ignore wins, instrument wins, or it's undefined.

tool/internal/instrument/testdata/golden/ignore-directive/source.go
```diff
@@ -0,0 +1,11 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+//otelc:ignore
```
darccio, line 6: Both fixtures only apply directives to plain top-level functions. `FuncHasDirective` operates on `*dst.FuncDecl` generically, so it should work identically for a method with a receiver, but nothing here proves it. A small additional case with `//otelc:ignore` on a method would close that gap.

tool/internal/instrument/testdata/golden/ignore-directive/hook.go
```diff
@@ -0,0 +1,13 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package testdata
+
+import (
+	_ "unsafe"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/inst"
```
darccio, line 9: This imports `.../pkg/inst`, which doesn't exist in this repo (only `pkg/hook` does) — same import is duplicated in `ignore-file-directive/hook.go`. It doesn't break anything since these `hook.go` fixtures aren't compiled against a real import path by the golden harness, and it's not something this PR invented: 7 other existing fixtures (e.g. `target-glob-*`, `is-test-filter-*`) already use this same nonexistent path, while the other 28 correctly use `pkg/hook`. Since this PR is touching new fixtures anyway, might be worth using the real `pkg/hook` import here to avoid perpetuating the inconsistency for future readers.

tool/internal/instrument/testdata/golden/ignore-file-directive/source.go
```diff
@@ -0,0 +1,13 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//otelc:ignore
```
darccio, line 4: This fixture only exercises the layout where a blank line separates `//otelc:ignore` from `package main` (keeping it out of the package doc-comment slot). If a user omits that blank line, it's untested whether `dst` still places the comment in `file.Decs.Start` for `FileLeadHasDirective` to find — worth a second fixture (or a unit test closer to `FileLeadHasDirective`) covering the no-blank-line layout, since that's likely to be the first thing a real user tries.

darccio (commented):
Review notes on the golden test fixtures for the two new directives. Non-blocking, flagging coverage gaps for discussion.

tool/internal/ast/directive.go
darccio, line 160: This PR introduces two new source-level directives (`//otelc:ignore`, `//otelc:instrument`) but doesn't touch anything under `docs/` — I checked the current `docs/rules.md`, `docs/instrument-guide.md`, and `docs/troubleshooting.md` (via this branch, which already has upstream `main` merged in) and none of them mention either directive. This is new syntax users have to type correctly by hand (no space after `//`, a blank line required before `package` for the file-level form to be recognized, and non-obvious precedence when both directives apply to the same function) — exactly the kind of feature that benefits from a docs mention. The PR checklist has "Documentation updated (if applicable)" checked, but I don't see a corresponding docs change. Worth adding a short section (e.g. to `docs/rules.md` or `docs/instrument-guide.md`) covering syntax, placement rules, and the override/precedence behavior.

darccio:
@pradhyum6144 Please address the comments.

PR #470, opened 2026-05-06
fix(pkg/instrumentation/databasesql): traces drop server.address for standard DSN formats 

### LINKED ISSUE #460: fix(pkg/instrumentation/databasesql): traces drop server.address for standard DSN formats
While looking at the databasesql instrumentation, I found that parse.go doesn't handle several standard connection string formats, which causes spans to silently lose server.address and db.name attributes.

Three specific issues:

- parsePostgres() function only handles the `postgres://` url scheme and the libpq key value format `(host=localhost port=5432 dbname=mydb)` fails entirely, both pq and pgx use this as a primary connection method.

- parseMySQL() function requires parentheses around the host address. DSNs like `user:pass@/dbname` return "invalid MySQL DSN" even though go sql driver's mysql supports this and defaults to localhost:3306.

- ParseDbName() function only searches backward for /. key value patterns like `dbname=mydb (libpq)` and `Database=myDB (SQL Server)` return empty string.

when any of these fail, client.go falls back to "unknown" for the address and empty string for the db name.
I'm happy to open a PR with fixes and tests. No tests exist for this package currently so yeah wanted to confirm first if this is a known limitation or intentional before putting in the work

alaotach:
The DSN parsing in parse.go didn't handle several standard connection string formats causing spans to silently lose server.address, port, and namespace attributes for perfectly valid DSNs

## What changed:
- parsePostgres() - added fallback for libpq key value format (host=localhost port=5432 dbname=mydb). Previously only postgres:// URL scheme worked. Both pq and pgx use this format heavily.
- parseMySQL() - added fallback to localhost:3306 when no quoted address is present. go-sql-driver/mysql allows user:pass@/dbname and treats it as local, but the parser was returning "invalid MySQL DSN" for these.
- ParseDbName() - added regex based extraction for dbname= and Database= key value patterns before falling back to the / search. sql server and libpq style conn strings were returning empty db name.

Integration tests added for all three cases following existing test structure. Test app driver registration extended to cover the aliases these tests need (mysql, postgres, sqlserver etc).
Before this, any app using these DSN formats would get server.address="unknown" and empty namespace in their traces which defeats the point of the instrumentation.

closes #460


pkg/instrumentation/databasesql/parse.go
```diff
@@ -7,10 +7,17 @@ import (
 	"errors"
 	"fmt"
 	nurl "net/url"
+	"regexp"
 	"strings"
 )
 
+var dbNamePattern = regexp.MustCompile(`(?i)(^|[\s;?&])(?:dbname|database)\s*=\s*('([^']*)'|"([^"]*)"|[^;\s&]+)`)
```
txabman42, line 15: This regex is too tricky, e.g. from `user:pass@tcp(host)/mydb?timeout=30s&database=other` matches `database=other`

what about making `parseDSN` return also the dbname?

test/apps/dbclient/main.go
```diff
@@ -25,6 +25,11 @@ var (
 
 func init() {
 	sql.Register("testdb", &testDriver{})
+	sql.Register("mysql", &testDriver{})
```
txabman42, line 28: What's the value of these new registrations in the test?

pkg/instrumentation/databasesql/parse.go
```diff
@@ -7,10 +7,17 @@ import (
...
+
 func ParseDbName(dsn string) string {
```
txabman42, line 17: Could we add some unit tests to cover non-happy path scenarios?

alaotach:
hey I’ve fixed the issues and the libpq host parsing problem, the pr should be safe to merge. 

please take a look and let me know if you want any changes

y1yang0:
@NameHaibinZhang Please help review this when you have time. Thanks.

txabman42 (changes requested):

txabman42:
Closing, already solved by https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/474

PR #473, opened 2026-05-07
chore(test): isolate golden source fixtures

vyagh:
## Description
Follow up to #466. This trims the golden test fixtures so each case only keeps the source it actually needs.
The golden files were regenerated after that.

refs : #411 

  ## Checklist

  - [x] Tests pass: `make test`
  - [x] Documentation updated (not applicable)


vyagh:
@y1yang0

txabman42 (approved):
LGTM! TY

tool/internal/instrument/testdata/golden/call-rule-only/source.go
```diff
@@ -5,36 +5,6 @@ package main
...
-
-func (g *GenStruct[T]) GenericMethod(p1 T, p2 string) (T, error) {
-	return p1, nil
-}
-
-func EllipsisFunc(p1 ...string) {}
-
-func UnderscoreFunc(_ int, _ float32) {}
-
 func Wrapper(size uintptr) uintptr {
```
y1yang0, line 8: Found an issue unrelated to this PR(no changes are needed in this PR): the wrap_sizeof_call test rule is missing the path field. We need this field to specify where to locate the Wrapper.

Tracked in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/475

tool/internal/instrument/testdata/golden/invalid-receiver/source.go
```diff
@@ -3,43 +3,8 @@
 
 package main
 
-var GlobalVar = "original"
-
 type T struct{}
 
 func (t *T) Func1(p1 string, p2 int) (float32, error) {
 	return 0.0, nil
 }
```
y1yang0: Should be removed as well?

y1yang0 (commented):
Otherwise LGTM, thanks for doing this.

y1yang0 (approved):

vyagh:
i'll address the changes, yang 
please only merge after that :)

tool/internal/instrument/testdata/golden/invalid-receiver/source.go
vyagh: I think we should keep this, I tried to remove this but without this it just fails because "Func1" does not exist at all

amazingakai: `invalid-receiver/rules.yml` contains `recv: *NonExistent` so it shouldn't match at all in first place (which I don't think it does since there is no .golden file generated for this case)

vyagh: removed it :)

PR #474, opened 2026-05-07
fix(databasesql): expand DSN parsing to cover libpq KV, MySQL no-parens, and SQLite URI formats

### LINKED ISSUE #460: fix(pkg/instrumentation/databasesql): traces drop server.address for standard DSN formats

pradhyum6144:
## Summary

Fixes #460.

Three DSN formats were silently producing empty `db.name` / `server.address` attributes:

| Driver | Format | Example |
|---|---|---|
| `postgres` / `pgx` | libpq key=value | `host=db.example.com port=5432 dbname=prod` |
| `mysql` | host not parenthesised | `user:pass@tcp:3306/mydb` |
| `sqlite3` | `file:` URI scheme | `file:test.db?cache=shared` |

## Changes

- **New `dsnparse` sub-package** (`pkg/instrumentation/databasesql/dsnparse/`) — isolates all parsing logic so it can be compiled and tested without the otelc-injected fields on `sql.DB`.
- **`ParseDSN(driver, dsn string) DSNInfo`** — driver-dispatched parser returning structured `{Host, Port, DBName}`. Supports `postgres`, `pgx`, `postgresql`, `mysql`, `sqlite3`, `sqlite`, `sqlserver`, `mssql`, `clickhouse`, `godror`, `oracle`, `oci8`, `go-oci8`.
- **PostgreSQL**: handles both RFC 3986 URL and libpq `key=value` connection strings (including single-quoted values with backslash escapes).
- **MySQL**: handles both the standard `proto(host:port)/dbname` form and the non-standard form without parentheses; Unix socket paths (`unix(/tmp/mysql.sock)`) return an empty port.
- **SQLite**: extracts `DBName` from `file:` URIs, `:memory:`, and bare filenames/paths.
- **SQL Server, ClickHouse, Oracle**: URL and ADO.NET / `user@host/service` forms.
- **`parse.go`** in `package db` is now a thin shim that re-exports `ParseDSN`, `ParseDbName`, and the internal `LegacyParseDSN` shim unchanged.
- **60+ table-driven tests** covering URL, key-value, and edge cases for each driver; three sub-tests are explicitly named `*PreviouslyFailing` to document the regression.

## Test plan

- [ ] `go test -C pkg/instrumentation/databasesql ./dsnparse/...` — all tests pass
- [ ] Verify `db.name` and `server.address` attributes appear in traces when connecting with each of the three previously-failing DSN formats


txabman42:
@pradhyum6144 duplicated, there is already a [PR](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/470) trying to solve https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/460.

Since you already have context about the problem, feel free to review the other PR and align approaches.

kakkoyun (approved):

PR #476, opened 2026-05-08
fix(tool): remove unused after trampoline declarations

### LINKED ISSUE #192: Remove unused functions made redundant by optimizer
Currently, optimize.go implements three types of optimizations. After these optimizations are applied, some functions may become redundant and can be removed.

For instance, consider the following optimization:
```
	if ctx, skip := otel_trampoline_before(&arg); skip {
	    otel_trampoline_after(ctx, &retval)
	    return ...
	} else {
	    ;
	    ...
	}
=>
	if false {
	    ;
	} else {
	    defer otel_trampoline_after(&HookContext{...}, &retval)
	    ...
	}
```

As a result of this transformation, the otel_trampoline_before function is no longer called, allowing us to remove its definition entirely.

In addition to this case, the other two optimizations can create similar opportunities for dead code elimination. This helps to reduce the final binary size and improve instruction cache performance.

vyagh:
## Description
Remove unused generated `OtelAfterTrampoline_*` functions after trampoline optimization. This only removes the declaration when there are no calls left in the instrumented file. If a skip or defer pathstill calls the after trampoline, it stays

  Fixes #192

  ## Checklist

  - [x] Tests pass: `make test`


tool/internal/instrument/optimize.go
```diff
@@ -178,6 +178,50 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+			return false
+		}
+		if callExpr, ok := node.(*dst.CallExpr); ok && callsFuncName(callExpr.Fun, name) {
+			found = true
+			return false
+		}
+		return true
+	})
+	return found
+}
```
amazingakai: Do we need this check, doesn't `removedOnExit` already mean there is no after hook?

tool/internal/instrument/optimize.go
y1yang0, line 364: Maybe simpliy do removal here?
```
      if rule.After == "" {
			err := removeAfterTrampolineCall(tjump)
			if err != nil {
				return err
			}
             removeAfterTrampoline(tjump) <---
			removedOnExit = true
		}
```

tool/internal/instrument/optimize.go
```diff
@@ -178,6 +178,50 @@ func removeBeforeTrampolineCall(targetFile *dst.File, tjump *TJump) error {
...
+
+func removeUnusedAfterTrampolineDecl(targetFile *dst.File, tjump *TJump) error {
+	afterTrampolineName := makeName(tjump.rule, tjump.target, false)
+	if hasFuncCall(targetFile, afterTrampolineName) {
```
y1yang0: Since after-trampoline-call is removed, this makes no sense?

amazingakai: Non-blocking nit: would it make sense to move the after trampoline removal logic directly inside `removeAfterTrampolineCall` for symmetry with `removeBeforeTrampolineCall`?

kakkoyun:
@vyagh Friendly ping. Would you like to address the comments?

vyagh:
thanks for the ping @kakkoyun :)
sorry got a bit occupied, I'll look into this asap!

vyagh:
Pushed a smaller cleanup based on the feedback, and golden files are updated + make all passes locally

tool/internal/instrument/optimize.go
y1yang0: This one is unaddressed

vyagh: my bad

kakkoyun (approved):
LGTM

tool/internal/instrument/optimize.go
```diff
@@ -353,17 +393,8 @@ func (ip *InstrumentPhase) optimizeTJumps() error {
...
-			if err != nil {
-				return err
-			}
-			if canFlattenTJump(hookFunc) {
-				err1 := flattenTJump(tjump, removedOnExit)
-				if err1 != nil {
-					return err1
-				}
-			}
+		if err := flattenTJumpIfPossible(tjump, removedOnExit); err != nil {
```
y1yang0: Unrelated changes? Please restore it.

vyagh: I restored it.

y1yang0 (approved):

amazingakai (approved):
Awesome :heart:

PR #478, opened 2026-05-09
fix(tool): validate OTELC_RULES and --rules rule paths for security

tryh4rd-26:
## Description

I added path validation before loading rule files so rules must live inside the working directory unless explicitly opted in via `OTELC_ALLOW_EXTERNAL_RULES=1`.

- Added `IsPathInside()` in `util/shared.go`, uses `filepath.Rel` to check containment after resolving absolute paths
- Added `ValidateRulesPath()` as a single shared helper so both the env var path (`loadRules`) and the config file path (`loadCustomRules`) go through the same check, no duplication
- `ValidateRulesPath` returns an explicit error if `OTELC_WORK_DIR` is unset, instead of silently returning false
- Added `#nosec G304` on the `os.ReadFile(rulePath)` call since the path is validated immediately above it
- Updated existing tests that write rules to temp dirs to set `OTELC_ALLOW_EXTERNAL_RULES=1` so they don't break under the new validation
- Fixed test indentation in `TestLoadRulesEnvValidation` and fixed `gofumpt` alignment in the const block

 **Assisted-by: GitHub Copilot** 

## Motivation

Without this, `OTELC_RULES` and `--rules` accept any path on the filesystem, so a misconfigured or tampered env var could silently load rules from anywhere. Defaulting to the working directory with an explicit opt-in should be the right call.

---

## Checklist
- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -324,7 +324,13 @@ func loadCustomRules(ruleConfig string) ([]rule.InstRule, error) {
 	for path := range ruleFiles {
 		path = strings.TrimSpace(path)
 
-		// Get all rule files from path (file or directory)
+		// Guardrail: keep rules inside working directory unless opt-in is set.
+		if err := util.ValidateRulesPath(path); err != nil {
+			return nil, err
+		}
+
+		// Get all rule files from path (file or directory).
+		// IsPathInside uses filepath.Abs internally which handles path normalization.
 		files, err := ruleFromDir(path)
 		if err != nil {
 			return nil, err
@@ -356,7 +362,20 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
 	rulePath := os.Getenv(util.EnvOtelcRules)
 	if rulePath != "" {
 		sp.Debug("rules source: environment variable %s (%s)", util.EnvOtelcRules, rulePath)
-		content, err := os.ReadFile(filepath.Clean(rulePath))
+
+		// Guardrail: keep rules inside working directory unless opt-in is set.
+		if err := util.ValidateRulesPath(rulePath); err != nil {
+			return nil, err
+		}
+		// Warn if loading from external path (user has opted in).
+		wd := util.GetOtelcWorkDir()
+		if wd != "" && !util.IsPathInside(wd, rulePath) {
+			sp.Warn("loading rules from external path (opt-in)", "path", rulePath)
+		}
+
+		// Path is validated by ValidateRulesPath() call above; safe for use.
+		// #nosec G304
+		content, err := os.ReadFile(rulePath)
 		if err != nil {
 			return nil, ex.Wrapf(err, "failed to read %s from env variable", rulePath)
 		}
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -374,8 +374,7 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
 		}
 
 		// Path is validated by ValidateRulesPath() call above; safe for use.
-		// #nosec G304
-		content, err := os.ReadFile(rulePath)
+		content, err := os.ReadFile(rulePath) //nolint:gosec // validated above
 		if err != nil {
 			return nil, ex.Wrapf(err, "failed to read %s from env variable", rulePath)
 		}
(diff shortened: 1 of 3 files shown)


amazingakai:
Thank you for the contribution! I had one question:

When using `--rules`, the user is already explicitly providing the rules path. And for `OTELC_RULES`, if an attacker is already able to tamper with that variable, wouldn't they also be able to set `OTELC_ALLOW_EXTERNAL_RULES=1` at the same time? Similarly the attacker can also modify `OTELC_WORK_DIR` which would make the validation pass.
I may be missing part of the thread model here, but I am not entirely sure how this validation helps.

tryh4rd-26:
> Thank you for the contribution! I had one question:
> 
> When using `--rules`, the user is already explicitly providing the rules path. And for `OTELC_RULES`, if an attacker is already able to tamper with that variable, wouldn't they also be able to set `OTELC_ALLOW_EXTERNAL_RULES=1` at the same time? Similarly the attacker can also modify `OTELC_WORK_DIR` which would make the validation pass. I may be missing part of the thread model here, but I am not entirely sure how this validation helps.


Yeah, you're right. This doesn't protect against an attacker who can fully control the process environment since they could just set `OTELC_ALLOW_EXTERNAL_RULES` and `OTELC_WORK_DIR` alongside `OTELC_RULES` anyway.

The intent is narrower than I made it sound:
- Prevent accidental external rule loading from misconfiguration
- Reduce confused-deputy style mistakes in normal developer/CI usage where env vars get inherited unintentionally
- Make external rule loading an explicit opt-in rather than silent default behavior

Should have framed this as a safety guardrail, not a hard security boundary. I'll update the PR description and comments to reflect that.

tool/internal/setup/match.go
```diff
@@ -356,7 +362,20 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
+		}
+		// Warn if loading from external path (user has opted in).
+		wd := util.GetOtelcWorkDir()
+		if wd != "" && !util.IsPathInside(wd, rulePath) {
+			sp.Warn("loading rules from external path (opt-in)", "path", rulePath)
+		}
+
+		// Path is validated by ValidateRulesPath() call above; safe for use.
+		// #nosec G304
+		content, err := os.ReadFile(rulePath)
```
amazingakai:
```suggestion
		content, err := os.ReadFile(rulePath) //nolint:gosec // validated above
```

tool/internal/setup/match.go
```diff
@@ -324,7 +324,13 @@ func loadCustomRules(ruleConfig string) ([]rule.InstRule, error) {
...
 		path = strings.TrimSpace(path)
 
-		// Get all rule files from path (file or directory)
+		// Guardrail: keep rules inside working directory unless opt-in is set.
+		if err := util.ValidateRulesPath(path); err != nil {
+			return nil, err
+		}
+
+		// Get all rule files from path (file or directory).
+		// IsPathInside uses filepath.Abs internally which handles path normalization.
```
amazingakai, line 333: I'm still not convinced we need this check here, `--rules` flag is already explicit. Take my comment as non-blocking, let's wait for a maintainer here.

tool/internal/setup/match.go
```diff
@@ -356,7 +362,20 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
-		content, err := os.ReadFile(filepath.Clean(rulePath))
+
+		// Guardrail: keep rules inside working directory unless opt-in is set.
+		if err := util.ValidateRulesPath(rulePath); err != nil {
+			return nil, err
+		}
+		// Warn if loading from external path (user has opted in).
+		wd := util.GetOtelcWorkDir()
+		if wd != "" && !util.IsPathInside(wd, rulePath) {
+			sp.Warn("loading rules from external path (opt-in)", "path", rulePath)
```
amazingakai, line 373: Since `sp.Warn` logs to `.otelc-build/debug.log`, this warning may be missed by a lot of people, but I don't see any code logging to stdout/stderr directly so maybe this is fine.

tryh4rd-26: Agreed, will leave as-is since there's no precedent for direct stderr logging here.

y1yang0:
Thanks for contribution. But I agree with @amazingakai .  It's not worth adding extra complexity for a low-probability misconfiguration, because:

- `-rules` is rarely used; most users expect out-of-the-box rules rather than custom ones.
- Misconfigurations are fairly easy to catch, as .otelc-build/debug.log already provides comprehensive logging.

PR #481, opened 2026-05-09
chore(integration): add DSN parsing coverage for db client instrumentation

### LINKED ISSUE #480: Add unit tests for databasesql DSN parsing helpers
## Problem

The DSN parsing helpers in `pkg/instrumentation/databasesql/parse.go` currently do not have direct unit test coverage.

The existing integration tests use a mock driver and do not exercise the dialect-specific parsing logic used for common databases like MySQL, PostgreSQL, SQL Server, and SQLite. As a result, behaviors such as `server.address` and database name extraction are not directly validated by standalone tests.

## Proposed Change

I’m planning to send a small PR adding table-driven unit tests for:

* `parseDSN`
* `ParseDbName`

The initial scope will focus only on common happy-path DSN formats to keep the change small and review-friendly.

pushkarscripts:
## Description

Add table-driven unit tests for the DSN parsing helpers in `pkg/instrumentation/databasesql/parse.go`.

Covers common parsing cases for:
* MySQL
* PostgreSQL
* SQL Server
* SQLite

## Motivation

The DSN parsing helpers currently do not have direct unit test coverage.

This PR adds foundational coverage for `parseDSN` and `ParseDbName` to validate extraction of database names and `server.address` values from common DSN formats.

Fixes #480 

---

## Checklist

- [x] Tests pass: `make test`


amazingakai (commented):
Thanks for the contribution!

`pkg/instrumentation/databasesql` itself currently does not compile under normal `go test` runs because we inject additional fields into `sql.DB` during instrumentation (`Endpoint`, `DriverName`, `DSN`, `DbName`, etc). So right now only the `.../databasesql/semconv` package tests are actually running successfully.

It might be nicer long-term if we instead kept this metadata in a separate structure/map associated with `sql.DB` instances rather than injecting fields directly into `sql.DB`. That would allow us to add unit tests for `.../databasesql`. @NameHaibinZhang  would appreciate your opinion here.

pushkarscripts:
> Thanks for the contribution!
> 
> `pkg/instrumentation/databasesql` itself currently does not compile under normal `go test` runs because we inject additional fields into `sql.DB` during instrumentation (`Endpoint`, `DriverName`, `DSN`, `DbName`, etc). So right now only the `.../databasesql/semconv` package tests are actually running successfully.
> 
> It might be nicer long-term if we instead kept this metadata in a separate structure/map associated with `sql.DB` instances rather than injecting fields directly into `sql.DB`. That would allow us to add unit tests for `.../databasesql`. @NameHaibinZhang would appreciate your opinion here.

Thanks for clarifying

I initially treated `parse.go` as an isolated utility layer, but I understand the concern around the injected `sql.DB` fields and package-level testing now.

The separate metadata structure approach makes sense. I’ll wait for maintainer feedback as well.

NameHaibinZhang:
A separate map could potentially introduce memory management issues, such as how to properly release entries and avoid leaks. Based on our previous experience, embedding the metadata directly in the DB struct is more appropriate. I'd prefer not to introduce a new map for this purpose.The testing concern can be addressed by adding these tests to the integration test suite instead. Would you consider moving these tests to the integration tests? @pushkarscripts

pushkarscripts:
> A separate map could potentially introduce memory management issues, such as how to properly release entries and avoid leaks. Based on our previous experience, embedding the metadata directly in the DB struct is more appropriate. I'd prefer not to introduce a new map for this purpose.The testing concern can be addressed by adding these tests to the integration test suite instead. Would you consider moving these tests to the integration tests? @pushkarscripts

Alright! I’ll move the coverage to the integration test suite instead and update the PR accordingly.

pushkarscripts:
I've updated the PR to move the DSN coverage into the integration test suite as suggested.

Also added coverage for the different DSN parsing paths through the dbclient integration flow. Let me know if any further changes are needed.

kakkoyun (approved):

kakkoyun:
@pushkarscripts Tests failures look legit. Could you have a look?

test/integration/db_client_test.go
```diff
@@ -184,3 +184,74 @@ func TestDBClient(t *testing.T) {
...
+			wantAddr:   "sqlite3",
+			wantPort:   0,
+			wantDb:     "", // Current ParseDbName logic returns empty if no '/' is found
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f := testutil.NewTestFixture(t)
+			f.BuildAndRun("dbclient",
```
txabman42: After these [changes](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/540), now we could we use the binary directly that `TestMain` already gives prebuilt.

And then just sth like:

```go
f.Run("dbclient",
    "-driver="+tt.driverName,
    "-dsn="+tt.dsn,
    "-op=ping",
)
```

txabman42 (approved):
LGTM! Lets just align with the new testing approach, more efficient.

test/integration/db_client_test.go
pushkarscripts: Thanks! I'll take a look and update it to follow the newer test pattern.

kakkoyun:
@pushkarscripts I'd be happy to merge this when the CI is green.

NameHaibinZhang (approved):

Issue #475, opened 2026-05-08
The wrap_sizeof_call test rule is missing the `path` field


y1yang0:
The wrap_sizeof_call test rule is missing the `path` field. We need this field to specify where to locate the `Wrapper` function

https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/8c7600d04fcf74344775545e47e1531e7cff4672/tool/internal/instrument/testdata/golden/call-rule-only/rules.yml#L1-L4

wrap_sizeof_with_strings has the same issue. 

We should do a thorough review to catch all missing path fields.

amazingakai:
@y1yang0 `InstCallRule` does not support the `path` field right now. Should we support it across more rules (like call rules, raw rules, declaration rules, etc), or should `Wrapper` resolution instead happen through the existing `imports` field?

My understanding is that we would then mainly need a way for `compileArgs/createImportCfg` in golden tests to compile and resolve/import wrapper packages relative to the test case itself when used.

Although I am not entirely sure there is much value in doing that for golden tests specifically, since we would mostly just be adding extra test infrastructure to wire imports from testcase-local packages without gaining much additional coverage from it.

pm-ju:
hey @y1yang0 is this issue solved already or anyone working on this?

txabman42:
I agree with @amazingakai 

IMO we should close https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/492 and https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/482.

We already have the import mechanism. I guess the main problem is naming and the diffs between rules. Quite related to previous discussions (https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/377).

Regarding testing, we are testing inline targets. The call in another package is missing. As that's the real production scenario, I think it's worth testing --> [PR](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/496)

cc @y1yang0

ecleon26:
@txabman42 Looks like the right direction to me. Using `imports` keeps the call/raw/decl/directive rule behaviour, without adding another meaning to `path`

One minor robustness suggestion: in buildTestcaseHelpers, after go list -export -json, perhaps check that ImportPath and Export are both non-empty before appending to helpers. If `Export` is empty for some reason, the subsequent compile failure would be a bit harder to link back to helper package discovery.

[...]

PR #482, opened 2026-05-10
fix(tool): support path helpers for call rules

ecleon26:
## Description

Adds `path` support to call rules so replacement wrappers can reference helper functions that live outside the target source file.

This change:
- adds `Path` parsing to `InstCallRule`
- detects unqualified helper calls used by call-rule `replace` / `append_args`
- materializes matching helper files from the configured `path` into the target package during instrumentation
- updates importcfg for helper-file imports
- updates the affected golden fixtures, including `wrap_sizeof_call` and `wrap_sizeof_with_strings`
- documents the new call-rule `path` behavior

## Motivation

Call-rule YAML fixtures can specify wrapper expressions such as `Wrapper({{ . }})`, but `InstCallRule` previously had no `Path` field. As a result, adding `path:` to YAML would be silently ignored, and call-rule replacements could only reference helpers already defined in the target source file or symbols available through imports.

This fixes the missing code path so helper functions can be located and compiled into the target package when a call rule needs them.

Fixes #

---

## Checklist

- [x] Tests pass: `make test`


ecleon26:
@y1yang0

amazingakai:
Thank you for the contribution.

We already support helper usage through the existing `imports` field. For example, if we want to use our own wrapper/helper code around a function, we can define it inside `github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/...`, add it through `imports`, and then reference it directly inside `replace`.
From my understanding, the main missing piece right now is support for this in golden tests, since we would need `compileArgs/createImportCfg` to compile and resolve those helper imports correctly there as well (please see my [comment](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/475#issuecomment-4402774316) on the original issue).

ecleon26:
That makes sense to me. I agree that broadening `path` to call/raw/decl rules may blur the existing rule semantics, since those rules already have an `imports` mechanism for external helper code.

For call rules specifically, it seems cleaner to reference helpers through an imported package alias in `replace`, for example `wrapper.Wrapper({{ . }})`, and declare that package through `imports`. That keeps helper resolution explicit and consistent with raw/decl-style expression injection.

So I think the better fix may be limited to golden-test infrastructure: either teach `compileArgs` / `createImportCfg` how to resolve testcase-local helper packages, or avoid testcase-local wrapper imports if that adds too much test-only complexity. I do not think we should add `path` support across more rule types unless we want a broader rule-model change.

kakkoyun:
@ecleon26 Would you like to rebase and revive this PR?

tool/internal/rule/call_rule.go
```diff
@@ -66,6 +67,13 @@ type InstCallRule struct {
 	// When set and the call is ellipsis, an IIFE wrapper is generated.
 	// When unset and the call is ellipsis, the call is skipped with a warning.
 	VariadicType string `json:"variadic_type" yaml:"variadic_type"`
+
+	// Path is the module path where helper code referenced by the replacement is located.
+	Path string `json:"path" yaml:"path"`
+
+	// ResolvedPath is the local filesystem path of the package directory resolved
+	// from Path. Populated during the setup phase when Path is non-empty.
+	ResolvedPath string `json:"resolved_path" yaml:"-"`
```
darccio, line 76: `ResolvedPath` is `yaml:"-"` but still has an explicit `json:"resolved_path"` tag, so it round-trips through the JSON-serialized matched-rule-set file that setup hands to the instrument phase. That's intentional and matches `InstFileRule`/`InstFuncRule`. One consequence: if a rule author's own JSON (rather than YAML) input ever included a `resolved_path` key, it would be silently honored instead of being computed by setup. Worth confirming that JSON is never an accepted user-facing input format for rules, or explicitly ignoring the field on read.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+var nonIdentifierChars = regexp.MustCompile(`[^A-Za-z0-9_]+`)
+
+func callRuleHelperOutputName(ruleName, file string) string {
+	base := filepath.Base(file)
+	ext := filepath.Ext(base)
+	name := strings.TrimSuffix(base, ext)
+	rulePart := nonIdentifierChars.ReplaceAllString(ruleName, "_")
+	filePart := nonIdentifierChars.ReplaceAllString(name, "_")
+	return fmt.Sprintf("otelc.%s.%s.go", rulePart, filePart)
+}
```
darccio, line 270: Helper output files are named `otelc.<ruleName>.<helperFileBase>.go` — the rule name is baked into the filename. If two distinct call rules both reference a helper file that lives in the same source path (e.g. two rules pointing at the same `path:` directory and both needing `Wrapper`), each rule materializes its own copy of that file into the target package under a different generated name, and the target package ends up with `Wrapper` declared twice — a compile-time redeclaration error.

File rules avoid this because their output name (`apply_file.go`'s `otelc.<fileBase>.go`) has no rule prefix, so `!util.PathExists(newFile)` naturally dedups repeat references to the same file. Call rules don't get that dedup here since the name varies per rule.

Is sharing one helper file across multiple call rules a supported scenario? If so, this needs either a rule-independent output name (with a correctness argument for why that's safe) or an explicit dedup/collision check.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+
+func removeLocalFuncNames(names map[string]bool, root *dst.File) {
+	for _, decl := range root.Decls {
+		funcDecl, ok := decl.(*dst.FuncDecl)
+		if !ok || funcDecl.Recv != nil {
+			continue
+		}
+		delete(names, funcDecl.Name.Name)
+	}
+}
```
darccio, line 209: `removeLocalFuncNames` only inspects `root`, the current target file, for existing top-level function declarations. It doesn't look at sibling files in the same package. If a helper function with the same name is already defined elsewhere in the target package (a plausible case if the `path:` directory overlaps with the package's own layout, or a prior file rule already introduced it), this rule would still materialize and add a second declaration, causing a package-level redeclaration at compile time.

Worth confirming whether this is meant to be scoped to just the current file, or whether it should also check already-materialized/known package declarations.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+			return true
+		}
+		ident, ok := call.Fun.(*dst.Ident)
+		if !ok || isBuiltinCall(ident.Name) {
+			return true
+		}
+		names[ident.Name] = true
+		return true
+	})
+}
```
darccio, line 189: `collectUnqualifiedCallNames` only detects names used in call position (`ident, ok := call.Fun.(*dst.Ident)`). A helper referenced by value rather than invoked directly — e.g. passed as a callback (`Register(myHelper)`) or assigned to a var — won't be picked up here, so it won't be materialized, and the generated file will fail to compile with `undefined: myHelper`.

This likely fails loudly rather than silently, so it may be an acceptable, documented limitation (similar to the existing "must produce a Go call expression" constraint on `replace`). Worth a doc note either way so it isn't rediscovered as a bug report.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+
+	files, err := callRuleHelperFiles(r.ResolvedPath, helperNames)
+	if err != nil {
+		return ex.Wrapf(err, "finding helper files for call rule %s at path %s", r.Name, r.Path)
+	}
+	for _, file := range files {
+		if err = ip.addCallRuleHelperFile(ctx, r, file, root.Name.Name); err != nil {
+			return err
+		}
+	}
```
darccio, line 144: Helper discovery only considers unqualified names referenced directly by the rule's `replace`/`append_args` expressions (via `callRuleHelperNames`). It never inspects the *body* of a matched helper file to see if that helper itself calls another unqualified helper defined in a sibling file under the same `path:`. If a user splits helper code across multiple files that call each other, only the directly-referenced file gets materialized, and the emitted package fails to compile with "undefined" for the second-order helper.

Since call rules already parse each candidate file's declarations (`callRuleHelperFiles`), extending name discovery to be transitive — rescanning newly matched files for further unqualified calls — would close this gap. Alternatively, if only single-file helpers are meant to be supported, that constraint should be documented and probably enforced with a clearer error than a raw Go compile failure.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+		}
+		delete(names, funcDecl.Name.Name)
+	}
+}
+
+func callRuleHelperFiles(resolvedPath string, names map[string]bool) ([]string, error) {
+	files, err := util.ListFiles(resolvedPath)
+	if err != nil {
+		return nil, err
+	}
```
darccio, line 215: `util.ListFiles(resolvedPath)` walks recursively (see `tool/util/sys.go`'s `filepath.Walk`-based implementation). If `ResolvedPath` contains a subdirectory that's actually a different Go package (e.g. an internal helper package under the same tree) and it happens to define a same-named function, that file would be matched, have its package renamed to the target, and be compiled in — silently pulling in unrelated code.

This behavior is inherited from the existing `apply_file.go` file-rule path (same `ListFiles` call), so it's not a new regression introduced by this PR, but it's now reachable from a new direction (`path:` on call rules) and worth a second look — e.g. should discovery be limited to files directly in `resolvedPath`, non-recursively?

tool/internal/instrument/apply_call.go
darccio, line 331: `applyCallAppendArgs` returns `true` (i.e. "this rule modified the file, proceed to materialize helpers") as soon as any call matches the rule — regardless of whether `appendCallArgs` itself succeeded. Its error is only logged via `ip.Warn` (line ~327) and swallowed, not propagated.

Contrast with `applyCallReplace`, whose error return aborts `applyCallRule` entirely (lines 34-38 above). So a broken `append_args` entry (e.g. an ellipsis call missing `variadic_type`) still triggers `applyCallRuleHelpers`, which then tries to resolve helper names from expressions that were never actually applied to the call site.

Is the softer handling for `append_args` errors intentional (e.g. because a single malformed call shouldn't block others), or should this propagate like `replace` does?

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+	for _, arg := range r.AppendArgs {
+		expr, err := parseGoExpression(arg)
+		if err != nil {
+			return nil, err
+		}
+		collectUnqualifiedCallNames(expr, names)
+	}
+	return names, nil
+}
```
darccio, line 174: `callRuleHelperNames` recompiles the `replace` template from scratch via `newCallTemplate` + `compileExpression`, using a synthetic placeholder call (`pkg.FuncName(...)`), even though `applyCallReplace` already compiled and applied this exact template earlier in the same `applyCallRule` invocation (see lines 32-38 above).

Not a correctness issue — the template is pure and the synthetic call is only used to extract referenced names — but it is redundant compilation work per call rule. Could potentially be threaded through from the first compilation instead of recomputing.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+	newFile := filepath.Join(ip.workDir, callRuleHelperOutputName(r.Name, file))
+	if !util.PathExists(newFile) {
+		if err = ast.WriteFile(newFile, root); err != nil {
+			return ex.Wrapf(err, "writing call rule helper file %s", newFile)
+		}
+		ip.keepForDebug(newFile)
+	}
+	if !ip.hasCompileArg(newFile) {
+		ip.addCompileArg(newFile)
+	}
```
darccio, line 300: The helper file is only written when `!util.PathExists(newFile)`, but it's *always* added to compile args once matched (subject only to the in-run `hasCompileArg` dedup). If `workDir` were ever reused across separate instrumentation invocations rather than being fresh per run, a stale file from a prior run could silently be picked up and compiled instead of being regenerated.

Worth confirming `workDir` is guaranteed fresh per run elsewhere in the pipeline — if so this is a non-issue, just noting the implicit assumption.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
...
+	return matched, nil
+}
+
+func mapsKeys(m map[string]bool) []string {
+	keys := make([]string, 0, len(m))
+	for key := range m {
+		keys = append(keys, key)
+	}
+	return keys
+}
```
darccio, line 259:
```suggestion
```

`mapsKeys` reimplements what `slices.Sorted(maps.Keys(m))` (stdlib since Go 1.23, and this module targets `go 1.25.0`) would give directly. Not submitting this as a one-click suggestion since it'd also require adding `maps` and `slices` imports — just flagging as a simplification opportunity.

docs/rules.md
```diff
@@ -700,11 +700,12 @@ This rule wraps function calls at call sites with instrumentation code. Unlike t
...
+| --------------- | ---------- | -------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
+| `replace`       | string     | No (one of `replace`/`append_args` required) | Replace string with `{{ . }}` placeholder for the original call. Must produce a Go call expression.                    |
+| `append_args`   | `[]string` | No (one of `replace`/`append_args` required) | Go expression strings appended as additional arguments to the matched call                                             |
+| `variadic_type` | string     | No                                           | Element type for the ellipsis IIFE wrapper (e.g. `grpc.DialOption`). Required when any matched call uses `...` spread. |
+| `path`          | string     | No                                           | Import path or local path containing helper functions referenced by unqualified calls in `replace` or `append_args`.   |
```
darccio, line 708: This row documents `path` as "Import path **or local path**", but I don't see a local-filesystem-path resolution mode in `tool/internal/setup/store.go` — `ruleGoPathResolver.resolve` always resolves `Path` via `packages.Load` as a Go import path. The relative-looking values used in golden fixtures (e.g. `testdata/golden/call-rule-only`) only work because the test harness (`writeMatchedJSON` in `instrument_test.go`) bypasses real resolution and copies `Path` straight into `ResolvedPath` — that's a test-only shortcut, not a production capability.

Notably, the pre-existing `path` field for function hook rules is documented a few hundred lines up as just "The import path..." (no "or local path"), even though it relies on the identical test-only convention. Suggest aligning this row with that wording unless there's a real local-path resolution mode elsewhere I'm missing — worth a quick check either way so the docs don't promise something the resolver doesn't do.

tool/internal/instrument/testdata/golden/call-rule-with-import/rules.yml
```diff
@@ -5,5 +5,6 @@ wrap_sizeof_with_strings:
   do:
     - wrap_call:
         replace: "(func() uintptr { fmt.Println(\"Wrapped!\"); return {{ . }} })()"
+        path: testdata/golden/call-rule-with-import
```
darccio, line 8: This fixture's `replace` only calls `fmt.Println(...)`, which is a qualified selector (`fmt.Println`), not a bare identifier — `collectUnqualifiedCallNames` won't register it as a helper name, so `callRuleHelperNames` returns empty and `applyCallRuleHelpers` short-circuits before ever touching `ResolvedPath`. There's no corresponding helper file or golden output added for this directory, which is consistent with that: this `path:` addition doesn't appear to exercise any new behavior.

If the intent is to cover "a call rule sets `path` but needs no helpers, so it's a harmless no-op," that's a reasonable case to pin down — but as-is it reads as an unexplained/inert change. A one-line YAML comment (or a note in the rule name) would make the intent clear instead of leaving it to be inferred from the absence of a helper golden file.

tool/internal/setup/store.go
```diff
@@ -14,71 +14,103 @@ import (
...
+	}
 
-		for _, funcRule := range ruleset.AllFuncRules() {
-			dir, err := resolve(funcRule.Path)
-			if err != nil {
-				return err
-			}
-			funcRule.ResolvedPath = dir
+	for _, callRule := range ruleset.AllCallRules() {
+		if callRule.Path == "" {
```
darccio, line 85: This guard checks the exact empty string, while the equivalent guard in the instrument phase (`tool/internal/instrument/apply_call.go`, in `applyCallRuleHelpers`) uses `strings.TrimSpace(r.Path) == ""`. A call rule with a whitespace-only `path` (e.g. `path: "  "`) would pass this check, get forwarded to `resolve(" ")`, and fail there with a `packages.Load` error — a confusing error for what's effectively an unset field, since the instrument-phase guard would have treated the same value as absent.

```suggestion
		if strings.TrimSpace(callRule.Path) == "" {
```

(Needs a `strings` import added to this file if adopted.) Aligning both guards to the same check keeps "path is unset" consistent across the setup/instrument boundary.

tool/internal/instrument/apply_call.go
```diff
@@ -104,6 +115,193 @@ func (*InstrumentPhase) applyCallReplace(
 	return true, nil
 }
 
+func (ip *InstrumentPhase) applyCallRuleHelpers(
+	ctx context.Context,
+	r *rule.InstCallRule,
+	root *dst.File,
+) error {
+	if strings.TrimSpace(r.Path) == "" {
```
darccio, line 123: For symmetry with the setup-phase guard (`tool/internal/setup/store.go`, in `resolveRuleSet`, which currently checks `callRule.Path == ""` exactly, not trimmed): if that guard is updated to trim as well, this one already does — just flagging the pairing so both stay in sync if either is touched later. Right now a whitespace-only `path` slips past setup's exact-match guard and fails during `packages.Load` resolution instead of being treated as unset here.

darccio (changes requested):

darccio:
@ecleon26 Please address my comments.

PR #486, opened 2026-05-11
fix(instrument): correct applyCallAppendArgs match accounting

### LINKED ISSUE #485: applyCallAppendArgs returns true unconditionally, bypassing unmatched-rule assertion
### Describe the bug

`applyCallAppendArgs` in `tool/internal/instrument/apply_call.go` returns `true` whenever `r.AppendArgs` is non-empty, even if no matching calls are found in the AST.

`applyCallRule` uses the return values from `applyCallAppendArgs` and `applyCallReplace` to enforce this assertion:

```go
util.Assert(appendModified || replaceModified, "call rule did not match any call")
```

Because `applyCallAppendArgs` always reports success, append-only rules can bypass this assertion even when zero calls match.

### To Reproduce

Create a call rule with `append_args` set and apply it to a source file containing no matching calls.
`applyCallAppendArgs` returns `true`, causing the assertion in `applyCallRule` to pass even though the rule matched nothing.

### Expected behavior

`applyCallAppendArgs` should return `false` when no matching calls are found so unmatched rules are detected consistently.

### Additional context

`applyCallReplace` already returns `false` when no replacements are generated (`len(replacements) == 0`), so this appears to be inconsistent return-value behavior between the two call instrumentation paths.

The fix is a one-line change

```go
return len(matchingCalls) > 0
```

I can open a PR with the fix and a regression test for the same

beep-boopp:
## Description

`applyCallAppendArgs` returns `true` when `AppendArgs` is set, regardless of whether any calls actually matched. This means the `util.Assert` in `applyCallRule` never fires for append-only rules that match nothing.

Changed `return true` to `return len(matchingCalls) > 0`. Added a test that verifies the function returns false when AppendArgs is set but no calls match in the AST.

## Motivation

`applyCallReplace` already handles this correctly — it returns `false` when `len(replacements) == 0`. This just makes `applyCallAppendArgs` consistent.

Fixes #485 

---

## Checklist
- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/instrument/apply_call.go
+++ b/tool/internal/instrument/apply_call.go
@@ -120,7 +120,7 @@ func (ip *InstrumentPhase) applyCallAppendArgs(
 		}
 	}
 
-	return true
+	return len(matchingCalls) > 0
 }
 
 // appendCallArgs appends the expressions from r.AppendArgs to the call's argument list.
--- a/tool/internal/instrument/apply_call_test.go
+++ b/tool/internal/instrument/apply_call_test.go
@@ -445,3 +445,29 @@ func TestMatchesCallRule_ImportAliasFromGopkgIn(t *testing.T) {
 
 	assert.True(t, matches)
 }
+
+func TestApplyCallAppendArgs_NoMatchReturnsFalse(t *testing.T) {
+	// A file with no matching calls should cause applyCallAppendArgs to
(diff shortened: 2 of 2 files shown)


y1yang0 (approved):
Good catch

NameHaibinZhang:
LGTM

NameHaibinZhang (approved):

PR #491, opened 2026-05-12
fix(instrument): propagate wrap errors from applyCallReplace

### LINKED ISSUE #490: applyCallReplace swallows wrap errors, returns false/nil instead of propagating failure
### Describe the bug

`applyCallReplace` in `tool/internal/instrument/apply_call.go` logs a warning when `compileExpression` fails for a matched call, but it never actually propagates the error. 

If matched calls fail to wrap, the function returns `false, nil` — which is indistinguishable from "no calls matched." Because of this, `applyCallRule` can't tell the difference between "nothing matched" and "everything matched but the wrapping broke." This leads to either a misleading assertion crash (`"call rule did not match any call"`) or a silent success with zero instrumentation actually applied.

### To Reproduce

Apply a call rule with a replace template that passes `newCallTemplate` but causes `compileExpression` to fail when evaluated against the matched AST node. 
`applyCallReplace` logs the warning but returns `false, nil`. The caller treats it as a clean no-match.

### Expected behavior

`applyCallReplace` should return an error if calls were matched but failed to wrap, rather than swallowing the error and returning `false, nil`.

### Additional context

The existing warning log is useful, but the wrapping failure needs to actually propagate as an error. Right now an empty `replacements` map is the only signal being checked, which creates this ambiguity.

I can open a PR shortly with the fix (collecting the wrap errors and returning them) and a regression test to validate it.

beep-boopp:
## Description

`applyCallReplace` now tracks errors from `compileExpression` failures during the AST walk. If all matched calls fail to wrap, the collected errors are returned instead of `false, nil`, so the caller can distinguish "nothing matched" from "everything matched but wrapping broke."

Added a regression test (`TestApplyCallRule_WrapFailureReturnsError`) that uses a template which passes `newCallTemplate` validation but fails at `compileExpression` — confirms `applyCallRule` now returns an error instead of silently succeeding.

## Motivation

Fixes #490

---

## Checklist
- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/instrument/apply_call.go
+++ b/tool/internal/instrument/apply_call.go
@@ -43,7 +43,7 @@ func (ip *InstrumentPhase) applyCallRule(ctx context.Context, r *rule.InstCallRu
 // applyCallReplace applies replacement wrapping to all matching calls in root using a
 // two-pass approach to avoid re-matching wrapped nodes.
 // Returns true if any replacement was made.
-func (ip *InstrumentPhase) applyCallReplace(
+func (*InstrumentPhase) applyCallReplace(
 	r *rule.InstCallRule,
 	root *dst.File,
 	importAliases map[string]string,
@@ -56,7 +56,11 @@ func (ip *InstrumentPhase) applyCallReplace(
 	// Pass 1: collect matching calls and pre-compute replacements to avoid
 	// re-matching the original call pointer inside its own wrapper.
 	replacements := make(map[*dst.CallExpr]dst.Expr)
+	var wrapError error
 	dst.Inspect(root, func(node dst.Node) bool {
+		if wrapError != nil {
+			return false
+		}
 		call, ok := node.(*dst.CallExpr)
 		if !ok {
 			return true
@@ -66,12 +70,16 @@ func (ip *InstrumentPhase) applyCallReplace(
 		}
 		wrapped, wrapErr := tmpl.compileExpression(call)
 		if wrapErr != nil {
-			ip.Warn("Failed to wrap call", "error", wrapErr)
-			return true
+			wrapError = wrapErr
+			return false
 		}
 		replacements[call] = util.AssertType[dst.Expr](dst.Clone(wrapped))
 		return true
 	})
+
+	if wrapError != nil {
+		return false, ex.Wrapf(wrapError, "failed to wrap matched call")
+	}
 
 	if len(replacements) == 0 {
 		return false, nil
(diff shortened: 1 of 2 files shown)


tool/internal/instrument/apply_call.go
```diff
@@ -67,13 +68,17 @@ func (ip *InstrumentPhase) applyCallReplace(
...
 			return true
 		}
 		replacements[call] = util.AssertType[dst.Expr](dst.Clone(wrapped))
 		return true
 	})
 
 	if len(replacements) == 0 {
+		if len(wrapErrors) > 0 {
+			return false, ex.Newf("all %d matched call(s) failed to wrap: %v", len(wrapErrors), ex.Join(wrapErrors...))
+		}
```
y1yang0: It seems that some compileExpression failures are now being treated as successes, i.e., they don't return an error?

beep-boopp: oh so I see it, the fix basically only resolved a complete failure. In the event of a partial one we simply drop the failed wraps.
Would u prefer a warning log for partial failures of should kft error out entirely?

beep-boopp: For partial failures, I am considering adding this right after the total failure check, before Pass 2
if len(wrapErrors) > 0 {
    ip.Warn("Some matched calls failed to wrap", "succeeded", len(replacements), "failed", len(wrapErrors), "errors", ex.Join(wrapErrors...))
}

y1yang0: Maybe we can fast-fail when we see compileExpression failure? Given that compileExpression failure is a critical bug

```
	replacements := make(map[*dst.CallExpr]dst.Expr)
	var wrapError error
	dst.Inspect(root, func(node dst.Node) bool {
		call, ok := node.(*dst.CallExpr)
		if !ok {
			return true
		}
		if !matchesCallRule(call, r, importAliases) {
			return true
		}
		wrapped, wrapErr := tmpl.compileExpression(call)
		if wrapErr != nil {
			wrapError = err 
			return false // fast-fail here
		}
		replacements[call] = util.AssertType[dst.Expr](dst.Clone(wrapped))
		return true
	})
    if wrapError !=nil {
			return false, ex.Newf("all %d matched call(s) failed to wrap: %v", len(wrapErrors), ex.Join(wrapErrors...))
	if len(replacements) == 0 {
		...
		}
```

beep-boopp: Makes sense. Fast-fail is quicker and cleaner. I'll update the pr shortly.

y1yang0 (approved):
LGTM. Thanks for doing this.

PR #492, opened 2026-05-12
fix(test): teach golden test infra to resolve non-stdlib rule imports

gyanranjanpanda:
This PR extends call-wrapping instrumentation rules to support a `path` field so call-rule templates can reference wrapper/helper functions supplied from auxiliary Go files, and updates the relevant golden test data and documentation.

### Why this is needed
In complex call-rule setups, we want to specify a custom `path` in our rule rules so the golden test utility can correctly resolve and load the non-stdlib helper modules.

### What has been done
- Added `Path` field to `InstCallRule` struct and added a YAML round-trip unit test to verify serialization/deserialization.
- - Integrated the `Path` field into the instrumentation phase so that it loads `//go:build ignore` helper files from the call-rule `path` into the target compilation, mimicking the behavior of file-based rules.
- - Updated golden test fixtures for the call-rule-only scenario, moving the `Wrapper` function into an auxiliary `hook.go` file to match this pattern.
- - Enhanced the `golden_test.go` test suite to locate external rule source directories using the package path if they are not present in the local workspace, improving the robustness of the tests.

### DIFF
--- a/tool/internal/instrument/instrument_test.go
+++ b/tool/internal/instrument/instrument_test.go
@@ -78,7 +78,7 @@ func runTest(t *testing.T, testName string) {
 	ruleSet := loadRulesYAML(t, testName, sourceFile)
 	writeMatchedJSON(ruleSet)
 
-	args := compileArgs(tempDir, sourceFile)
+	args := compileArgs(t, tempDir, sourceFile, ruleSet)
 	err := Toolexec(ctx, args)
 
 	if testName == invalidReceiver {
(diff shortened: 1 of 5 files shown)


gyanranjanpanda:
@y1yang0  could u review this waiting for your feed back

amazingakai (commented):
Duplicate of #482. Please also see my comment [there](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/482#issuecomment-4427812012) and [on the original issue](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/475#issuecomment-4402774316).

kakkoyun:
> Duplicate of #482. Please also see my comment [there](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/482#issuecomment-4427812012) and [on the original issue](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/475#issuecomment-4402774316).

@gyanranjanpanda This comment is important and needs to be addressed.

gyanranjanpanda:
Hi @amazingakai @kakkoyun — I've completely reworked this PR based on your feedback.

Instead of adding a path field to InstCallRule, I've fixed the golden test infrastructure so that call rules can properly use the existing imports mechanism for non-stdlib helper packages — which was the actual gap you identified.

What changed:

Extended createImportCfg to resolve rule-declared imports (not just stdlib) via go list
Added a new call-rule-local-import golden test that demonstrates the correct pattern: imports + qualified call (wrapper.Sizeof({{ . }}))
Added a testhelper/wrapper package as a reusable test helper
All 29 integration tests pass, including the new one
This is a different (and I believe more complete) approach than #482 — that PR added path support to call rules, which you flagged as blurring rule semantics. This PR instead fixes the underlying infra gap so helpers work through imports as intended.

Happy to address any further feedback!

amazingakai:
@gyanranjanpanda This was already addressed in #496

gyanranjanpanda:
Thanks for the clarification @amazingakai. Since there are similar PRs already, should I close this one, or is there any further improvement/change you’d suggest I work on here?

amazingakai:
@gyanranjanpanda Probably yes, but since #475 is still open, let's also wait for Yi Yang's response there first.

PR #495, opened 2026-05-12
fix(tool): refresh matched rules after dependency graph changes

### LINKED ISSUE #494: Building projects with older grpc versions than instrumentation dependencies causes build failures
### Describe the bug

Building a project that depends on an older `google.golang.org/grpc` version than the one used by `pkg/instrumentation/grpc` can fail during instrumentation.

`otelc` setup initially matches rules against the user's current dependency graph. Later, after instrumentation dependencies are added and go mod tidy runs, the dependency graph may change (for example grpc being bumped).

However, the previously matched rules are still stored using source file paths from the old dependency graph. During instrumentation, this causes stale file paths to be used against newer compile targets and results in build failures.

### Environment

 - OS: Fedora Linux
 - Go Version: 1.25.9
 - Version: main

### To Reproduce

1. Create a project using `google.golang.org/grpc` older than v1.80.0 (version from pkg/instrumentation/grpc)
2. Run `otelc go build`
3. Build will fail

### Expected behavior

Instrumentation should refresh dependency matching after syncing instrumentation dependencies so matched source file paths always correspond to the final resolved dependency graph.

### Additional context

This happens because `pkg/instrumentation/grpc` introduces newer grpc dep during setup, causing `go mod tidy` to update the dependency graph after rule matching has already completed.

amazingakai:
## Description

Matches rules in multiple setup passes to ensure matched rules are refreshed after dependency graph of a project changes

## Motivation

Fixes build failures caused by instrumentation dependencies updating the user's dependency graph (via go mod tidy), which could make previously matched rules stale

Fixes #494

---

## Checklist

- [x] Tests pass: `make test`


amazingakai:
This slows down the setup time a bit because it it has run the setup in 2-3 passes now. I will try to investigate this a bit more to see if I missed something that could be optimized but I think this is reasonable. Integration tests on windows are failing due to timeout, we might need to increase the timeout for integration tests

amazingakai:
Done. Since first pass always added otel as a dependency, it was trying to instrument that as well which slowed down the build time.

kakkoyun:
@amazingakai This needs a rebase. Is it ready to review otherwise?

amazingakai:
@kakkoyun Done. Ready for review :)

txabman42 (commented):
This PR solves the problem.

However, it adds a patch over a design (root) problem + extra complexity. I prefer to evaluate the problem and see if we can solve it without patches.

Current approach:
```
findDeps  →  matchDeps   →  syncDeps (mutates go.mod)  →  store
   ^                              |
   |   paths captured here        |   graph changes here
   +------------------------------+
                stale
```

Some ideas:

### Option 1
Easy + solves the problem --> not definitive design

Do `go mod tidy` before the `findDeps`.  We can following step before `findDeps`: `pinGraph` which adds all rule replaces, then go mod tidy.

This solves the problem, as the build plan from `findDeps` see final graphs with the replaces included.
Cons --> we are adding replaces for all instrumentation packages to our `go.mod`. Not a performance drama, but you know, not the cleanest solution.

### Option 2
Requires more refactor --> clean design

Setup phase --> remove the absolute paths from `match.json`, just keep the relation between the rule name, the target and the symbol to be instrumented. InstRuleSet would look like:

```go
type InstRuleSet struct {
    ImportPath  string
    PackageName string
    FuncRules   []*InstFuncRule    // no path keys
    StructRules []*InstStructRule
    CallRules   []*InstCallRule
    ...
}
```

We avoid finding go sources at setup phase.

Toolexec phase --> we need to make generic the InstFuncRule approach (`AddCallRule`) --> check per package file if the rule matches

So:
- setup --> matches a rule to a target package
- toolexec --> check per package file if the rule applies there

No one knows better than the compiler about the final files to be used.

amazingakai:
1. I don't think option 1 would scale very well as our integrations grow. Running `go mod tidy` upfront could become increasingly expensive since it may try to download and resolve dependencies for all instrumentation packages even if most of them are not actually needed for the current build.

2. I actually tried something similar to option 2, but I think there are a few more concerns there as well. After bumping a dependency version, the originally matched function may no longer exist or its signature may have changed. So I think the whole matched ruleset becomes invalidated, not just the resolved file paths. I think this would effectively require us to move the actual rule matching into the instrumentation/toolexec phase and keep setup limited to matching based only on targets/packages.

I just realized one possible improvement in current PR could be:
* detect whether one of the instrumented dependencies was actually bumped during `syncDeps`
* only rerun setup in that case

That way, in many cases we would still only run a single setup pass. What do you think?

txabman42:
1. I agree, it's not the best solution in the long term
2. > I think this would effectively require us to move the actual rule matching into the instrumentation/toolexec phase and keep setup limited to matching based only on targets/packages.

    That's what I was proposing. Just as an idea, but maybe worth exploring. Things like"originally matched function may no longer exist" shouldn't be a problem. If the rule matches X version, then we ensure it will work.

> detect whether one of the instrumented dependencies was actually bumped during syncDeps

Yes, that would be a good improvement for the current solution. Take the Option 2 as a non-blocking idea, I'm interested in defining if there could be a better design for this problem. Probably worth a new issue.

amazingakai:
> Things like "originally matched function may no longer exist" shouldn't be a problem. If the rule matches X version, then we ensure it will work.

Yes, but then the version would also need to be checked during the instrument phase. I think the current design would still be more efficient with `isSetup` implemented, since the instrument phase can operate only on already matched rules per file instead of trying to match every rule in the ruleset one by one for all files.

> Yes, that would be a good improvement for the current solution.

Done.

txabman42:
>  I think the current design would still be more efficient with isSetup implemented, since the instrument phase can operate only on already matched rules per file instead of trying to match every rule in the ruleset one by one for all files.

- Toolexec already filters by `-p` before doing any AST work, so packages without rules should cost nothing.
- For the matched packages, the AST is even smaller, as AST just gets the files from the compile, not the whole package as in setup
- `isSetup` becomes cleaner, instead of trying to answer "_do all these absolute paths still exist_" it answer "_does the rule still match the resolved go.mod_"?
- The proposed solution would imply parsing more files per instrumented package. However, aggregate parsing is roughly the same as the current setup work. This approach should be moving the load from setup to toolexec.

The required refactor is not small, but I still think it's an idea to explore. Out of this PR scope.

amazingakai:
Interesting direction. We should probably document the pros/cons of both approaches and see which one ends up cleaner/more efficient overall. Definitely feels worth exploring. Do you want to open an issue for it or should I?

txabman42:
> We should probably document the pros/cons of both approaches and see which one ends up cleaner/more efficient overall.

Yes, that it. Lets explore if its worth a refactor or not.

@amazingakai feel free to open an issue for that TY

darccio:
@amazingakai Can we rebase this PR?

amazingakai:
@darccio #584 fixes the issue, so I'll close this PR once #584 is merged.

kakkoyun:
@amazingakai @txabman42 checking in on #494. #584's pin-stage `go mod tidy` is meant to fix it (Option 1 from this thread), and this PR also closes #494, so there are two closers for one issue. Since #584 is mid-rebase and getting split, I'd lean toward pulling the focused #494 fix out as a small standalone PR so it can land on its own. If #584's pin path is the chosen fix instead, let's close this one. Either way, let's pick one. Note this branch is about 4 weeks stale and now conflicting.

amazingakai:
I prefer #584's pin-stage approach as the fix, so I'll close this PR.

PR #496, opened 2026-05-12
chore(test): add call rule external wrapper test

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/475


tool/internal/instrument/testdata/golden/call-rule-external-wrapper/rules.yml
```diff
@@ -0,0 +1,6 @@
+wrap_sizeof_external:
+  target: main
+  function_call: unsafe.Sizeof
+  replace: "wrapper.Wrapper({{ . }})"
+  imports:
+    wrapper: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/internal/instrument/testdata/golden/call-rule-external-wrapper/helpers/wrapper"
```
amazingakai: Should we add more rule types here instead of just `function_call`. Since every rule's apply func calls `ip.addRuleImports` manually, I think it is worth testing for all.

amazingakai (approved):
Looks good, thanks for taking care of this!

tool/internal/instrument/testdata/golden/call-rule-external-wrapper/rules.yml
txabman42: Good idea, added 50832c5978d1e069da354bf6aa169fdbeab2e5ef

kakkoyun (approved):
LGTM

PR #499, opened 2026-05-13
fix(cli): wire --debug flag to logger level 

### LINKED ISSUE #497: --debug flag defined but never wired to logger level
### Describe the bug

The `--debug` / `-d` flag is registered in the CLI and shows up in `--help`, but `initLogger` never reads it. The log handler is hardcoded to `slog.LevelInfo` regardless of whether `--debug` is passed.

All 13 `Debug()` calls across `setup/match.go`, `setup/find.go`, `instrument/toolexec.go`, and `instrument/match.go` are silently filtered out. There's no way to enable debug output from the CLI.

Additionally, all log output (including debug) only goes to `debug.log` in the work directory — even if the flag worked, a user would have to dig through a temp directory to see it.

### To Reproduce

```bash
otelc --debug go build ./...
# vs
otelc go build ./...
# identical behavior — debug flag has no effect
```

`initLogger` at `tool/cmd/main.go:121` hardcodes the level:

```go
handler := slog.NewTextHandler(writer, &slog.HandlerOptions{
    Level: slog.LevelInfo,  // never changes, --debug is not read
})
```

The flag has actually never worked — it was originally introduced in #50 as `"-debug"` (extra dash, would render as `---debug`). #281 fixed the typo but didn't catch that it was never wired to the logger.

### Expected behavior

`otelc --debug` should set the logger to `slog.LevelDebug` so existing `Debug()` calls actually produce output. Debug output should also be teed to stderr so it's visible in the terminal, consistent with how `go build -x` and other Go toolchain tools surface verbose output.

### Additional context

Stumbled across this while looking at CLI output behavior. All the debug instrumentation is already written — rule source resolution, CGO mappings, matched rule sets, import additions — it's just unreachable.

I can open a PR shortly with the fix.

beep-boopp:
## Description

The `--debug` flag is defined in the CLI but `initLogger` never reads it — the log level is hardcoded to `slog.LevelInfo` regardless. This means all `Debug()` calls across setup and instrument phases are silently filtered out.

This wires `cmd.Bool("debug")` to set `slog.LevelDebug` when the flag is active, Debug output goes to debug.log in the work directory, consistent with existing logging behavior.

## Motivation

The flag has never actually worked, it was introduced in #50 as `"-debug"` (triple dash), #281 fixed the typo but the wiring to the logger was never added. There are 13 existing `Debug()` calls across the codebase that are currently unreachable.

Fixes #497

---

## Checklist
- [x] Tests pass: `make test`

### DIFF
--- a/tool/cmd/main.go
+++ b/tool/cmd/main.go
@@ -38,6 +38,7 @@ func main() {
 			&cli.BoolFlag{
 				Name:    "debug",
 				Aliases: []string{"d"},
+				Sources: cli.EnvVars(util.EnvOtelcDebug),
 				Usage:   "Enable debug mode",
 				Value:   false,
 			},
@@ -111,21 +112,29 @@ func initLogger(ctx context.Context, cmd *cli.Command) (context.Context, error) 
 	}
 
 	logFilename := filepath.Join(buildTempDir, debugLogFilename)
-	writer, err := os.OpenFile(logFilename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
+	logFile, err := os.OpenFile(logFilename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
 	if err != nil {
 		return ctx, ex.Wrapf(err, "failed to open log file %q", logFilename)
 	}
 
+	level := slog.LevelInfo
+	if cmd.Bool("debug") {
+		level = slog.LevelDebug
+		if setErr := os.Setenv(util.EnvOtelcDebug, "1"); setErr != nil {
+			return ctx, ex.Wrapf(setErr, "set %s", util.EnvOtelcDebug)
+		}
+	}
+
 	// Log timestamps and levels are omitted: they add noise when correlating
 	// with Go toolchain output and the log file is for human debugging only.
-	handler := slog.NewTextHandler(writer, &slog.HandlerOptions{
+	handler := slog.NewTextHandler(logFile, &slog.HandlerOptions{
 		ReplaceAttr: func(_ []string, a slog.Attr) slog.Attr {
 			if a.Key == slog.TimeKey || a.Key == slog.LevelKey {
 				return slog.Attr{}
 			}
 			return a
 		},
-		Level: slog.LevelInfo,
+		Level: level,
 	})
 	logger := slog.New(handler)
 	ctx = util.ContextWithLogger(ctx, logger)
--- a/tool/cmd/main_test.go
+++ b/tool/cmd/main_test.go
@@ -1,0 +1,97 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"context"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/util"
+	"github.com/urfave/cli/v3"
+)
+
+func TestInitLogger(t *testing.T) {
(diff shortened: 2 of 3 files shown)


tool/cmd/main.go
```diff
@@ -111,21 +112,28 @@ func initLogger(ctx context.Context, cmd *cli.Command) (context.Context, error)
...
 	logFilename := filepath.Join(buildTempDir, debugLogFilename)
-	writer, err := os.OpenFile(logFilename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
+	logFile, err := os.OpenFile(logFilename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
 	if err != nil {
 		return ctx, ex.Wrapf(err, "failed to open log file %q", logFilename)
 	}
 
+	level := slog.LevelInfo
+	var w io.Writer = logFile
+	if cmd.Bool("debug") {
```
y1yang0, line 121: It seems the `-debug` flag is not passed to instrument phase. So, does this mean the flag only works in the setup phase and is actually ineffective in the instrument phase?"

beep-boopp: oh hey good catch , missed that, the subprocess inherits env vars but not the cli flags. I'll follow the same pattern as --stats(set OTELC_DEBUG=1 in the parent and read it in initLogger ) . Pushing the fix shortly

beep-boopp:
Hey @y1yang0 just a heads up with the stderr tee, --debug surfaces a pretty verbose existing log in find.go:122 that dumps the full build plan as a single field. Makes the terminal output noisy. Should I drop the stderr tee and keep this to just the level fix, or is the tee worth keeping?

y1yang0:
I prefer keeping it as the fix only

tool/cmd/main.go
y1yang0, line 42: No `Sources: cli.EnvVars(...)`?

beep-boopp:
I have added Sources: cli.EnvVars(...) and simplified the os.Getenv check. Integration test failure is a pre-existing timeout on Windows (same 600s timeout reproduces on main without my changes, tested that locally).

tool/cmd/main.go
```diff
@@ -111,21 +112,29 @@ func initLogger(ctx context.Context, cmd *cli.Command) (context.Context, error)
...
+		level = slog.LevelDebug
+		if setErr := os.Setenv(util.EnvOtelcDebug, "1"); setErr != nil {
+			return ctx, ex.Wrapf(setErr, "set %s", util.EnvOtelcDebug)
+		}
+	}
+
 	// Log timestamps and levels are omitted: they add noise when correlating
 	// with Go toolchain output and the log file is for human debugging only.
-	handler := slog.NewTextHandler(writer, &slog.HandlerOptions{
+	handler := slog.NewTextHandler(logFile, &slog.HandlerOptions{
```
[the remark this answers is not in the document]
beep-boopp: About this, we had earlier come to an agreement that we would scope this pr down to just the fix since the tee to stderr was getting a little too verbose...i will update the pr title and description to match the changes committed

beep-boopp:
Hey @y1yang0  all review feedback has been addressed, PTAL when you have the time thank you.

y1yang0:
Please fix test failure

beep-boopp:
Hey @y1yang0  so  i added unit tests for initLogger covering default level, debug flag , env var propagation and cli.envVars sourcing. Remaining codecov gap is error return paths....that i cant trigger without mocking stdlib but atleast integration tests are passing now.

do let me know if any changes are to be made....

y1yang0 (approved):

PR #501, opened 2026-05-13
feat(gin): add gin-gonic/gin server instrumentation

### LINKED ISSUE #479: Instrumentation Request for gin-gonic/gin

- [ ] I have searched the existing issues

### Package

github.com/gin-gonic/gin

### Version

v1.10.0 (current stable; needs >= v1.4.0 for FullPath())

### Environment

Go HTTP servers, mostly containerized. Gin is pretty widely used this would cover a lot of ground.

### Use-case

REST APIs built with gin. The net/http hook right now gives spans like "GET" with url.path=/users/42 but there's no way to get the actual route template. With gin, c.FullPath() returns /users/:id which is what you do want as http.route for grouping requests properly in dashboards. Without it every unique URL becomes its own series.
The hard part is that c.FullPath() is only populated after gin's router matches the request so a gin-specific hook is needed on top of net/http.

### Telemetry

Server spans following HTTP semconv v1.37.0, same attributes as the existing nethttp server hook http.request.method, http.response.status_code, url.path, server.address, etc. The key addition would be http.route from c.FullPath().

Span name would go from just "GET" to "GET /users/:id" after the route is resolved. 5xx responses would set span status to Error. Metrics deferred to a follow-up.

### Anything else?

While looking into this I checked how loongsuite and dd-trace-go handle gin:

- loongsuite (alibaba/loongsuite-go-agent/pkg/rules/gin/) hooks (*Context) Next and uses GLS to find the span. Since otelc uses context.Context instead of GLS, that approach doesn't directly apply here.
- dd-trace-go (DataDog/dd-trace-go/contrib/gin-gonic/gin/) adds a runtime middleware but that requires the user to change their code.

What I'm thinking is two hooks in one YAML file (similar to how redis/v9/v9.yaml has multiple rules):
1. (*Engine).ServeHTTP start the span, wrap ResponseWriter, pass context through r.WithContext(ctx)
2. (*Context).Next at this point routing is done, so read c.FullPath() and update the span name

I've been experimenting locally and have something that seems to work with basic unit tests. Happy to share it once we agree on the approach
Also not sure if this needs an ADR since it's the first case of two hooks coordinating through request context.

### Tip

waqar2403:
## Description

Adds compile-time instrumentation for `gin-gonic/gin` HTTP servers using a single-hook enrichment approach.

Instead of creating a separate span, we hook `(*gin.Context).Next` and enrich the span already created by the `net/http` server hook. By the time `Next` is called, gin's router has matched the request and populated `c.FullPath()`, so we update the span name from `"METHOD"` to `"METHOD /route/pattern"` and record the `http.route` attribute.

This keeps the telemetry output aligned with `otelgin` conventions: same span name format, same route attribute, one span per request.

Key design decisions:
- **Single hook only**: `net/http`'s `serverHandler.ServeHTTP` already fires for gin (since `gin.Engine` implements `http.Handler`), so no need for a second span-creation hook
- **Idempotency guard**: `c.Next()` is called by each middleware in the chain, so the hook uses `c.Set`/`c.Get` to ensure the span is only updated once per request
- **Empty route handling**: If no route matched (404), the span name stays as just the HTTP method

Resolves #479

## Checklist

- [x] Tests pass: `make test`


NameHaibinZhang (commented):
## Review Summary

Nice approach — enriching the existing `net/http` span instead of creating a duplicate is the right call. The idempotency guard via `c.Set`/`c.Get` handles the multi-middleware `c.Next()` problem cleanly.

### Issues to address

1. **Integration test hardcodes port 8080** — Unlike `http_server_test.go` which parameterizes the port, the gin test uses a fixed port. This can cause flaky failures if the port is occupied. Suggest passing `-port=<unused>` like other tests:
   ```go
   f.BuildAndStart("ginserver", "-port=8090")
   ```

2. **No integration test for the 404/no-route case** — The unit test covers empty route, but an end-to-end test verifying that a request to an unregistered path still produces a span named just `"GET"` (without route pattern) would strengthen regression protection.

3. **Heavy transitive dependency tree** — The `go.mod` pulls in prometheus, grpc-gateway, autoexport, and many OTLP exporters indirectly. For a hook that only needs `otel/trace` + `otel/semconv`, this seems excessive. This may be inherited from `shared`/`pkg` and not specific to this PR, but worth investigating whether the dependency graph can be trimmed.

### Nits

- PR scope in title: `feat(instrumentation)` is broad — `feat(gin)` would be more specific.
- Consider documenting `"otel.gin.route.set"` as a reserved key in gin context, to avoid potential collision with user middleware.

### Overall

Solid contribution. The design is sound, tests cover the key scenarios, and it aligns well with `otelgin` conventions. Addressing the hardcoded port should be straightforward. 👍

waqar2403:
Thanks @NameHaibinZhang for the review. Here are the fixes that I have made.

- Reordered the gate in BeforeNext so routeSetKey is set only after the IsRecording check passes. Earlier a non recording first call would burn the gate and block any later recording span on the same request from being enriched. Rare in practice but added TestBeforeNext_NonRecordingSpanDoesNotBurnGate to lock it in.
- Changed the four OTEL_GO_ENABLED_INSTRUMENTATIONS and OTEL_GO_DISABLED_INSTRUMENTATIONS values from "GIN" to "gin" to match the doc on shared.Instrumented and the nethttp grpc redis tests. Both work since parsing lowercases anyway, this is just for consistency.
- Passing -port=8090 to ginserver in the integration test instead of relying on the default 8080. Same pattern as http_server_test.go.
- Updated the godoc on routeSetKey to mark it as reserved so user middleware does not collide. The "otel." prefix already signalled it but explicit is better.
- Renamed the PR title to feat(gin) as suggested.
- Added TestGinServer_UnregisteredRoute for the no route case. It hits an unregistered path and asserts the span name stays as plain GET http.route is not set and url.path is the literal path. Mainly a cardinality guard so a future regression cannot turn every scanner hit into a unique span name.

- On the dependency tree I am looking into that separately and will share my findings once done.

let me know if anything else needs change.

pkg/instrumentation/gin/server/server.yaml
```diff
@@ -0,0 +1,6 @@
+gin_context_next_hook:
+  target: github.com/gin-gonic/gin
+  func: Next
+  recv: "*Context"
+  before: BeforeNext
```
txabman42: Could be interesting to have an `AfterNext` to catch all the `c.Errors()`, similar to [go-contrib approach](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ba143fc95a5e03e235d699721b8e9c62e3bfa2a7/instrumentation/github.com/gin-gonic/gin/otelgin/gin.go#L121)

txabman42 (changes requested):

pkg/instrumentation/gin/server/server.yaml
waqar2403: Hi @txabman42  
Added `AfterNext` hook that captures `c.Errors()` after `(*gin.Context).Next` returns — sets span status to Error and records each error as an exception event via `span.RecordError()`, matching the otelgin approach. Uses the same flag-based idempotency as `BeforeNext` to handle multiple middleware calling `c.Next()`. Also added a `/error` route to the test app and `TestGinServer_GinError` integration test that exercises `c.Error()` on a 200 OK response to verify gin-level errors are captured independently of HTTP status.

txabman42: As you correctly did in the BeforeNext, we need to handle in `AfterNext` the case where multiple middlewares call `c.Next()`.

We can count in the BeforeNext the number of middlewares and use it to iterate over all errors in the outer middleware.

waqar2403: Switched from the flag based approach to a depth counter as you suggested. BeforeNext increments the counter for each c.Next() call and AfterNext decrements it, only recording errors when the count reaches zero at the outermost middleware return. This way errors added by middleware after c.Next() returns are also captured.

waqar2403:
@txabman42 I have improved the gin integration tests so they share a single build of the test app instead of building it 4 times. This saves about 90 seconds on the Windows job and keeps the suite well inside the ten minute timeout.

test/integration/gin_server_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/test/testutil"
+)
+
+func TestGinServer(t *testing.T) {
+	build := testutil.NewTestFixture(t, testutil.WithoutCollector())
```
txabman42: No needed, f := testutil.NewTestFixture(t) is done inside each run

txabman42 (approved):
LGTM! ❤️ 

Just a minor nitpick

test/integration/gin_server_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/test/testutil"
+)
+
+func TestGinServer(t *testing.T) {
```
txabman42, line 21: Regarding i[ntegration test fail](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/26028717488/job/76621954931?pr=501):

Couple of ideas to evaluate:
- We are not seeing why the app crashed, it seems `cmd.Stdout` is nil (defined in Start).
- Why can verify if the port is already binded during start. If so, we can improve the cleanup phase so it doesn't happen or use different ports.

txabman42 (changes requested):

test/integration/gin_server_test.go
waqar2403: Thanks for helping. I need a bit of time to figure out the right approach. Will push an update soon.

waqar2403: Sorry for the late reply @txabman42  got caught up with a few things !!!
Fixed:
- cmd.Stdout nil: Start() now captures stdout/stderr into a bytes.Buffer logs it via t.Logf on test failure and calls cmd.Wait() after Kill() to flush everything. No more crashes.
- Port reuse: Each test case gets its own port (8090-8093) avoids port reuse between subtests.
Also removed the outer build fixture per your other comment each subtest calls BuildAndStart independently now.

waqar2403:
@txabman42 The port fix and stdout capture both work all 4 gin subtests pass on Windows CI.

The remaining CI failure is a timeout. I followed the BuildAndStart-per-subtest convention (matching grpc_server_test, http_server_test, etc.) but 4 instrumented builds at ~38s each adds 151s on Windows. The existing suite already takes ~474s so the total exceeds the 10-minute limit and TestGRPCClient gets killed.

I think the simplest fix is bumping the Makefile timeout from 10m to 15m. This is a scaling issue that'll come up as more integration tests get added anyway. Can do this if you agree or open to other ideas.

txabman42:
@waqar2403 lets reduce the integration test times https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/540

pkg/instrumentation/gin/server/gin.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/shared"
+)
+
+const instrumentationKey = "GIN"
+
+var logger = shared.Logger()
+
+// ginEnabler controls whether gin instrumentation is enabled.
+type ginEnabler struct{}
```
txabman42: Although the idea could be good, I'd remove this. In case we want an ON/OFF lever, we'd need to think in a common approach for all instrumentation pkgs.

kakkoyun:
@waqar2403 This needs a rebase after #540 is merged.

waqar2403:
> @waqar2403 This needs a rebase after #540 is merged.

@kakkoyun Done, rebased on top of [#540](https://github.com/waqar2403/opentelemetry-go-compile-instrumentation/issues/540) and adopted the TestMain/FreePort pattern.

pkg/instrumentation/gin/server/gin.go
waqar2403: Removed

NameHaibinZhang (approved):

NameHaibinZhang:
LGTM

txabman42 (approved):
LGTM! Well done ❤️

PR #502, opened 2026-05-13
fix(setup): warn when instrumentation hooks bump consumer versions

### LINKED ISSUE #489: setup: silently bumps consumer's go version and OTel dependencies
## Problem

`setup` phase silently bumps the consumer's `go` directive and OTel dependency versions.

Nothing in the setup output tells the user this happened, and since `go.mod` is restored by cleanup after a successful build, the user may never notice. We should advise users which versions are going to be used during the build.

The root cause is that every hook module in `pkg/instrumentation/**/go.mod` pins `go 1.25.0` and `otel v1.43.0`, the maintainer's current dev version, not the minimum required. When `syncDeps` adds `replace` directives and runs `go mod tidy`, Go's MVS raises the consumer to match.

## Proposed fix

**A.** Lower the declared floor in hook modules to the minimum `go` version and OTel version the hook code actually requires (likely `go 1.23` / `otel v1.38`). Add a CI matrix to prevent the floor from silently drifting back up.

**B.** In `syncDeps`, before running `go mod tidy`, compare the consumer's `go.mod` against each matched hook and emit a visible warning listing any version that would be raised (old → new, which hook forced it).

gyanranjanpanda:
## Problem

The `syncDeps` function silently bumps the consumer's `go` directive and OTel dependency versions during setup. Because `go.mod` is restored by cleanup after a successful build, the user never notices these version changes happened.

## Solution

This implements **Part B** of the proposed fix from #489: before running `go mod tidy`, we snapshot the consumer's `go` directive and direct dependency versions. After tidy, we compare and emit `Warn`-level log messages for every version that was raised.

### Example output

```
WARN go directive bumped by instrumentation hooks from=1.22.0 to=1.25.0
WARN dependency version bumped by instrumentation hooks module=go.opentelemetry.io/otel from=v1.38.0 to=v1.43.0
```

### `tool/internal/setup/sync.go`

| Symbol | Purpose |
|--------|---------|
| `versionSnapshot` | Struct holding `goVersion` and `deps map[string]string` |
| `snapshotVersions()` | Captures pre-mutation state from the parsed `modfile.File` |
| `warnVersionBumps()` | Re-parses go.mod post-tidy, compares against snapshot, emits warnings |

**Design decisions:**
- Go directives use bare versions (`1.25.0`) while `semver.Compare` needs a `v` prefix -- we prepend `v` before comparison
- - Only **direct** (non-indirect) dependencies are tracked to avoid noise
- - `warnVersionBumps` is non-fatal: parse errors are logged as warnings, not returned as errors
- - Uses `semver.IsValid` guards to skip malformed versions gracefully
### `tool/internal/setup/sync_test.go`

**6 new tests** covering:
- `TestSnapshotVersions` -- direct deps captured, indirect excluded
- - `TestSnapshotVersions_MinimalGoMod` -- empty deps, go version captured
- - `TestWarnVersionBumps_GoVersionRaised` -- go directive bump produces warning
- - `TestWarnVersionBumps_DepVersionRaised` -- OTel dep bump produces warning
- - `TestWarnVersionBumps_NoChange` -- no warnings when versions match
- - `TestWarnVersionBumps_MissingFile` -- graceful handling of missing go.mod
## What's NOT in this PR

**Part A** (lowering hook module floors from `go 1.25.0` / `otel v1.43.0` to true minimums) is a separate effort requiring CI matrix testing across Go versions.

Fixes #489

### DIFF
--- a/tool/internal/setup/sync.go
+++ b/tool/internal/setup/sync.go
@@ -5,6 +5,8 @@ package setup
 
 import (
 	"context"
+	"fmt"
+	goversion "go/version"
 	"os"
 	"path/filepath"
 	"strings"
@@ -71,14 +73,13 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error) 
 	return false, nil
 }
 
+// versionSnapshot records go directive and direct dep versions before tidy.
 type versionSnapshot struct {
 	goVersion string
-	deps      map[string]string // direct deps only
+	deps      map[string]string
 }
 
-// snapshotVersions records the go directive and direct dep versions
-// so we can detect what go mod tidy bumped.
-func snapshotVersions(mf *modfile.File) versionSnapshot {
+func snapshotVersion(mf *modfile.File) versionSnapshot {
 	snap := versionSnapshot{
 		deps: make(map[string]string),
 	}
@@ -93,42 +94,25 @@ func snapshotVersions(mf *modfile.File) versionSnapshot {
 	return snap
 }
 
-// warnVersionBumps diffs the go.mod after tidy against the pre-mutation
-// snapshot and logs any version that MVS raised.
-func (sp *SetupPhase) warnVersionBumps(goModPath string, before versionSnapshot, hookModules []string) {
+func (sp *SetupPhase) warnVersion(goModPath string, before versionSnapshot) {
 	after, err := parseGoMod(goModPath)
 	if err != nil {
-		sp.Warn("unable to check for version bumps after go mod tidy", "error", err)
+		sp.Error("unable to check for version bumps after go mod tidy", "error", err)
 		return
 	}
 
-	hooks := strings.Join(hookModules, ", ")
-
-	// go directives are bare ("1.25.0"), semver wants "v" prefix
+	// Go directives use Go toolchain syntax ("1.21"), not module semver.
 	if after.Go != nil && before.goVersion != "" {
-		oldGo := "v" + before.goVersion
-		newGo := "v" + after.Go.Version
-		if semver.IsValid(oldGo) && semver.IsValid(newGo) && semver.Compare(newGo, oldGo) > 0 {
-			sp.Warn("go directive bumped by instrumentation hooks",
-				"from", before.goVersion, "to", after.Go.Version,
-				"hooks", hooks)
+		if goversion.Compare("go"+after.Go.Version, "go"+before.goVersion) > 0 {
+			sp.Warn(fmt.Sprintf("Bumped go version (%s -> %s)", before.goVersion, after.Go.Version))
 		}
 	}
 
 	for _, req := range after.Require {
-		if req.Indirect {
-			continue
-		}
-		oldVer, tracked := before.deps[req.Mod.Path]
-		if !tracked {
-			continue
-		}
-		if semver.IsValid(oldVer) && semver.IsValid(req.Mod.Version) &&
-			semver.Compare(req.Mod.Version, oldVer) > 0 {
-			sp.Warn("dependency version bumped by instrumentation hooks",
-				"module", req.Mod.Path,
-				"from", oldVer, "to", req.Mod.Version,
-				"hooks", hooks)
+		if oldVer, tracked := before.deps[req.Mod.Path]; tracked && !req.Indirect {
+			if semver.Compare(req.Mod.Version, oldVer) > 0 {
+				sp.Warn(fmt.Sprintf("Bumped dependency %s (%s -> %s)", req.Mod.Path, oldVer, req.Mod.Version))
+			}
 		}
 	}
 }
@@ -153,10 +137,8 @@ func (sp *SetupPhase) syncDeps(ctx context.Context, matched []*rule.InstRuleSet,
 		return err
 	}
 
-	// Record pre-mutation versions to detect bumps after tidy.
-	before := snapshotVersions(modfile)
+	before := snapshotVersion(modfile)
 	replaces := make([]*replaceDirective, 0)
-	hookModules := make([]string, 0, len(rules))
 	for _, m := range rules {
 		util.Assert(strings.HasPrefix(m.Path, util.OtelcRoot), "sanity check")
 		oldPath := m.Path
@@ -168,7 +150,6 @@ func (sp *SetupPhase) syncDeps(ctx context.Context, matched []*rule.InstRuleSet,
 			newPath:    newPath,
 			newVersion: "",
 		})
-		hookModules = append(hookModules, oldPath)
 	}
 
 	// Add replace directive for special pkg module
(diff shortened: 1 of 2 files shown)


gyanranjanpanda:
could @txabman42 review this

tool/internal/setup/sync.go
```diff
@@ -148,6 +216,7 @@ func (sp *SetupPhase) syncDeps(ctx context.Context, matched []*rule.InstRuleSet,
 		if err != nil {
 			return ex.Wrapf(err, "running go mod tidy in %s", moduleDir)
 		}
+		sp.warnVersionBumps(goModFile, before, hookModules)
```
y1yang0: Please add oneline commet, as this is not very intuitive(self-explanatory)

tool/internal/setup/sync.go
```diff
@@ -70,6 +71,68 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+		if !req.Indirect {
+			snap.deps[req.Mod.Path] = req.Mod.Version
+		}
+	}
+	return snap
+}
+
+// warnVersionBumps diffs the go.mod after tidy against the pre-mutation
+// snapshot and logs any version that MVS raised.
+func (sp *SetupPhase) warnVersionBumps(goModPath string, before versionSnapshot, hookModules []string) {
```
y1yang0: I suggest:

- Remove hookModule, we only need to inform the user which dependencies in go.mod have been upgraded.
- Use consistent warning messages, for example:
```
"Bumped go version (v%s -> v%s)"
"Bumped dependency %s (v%s -> v%s)"
```
- Use consistent function names, like `snapshotVersion` and `warnVersion`

tool/internal/setup/sync.go
```diff
@@ -70,6 +71,68 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+	after, err := parseGoMod(goModPath)
+	if err != nil {
+		sp.Warn("unable to check for version bumps after go mod tidy", "error", err)
```
y1yang0: This is error, not a warning

tool/internal/setup/sync.go
```diff
@@ -70,6 +71,68 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+		newGo := "v" + after.Go.Version
+		if semver.IsValid(oldGo) && semver.IsValid(newGo) && semver.Compare(newGo, oldGo) > 0 {
+			sp.Warn("go directive bumped by instrumentation hooks",
+				"from", before.goVersion, "to", after.Go.Version,
+				"hooks", hooks)
+		}
+	}
+
+	for _, req := range after.Require {
+		if req.Indirect {
```
y1yang0: Maybe simply?
```
for req:= range after.Requires{ if oldVer,tracked:= before.deps[req.Mod.Path];tracked { .... } }
```

tool/internal/setup/sync.go
```diff
@@ -70,6 +71,68 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+			continue
+		}
+		oldVer, tracked := before.deps[req.Mod.Path]
+		if !tracked {
+			continue
+		}
+		if semver.IsValid(oldVer) && semver.IsValid(req.Mod.Version) &&
```
y1yang0: `semver.IsValid` make no sense?

gyanranjanpanda:
i adressed all issue @y1yang0  u mention could u verify this if any thing i  can change i am willing to do

tool/internal/setup/sync.go
```diff
@@ -70,6 +73,50 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+		}
+	}
+	return snap
+}
+
+func (sp *SetupPhase) warnVersion(goModPath string, before versionSnapshot) {
+	after, err := parseGoMod(goModPath)
+	if err != nil {
+		sp.Error("unable to check for version bumps after go mod tidy", "error", err)
+		return
```
y1yang0: I mean returning an error `return ex.Newf()` instead of just logging

tool/internal/setup/sync.go
```diff
@@ -70,6 +73,50 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+
+	// Go directives use Go toolchain syntax ("1.21"), not module semver.
+	if after.Go != nil && before.goVersion != "" {
+		if goversion.Compare("go"+after.Go.Version, "go"+before.goVersion) > 0 {
+			sp.Warn(fmt.Sprintf("Bumped go version (%s -> %s)", before.goVersion, after.Go.Version))
+		}
+	}
+
+	for _, req := range after.Require {
+		if oldVer, tracked := before.deps[req.Mod.Path]; tracked && !req.Indirect {
```
y1yang0: Checking `req.Indirect` makes no sense ? as snapshotVersion already checks that.

txabman42 (approved):
LGTM once @y1yang0 comments are resolved!

gyanranjanpanda:
i adressd the latest  @y1yang0  issue could u review this

tool/internal/setup/sync.go
```diff
@@ -70,6 +73,50 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+func (sp *SetupPhase) warnVersion(goModPath string, before versionSnapshot) error {
+	after, err := parseGoMod(goModPath)
+	if err != nil {
+		return ex.Wrapf(err, "unable to check for version bumps after go mod tidy")
+	}
+
+	// Go directives use Go toolchain syntax ("1.21"), not module semver.
+	if after.Go != nil && before.goVersion != "" {
+		if goversion.Compare("go"+after.Go.Version, "go"+before.goVersion) > 0 {
+			sp.Warn(fmt.Sprintf("Bumped go version (%s -> %s)", before.goVersion, after.Go.Version))
```
amazingakai, line 106: `sp.Warn` logs to `.otelc-build/debug.log`. I suggest we write these warnings to stderr.

tool/internal/setup/sync.go
```diff
@@ -70,6 +73,50 @@ func addReplace(modfile *modfile.File, replace *replaceDirective) (bool, error)
...
+		}
+	}
+
+	for _, req := range after.Require {
+		if oldVer, tracked := before.deps[req.Mod.Path]; tracked {
+			if semver.Compare(req.Mod.Version, oldVer) > 0 {
+				sp.Warn(fmt.Sprintf("Bumped dependency %s (%s -> %s)", req.Mod.Path, oldVer, req.Mod.Version))
```
amazingakai, line 113: Same^

tool/internal/setup/sync_test.go
```diff
@@ -225,3 +226,184 @@ go 1.21
...
+		goVersion: "1.21",
+		deps:      map[string]string{},
+	}
+
+	require.NoError(t, sp.warnVersion(gomodPath, before))
+
+	logged := buf.String()
+	assert.Contains(t, logged, "Bumped go version")
+	assert.Contains(t, logged, "1.21 -> 1.25.0")
+}
```
amazingakai, line 321: Repeated test?

gyanranjanpanda:
Thanks, fixed. I removed the repeated standalone Go language-version test and folded it into `TestWarnVersion_GoVersionRaised` as a table-driven case. This keeps coverage for both `1.22.0 -> 1.25.0` and `1.21 -> 1.25.0` while avoiding duplicate setup/assertions.

Verified with:

```sh
go test ./tool/internal/setup
could @amazingakai  verify now

y1yang0 (approved):

PR #506, opened 2026-05-15
feat(setup): warn when no instrumentation rules match

### LINKED ISSUE #505: feat(cli): warn when no instrumentation rules match during build
## Problem

I noticed this while working on the `--debug` flag fix (#499) and reading through the match pipeline.

When instrumentation rules match nothing, `otelc` completes successfully with no user-visible feedback. This results in an uninstrumented binary without any indication that rules were ineffective.

## Reproduction

1. Create a rule with an incorrect target (e.g. typo in package path — `net/htp` instead of `net/http`)
2. Run: `otelc --rules broken.yaml go build -o app .`

## Observed behavior

- Exit code 0
- No output on stdout/stderr
- `.otelc-build/matched.json` contains `[]`
- `debug.log` shows `"Found available rules"` → `"Stored matched sets"` with no indication that zero matches occurred

Even in debug mode, a zero-match outcome is only observable by the *absence* of `"Match ..."` log lines.

## Expected behavior

Users should receive an explicit warning when no rules match, since this commonly indicates misconfiguration (typos, version mismatches, renamed functions, outdated targets).

## Proposal

Emit a warning to stderr when `matchDeps` returns empty, e.g.:

 0 rules matched — no instrumentation will be applied.

This is a small, low-risk change that provides immediate feedback to users. Happy to open a PR.

This could also open the door to something like otelc list down the line,  a read-only command to preview what rules match without building, but that's a separate discussion.

beep-boopp:
## Description

Emit a warning to stderr when `matchDeps` returns zero matches. Previously, a build with no matching rules would succeed silently — no stdout, no stderr, and the only evidence lived in `debug.log` inside the work directory.

Now the user sees:
Warning: no instrumentation will be applied

## Motivation

Discovered while working on the `--debug` flag fix (#499). Even with debug logging wired up, there's no feedback when rules match nothing — the build just succeeds and produces an uninstrumented binary.

This is easy to hit in practice: a typo in a target path, a renamed function, a version range that no longer applies. The user only finds out when traces don't show up in their collector.

Fixes #505
---

## Checklist

- [x] Tests pass: `make test`


tool/internal/setup/setup.go
```diff
@@ -32,6 +34,14 @@ func (sp *SetupPhase) Error(msg string, args ...any) { sp.logger.Error(msg, args
 func (sp *SetupPhase) Warn(msg string, args ...any)  { sp.logger.Warn(msg, args...) }
 func (sp *SetupPhase) Debug(msg string, args ...any) { sp.logger.Debug(msg, args...) }
 
+// warnIfNoMatches emits a warning to stderr when no instrumentation rules matched
+func (sp *SetupPhase) warnIfNoMatches(matched []*rule.InstRuleSet, w io.Writer) {
+	if len(matched) == 0 {
+		_, _ = fmt.Fprintf(w, "Warning : 0 rules matched - no instrumentation will be applied.\n")
```
amazingakai: We can use `fmt.Fprint` here.

amazingakai (approved):
Looks good.

beep-boopp:
Hey @txabman42 PTAL at this whenever you have the time. As it stands the warning fires during setup (before the build). Let me know if placement works or if you'd prefer post-build.

tool/internal/setup/setup.go
txabman42:
```suggestion
		_, _ = fmt.Fprintf(w, "Warning: 0 rules matched - no instrumentation will be applied.\n")
```

txabman42 (approved):
LGTM! TY ❤️ 

Just fix those couple of nitpicks before merging

tool/internal/setup/setup.go
y1yang0: Maybe simply `Warning: no instrumentation will be applied`

tool/internal/setup/setup.go
```diff
@@ -188,6 +198,7 @@ func Setup(ctx context.Context, cmd *cli.Command) error {
 	if err != nil {
 		return ex.Wrapf(err, "matching dependencies to hook rules")
 	}
+	sp.warnIfNoMatches(matched, os.Stderr)
```
y1yang0: I don't think it's necessary to insert such a fine-grained function into the main flow (Setup). The main flow is expected to be a high-level overview of the major stages (e.g., find, extract, match).

Perhaps a better place for this logic would be at the end of matchDeps. This would also eliminate the need for a separate warnIfNoMatches function. For example:
```go
func (sp *SetupPhase) matchDeps(...) {
	....
	if len(matched)==0 {
		.... // print to stderr and log files
	}
	return matched, nil
}
```

beep-boopp:
On second thought it made sense..so refactored per @y1yang0's suggestion, moved the warning into matchDeps directly and simplified the message. @txabman42 PTAL when you get a chance.

y1yang0 (approved):

PR #509, opened 2026-05-15
feat(databasesql): add DSN parser registry with extensibility and tests.

### LINKED ISSUE #508: DSN parser lacks extensibility and silently produces empty addresses for unknown drivers
Severity: Low
pkg/instrumentation/databasesql/parse.go:33-44
Any driver not in this list (e.g., pgx, lib/pq, sqlserver, oracle) silently returns an empty address. The caller (DbClientRequestTraceAttrs) then emits server.address: "" — an empty required attribute — with no error or warning. As the mentorship adds new database integrations (MySQL demo #168, others), this list will need frequent manual updates, making it a maintenance bottleneck.

Expected Behavior

Unknown drivers should return an explicit error or a best-effort parse, and the caller should log a warning when the address cannot be determined. A registry-based or pluggable parser would scale better.

karthik120710:
Summary                                                                                                                                             

  - Replaced the hardcoded switch in parseDSN with a sync.RWMutex-protected registry of DSNParser functions, making the DSN parsing layer pluggable     
  without modifying core code.
  - Added RegisterDSNParser(driverName, parser) as a public API so external drivers and future integrations can register their own parsers (e.g., from  
  init()).                                                                                                                                              
  - Registered all previously hardcoded drivers at init() time, including newly added aliases: pgx, lib/pq, sqlserver, oracle, go-oci8, and oci8.
  - Added a bestEffortParse fallback (standard URL parsing) for unrecognized drivers, replacing the prior silent empty-address behavior.                
  - Added a Warn log when DSN parsing fails so server.address omission is surfaced rather than silently emitted as "unknown".                           
  - Expanded DbClientRequestTraceAttrs system-name mapping to cover mariadb, pgx, lib/pq, clickhouse, oracle, and mssql/sqlserver — previously these all
   fell through to other_sql.                                                                                                                           
  - Bumped go.opentelemetry.io/otel/log, sdk/log, and related packages to v0.19.0.                                                                      
  - Added a test case for the clickhouse driver system-name mapping and a separate case confirming unknown drivers fall back to other_sql.              

Fixes #508

  Test plan                                                                                                                                             
                                                                                                                                                        
  - Run go test ./pkg/instrumentation/databasesql/... — all existing and new semconv tests pass.                                                        
  - Verify a custom driver can be registered via RegisterDSNParser in an init() and is correctly used at hook time.
  - Confirm that an unregistered driver with a standard URL DSN gets a best-effort address instead of an empty string.                                  
  - Confirm a warning is logged when DSN parsing fails.  
  #508

### DIFF
--- a/instrumentation/database/sql/client.go
+++ b/instrumentation/database/sql/client.go
@@ -43,6 +43,10 @@ func beforeOpenInstrumentation(ictx hook.HookContext, driverName, dataSourceName
 	info := ParseDSN(driverName, dataSourceName)
 	addr := info.Addr()
 	if addr == "" {
+		// Surface the omission rather than silently emitting server.address:
+		// "unknown". The DSN is intentionally not logged as it may carry
+		// credentials.
+		logger.Warn("could not determine server.address from DSN", "driver", driverName)
 		addr = "unknown"
 	}
 	dbName := info.DBName
--- a/instrumentation/database/sql/dsnparse/parse.go
+++ b/instrumentation/database/sql/dsnparse/parse.go
@@ -6,6 +6,7 @@ package dsnparse
 import (
 	nurl "net/url"
 	"strings"
+	"sync"
 )
 
 // DSNInfo holds the parsed server address components and database name from a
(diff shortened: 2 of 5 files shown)


karthik120710:
hi @txabman42 i have added  pluggable DSN parser registry and expand semconv driver mapping can u review my pr.

waqar2403:
Hey @karthik120710, nice work a few things I noticed while going through the diff

Build issue: The refactor removes ParseDbName from parse.go but client.go:47 still calls it (dbName := ParseDbName(dataSourceName)). This would fail to compile. Was the removal accidental during the registry refactor?

Test coverage: The new tests cover the semconv mapping changes (clickhouse, unknown driver) but there are no tests for RegisterDSNParser itself, bestEffortParse, or the new aliases like pgx and lib/pq going through parsePostgres. Since the registry is the main change here, a basic test showing registration and lookup works would help.

Minor: The PR title has a leading space which is causing the conventional commit check to fail. Trimming it should fix the CI.

karthik120710:
Hi @waqar2403 I have add units test and updated process title , but still unit test are not running .

waqar2403:
@karthik120710 good progress on the registry refactoring. Two things here:

Tests not running Your go.mod bumps a few OTel indirect deps (otlploghttp, otel/log, sdk/log) but the go.sum likely needs regeneration. Try this from the module root:
```
cd pkg/instrumentation/databasesql
go mod tidy
go test ./
```

If there are still import resolution errors after that, check if testify is listed under `require` in your go.mod it should already be there from the existing db_test.go usage.

Minor style note on RegisterDSNParse Right now `RegisterDSNParser` is exposed as a mutable variable

```go
var RegisterDSNParser = dsnparse.RegisterDSNParser
```

This means any consumer can accidentally overwrite it like `db.RegisterDSNParser = nil` A simple wrapper function removes that risk

```go
func RegisterDSNParser(driverName string, parser dsnparse.DSNParser) {
    dsnparse.RegisterDSNParser(driverName, parser)
}
```
Not a blocker just something to consider for safety.
Otherwise the internal/dsnparse package structure, the registry with sync.RWMutex, the BestEffortParse fallback, and the new semconv driver mappings all look good to me.

pkg/instrumentation/databasesql/internal/dsnparse/parse.go
```diff
@@ -0,0 +1,275 @@
...
+import (
+	"errors"
+	"fmt"
+	nurl "net/url"
+	"strings"
+	"sync"
+)
+
+// DSNParser parses a driver-specific DSN and returns the server address (host:port).
+type DSNParser func(dsn string) (addr string, err error)
```
kakkoyun: I think I would prefer having an interface for parser, and implementations in struct.

pkg/instrumentation/databasesql/internal/dsnparse/parse.go
```diff
@@ -0,0 +1,275 @@
...
+
+var (
+	parserMu       sync.RWMutex
+	parserRegistry = map[string]DSNParser{}
```
kakkoyun: We can have another struct type for registry and we don't need package level locks.

pkg/instrumentation/databasesql/internal/dsnparse/parse.go
```diff
@@ -0,0 +1,275 @@
...
+// It is safe to call from package init() functions.
+func RegisterDSNParser(driverName string, parser DSNParser) {
+	parserMu.Lock()
+	defer parserMu.Unlock()
+	parserRegistry[driverName] = parser
+}
+
+func init() {
+	// Register all built-in DSN parsers.
+	RegisterDSNParser("mysql", ParseMySQL)
```
kakkoyun: We are constantly locking and unlocking here. We can have a batch API where lock only once.

pkg/instrumentation/databasesql/internal/dsnparse/parse.go
```diff
@@ -0,0 +1,275 @@
...
+		return parser(dsn)
+	}
+
+	// Best-effort: try standard URL parsing for drivers not in the registry.
+	return BestEffortParse(dsn)
+}
+
+// BestEffortParse attempts to extract a host:port from a DSN using standard URL parsing.
+// It is used as a fallback for drivers that have no registered parser.
+func BestEffortParse(dsn string) (string, error) {
```
kakkoyun: We don't need to export this method.

pkg/instrumentation/databasesql/internal/dsnparse/parse.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func ParseClickHouse(dsn string) (addr string, err error) {
+	// ClickHouse DSN formats:
+	// tcp://host:port?database=dbname&username=user&password=pass
+	// http://host:port?database=dbname
+	// clickhouse://host:port/database?username=user&password=pass
+	u, err := nurl.Parse(dsn)
+	if err != nil {
+		return "", err
```
kakkoyun: We shouldn't just return the net/url error here. It exposes the rawURL and it means we can leak the sensitive information to some external service if somebody tries to log or send the error.

See: https://cs.opensource.google/go/go/+/refs/tags/go1.26.3:src/net/url/url.go;l=398

kakkoyun (changes requested):
Thanks for the initiative. 

This requires a bit of improvement. Please see my comments.

karthik120710:
hi @kakkoyun i have made the changes can u review my pr.

pkg/instrumentation/databasesql/internal/dsnparse/parse.go
```diff
@@ -0,0 +1,336 @@
...
+
+// DSNParserFunc is an adapter to allow use of ordinary functions as DSNParsers.
+type DSNParserFunc func(dsn string) (addr string, err error)
+
+// Parse calls f(dsn).
+func (f DSNParserFunc) Parse(dsn string) (string, error) { return f(dsn) }
+
+// sanitizeURLError strips the raw URL from *url.Error to prevent leaking
+// credentials or other sensitive data embedded in DSNs via error messages.
+func sanitizeURLError(err error) error {
```
kakkoyun: Could we add some test cases to prove we don't leak any credentials?

kakkoyun (commented):
This looking better. Closer to done.
We need to add some test cases and make sure CI is green.

darccio:
@karthik120710 Will you be able during this month to apply @kakkoyun's feedback, fix the types detected by the CI check `Check typos` and fix the conflicts? Thanks!

vyagh:
I'll be happy to take it to the line if it's stalled.
@karthik120710  @darccio whatever works.

darccio:
@vyagh We'd appreciate it! Thanks!

karthik120710:
hi @kakkoyun i have added tests verify sanitizeURLError which dosent leak credentionls.

pkg/instrumentation/databasesql/internal/dsnparse/parse.go
karthik120710: i have addeed test case

darccio:
@karthik120710 Check the typos, please. https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/27828055364/job/82922801421?pr=509#step:3:41

kakkoyun:
@karthik120710 this is in good shape. The parser interface, the registry struct (no more package-level locks), and the `sanitizeURLError` redaction with its leak tests all cover the earlier feedback. Two things left:

1. Rebase onto main. `pkg/instrumentation` moved to top-level `instrumentation/` in #597, so the branch is conflicting.
2. Get `Check Typos` green (the only failing check).

Ping me once it's rebased and CI is green and I'll re-review and resolve the threads. #508 is a v1 GA blocker, let's get it in.

darccio:
@vyagh I took over the PR to unblock v1.

kakkoyun (approved):
LGTM. Thanks!

vyagh:
thanks @darccio

Issue #512, opened 2026-05-17
isSetup() is unimplemented and hard-codes false


karthik120710:
Description:                                                                                                                                          
  ## Description                                                                                                                                        
                
  `isSetup()` in `tool/internal/setup/setup.go` is declared but never implemented:                                                                      
                                                                                                                                                       
  The function is currently unreferenced, so it causes no runtime failure. However, its existence suggests a guard was planned to avoid re-running setup
   unnecessarily. If it is ever wired in as a guard, it will always report "not set up", causing redundant or incorrect re-initialization on every      
  build.

  Expected Behavior

  isSetup() should inspect actual state (e.g., check for presence of previously written files or a marker) and return true when setup has already been  
  completed.

kakkoyun:
Duplicate of #161

PR #513, opened 2026-05-17
fix(setup): implement isSetup() to check for matched.json marker

karthik120710:
Description:                                                                                                                                          
  ## Summary                                                                                                                                          
                                                                                                                                                        
  Fixes the unimplemented `isSetup()` guard in `tool/internal/setup/setup.go`.                                                                        
                                                                                                                                                        
  Resolves #<issue-number>
                                                                                                                                                        
  ## Changes                                                                                                                                          
            
  - Implemented `isSetup()` to check for the presence of `matched.json` (written by `store()` at the end of a successful `Setup` run) instead of always
  returning `false`                                                                                                                                     
  - Added unit tests for both the "not set up" and "already set up" cases
                                                                                                                                                        
  ## Why                                                                                                                                                
        
  Without this fix, the guard at `Setup()` line 149 would never short-circuit, causing setup to re-run on every build once the guard is wired in —      
  resulting in redundant re-initialization.                                                                                                             
                                           
  ## Test Plan                                                                                                                                          
                                                                                                                                                      
  - [ ] `go test ./tool/internal/setup/... -run TestIsSetup` passes
  - [ ] Full setup package tests pass: `go test ./tool/internal/setup/...`                                                                              
  
  Notes from the guidelines:                                                                                                                            
  - No changelog entry needed — this is a bug fix for an internal/unshipped guard with no user-facing behavior change. You can prefix the title with  
  [chore] if reviewers flag it, or ask a maintainer to add the Skip Changelog label.                                                                    
  - The title follows the [component] description format required by the guidelines.    
  #512

### DIFF
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -43,10 +43,12 @@ func (sp *SetupPhase) keepForDebug(srcPath string) {
 	}
 }
 
-// This function can be used to check if the setup has been completed.
+// isSetup reports whether setup has already been completed for the current
+// build by checking for the presence of the matched-rules file written by
+// store() at the end of a successful Setup run.
 func isSetup() bool {
-	// TODO: Implement Task
-	return false
+	_, err := os.Stat(util.GetMatchedRuleFile())
+	return err == nil
 }
 
 // flagsWithPathValues contains flags that accept a value from "go build" command.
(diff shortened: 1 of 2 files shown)


karthik120710:
Hi @txabman42 @kakkoyun  can u review my pr

y1yang0 (approved):

txabman42 (changes requested):
Checking the file existence is not enough. The setup file might be related to a previous configuration, making not possible to do the setup again.

This problem is being discussed in this [issue](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/161)

PR #517, opened 2026-05-18
fix(util): handle nil values in AssertType

pushkarscripts:
## Description

Improve `AssertType` nil safety and type mismatch diagnostics in `tool/util/assert.go`.

This updates type formatting to use `reflect.Type.String()` and prevents nil-related panics during failed assertions. Also adds minimal unit coverage for concrete and pointer type assertions.

## Motivation

`AssertType` currently calls `reflect.TypeOf(v).Name()` directly, which can panic for nil values and produces unclear diagnostics for composite types like pointers.

This change makes assertion failures safer and improves error messages for debugging type mismatches.

---

## Checklist

- [x] Tests pass: `make test`


tool/util/assert_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAssertType_Success(t *testing.T) {
```
txabman42, line 16: Lets also test what this PR tries to fix: _handle nil values in AssertType_.

txabman42 (changes requested):

tool/util/assert_test.go
pushkarscripts: > Lets also test what this PR tries to fix: _handle nil values in AssertType_.

Good point, I’ll add coverage for nil handling as well

tool/util/assert.go
```diff
@@ -18,10 +18,12 @@ func Assert(condition bool, message string) {
 func AssertType[T any](v any) T {
 	value, ok := v.(T)
 	if !ok {
-		actualType := reflect.TypeOf(v).Name()
-		var zero T
-		expectType := reflect.TypeOf(zero).String()
-		ex.Fatalf("Type assertion failed: %s, expected %s",
+		actualType := "nil"
```
kakkoyun, line 21: suggestion: We also have a new `TypeFor` API, it worth checking https://pkg.go.dev/reflect#TypeFor (I haven't check how it behaves with pointer types and nil).

nit: We can also strictly check if the type is a pointer type and then assume it could be nil.

pushkarscripts: > suggestion: We also have a new `TypeFor` API, it worth checking https://pkg.go.dev/reflect#TypeFor (I haven't check how it behaves with pointer types and nil).
> 
> nit: We can also strictly check if the type is a pointer type and then assume it could be nil.

sure, I’ll update it accordingly.

kakkoyun (approved):

txabman42 (approved):
LGTM! TY ❤️

PR #518, opened 2026-05-18
fix(tool): forward -C flag directory to package loading

### LINKED ISSUE #507: Support go build -C
`otelc go build -C <dir> [pkgs]` does not respect the `-C` flag.

`-C <dir>` it's never forwarded to its package loading step. `packages.Config` has no `Dir` set, so package patterns like `.` or `./...` are resolved against the cwd instead of `<dir>`.

Related to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/483/changes#r3234440469

Pulkit7070:
## Problem

`otelc go build -C <dir>` does not respect the `-C` flag. Package patterns like `.` or `./...` are resolved against the process working directory instead of `<dir>` because `packages.Config.Dir` was never set.

Steps to reproduce:
1. Create a Go module in `/some/dir`
2. Run `otelc go build -C /some/dir ./...` from a different directory
3. Package loading fails or instruments the wrong packages

Fixes #507

## Solution

Add `LoadPackagesInDir` to the `pkgload` package. It accepts an explicit `dir` string and sets `packages.Config.Dir`, so `go/packages` resolves patterns relative to that directory.

In `getBuildPackages`, extract the `-C` flag value from the build args using the new `extractFlagValue` helper and forward it to every `LoadPackagesInDir` call.

Existing callers (`ResolvePackageName`, `ResolveExportFiles`) continue to use `LoadPackages`, which delegates to `LoadPackagesInDir` with an empty dir.

## Changes

- `tool/internal/pkgload/pkgload.go`: add `LoadPackagesInDir(ctx, dir, mode, buildFlags, patterns...)`. `LoadPackages` is a thin wrapper passing empty dir.
- `tool/internal/setup/setup.go`: add `extractFlagValue` helper; update `getBuildPackages` to extract `-C` and pass it to `LoadPackagesInDir`.


y1yang0 (commented):
LGTM, this is surprisingly clean and elegant. Please also add an integration test

tool/internal/pkgload/pkgload.go
```diff
@@ -13,16 +13,31 @@ import (
...
 	mode packages.LoadMode,
 	buildFlags []string,
 	patterns ...string,
+) ([]*packages.Package, error) {
+	return LoadPackagesInDir(ctx, "", mode, buildFlags, patterns...)
+}
+
+// LoadPackagesInDir is like LoadPackages but resolves package patterns relative
+// to dir. If dir is empty, the current working directory is used.
+func LoadPackagesInDir(
```
y1yang0: This distinction between LoadPackages and LoadPackagesInDir seems premature.  I'd suggest we just keep LoadPackages for now

amazingakai (changes requested):
This currently does not work:
```bash
otelc go build -C ./somedir
```
because we pass `-C` after `-a -x -n` inside `listBuildPlan`.

From the debug log:
```
new="[go build -a -x -n -C ./somedir]"
```

But `go` requires `-C` to appear before build flags/packages, so this becomes invalid.

Would it be simpler to consume the `-C` flag early and just `chdir` into it at command start (with a deferred restore of the original cwd)?

kakkoyun:
> This currently does not work:
> 
> ```shell
> otelc go build -C ./somedir
> ```
> 
> because we pass `-C` after `-a -x -n` inside `listBuildPlan`.
> 
> From the debug log:
> 
> ```
> new="[go build -a -x -n -C ./somedir]"
> ```
> 
> But `go` requires `-C` to appear before build flags/packages, so this becomes invalid.
> 
> Would it be simpler to consume the `-C` flag early and just `chdir` into it at command start (with a deferred restore of the original cwd)?

@Pulkit7070 concerns of @amazingakai needs to be addressed here.

Pulkit7070:
Thanks for the thorough reviews. Both points are well-taken and I've addressed them in the latest push.

**@amazingakai** - you're right that the `cfg.Dir` approach was incomplete. The deeper problem is that `listBuildPlan` in `find.go` prepends `-a -x -n` to the remaining args, so `-C dir` would end up in the wrong position (`go build -a -x -n -C dir ./...`) and the `go` tool would reject it. The fix now uses `os.Chdir()` instead, and a new `consumeCFlag` helper strips `-C dir` from the args before they reach `listBuildPlan`, so the subprocess never sees the flag at all. The working directory change covers both the `go/packages` resolution step and the `listBuildPlan` dry-run. `getBuildPackages` also handles `os.Chdir()` for the case where it is called with the original args from the `GoBuild` cleanup defer.

**@y1yang0** - `LoadPackagesInDir` is fully reverted. `pkgload.go` is back to its original form with only `LoadPackages`, `ResolvePackageName`, and `ResolveExportFiles`. The `-C` handling lives entirely in `setup.go` where the flag is consumed.

Let me know if anything looks off - happy to iterate.

kakkoyun:
I merged #468 (first come, first served). This needs another rebase.

And let's wait for @amazingakai's review.

Pulkit7070:
Rebased on main. Also squashed the messy history (reverts, CLA retriggers, merge commit) down to a single clean commit.

y1yang0:
But why not fix listBuildPlan by inserting -C after go build to solve this problem?

Pulkit7070:
Good question. Fixing listBuildPlan alone is not sufficient because the -C flag affects two independent layers:

1. **go/packages resolution** - when getBuildPackages calls packages.Load, it resolves patterns like ./... relative to the process working directory. Repositioning -C in listBuildPlan does not help here at all.
2. **listBuildPlan dry-run** - the subprocess command needs -C in the correct position.

To cover both with the cfg.Dir approach, we would need to thread the extracted dir value into every packages.Config construction, not just listBuildPlan.

os.Chdir handles both layers with a single change: the process working directory shifts, so go/packages resolves correctly and the dry-run subprocess inherits the same cwd. consumeCFlag strips -C from args so listBuildPlan never sees it and the subprocess flag ordering stays valid.

Also, amazingakai explicitly suggested the os.Chdir approach in their CHANGES_REQUESTED review, which is still open. Happy to switch if you and amazingakai align on the listBuildPlan approach instead.

y1yang0:
But if we are to use os.chdir directly, we would essentially need to mirror the behavior of the Go toolchain as implemented here:

https://github.com/golang/go/blob/03d1f8efc82678bba84d5e50d9144cfc6847a1f9/src/cmd/go/main.go#L383

Pulkit7070:
You are right. Go's handleChdirFlag is strictly positional - it only looks for -C at the argument immediately after the subcommand, and removes it from os.Args so a reinvoked child toolchain does not process it again. Our consumeCFlag scans the full arg list, which does not match those semantics.

Given this, I think the cleaner fix is the one you originally suggested: fix listBuildPlan to insert -C dir right after go build, and set cfg.Dir on the packages.Config. No process-global chdir, no positional constraint to enforce, and the behavior is explicit at each call site.

Would you and @amazingakai prefer we switch to that approach? Happy to implement it - just want both reviewers to align before we push another redesign.

y1yang0:
Apologies for any confusion. I'd like to clarify my thought process. Initially, it seemed to me that both potential fixes were reasonable:

- The os.chdir approach, where subsequent commands are unaware of the -C flag. This seemed simpler.
- Passing -C to LoadPackages and fixing the argument order in listBuildPlan (go build -C dir {remaining}). This seemed more explicit and self-contained.

However, after reviewing the Go toolchain's source code, I see its logic is essentially to consume the -C flag and then call os.Chdir. This strongly suggests that the first approach is indeed the more correct and idiomatic one.

I'm leaning this way now. Do other reviewers have any further thoughts on this?

Pulkit7070:
Thanks for the clarification. Agreed - Go's own toolchain doing os.Chdir is a strong signal that it is the right approach.

The one remaining gap in our current implementation: our consumeCFlag scans the entire args list for -C, but Go's handleChdirFlag only looks for it at the specific position right after the subcommand. We should tighten that to match Go's positional semantics before merging.

Waiting on @amazingakai to weigh in, then we will implement with that correction.

amazingakai:
@Pulkit7070 I think it completely makes sense to match go's behavior here. We should consume -C only if it appears as the first argument after otelc go build/install or otelc setup. Also, after looking at the go toolchain's code I realized we don't need to restore to the original directory with defer, changing the process' dir doesn't change the shell's dir, so restoring to the original dir would do nothing here.

Also, something probably went wrong with the rebase here, this PR is undoing some changes I did in #468 and also removing some unrelated comments. Please fix those.

Pulkit7070:
Done. Three things addressed in the latest push:

1. **Positional constraint**: replaced consumeCFlag (full scan) with consumeCFlagPositional that only consumes -C when it is the first argument, matching Go toolchain semantics.
2. **No defer restore**: removed the defer os.Chdir(orig) as you noted it has no effect on the parent shell.
3. **#468 regression fixed**: the previous squash was sitting on top of pre-#468 code. Rebased cleanly on current main - all of #468s changes are intact and the only diff now is +26 lines in setup.go and +115 lines in setup_test.go.

tool/internal/setup/setup.go
```diff
@@ -75,6 +75,23 @@ var flagsWithPathValues = map[string]bool{
...
+func consumeCFlagPositional(args []string) (string, []string) {
+	if len(args) == 0 {
+		return "", args
+	}
+	if strings.HasPrefix(args[0], "-C=") {
+		return strings.TrimPrefix(args[0], "-C="), args[1:]
+	}
+	if args[0] == "-C" && len(args) > 1 {
+		return args[1], args[2:]
+	}
```
amazingakai, line 97: https://github.com/golang/go/blob/03d1f8efc82678bba84d5e50d9144cfc6847a1f9/src/cmd/go/main.go#L395
Also supports --C

amazingakai (changes requested):
We also require `consumeCFlagPositional` in `GoBuild`, otherwise the build fails, but this could be a little tricky.
We have to handle the following cases:

* `otelc setup -C ...` -- already handled
* `otelc go build -C ...` -- after consuming the -C flag we also have to trim it from the args. We might need to change the signature of Setup a little here and accept args + flags instead of cmd. @y1yang0, would appreciate your opinion here as well.
* `go build` also supports both `go -C ./dir build` and `go build -C ./dir,` so we have to check both cases and allow -C to appear immediately before or after the build/install arg.

Pulkit7070:
Good catch. Tracing through the two cases:

**go build -C dir ./...**: cmd.Args().Tail() = [-C, dir, ./...], consumeCFlagPositional picks it up at position 0 - already handled.

**go -C dir build ./...**: GoBuild fails immediately at the cmd.Args().First() != "build" check before Setup is even called. And even if we fixed that, Setup does cmd.Args().Tail() which gives ["dir", "build", "./..."] - the -C is gone.

The fix needs to happen in GoBuild before any arg inspection. Since cmd cannot be mutated, the cleanest path is the one you suggested: change Setup to accept an explicit args slice instead of cmd, so GoBuild can pre-process -C in both positions, chdir, strip it, then pass the cleaned args down.

Rough shape:



consumeCFlagFromArgs would check position 0 for -C (before subcommand) and also position 1 for the go build -C case after subcommand trimming.

Does this match what you had in mind? Happy to implement once you and @y1yang0 agree on the shape.

Pulkit7070:
Gentle ping @amazingakai @kakkoyun - the syntax error in the test file has been fixed in d1aea36, though it looks like the full CI suite did not trigger for that commit. Happy to rebase or push an empty commit to kick it off if needed. Also still waiting on alignment for the GoBuild -C handling approach before implementing - let me know how you would like to proceed.

amazingakai:
> go build -C dir ./...: cmd.Args().Tail() = [-C, dir, ./...], consumeCFlagPositional picks it up at position 0 - already handled.

This does not work because `Setup` only modifies a local copy of the args. The original `GoBuild` still sees `-C` in the argument list and passes it to `go build`, which then causes build failure.

Pulkit7070:
Thanks for the detailed diagnosis. You are right - Setup consumed -C from a local copy of args and called os.Chdir, but BuildWithToolexec re-read cmd.Args().Slice() independently, so -C was still present and ended up in the wrong position in the toolexec invocation.

Fixed in the latest commit (94ec34e): BuildWithToolexec now calls consumeCFlagPositional on args[1:] before constructing newArgs, stripping -C once Setup has already applied the directory change via os.Chdir.

Pulkit7070:
Thanks for the detailed review @amazingakai. I've rebased on main and addressed both points.

**Three -C positions now handled**, matching Go's own toolchain:
- `otelc setup -C ./dir`: -C at position 0 (was already working)
- `otelc go build -C ./dir ...`: -C after build/install (was already working)
- `otelc go -C ./dir build ...`: new, -C before build/install

**Implementation:**
1. `Setup` strips -C-before-build first, then trims build/install, then consumes -C-after-build. One `os.Chdir` runs from whichever position matched.
2. `GoBuild` looks past a leading -C when validating the `build`/`install` dispatch (the actual chdir still happens in Setup).
3. Extracted `stripCFlag()` helper used in `BuildWithToolexec` so -C never leaks into the forwarded `go build -toolexec=...` command (which is what was producing the invalid `go build -a -x -n -C ./somedir` you flagged).

**Tests:** added `TestStripCFlag` covering all six placement combinations; existing `TestConsumeCFlagPositional` and `TestGetPackagesWithCFlag` still pass. The three pre-existing failures (`TestRuleFilesFromDir`, `TestMultipleRuleFiles`, `TestLoadDefaultRules`) are unrelated, they depend on a missing `otelc-pkg.gz` build artifact and fail on upstream main as well.

PTAL when you have a moment.

amazingakai (commented):
Right now this strips the flag from `args` in `BuildWithToolexec` and lets `Setup` handle the `Chdir`, which works. Would it be simpler to consume the arguments once in `GoBuild()` and pass the already-parsed values to `Setup` and `BuildWithToolexec`?

`Setup` would still need its own parsing path because `otelc setup` can be invoked independently of `otelc go build`.

More generally, I wonder if it would make sense to separate command parsing from setup execution. For example, having a `SetupCommand()` responsible for consuming cli flags, while `Setup()` simply accepts the parsed flags and args as values instead of a `cli.Command`.
(Similarly, `BuildWithToolexec()` could accept parsed values as well.)

I would appreciate opinions from other reviewers on this^

tool/internal/setup/setup.go
amazingakai: This is still unaddressed.

kakkoyun:
@Pulkit7070 Do you want to continue working on this?

PR #520, opened 2026-05-18
chore: validate exported span output in basic integration test

### LINKED ISSUE #519: Validate exported span output in basic integration test
## Problem

The current basic integration test verifies that the basic demo builds, runs, and emits several hook-related log messages. But, it can be strengthened by validating the actual exported OpenTelemetry span output.

The basic demo already emits a `hello-world` span in JSON form. The test could parse this already-emitted span output and verify that the demo produces meaningful telemetry, not only hook log output.

This change would not add additional demo output. It would only validate the already-emitted span output from the basic demo.

## Proposed solution

Add a small helper in `test/integration/basic_test.go` that:

- scans the demo output line by line,
- parses JSON lines where possible,
- finds the exported span with `Name == "hello-world"`,
- verifies that `SpanContext.TraceID` and `SpanContext.SpanID` are non-empty and non-zero,
- verifies that `InstrumentationScope.Name` contains the expected `helloworld` instrumentation package,
- verifies expected demo attributes, such as:
  - `hello.path = /api/hello`
  - `hello.param.name = world`
  - `hello.status = 200`

This would make the basic integration test assert that the demo not only runs and executes hooks, but also emits the expected OpenTelemetry span telemetry.

## Alternatives

A simpler alternative would be to use substring assertions such as checking if the output contains `"Name":"hello-world"` or `"TraceID"`. However, parsing the JSON output would be more robust because it can verify the actual span structure, non-zero IDs, instrumentation scope, and attribute key/value pairs without relying on a full snapshot of the output.

I would avoid adding new log lines or additional demo output, since the current demo is already fairly verbose.

## Additional context

I noticed this while running the basic demo locally with:

```bash
make build
cd demo/app/basic
../../../otelc go build
./basic
```

fmutlu68:
## Description

This PR strengthens the existing basic integration test by validating the already-emitted OpenTelemetry span output from the basic demo.

It adds a small JSON parsing helper that scans the demo output, finds the exported `hello-world` span, and verifies:

- non-empty and non-zero `TraceID`
- non-empty and non-zero `SpanID`
- the expected `helloworld` instrumentation scope
- expected demo span attributes:
  - `hello.path = /api/hello`
  - `hello.param.name = world`
  - `hello.status = 200`

This does not add any new demo output. It only validates the span output that the basic demo already emits.

## Motivation

The basic integration test already verifies that the demo builds, runs, and emits hook-related output. However, validating the exported span itself makes the test assert that the demo produces meaningful OpenTelemetry telemetry, not just hook log messages.

This improves regression coverage for the simplest end-to-end instrumentation path.

Fixes #519

## Testing

- `make format`
- `make lint`
- `make test`
- `make build`
- `go -C test test -tags integration ./integration -run TestBasic -count=1 -v`

---

## Checklist

- PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- Code formatted: `make format`
- Linters pass: `make lint`
- Tests pass: `make test`
- Tests added for new functionality
- Tests follow [testing guidelines](docs/testing.md)

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -6,8 +6,10 @@
 package test
 
 import (
+	"encoding/json"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM.

Thanks for the contribution.

PR #526, opened 2026-05-19
test: cover latest library build dependency discovery

### LINKED ISSUE #525: Add unit coverage for LatestLibBuild dependency discovery

The LatestLibBuild workflow depends on helper logic that discovers which direct module dependencies are covered by instrumentation rule targets.

Currently, this logic is exercised through the LatestLibBuild workflow, but the lower-level dependency matching behavior in `test/testutil/latestlib.go` does not appear to have focused unit coverage.

If this matching logic regresses, the workflow may either skip dependencies that should be tested against `@latest` or attempt to update unrelated dependencies.

Add focused unit tests for the dependency discovery and target matching logic in `test/testutil/latestlib.go`, covering:

- exact module target matches,
- module-to-subpackage target matches,
- prefix false positives,
- unrelated targets,
- indirect dependency filtering,
- local replace filtering.

An alternative would be to rely only on the existing LatestLibBuild workflow tests. However, focused unit tests would make the dependency matching behavior easier to validate and debug without running the broader workflow.

**Additional context**

This is related to the LatestLibBuild compatibility workflow. The goal is to improve confidence that instrumented dependencies are selected correctly before they are bumped to `@latest`.

fmutlu68:
## Description

This PR adds focused unit coverage for the helper logic used by the LatestLibBuild workflow.

It verifies that instrumentation targets are matched to module dependencies correctly and that indirect or locally replaced dependencies are ignored.

## Motivation

LatestLibBuild tests rely on dependency discovery to decide which instrumented libraries should be upgraded to `@latest`. Covering this logic helps prevent regressions in the compatibility test workflow.

Fixes #525

## Testing

- `go -C test test ./testutil`
- `go -C test test ./...`
- `make format`
- `make lint`
- `make test`

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ SHELL := /bin/bash
         build-demo build-demo-grpc build-demo-http format/go format/yaml lint/go lint/yaml \
         lint/action lint/makefile lint/license-header lint/license-header/fix lint/dockerfile actionlint yamlfmt gotestfmt ratchet ratchet/pin \
         ratchet/update ratchet/check golangci-lint embedmd checkmake hadolint help docs check-embed check-api-sync check-golden-files \
-        test-unit/update-golden test-unit/tool test-unit/pkg test-unit/demo \
+        test-unit/update-golden test-unit/tool test-unit/pkg test-unit/demo test-unit/helper \
         test-unit/coverage test-unit/tool/coverage test-unit/pkg/coverage \
         test-integration/coverage test-e2e/coverage \
         registry-diff registry-check registry-resolve weaver-install tidy/test-apps \
@@ -375,7 +375,7 @@ benchmark/threshold: build ## Enforce absolute otelc overhead ceiling (fails if 
 test: ## Run all tests (unit + integration + e2e)
 test: test-unit test-integration test-e2e
 
-test-unit: test-unit/tool test-unit/pkg test-unit/demo ## Run all unit tests (tool + pkg + demo)
+test-unit: test-unit/tool test-unit/pkg test-unit/demo test-unit/helper ## Run all unit tests (tool + pkg + demo + test helpers)
 
 .ONESHELL:
 test-unit/update-golden: ## Run unit tests and update golden files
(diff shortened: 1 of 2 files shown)


fmutlu68:
The failed integration job appears unrelated to this PR. The new tests added here are under `test/testutil` and passed locally with:

- `go -C test test ./testutil`
- `go -C test test ./...`

The failing job is `TestGRPCServerTelemetryFlushOnSignal` on Ubuntu 22.04 arm64, timing out while waiting for TCP readiness at `localhost:50052`. This seems to happen before the span-flush assertion path.

Would it be possible to rerun the failed integration job, @kakkoyun?

test/testutil/latestlib_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCoversAnyTarget(t *testing.T) {
```
txabman42, line 14: IINM, these tests are not being executed during CI. We'd need to target `/test` folder for `test-unit`.

txabman42 (changes requested):

test/testutil/latestlib_test.go
fmutlu68: Thanks, that makes sense. I kept the tests under `test/testutil`, since they exercise helpers from that package, and added a separate `test-unit/test` target so these tests are included in the unit test workflow.
Verified with these:

- `make test-unit/test`
- `make test-unit`
- `make format`
- `make lint`
- `make test`

Makefile
```diff
@@ -418,6 +418,13 @@ test-unit/pkg: package ## Run unit tests for pkg modules only
 		go test -C "$$moddir" -v -shuffle=on -timeout=5m -count=1 ./... 2>&1 | tee -a ./gotest-unit-pkg.log; \
 	done
 
+.ONESHELL:
+test-unit/test: ## Run unit tests for test helper packages
```
```suggestion
test-unit/helper: ## Run unit tests for test helper packages
```

fmutlu68: Thanks, `test-unit/helper` is definitely clearer here. I updated the target name from `test-unit/test` to `test-unit/helper` and ran the relevant tests again.

txabman42 (approved):
LGTM! TY ❤️

PR #529, opened 2026-05-20
test(latestlibbuild): check supported version ranges when discovering deps

amazingakai:
## Description

Skip bumping dependencies in `TestLatestLibBuild` when the latest upstream version falls outside the supported instrumentation version ranges.

## Motivation

Required for #426 as bumping `k8s.io/client-go` to `@latest` would cause CI failures even when newer upstream versions are explicitly unsupported by the instrumentation rules.

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format`
- [x] Linters pass: `make lint`
- [x] Tests pass: `make test`
- [ ] Tests added for new functionality
- [ ] Tests follow [testing guidelines](docs/testing.md)
- [ ] Documentation updated (if applicable)

PR #530, opened 2026-05-20
fix(setup): truncate extracted files and add extraction tests

amazingakai:
## Description

Truncate extracted files before overwriting them during instrumentation package extraction and add unit tests for extraction behavior.

## Motivation

Found this while working on #529. Previously extracted files were overwritten without truncation, which could leave stale trailing bytes when the new content was shorter than the existing file.

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/extract_test.go
+++ b/tool/internal/setup/extract_test.go
@@ -1,0 +1,202 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package setup
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNormalizePath(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{
+			name: "pkg_temp dir",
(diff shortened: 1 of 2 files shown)


tool/internal/setup/extract_test.go
```diff
@@ -0,0 +1,202 @@
...
+	err = extractGZip(tarBuf.Bytes(), tmpDir)
+	require.NoError(t, err)
+
+	bs, err := os.ReadFile(filepath.Join(tmpDir, "pkg", "test.yaml"))
+	require.NoError(t, err)
+
+	require.Equal(t, content, bs)
+}
+
+func TestExtractGZip_SkipsZipSlip(t *testing.T) {
```
pushkarscripts, line 175: I was wondering if it makes sense to also test absolute paths like `/tmp/evil.yaml` (or Windows-style ones), or if those are already handled here?

amazingakai: Good idea, done.

NameHaibinZhang (approved):

NameHaibinZhang:
LGTM

tool/internal/setup/extract_test.go
kakkoyun: We can merge this after this one is addressed.

txabman42 (approved):
LGTM! Nice catch!!!

PR #560, opened 2026-06-03
fix(tool): disambiguate trampoline names across do-sequence entries

### LINKED ISSUE #544: bug(tool/instrument): do sequence on a single func emits duplicate trampoline names
Follow-up from #377 (review-summary point from @amazingakai).

## Context

The `do:` field accepts a sequence of modifier maps, e.g.:

```yaml
some_rule:
  target: example.com/svc
  where:
    func: Open
  do:
    - inject_hooks: { before: BeforeOpen, ... }
    - inject_hooks: { after: AfterOpen, ... }
```

`normalize.go` currently expands this into multiple flat rules with the same target/selector. The instrumenter then tries to emit two trampoline functions for the same target function, both with the same generated name, and the compile step fails (or, worse, silently picks one).

## Problem

Multiple hooks / advice chains on a single function is a common AOP need; the schema declares it valid but the implementation can't honour it.

## Proposed work

- Trampoline-name generation must include a discriminator per `do` entry index (e.g. `__hook_<rule>_<index>`).
- The order of execution within the trampoline must match the `do` sequence order.
- Add a golden test fixture covering the multi-hook case.

## References

- #377 review summary from @amazingakai

kakkoyun:
## Problem

A `do:` sequence that applies several modifiers to the same function produced
trampoline and `HookContextImpl` declarations with identical names, so
compilation failed with redeclaration errors.

Root cause: names came from `util.CRC32(rule.String())`, and `String()` returned
only the rule name, which is the same for every `do` entry expanded from one rule.

## Approach

Trampoline and `HookContextImpl` names now derive from what the rule does, not
its name or position.

- `tool/internal/rule/func_rule.go` adds `InstFuncRule.Identity()`: a CRC32 over
  the rule's instrumentation content (target, version, func, recv, before/after
  hooks, path, signature filters). It uses length-prefixed encoding (`len:value`)
  rather than a delimiter, so it stays injective for any content. Type names can
  contain commas or `|`, so no separator is safe to rely on.
- The name-generation sites (`apply_func.go`, `optimize.go`, `trampoline.go`) call
  `rule.Identity()` instead of `util.CRC32(rule.String())`. Identity is the single
  source of every suffix.
- De-duplication (`apply_func.go`, `toolexec.go`): rules with the same identity
  emit their trampoline once instead of redeclaring it. Imports are applied for
  every matching rule first, so a de-duplicated rule never drops imports its hook
  needs.

Do-sequence execution order stays correct through insertion order: the
normalizer expands entries in order and the instrumenter applies them in that
order. Explicit, controllable ordering is deferred (see #583).

## Why content, not position

An earlier revision used a `name#<index>` suffix. As raised in review, that still
collides: `my_hook` `do[1]` renders `my_hook#1`, and a rule named `my_hook#1` at
`do[0]` renders the same string. Content-derived identity avoids this, because the
inputs are full rule bodies rather than a rendered string. The only remaining
collision is between byte-identical rules, which are the same rule and
de-duplicate.

## Testing

- `multiple-hooks-single-func/`: a before-hook (`do[0]`) and an after-hook
  (`do[1]`) on one function emit two distinct trampoline sets in `do` order.
- `dedup-identical-rules/`: two identical rules with different names on one
  function collapse to one trampoline set and compile.
- `TestInstFuncRule_Identity`: do-sequence modifiers stay distinct, the
  `my_hook#1` collision is gone, identical content shares one identity, and
  signature filters participate.
- Func-rule goldens regenerated for the content-derived suffix. Struct, call,
  decl, and raw goldens are untouched.

Verified with `go build ./tool/...`, `make test-unit/tool` (`-race`),
`make check-golden-files`, and `make lint`.

Fixes #544.


amazingakai (commented):
Not a blocking comment, but I wonder if we could derive the trampoline/hookcontext identifier from the rule content itself instead of an index.

The index approach can still collide in cases like:
```yaml
my_hook:
  target: main
  where:
    func: Func1
  do:
    - inject_hooks:
        before: H1Before
        path: "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/myinst"
    - inject_hooks:
        after: H1After
        path: github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/myinst

my_hook_1:
  name: "my_hook#1"
  target: main
  where:
    func: Func1
  do:
    - inject_hooks:
        before: H2Before
        path: github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/myinst
```

In this example, the second `do` entry of `my_hook` and the first `do` entry of `my_hook_1` would both end up producing `my_hook#1`. Admittedly this is a pretty contrived example and most users probably won't write rules like this, but the collision is still possible.

Instead, we could compute a CRC32 from the flat rule content in NewInstFuncRule() and use that when generating trampoline/hookcontext names. That would make the identifiers stable regardless of ordering and avoid collisions caused by generated indices. The only remaining collision case would be when two rules have identical contents, which seems reasonable since they are effectively the same rule. WDYT?

txabman42 (changes requested):
I think @amazingakai idea should be blocking; I'm aligned with it.

Rule name should be based on what it does, not on the position.

amazingakai:
> Rule name should be based on what it does, not on the position.

Right, but we don't necessarily need to append the CRC32 to the rule name. We could store it as an additional field on `InstFuncRule` and use that as a suffix wherever we currently do `util.CRC32(rule.String())`.

tool/internal/rule/func_rule.go
```diff
@@ -86,3 +95,15 @@ func (r *InstFuncRule) validate() error {
...
+// HookContext names. It must differ between two rules that target the same
+// function via separate do-sequence modifiers, otherwise their generated
+// declarations collide. The do-index suffix is only appended for index > 0 so
+// that single-modifier and legacy rules keep their historical names.
+func (r *InstFuncRule) String() string {
+	if r.DoIndex == 0 {
+		return r.Name
+	}
+	return fmt.Sprintf("%s_%d", r.Name, r.DoIndex)
+}
```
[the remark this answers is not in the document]
kakkoyun: Addressed in 554f605: the rendered `_<index>`/`#<index>` suffix is removed entirely. Trampoline/HookContext identity now comes from `InstFuncRule.Identity()` — a CRC32 over the rule's content (target/func/recv/hooks/path/signature) plus its application index — so a rule literally named `<base>_<n>` can no longer stringify-collide with `<base>` at index n.

kakkoyun:
Thanks @amazingakai @txabman42. The issue addressed in 554f605.

amazingakai (approved):
Looks Good ❤️

tool/internal/rule/func_rule.go
```diff
@@ -60,6 +62,20 @@ type InstFuncRule struct {
...
+	// that hooks run in the sequence the user declared. It is not part of the
+	// user-facing schema.
+	//
+	// It is deliberately NOT part of Identity: trampoline/HookContext names are
+	// content-derived, and application order does not change what a rule does.
+	// Order is preserved by the sequence in which the expanded rules are applied,
+	// not by the generated name. The field is retained so that ordering is
+	// available to downstream and future operations without coupling it to name
+	// generation.
+	ApplicationIndex int `json:"application_index,omitempty" yaml:"application_index,omitempty"`
```
y1yang0: `Order` is more simply and consistency

tool/internal/rule/func_rule.go
```diff
@@ -86,3 +102,48 @@ func (r *InstFuncRule) validate() error {
...
+		}
+		return strconv.Itoa(len(xs)) + ";" + strings.Join(encs, "")
+	}
+	encSig := func(s *FuncSignature) string {
+		if s == nil {
+			return "-" // absent: distinct from a present-but-empty signature
+		}
+		return "+" + encList(s.Args) + encList(s.Returns)
+	}
+	parts := []string{
```
y1yang0, line 131: New fields added in the future may also factor into this calculation, we should briefly mention this in the InstFuncRule comment for reference.

tool/internal/rule/normalize.go
```diff
@@ -14,6 +14,15 @@ import (
 const (
 	KeyWhere = "where"
 	KeyDo    = "do"
+
```
y1yang0: It seems like `ApplicationIndex` and `identity` are two separate concepts—one ensures hook order, while the other prevents conflicts. Should they be split into two separate patches?

y1yang0 (approved):

y1yang0 (changes requested):
manually withdraw approval

y1yang0 (approved):

tool/internal/rule/normalize.go
kakkoyun: Split done. Removed `application_index` from this PR since nothing consumed it yet; identity stays content-only. Do-sequence execution order is still preserved by insertion order. Explicit, controllable ordering is tracked in #583.

kakkoyun:
@txabman42 waits for your blessing.

darccio (commented):
LGTM

darccio (approved):

NameHaibinZhang:
LGTM

NameHaibinZhang (approved):

txabman42 (approved):
LGTM! GJ ❤️

PR #563, opened 2026-06-03
test(apps/k8sclient): enable parallelization and force immediate deletion of pods

amazingakai:
## Description

This PR enables `t.Parallel()` for the k8s integration test and forces immediate pod deletion (`GracePeriodSeconds: 0`) to avoid deletion timeouts.

I tested this both locally and in GitHub Actions on my fork:
https://github.com/amazingakai/opentelemetry-go-compile-instrumentation/actions/runs/26896875910

The test appears to be more stable with this approach.

---

## Checklist

* [x] PR title follows [[conventional commits](https://www.conventionalcommits.org/)](https://www.conventionalcommits.org/) format
* [x] Code formatted: `make format`
* [x] Linters pass: `make lint`
* [x] Tests pass: `make test`
* [ ] Tests added for new functionality
* [ ] Tests follow [testing guidelines](docs/testing.md)
* [ ] Documentation updated (if applicable)

PR #574, opened 2026-06-04
test(integration/dbclient): update DB client DSN parsing expectations

amazingakai:
This PR updates the test expectations to match the current DSN parsing logic.

#481 was written before #474 merged and expected the previous DSN parsing behavior.

After both PRs merged, the integration test expectations became stale and started failing in CI.


kakkoyun (approved):

kakkoyun:
@amazingakai I assume this was the culprit broken the CI.

amazingakai:
@kakkoyun Yep.

The CI is still failing, but it is unrelated to this PR. `TestK8SClient` is a bit flaky at the moment and should be fixed by #563 

Rerunning the integration test may make it pass. The failure depends on timing around K3S pod deletion event, so `TestK8SClient` can currently pass or fail depending on when the pod gets deleted.

PR #579, opened 2026-06-05
doc(adr): Instrumentation Ownership and Compatibility

### LINKED ISSUE #500: Meta: compatibility with OTEL reference instrumentations
## Goal

Users should not see meaningful behavioral differences when instrumenting an application with `otelc` instead of the OTEL reference instrumentation for the same package.

For many packages, the reference implementation is likely the corresponding package under [`opentelemetry-go-contrib/instrumentation`](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation), for example `otelhttp`, `otelgrpc`, `otelgin`, etc. In other cases, the reference could be another package or approach accepted by the OTel Go SIG. Part of this work is to identify what the reference should be for each supported library.

The main goal is to stay aligned with the OTel ecosystem over time, ideally by delegating to existing OTEL instrumentation code where possible instead of maintaining separate behavior in this repository.

## Context

Recent discussion in #202 and #459 suggests that rules such as function-call wrapping and `wrap_struct_literal` could let us inject existing `go-contrib` wrappers into user code. That would help us reuse OTEL instrumentation behavior directly and reduce drift.

However, before scaling this approach package by package, we should understand whether it works as a general strategy.

## Questions to answer

- Which OTel instrumentation implementation should be treated as the reference for each supported package?
- Can current rule types plus proposed rules like `wrap_struct_literal` cover most reference instrumentations?
- Where do we still need new join-points, filters, or advice operators?
- Can we avoid changes in other OTel repositories, especially `opentelemetry-go-contrib`?
- If external changes are needed, what are they and why?
- How do we avoid duplicate spans/metrics when both compile-time instrumentation and manually configured OTel instrumentation are present?
- How do we test parity and detect drift over time?

## Proposed checklist

- [ ] Audit packages under `opentelemetry-go-contrib/instrumentation`.
- [ ] For each package, identify the reference implementation and API shape.
- [ ] Map each package to the rule types needed to reuse/delegate to that implementation.
- [ ] List blockers where current/proposed rule types are not enough.
- [ ] Decide whether blockers should be solved in this repo or require upstream OTel changes.
- [ ] Draft an ADR using [`docs/adr/template.md`](docs/adr/template.md) with the chosen strategy.
- [ ] Define a parity test approach comparing `otelc` output with manually configured reference instrumentation.
- [ ] Create focused follow-up issues for concrete blockers or package groups.

txabman42:
Closes: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/500


NameHaibinZhang (approved):

NameHaibinZhang:
LGTM

kakkoyun (approved):
LGTM.

Thanks for documenting this.

PR #580, opened 2026-06-10
refactor(tool): centralize function rule lookup

darccio:
## Description

This PR centralizes function-rule lookup logic in `ast.FindFuncDecl`.

The new helper matches function declarations using the rule target function name, receiver, and optional signature filters. The setup and instrumentation phases now use this shared helper instead of duplicating the lookup and signature-filter checks independently.

This also adds tests covering:

- rule-aware function declaration lookup
- setup matching with function signature filters
- instrumentation behavior when signature filters do not match

## Motivation

Function rules with signature filters should be matched consistently across setup-time rule selection and instrumentation-time application.

Before this change, both phases performed their own combination of function lookup and signature-filter matching. Centralizing that logic reduces duplication and makes future changes to function-rule matching less error-prone.

Follows up on #388.

---

## Checklist

- [x] Tests pass: `make test`


kakkoyun (approved):

tool/internal/instrument/apply_func.go
```diff
@@ -347,18 +347,12 @@ func (ip *InstrumentPhase) parseFile(file string) (*dst.File, error) {
...
-	ok, err := ast.FuncDeclMatchesFilters(funcDecl, rule)
+	funcDecl, ok, err := ast.FindFuncDeclForRule(root, rule)
 	if err != nil {
 		return err
 	}
 	if !ok {
-		ip.Warn("Skipping func rule: signature filters did not match",
-			"rule", rule.Name, "func", rule.Func)
-		return nil
+		return ex.Newf("can not find function %s", rule.Func)
```
txabman42, line 355: I would try to make FindFuncDeclForRule diff between `func not found` and `signature not match`. Same as before. Better for debugging.

txabman42 (approved):
LGTM! ❤️ Just minor nitpick

tool/internal/instrument/apply_func.go
darccio: Thanks @txabman42, I'll apply it!

y1yang0 (commented):
Perhaps we don't need to keep both `FindFuncDeclForRule` and `FindFuncDecl`—just `FindFuncDecl` should suffice.

darccio:
@y1yang0 Done in 2a274cadd89d2b1392cbbb9a8a06958e3b9a30ed. Let me know how it looks to you. I used generics because `FuncFuncDecl` is used for raw rules.

y1yang0 (approved):

PR #581, opened 2026-06-12
test: add versionmatrix to verify declared version range bounds

### LINKED ISSUE #578: Add a release-gating version-matrix test to verify supported ranges
### Background
We already test forward compatibility against new releases via LatestLibBuild and LatestLibRun (see [docs/testing.md](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/testing.md)). But these only test @latest; they never exercise the inside of a declared range like v1.2.3,v4.5.6. We assume those ranges are correct rather than verifying them (related: [#406](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/406)).

### Proposal
Add a release-gating matrix test that, for each instrumented library's declared version range, runs the existing integration suite against a small matrix of concrete in-range versions (at least lower bound + upper bound). 

Define when to run it. It could run as a release gate before each cycle, for example.

Giving a clear pass/fail that every declared range is actually exercised end to end. A failure means the declared range is incorrect, with the same range-splitting remediation as in the existing latest-lib tests.

waqar2403:
## Description

This adds the VersionMatrix test tier. LatestLibBuild and LatestLibRun only exercise `@latest` so the declared version ranges in the rules were assumed correct but never verified. This tier pins each test app's instrumented deps to the bounds of the declared ranges and runs the existing integration suite once per bound:

1. Lower pass: pins each dep to the lowest declared lower bound.
2. Upper pass: pins each dep to the highest published release covered by any declared range.

Ranges are half open, the cap is the first unsupported version. So for k8s client go `v0.34.0,v0.36.0` the upper bound tested is v0.35.5, not v0.36.0 itself. For open ended ranges the upper bound is the latest release so we keep testing latest as discussed in the issue. It runs weekly plus `workflow_dispatch` same cadence as LatestLibRun and opens an issue labelled `versionmatrix-failure` when it fails on main.

Two notes:

1. The lower pass currently fails for the otel rules, which I believe is the test doing its job. `hooks.yaml` declares v1.39.0 but `pkg/instrumentation/shared/go.mod` requires otel v1.43.0, and the shared module is injected into every instrumented build, so any app pinned below v1.43.0 gets forced up by MVS and otelc fails on the version mismatch. The declared v1.39.0 floor is not buildable as declared. The fix is either raising the bound in `hooks.yaml` to v1.43.0 or lowering the shared and hook module requirements. I would lean towards raising the bound since fixing the declared range is the remediation the issue describes happy to do that in this PR or as a follow up whichever you prefer.
2. The `versionmatrix-failure` label needs to be created in the repo before the notify job can open issues.

Fixes #578


docs/testing.md
```diff
@@ -133,6 +137,27 @@ A failure means that the latest release of an upstream library introduced a **ru
...
+
+| Category | Location | Build Tag | Scope |
+| :------- | :------- | :-------- | :---- |
+| VersionMatrix | `test/versionmatrix/` | `versionmatrix` | Run integration suite at the declared lower and upper bound of each library's version ranges. |
+
+Version ranges are half-open (`v0.34.0,v0.36.0` supports `>= v0.34.0` and `< v0.36.0`), so the upper bound tested for a capped range is the newest release **below** the cap, and the upper bound of an open-ended range is the latest release — the latter deliberately overlaps LatestLibRun. Note that a capped range whose library has already released past the cap (e.g. `v0.34.0,v0.36.0` when `@latest` is `v0.36.1`) is skipped by both latest-lib tests, so this test is the only one exercising it at all.
```
txabman42: I would also mention that if there are two ranges, we dont take lower/upper limits per rule, but per dependency.

txabman42: Nitpick: If upperbound returns latest and it fails, AFAIU its duplicated with `LatestLibRun` and would generate 2 issues. Could be interesting to skip those cases for matrix tests just to avoid duplicates.

txabman42 (changes requested):
Nice PR, just some minor changes.

Regarding:
> The lower pass currently fails for the otel rules

I would try to fix it here (bump lower version?) and see if that error disappears.

waqar2403:
@txabman42 Pushed the changes:

- Bumped the otel SDK floor in `hooks.yaml` up to v1.43.0, which is the version the injected instrumentation modules actually need. So the lower pass is green now and that build error is gone.
- Wrote it down in `docs/testing.md` that bounds are taken per dependency and not per rule.
- Added the skip for when the upper bound resolves to latest so the matrix doesn't end up duplicating LatestLibRun. With that in place otel only gets tested at its lower bound and k8s still gets its in-range upper bound (v0.35.x).

One thing worth flagging while we are here. Because bounds are per dependency, when a rule gets split into an old capped range plus a new open one (the v1 + v2 model) the split boundary, which is the new rule's floor doesn't get tested on its own. It is fine for the rules we have right now but if you want the split points covered I can go ahead and test each rule's lower bound too. Happy either way.

Also a heads-up the `versionmatrix-failure` label needs to be created in the repo first before the notify job can open any issues.

txabman42:
@waqar2403 yes, this could be a good moment to test always the limits per rule, not dependency. Now that we dont have too many, we cant push this and see how it scales this matrix tests.

waqar2403:
Done, pushed the per-rule version. It now takes the lower and upper bound of each rule's range separately and runs the integration suite once per distinct version (a tier), so the run count grows with the number of rules. For the rules we have today that's 3 tiers: k8s at v0.34.0, v0.35.0 (the second rule's floor, which wasn't being hit before) and v0.35.6, plus otel at v1.43.0. Upper bounds that resolve to latest are dropped so we don't duplicate LatestLibRun. Updated docs/testing.md to match.

instrumentation/go.opentelemetry.io/otel/hook/otelc.yaml
```diff
@@ -1,6 +1,8 @@
+# The otel SDK floor tracks the version the injected instrumentation modules
+# require (pkg/instrumentation/shared); raise it together with those modules.
 trace_context:
   target: go.opentelemetry.io/otel/sdk/trace
-  version: "v1.39.0"
```
txabman42, line 3: Updated from `v1.39.0` to `v1.43.0` as they were not supported by this rule.

txabman42 (approved):
LGTM! ❤️

test/versionmatrix/versionmatrix_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/test/testutil"
+)
+
+// appBoundVersions returns, per test-app directory, the per-rule boundary
+// versions of its instrumented direct dependencies (see testutil.BoundVersions).
+// Apps with no ranged dependencies are omitted.
+func appBoundVersions(t *testing.T) map[string]map[string][]string {
+	appsRoot := filepath.Join("..", "apps")
+	rulesRoot := filepath.Join("..", "..", "pkg", "instrumentation")
```
darccio: Shouldn't this use a different path?


```suggestion
	rulesRoot := filepath.Join("..", "..", "instrumentation")
```

darccio (changes requested):

test/versionmatrix/versionmatrix_test.go
darccio: I just confirmed that yes, it should be on `instrumentation`, not `pkg/instrumentation`: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/509#issuecomment-4798652261

.github/scripts/versionmatrix-issue.sh
```diff
@@ -0,0 +1,56 @@
...
+    --label "${LABEL}" \
+    --body "$(cat <<EOF
+The [VersionMatrix workflow](${RUN_URL}) failed on \`main\`.
+
+This means a declared version range in \`pkg/instrumentation/.../*.yaml\` does **not** work at one of its bounds: either the instrumented library version cannot be installed (another module in the build graph forces a different version) or the integration suite fails against it.
+
+## Remediation
+
+1. Identify the failing test and the library version that caused the break.
+2. Fix the rule's version range in the relevant \`pkg/instrumentation/.../*.yaml\` file: raise the lower bound to the oldest version that actually works, or cap the range below the first version that breaks.
```
darccio:
```suggestion
This means a declared version range in \`instrumentation/.../*.yaml\` does **not** work at one of its bounds: either the instrumented library version cannot be installed (another module in the build graph forces a different version) or the integration suite fails against it.

## Remediation

1. Identify the failing test and the library version that caused the break.
2. Fix the rule's version range in the relevant \`instrumentation/.../*.yaml\` file: raise the lower bound to the oldest version that actually works, or cap the range below the first version that breaks.
```

docs/testing.md
```diff
@@ -160,6 +164,27 @@ A failure means that the latest release of an upstream library introduced a **ru
...
+
+Per [ADR-0004](adr/0004-instrumentation-ownership-and-compatibility.md), the supported window per library is the last two major versions; that policy bounds the matrix this test has to cover.
+
+### What a failure means
+
+A failure means a declared version range does **not** actually work at one of its bounds: the bound cannot be installed (another module in the build graph, such as the instrumentation hook's own `go.mod`, forces a newer version), no published release is covered by the range at all, or the integration suite fails against the pinned version. The remediation is:
+
+1. Fix the rule's range in the relevant `pkg/instrumentation/.../*.yaml` file: raise the lower bound to the oldest version that actually works, or cap the range below the first version that breaks.
```
darccio:
```suggestion
1. Fix the rule's range in the relevant `instrumentation/.../*.yaml` file: raise the lower bound to the oldest version that actually works, or cap the range below the first version that breaks.
```

instrumentation/go.opentelemetry.io/otel/hook/otelc.yaml
```diff
@@ -1,6 +1,8 @@
+# The otel SDK floor tracks the version the injected instrumentation modules
+# require (pkg/instrumentation/shared); raise it together with those modules.
```
darccio:
```suggestion
# require (instrumentation/shared); raise it together with those modules.
```

darccio (changes requested):

darccio (approved):

PR #584, opened 2026-06-16
feat(tool): add import-based instrumentation configuration support

### LINKED ISSUE #494: Building projects with older grpc versions than instrumentation dependencies causes build failures

### LINKED ISSUE #567: feat: Remote integration support (a.k.a `otelc.tool.go` file support)
> **Original:** This file should help us to define the required dependencies for the instrumentations. Similar logic to [orchestrion.tool.go](https://github.com/DataDog/orchestrion/blob/main/instrument/orchestrion.tool.go).

## Summary

Add a user-facing, source-controlled file that **declares which instrumentations a project enables**, driving compile-time instrumentation from explicit imports rather than the current implicit all-or-nothing matching. This mirrors DataDog Orchestrion's `orchestrion.tool.go` mechanism and realizes the design already sketched in [`docs/ux-design.md`](docs/ux-design.md). 

**This will be eventually super useful to automagically add the remote integrations from any repo using go modules.**

We will support **both** filenames:
- **`otel.instrumentation.go`** — canonical name used throughout `docs/ux-design.md`.
- **`otelc.tool.go`** — alias matching this issue's title and the tool's `otelc.`-prefixed artifacts (`otelc-pkg.gz`, `otelc.runtime.go`, `.otelc-build/`).

The file follows the standard Go ["tools.go" idiom](https://pkg.go.dev/cmd/go#hdr-Tool_dependencies) — a `//go:build tools`-guarded file whose blank imports enumerate the instrumentation packages to activate:

```go
//go:build tools

package tools

import (
    _ "github.com/open-telemetry/opentelemetry-go/instrumentation/db"    // database instrumentation
    _ "github.com/open-telemetry/opentelemetry-go/instrumentation/http"  // net/http client + server
)
```

Because the file is guarded by a build tag, it never compiles into the binary — but `go mod` tracks those imports as real dependencies, giving reproducible, auditable builds.

### Today (all-or-nothing)

The setup phase (`tool/internal/setup/`) runs a dry build (`go build -a -x -n`, `find.go`), loads **every** embedded rule from `otelc-pkg.gz` (`loadDefaultRules`, `match.go`), and matches them against **every** discovered dependency. The only escape hatch is `--rules` / `OTELC_RULES` — a full *replacement* of the rule set. There is no declarative "enable these" surface; third-party packages cannot ship their own rules.

A runtime-only `OTEL_GO_ENABLED_INSTRUMENTATIONS` env var exists inside generated hook code (`pkg/instrumentation/shared/otel_setup.go`) but only gates telemetry emission at runtime — it has no effect on what the compiler instrumented.

### Reference: Orchestrion (import-driven, vendor-agnostic)

DataDog's [Orchestrion](https://github.com/DataDog/orchestrion) uses the same `.tool.go` idiom:
1. **Parse** `orchestrion.tool.go` with `go/parser` + `parser.ImportsOnly` (build tag intentionally ignored).
2. **Resolve** each import path to a package via `golang.org/x/tools/go/packages`.
3. **Recursively discover** each resolved package's `orchestrion.yml` (instrumentation config) and nested `orchestrion.tool.go` files, aggregating the full aspect set.
4. A **`pin` command** scaffolds/prunes/validates the file, wires `go.mod`, and runs `go mod tidy`. Auto-pin runs during a build if the file is absent.

This makes Orchestrion **vendor-agnostic**: any third-party package can ship instrumentation by including an `orchestrion.yml`, activated by a single blank import in the user's tool file.

### Target

Adopt the import-driven model so that:
- Projects **explicitly and reproducibly** declare their enabled instrumentations.
- Instrumentation config can be **vendored by any package** that ships an `otel.instrumentation.yml`, not only rules embedded in `otelc-pkg.gz`.

This also realizes the three configuration styles already designed in `docs/ux-design.md` §"Configuration Styles" (lines 151–181):

| Style | Mechanism | Go version |
|---|---|---|
| `otel.instrumentation.go` / `otelc.tool.go` | blank-import tools file | ≥ 1.23 |
| `.otel.yml` | YAML file next to `go.mod` | any (CI/CD friendly) |

### File discovery & parsing

- [ ] The tool locates a config file in the **module root** (directory containing `go.mod`), accepting **both** `otel.instrumentation.go` and `otelc.tool.go`. If both exist, behavior is explicitly defined (error or documented precedence), not a silent pick.
- [ ] Parsing uses `go/parser` with `parser.ImportsOnly` and **ignores the `//go:build tools` constraint** — the file is read even though it never compiles into the binary.
- [ ] All blank imports (`_ "path"`) are extracted. Handling of non-blank imports is documented (kept or rejected with a diagnostic).
- [ ] An absent, empty, or zero-import file is handled gracefully — no panic; falls through to the documented fallback behavior (criterion E).

### Import → config resolution (vendor-agnostic)

- [ ] Each extracted import path is resolved to a package using `golang.org/x/tools/go/packages` (consistent with existing setup-phase usage).
- [ ] For each resolved package, the tool discovers that package's `otel.instrumentation.yml` **and** any nested `otel.instrumentation.go` / `otelc.tool.go`, **recursively**, aggregating the resulting instrumentation config.
- [ ] Resolution works for packages **not** bundled in `otelc-pkg.gz` (third-party / external), proving the vendor-agnostic extension path.
- [ ] An imported package that contains no valid instrumentation config produces a clear, actionable diagnostic (and is a prune candidate per criterion D).

### Integration with the matching engine

- [ ] When a config file is present, the declared import set **scopes** which instrumentations are active, replacing all-or-nothing default matching.
- [ ] ~~Precedence across all config sources is explicitly defined, documented, and covered by tests: `otel.instrumentation.go` / `otelc.tool.go` imports → `go tool` deps → `.otel.yml` → `--rules` / `OTELC_RULES` → embedded defaults.~~
- [ ] Aggregated config feeds the existing `matchDeps` flow (`tool/internal/setup/match.go`); the `otelc.runtime.go` and `matched.json` outputs remain correct for the selected set.
- [ ] The existing "Warning: no instrumentation will be applied" path (`match.go:472`) remains correct and is not silently swallowed.

### `pin`-style command (scaffold / prune / validate)

- [ ] A new CLI subcommand (`otelc pin`) **creates** the config file if missing, with the correct `//go:build tools` tag, `package tools`, a descriptive header comment, and a `//go:generate` directive to re-run itself.
- [ ] Flags: `-prune` (default on) removes imports whose packages contain no valid config; `-validate` validates each discovered `*.instrumentation.yml` against the schema; `-generate` manages the `go:generate` directive. Defaults and behavior are documented.
- [ ] The command edits `go.mod` (adds `require` / tool deps, bumps the `go` directive if needed) and runs `go mod tidy`, reusing the replace + tidy logic in `tool/internal/setup/sync.go` where applicable. File writes are atomic (temp file + rename); comments and formatting are preserved (using `dave/dst`).
- [ ] Re-running `pin` on an already-pinned project is **idempotent** — it produces no spurious diffs.

### Auto-pin / clean-room fallback

- [ ] If no config file exists, `otelc go build` self-configures by analyzing the full dependency tree, preserving today's behavior and the clean-room usage mode described in `docs/ux-design.md` §"Clean-Room Usage".
- [ ] Auto-pin emits a clear, user-readable notice describing what it will create/modify (file, `go get`, `go mod tidy`) before acting.
- [ ] Clean-room mode leaves **no** persistent file-system changes after the build completes.

### Alternate config styles

- [ ] ~~`go tool` dependencies (go ≥ 1.24) are recognized as an equivalent declaration source, enabling the simpler `go tool otelc go build` invocation (`ux-design.md:167–170`).~~
- [ ] A `.otel.yml` file placed next to `go.mod` injects project-local instrumentation config using the same schema as `otel.instrumentation.yml`, applied only when building that module's own packages (not when the module appears as a dependency).

### Documentation & tests

- [ ] `docs/ux-design.md` is reconciled with shipped reality: document that both filenames are accepted; note any divergence between the doc's AoP `pointcut`/`advice` schema and the implemented `target`/`where`/`do` rule schema (`docs/rules.md`, ADR-0003).
- [ ] `docs/getting-started.md` and `docs/rules.md` show the new declarative workflow end-to-end.
- [ ] Unit tests cover: parsing (build-tag-ignored, blank imports, both filenames, malformed file), recursive config resolution, precedence ordering, and `pin` idempotency / prune / validate.
- [ ] An integration test (`test/`) builds a demo application whose enabled instrumentations come **only** from a checked-in `otel.instrumentation.go`, and asserts that the declared instrumentation **is** applied and that an un-declared instrumentation is **not** compiled in.

## Out of scope / follow-ups
- Removing `otelc-pkg.gz` embedding entirely (the declarative model may eventually supersede it).

## References

- Orchestrion: https://github.com/DataDog/orchestrion | `orchestrion pin` docs: https://datadoghq.dev/orchestrion/docs/getting-started/
- This repo: [`docs/ux-design.md`](docs/ux-design.md), [`docs/rules.md`](docs/rules.md), [`docs/adr/0003-structured-rule-schema.md`](docs/adr/0003-structured-rule-schema.md)
- Key source files: `tool/internal/setup/{find,match,sync,add}.go`, `tool/internal/rule/`, `tool/cmd/`

amazingakai:
## Summary

Adds support for import-based instrumentation configuration through `otel.instrumentation.go` (and the compatibility alias `otelc.tool.go`).

Instrumentations are now discovered from imported instrumentation packages rather than being loaded exclusively from the built-in rule set. Instrumentation packages may recursively compose other instrumentation packages through their own tool files.

To preserve the existing zero-configuration workflow, `otelc go build` continues to support clean-room usage. When no instrumentation file is present, `otelc` generates a temporary `otel.instrumentation.go` derived from the application's dependency graph for the duration of the build and removes it afterwards.

This PR also adds a new `otelc pin` command for generating, validating, and maintaining instrumentation configuration files.

As a side effect, dependency synchronization (`go mod tidy`) now happens during the pinning stage. This ensures rule matching always operates on the final dependency graph (fixing #494).

Closes #567 

Closes #494

## Future Work

#585 -  decouple built-in instrumentations from `otelc-pkg.gz`
#586 - move `pkg/instrumentation/basic` to demo applications
#587 - add support for `.otel.yml` instrumentation configuration


tool/util/sys.go
```diff
@@ -163,6 +163,52 @@ func WriteFile(filePath, content string) error {
 	return nil
 }
 
+// WriteFileAtomic writes data to a file atomically by first writing to a temporary file and then renaming it.
+// If the target file already exists, it preserves its permissions. Otherwise, it uses the provided permissions or defaults to 0644.
```
darccio, line 167: I think the second sentence should be rewritten to reflect the actual precedence (explicit > existing > default):

```suggestion
// WriteFileAtomic writes data to a file atomically by first writing to a temporary file and then renaming it.
// Permission precedence: explicit perm argument > existing file's permissions > default 0644.
```

tool/internal/setup/add.go
darccio, line 126: I think this should be updated to use `ast.WriteFileAtomic`. WDYT?

tool/internal/setup/config.go
```diff
@@ -0,0 +1,179 @@
...
+
+		for _, imp := range f.Imports {
+			if imp.Name == nil || imp.Name.Name != ast.IdentIgnore {
+				continue
+			}
+
+			importPath, unqoteErr := strconv.Unquote(imp.Path.Value)
+			if unqoteErr != nil {
+				return ex.Wrapf(unqoteErr, "failed to unquote import path %s in %s", imp.Path.Value, toolFile)
+			}
```
darccio, line 151: Nit:

```suggestion
			importPath, unquoteErr := strconv.Unquote(imp.Path.Value)
			if unquoteErr != nil {
				return ex.Wrapf(unquoteErr, "failed to unquote import path %s in %s", imp.Path.Value, toolFile)
			}
```

tool/internal/setup/config.go
```diff
@@ -0,0 +1,179 @@
...
+var ErrNotInstrumentation = errors.New("not an instrumentation package")
+
+func findToolFile(moduleDir string) (string, error) {
+	canonical := filepath.Join(moduleDir, ToolFileCanonical)
+	alias := filepath.Join(moduleDir, ToolFileAlias)
+
+	canonicalExists := util.PathExists(canonical)
+	aliasExists := util.PathExists(alias)
+
+	switch {
```
darccio, line 44: Wrapping with `ErrNotInstrumentation` here causes `walkInstrumentation` callers (in both `match.go` and `pin.go`) to treat this as a prunable import. They'll log a warning and continue rather than hard-failing. Since the error message says "only one instrumentation config file is allowed," a misconfigured instrumentation package would be silently skipped during `otelc go build`. Is this what we really want? If not, consider returning a plain `ex.Newf(...)` (not wrapped in `ErrNotInstrumentation`) so callers propagate it as a fatal error.

tool/internal/setup/match.go
```diff
@@ -421,14 +420,30 @@ func (sp *SetupPhase) loadRules() ([]rule.InstRule, error) {
...
+		toolFile, err := findToolFile(moduleDir)
+		if err != nil {
+			return nil, err
+		}
+		if toolFile != "" {
+			sp.Debug("rules source: instrumentation config file (%s)", toolFile)
+			toolFiles = append(toolFiles, toolFile)
+		}
+	}
+	return loadRulesFromToolFiles(ctx, toolFiles)
```
darccio, line 437: Should we implement here the fallback for `.otel.yml` files if `loadRulesFromToolFiles` return an empty slice?

tool/internal/setup/pin.go
```diff
@@ -0,0 +1,535 @@
...
+	// Nothing passed? Do not touch the directive at all.
+	if opts.Generate == nil {
+		return
+	}
+
+	comments := make([]string, 0, len(f.Decs.Start.All()))
+	for _, comment := range f.Decs.Start.All() {
+		// Drop any existing otelc pin directive. We will either
+		// regenerate it or remove it entirely.
+		const goGenerateDirectivePrefix = "//go:generate go run github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/cmd pin"
```
darccio, line 106: We should use `OtelcToolCmdRoot` here too.


```suggestion
		const goGenerateDirectivePrefix = strings.Join(args[:len(args)-1], "")
```

tool/internal/setup/pin.go
```diff
@@ -0,0 +1,535 @@
...
+		}
+
+		// Assume go.mod is in the same directory as the yaml.
+		goModPath := filepath.Join(filepath.Dir(path), "go.mod")
+		modFile, parseErr := parseGoMod(goModPath)
+		if parseErr != nil {
+			return ex.Wrapf(parseErr, "loading go.mod for %s", path)
+		}
+
+		loaded[modFile.Module.Mod.Path] = make([]yamlRule, 0, len(rules))
```
darccio, line 213: `loaded[modFile.Module.Mod.Path] = make([]yamlRule, 0, len(rules))` resets the slice unconditionally on every yaml file processed. If two rule files in the embedded `otelc-pkg.gz` ever share the same module root (e.g. `otelc.yaml` and `db.otelc.yaml` both in a module), the second file's walk iteration silently discards the first file's rules — only the last file processed for that module path survives.

The fix could be to replace the reset with an append:

```go
// before
loaded[modFile.Module.Mod.Path] = make([]yamlRule, 0, len(rules))
for _, r := range rules {
    ...
    loaded[modFile.Module.Mod.Path] = append(loaded[modFile.Module.Mod.Path], r)
}

// after
for _, r := range rules {
    if r.Target == "" {
        continue
    }
    loaded[modFile.Module.Mod.Path] = append(loaded[modFile.Module.Mod.Path], r)
}
```

This isn't triggered today since each instrumentation subpackage has exactly one rule file, but it's a silent footgun for anyone adding a second yaml to an existing module.

tool/internal/setup/pin.go
```diff
@@ -0,0 +1,535 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return loaded, nil
+}
+
+func ensureOtelcRequire(moduleDir, version string) (bool, error) {
+	if module.IsPseudoVersion(version) {
```
darccio, line 231: `module.IsPseudoVersion("v0.0.0")` returns false. This leads to the code below stamping `require ... v0.0.0`. Maybe we should guard `version == "v0.0.0"` along these pseudo-version check.

tool/internal/setup/pin.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	imports := matchInstrumentationImports(deps, ruleset)
+
+	// Nothing to instrument? Warn and skip generating tool file.
+	if len(imports) == 0 {
+		_, _ = fmt.Fprintln(os.Stderr, "Warning: no instrumentations matched")
+		logger.WarnContext(ctx, "no instrumentations matched, skipping generating tool file")
+		return &PinResult{AllDeps: deps}, nil
+	}
```
darccio, line 437: When `matchInstrumentationImports` returns no matches, `generatePinnedProjects` returns early with no tool file written and no error. The caller (`AutoPin`) already pre-computed `filesToRemove` before calling `Pin`, so the cleanup closure will attempt `os.Remove` on a file that was never created. This fails silently with a logged warning, which is benign.

The more user-visible consequence is that `loadRules` is then called with no tool files, returns zero rules, and the build proceeds completely uninstrumented with no diagnostic beyond the `Warning: no instrumentations matched` line on stderr. A user running `otelc go build` for the first time on a project with no matching dependencies would get a silent no-op build with no indication of why.

Consider surfacing this more clearly. For example, printing which dependencies were scanned and which instrumentation packages were considered, or at minimum making the warning more actionable (e.g. "no instrumentations matched for the current dependency graph; run `otelc pin --validate` to diagnose").

tool/internal/setup/pin.go
```diff
@@ -0,0 +1,535 @@
...
+		return nil, walkErr
+	}
+
+	for toolFile, prunedImports := range toolFiles {
+		if err := updateToolFile(ctx, toolFile, prunedImports, opts); err != nil {
+			return nil, err
+		}
+	}
+
+	return &PinResult{}, nil
```
darccio, line 408: `updatePinnedProjects` returns `&PinResult{}` with empty `AllDeps`. `Setup` seems to use this to decide whether to re-run `findDeps`. Since `AllDeps` is nil, it re-runs. `generatePinnedProjects` returns `AllDeps` populated. This asymmetry looks intentional (update path doesn't re-parse the build plan), but it means the update path always pays the `findDeps` cost again in Setup. Is this the intention?

tool/internal/setup/pin.go
```diff
@@ -0,0 +1,535 @@
...
+type PinResult struct {
+	AllDeps []*Dependency
+}
+
+// Pin generates or updates otel.instrumentation.go to pin instrumentation packages for the current module.
+// It also returns a PinResult, which may optionally include a list of all dependencies found in the project
+// (not just those with instrumentation). The PinResult can be used by later stages (setup) instead of
+// re-parsing compile commands again.
+func Pin(ctx context.Context, opts PinOptions) (*PinResult, error) {
+	pkgs, getErr := getBuildPackages(ctx, opts.Args)
```
darccio, line 477: `getBuildPackages` is called twice. Once here and another in `Setup` (line 231). It might have performance impact on large dependency graphs.

tool/internal/setup/setup.go
```diff
@@ -229,68 +233,57 @@ func Setup(ctx context.Context, cmd *cli.Command) error {
...
+
+	if deps == nil {
+		// Find all dependencies of the project being build
+		deps, err = findDeps(ctx, args)
+		if err != nil {
+			return ex.Wrapf(err, "finding dependencies")
+		}
+
+		// Extract the embedded pkg module into local directory
+		err = extractOtelcPkg()
```
darccio, line 261: `extractOtelcPkg` is called here on the update path (when AutoPin found an existing tool file and returned `nil` AllDeps). However, extraction is only needed by `syncDeps`, which adds `replace` directives pointing into the extracted pkg directory. The update path in `generatePinnedProjects` calls `syncDeps`, but `updatePinnedProjects` does not. So for a project with a committed `otel.instrumentation.go`, every `otelc go build` extracts the embedded gz unnecessarily.

This is not a correctness issue (extraction is idempotent), but it adds avoidable I/O on the hot path. Consider moving `extractOtelcPkg` to just before `syncDeps` in `generatePinnedProjects`, and removing it from this branch entirely.

test/integration/pin_test.go
```diff
@@ -0,0 +1,323 @@
+//go:build integration
```
darccio, line 1: There's no test for the recursive composition path: an instrumentation package whose tool file imports another instrumentation package, and verifying that both sets of rules are loaded.

darccio (changes requested):

tool/util/sys_test.go
```diff
@@ -426,3 +429,66 @@ func TestCopyFileSameFile(t *testing.T) {
...
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			path := filepath.Join(t.TempDir(), "test.txt")
+
+			if tt.initialData != nil {
+				require.NoError(t,
+					os.WriteFile(path, tt.initialData, tt.initialPerm),
+				)
+			}
```
darccio, line 473: The test creates the initial file with `os.WriteFile(path, data, 0o755)`, but `os.WriteFile` applies the process umask. On my machine umask is 0o077, so the file lands at 0o700. `WriteFileAtomic` then reads back 0o700 via `os.Stat`, correctly preserves it, and the test assertion fails because it expected 0o755.

```suggestion
			if tt.initialData != nil {
				require.NoError(t,
					os.WriteFile(path, tt.initialData, tt.initialPerm),
				)
			}
			require.NoError(t, os.Chmod(path, tt.initialPerm))
```

tool/internal/setup/pin.go
```diff
@@ -0,0 +1,535 @@
...
+			// Always recurse into nested tool files when they exist.
+			return true, nil
+		},
+	)
+	if walkErr != nil {
+		return nil, walkErr
+	}
+
+	for toolFile, prunedImports := range toolFiles {
+		if err := updateToolFile(ctx, toolFile, prunedImports, opts); err != nil {
```
txabman42, line 403: This always rewrites the `otel.instrumentation.go`. In case an import gets pruned or some formatting misalignment, the `otelc go build` would modify the committed `otel.instrumentation.go` file, which goes against the general rule of not modifying files.

tool/internal/setup/config.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, ex.Newf("errors loading package %s: %v", importPath, pkg.Errors)
+	}
+
+	pkgDir := pkgload.GetPackageDir(pkg)
+	if pkgDir == "" {
+		return nil, ex.Newf("could not determine directory for package %s", importPath)
+	}
+
+	// Instrumentation packages must be module roots.
+	goModPath := filepath.Join(pkgDir, "go.mod")
```
txabman42, line 86: We are forcing to have module at root level. I would make it more flexible, I think orchestrion also allows importing non-root modules cc @kakkoyun @darccio 

I'm thinking allowing cases like:

```
github.com/foo/otel/
├── go.mod
├── redis/
│   ├── redis.go
│   └── rules.otelc.yaml
└── kafka/
    ├── kafka.go
    └── rules.otelc.yaml
```

```go
import _ "github.com/foo/otel/redis"
```

txabman42 (changes requested):
I'm aligned with the 0005 adr. Also, the `go mod tidy` during a pin stage is a good fix for https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/494.

Couple of comments +

This PR does too many things.  Hard to review + multiple features added. We can split this into some few PRs:
    -  ADR --> so we can discuss and accept the ADR before the implementation
    - Rename  `*.yaml` -> `otelc.yaml` across instrumentation packages
    - `main.Version` -> `tool/util.Version` + .gitattributes
    - main implementation --> tool-file discovery, parsing, and rule loading
    -  otelc pin + auto-pin / clean-room

tool/internal/setup/config.go
kakkoyun: Yes, this should be allowed 👍 However, I think we should discourage patterns like this for the instrumentation we provide. I don't want community to create bundles of instrumentation. However, I can see several use case where these could depend each other.

docs/adr/0005-import-driven-instrumentation-selection.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Today, `otelc` loads all built-in instrumentation rules and matches them against the application's dependency graph. This makes instrumentation selection implicit: if a dependency is present and a matching rule exists, that instrumentation may be applied.
+
+This model works for the built-in rule set, but it does not scale well to external instrumentations. While `otelc` already supports replacing the default rules through `--rules`, that mechanism is intended for supplying an alternate rule set rather than declaratively enabling a set of instrumentations. Users cannot explicitly declare which instrumentations they want to enable at compile time, and third-party modules cannot easily distribute their own instrumentation packages without being embedded into `otelc` itself.
```
kakkoyun, line 13: 👍 💯

docs/adr/0005-import-driven-instrumentation-selection.md
```diff
@@ -0,0 +1,40 @@
...
+Instrumentation packages are resolved using `go/packages`. A package is considered an instrumentation package if it contains either an instrumentation tool file or one or more `*.otelc.yml` rule files. Instrumentation packages may themselves import other instrumentation packages, allowing instrumentation dependencies to be composed recursively.
+
+Rule discovery is scoped to the instrumentation package's own module root and does not recurse into nested Go modules.
+
+If no tool file exists, `otelc` preserves the current "clean-room" workflow by automatically generating a temporary instrumentation configuration from the application's dependency graph for the duration of the build. Users who want a persistent, source-controlled configuration can create one explicitly using `otelc pin`.
+
+## Consequences
+
+* Instrumentation selection becomes explicit and visible in source control, rather than being inferred solely from the dependency graph.
+* Third-party modules can distribute their own `otelc`-compatible instrumentations without changes to the `otelc` repository.
```
kakkoyun, line 37: This is our ultimate goal here. It deserve an emphasis.

docs/adr/0005-import-driven-instrumentation-selection.md
```diff
@@ -0,0 +1,40 @@
...
+* Instrumentation packages become regular Go module dependencies and participate naturally in `go mod tidy` and existing dependency management workflows.
+* The `otelc pin` command can create, update, and validate the tool file while preserving the existing zero-configuration experience through auto-pin.
+* Tradeoff: `otelc` now loads the built-in default rules by first generating a temporary `otel.instrumentation.go` when no user-provided tool file exists, adding an extra intermediate step to the default rule-loading process. This moves dependency synchronization (`go mod tidy`) into the pinning stage, ensuring that rule matching always operates on the final dependency graph and avoiding stale rule sets after dependency changes. As a side effect, built-in and third-party instrumentations now share the same loading path while preserving the existing zero-configuration workflow.
```
kakkoyun, line 40: 👍

kakkoyun (commented):
Thanks for taking care of this. Very decent PR. I need some more time to review this. 

There are already comments from @darccio and @txabman42 that needs to be addressed.

kakkoyun:
@amazingakai the split is landing well: ADR-0005 (#605), the `otelc.yaml` renames (#606), and the version-helpers move (#607) all merged today. This PR now needs a substantial rebase. It still carries those merged pieces (the ADR doc, the `otelc.yaml` files, `cmd_version.go`), and the `pkg/instrumentation` to `instrumentation/` move (#597) and the command-package move (#590) both touch it. Could you rebase it down to just the remaining core: tool-file discovery/parsing/rule loading, `otelc pin`/auto-pin, and `otel.instrumentation.go` support? If that's still large, splitting core from pin (per @txabman42) would make it easier to review.

After the rebase there are 19 unresolved threads to work through. The substantive ones from @darccio are in `pin.go`: the `AllDeps` nil/populated asymmetry (:408), `getBuildPackages` called twice (:477), the `module.IsPseudoVersion("v0.0.0")` guard (:231), the unconditional slice reset (:213), and the stray `extractOtelcPkg` on the update path (`setup.go:261`). @txabman42 also raised the root-module restriction (`config.go:86`) and the rewrite of the committed `otel.instrumentation.go` (`pin.go:403`).

This carries #567 (config model), which is a freeze item, so it's the highest priority of the v1 blockers. Happy to do a focused re-review once it's rebased.

amazingakai:
@kakkoyun I'm currently working on splitting out the core changes in #612, it's almost ready and I'll mark it ready for review soon. After #612 is merged, I'll rebase this PR and address the remaining review comments.

kakkoyun:
Closing this one for the sake of #612 (and others that have been already merged)

PR #594, opened 2026-06-22
test(integration): avoid building unused applications during targetet integration test runs

### LINKED ISSUE #588: test(integration): avoid building unused applications during targeted integration test runs
## Problem

The integration test suite currently builds test applications from `main_test.go` before running the tests. This improves performance when many tests reuse the same binaries, but it also means running a single test (for example `go test -run TestK8SClient`) still builds every other integration test application, even though they are not needed by the test.

This increases local iteration time and makes targeted test runs unnecessarily expensive.

## Proposal

Move application builds back to the tests that actually use them.

Tests that need the same executable multiple times can build it once and reuse it across subtests (`t.Run`), preserving the performance benefits of the current approach. At the same time, unrelated test applications would no longer be built when running a subset of the integration test suite.

This should retain the parallelization improvements from #540 while reducing unnecessary work during targeted test runs.

## Extra

Mongo client integration test currently does not use `t.Parallel()` and relies on `BuildAndRun()`, causing the test application to be rebuilt repeatedly.

As part of this work, this test could be updated to build the executable once and reuse it across subtests instead of rebuilding it repeatedly.

amazingakai:
## Description

Moves integration test application builds from `main_test.go` back into the tests that use them.

Tests that execute the same application multiple times now build the executable once and reuse it across subtests (`t.Run`), preserving the performance improvements from #540 while avoiding unnecessary builds when running a subset of the integration test suite.

## Motivation

Building test applications globally from `main_test.go` means running a single integration test still builds every application in the integration test suite, increasing local iteration time and making targeted test runs unnecessarily expensive.

This change keeps the existing parallelization benefits while ensuring only the applications required by the selected tests are built.

Fixes #588

---

## Checklist

- [x] Tests pass: `make test`


amazingakai:
cc @txabman42

txabman42 (approved):
LGTM! Nice improvement ❤️

PR #595, opened 2026-06-22
fix(util): allow ListFiles to walk hidden root directories

pushkarscripts:
## Description

`ListFiles` skips hidden directories during traversal by returning `filepath.SkipDir` when a directory name starts with `.`.

When the root directory passed to `ListFiles` is itself hidden, the walk terminates immediately and no files are returned, even though the caller explicitly requested traversal of that directory.

This change preserves the existing behavior for hidden subdirectories while allowing traversal of a hidden root directory passed directly to `ListFiles`.

## Motivation

`ListFiles` currently treats the root directory the same way as nested hidden directories.

As a result, calling `ListFiles` on a hidden directory returns an empty result because the walk is skipped before any files are visited. This behavior is unexpected when the hidden directory is explicitly provided by the caller.

This PR fixes that edge case and adds a regression test covering hidden root directory traversal.

---

## Checklist

- [ ] Linters pass: `make lint` *(typo check requires Docker)*
- [x] Tests pass: `make test`

### DIFF
--- a/tool/util/sys.go
+++ b/tool/util/sys.go
@@ -125,7 +125,7 @@ func ListFiles(dir string) ([]string, error) {
 			return ex.Wrap(err)
 		}
 		// Don't list files under hidden directories
-		if strings.HasPrefix(info.Name(), ".") {
+		if path != dir && strings.HasPrefix(info.Name(), ".") {
 			if info.IsDir() {
 				return filepath.SkipDir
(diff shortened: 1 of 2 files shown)


txabman42 (approved):
LGTM!

We shouldnt be affected by this, but a small safe improvement. TY ❤️

PR #604, opened 2026-06-24
feat(instrumentation): add OpenAI Go SDK instrumentation (v1/v2/v3)

NameHaibinZhang:
Add compile-time instrumentation for the OpenAI Go SDK supporting all three major versions. The middleware hooks into NewClient to intercept HTTP requests and produce OpenTelemetry spans following GenAI semantic conventions.

Supported operations:
- Chat completions (streaming and non-streaming)
- Text completions (streaming and non-streaming)
- Embeddings

Features:
- SSE streaming response parsing with first-token latency tracking
- 22-provider URL-based mapping (OpenAI, Azure, DeepSeek, Qwen, etc.)
- Request parameter extraction (model, temperature, top_p, penalties)
- Response metadata (token usage, finish reasons, response model)

Closes #327

## Motivation

Fixes #

---

## Checklist

- [x] Tests pass: `make test`


instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+		op := classifyOperation(req.URL.Path)
+		if op == opUnknown {
+			return next(req)
+		}
+
+		start := time.Now()
+		provider := getProviderName(req.URL.Host)
+		opName := operationName(op)
+
+		bodyBytes, err := io.ReadAll(io.LimitReader(req.Body, maxRequestBodySize))
```
kakkoyun: io.LimitReader(req.Body, maxRequestBodySize) caps the read at 1 MB, and the truncated bytes then become req.Body, which is what gets sent. LimitReader returns io.EOF at the cap rather than an error, so the check below does not catch it: a request over 1 MB (inline base64 images, long context, batched embeddings) is truncated and the API receives a malformed body. Could we parse a bounded copy for attributes (for example a TeeReader into a capped buffer) while forwarding the original body untouched? Same pattern in v2 and v3.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+		}
+
+		return resp, nil
+	}
+}
+
+func handleNonStreamingResponse(_ context.Context, resp *http.Response, span trace.Span, _ time.Time, op operationType) {
+	defer span.End()
+
+	bodyBytes, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBodySize))
```
kakkoyun: Same truncation on the response side: bodies over 4 MB are cut, and the truncated copy is handed back to the caller, which then sees malformed JSON or an unexpected EOF. Keeping the forwarded stream intact and bounding only the parsed copy would fix both directions. Applies to v2 and v3 as well.

instrumentation/github.com/openai/openai-go/v3/middleware.go
```diff
@@ -0,0 +1,343 @@
...
+			semconv.GenAIProviderName(provider),
+		}
+		spanAttrs = append(baseAttrs, spanAttrs...)
+
+		ctx := req.Context()
+		ctx, span := tracer.Start(ctx, spanName,
+			trace.WithSpanKind(trace.SpanKindInternal),
+			trace.WithAttributes(spanAttrs...),
+		)
+		req = req.WithContext(ctx)
```
kakkoyun, line 155: v1 and v2 call ctx = runtime.SuppressHTTPClientInstrumentation(ctx) just before this line; v3 does not, and it also does not import pkg/runtime. Without it, the net/http client hook still creates its own span, so v3 emits two spans per request. Adding the same call here brings v3 in line. This is the kind of drift that sharing the logic across versions would prevent (see the summary).

instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -0,0 +1,250 @@
...
+		r.teeReader = io.TeeReader(r.reader, r.logBuffer)
+	}
+
+	n, err = r.teeReader.Read(p)
+
+	if n > 0 {
+		r.processSSELines()
+	}
+
+	if err != nil && !r.done {
```
kakkoyun: done is read and set here and again in Close(), with no synchronization, and finalize()/span.End() can run from both. An http.Response body can be read on one goroutine and closed on another (a cancelled context, client teardown), which makes this a data race and a possible double End(). The existing tests read then close sequentially, so -race stays quiet. An atomic.Bool with CompareAndSwap(false, true) guarding finalize() would make it safe. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -0,0 +1,250 @@
...
+		r.done = true
+		r.finalize()
+	}
+	if r.reader != nil {
+		return r.reader.Close()
+	}
+	return nil
+}
+
+func (r *streamingReader) finalize() {
```
kakkoyun, line 82: finalize() and span.End() run only when Read hits EOF/error or Close() is called. If a caller stops reading and never closes the body, the span stays open and accumulates in the SDK. A goroutine waiting on ctx.Done() (once done is atomic) would bound the lifetime. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -0,0 +1,250 @@
...
+	r.span.End()
+}
+
+func (r *streamingReader) processSSELines() {
+	if r.logBuffer == nil || r.logBuffer.Len() == 0 {
+		return
+	}
+
+	data := r.logBuffer.Bytes()
+	r.lineBuffer.Write(data)
```
kakkoyun, line 109: lineBuffer accumulates across reads and is only drained when a newline appears, so an upstream that never emits one grows it without limit. A cap on lineBuffer.Len() with a drop-and-finalize on overflow would protect against a malformed or hostile stream. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+			model, spanAttrs = parseChatRequest(bodyBytes)
+		case opCompletion:
+			model, spanAttrs = parseCompletionRequest(bodyBytes)
+		case opEmbedding:
+			model, spanAttrs = parseEmbeddingRequest(bodyBytes)
+		}
+
+		spanName := opName + " " + model
+		baseAttrs := []attribute.KeyValue{
+			semconv.GenAISystem("openai"),
```
kakkoyun, line 142: gen_ai.system is set to the literal "openai" here, while gen_ai.provider.name a few lines down is detected from the host, so a DeepSeek or Azure call carries both and they disagree. gen_ai.system is deprecated in favor of gen_ai.provider.name, so dropping this line (or deriving it from the detected provider) resolves the conflict. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+			semconv.GenAIOperationName(opName),
+			semconv.GenAIRequestModel(model),
+			semconv.GenAIProviderName(provider),
+		}
+		spanAttrs = append(baseAttrs, spanAttrs...)
+
+		ctx := req.Context()
+		ctx, span := tracer.Start(ctx, spanName,
+			trace.WithSpanKind(trace.SpanKindInternal),
```
kakkoyun: The GenAI spec recommends CLIENT span kind for calls to a remote model; INTERNAL is for models running in-process. Since these go over the network, SpanKindClient matches the intent. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/semconv/genai.go
```diff
@@ -0,0 +1,97 @@
...
+	GenAIUsageInputTokensKey          = attribute.Key("gen_ai.usage.input_tokens")
+	GenAIUsageOutputTokensKey         = attribute.Key("gen_ai.usage.output_tokens")
+	GenAIUsageTotalTokensKey          = attribute.Key("gen_ai.usage.total_tokens")
+	GenAIProviderNameKey              = attribute.Key("gen_ai.provider.name")
+	GenAIRequestMaxTokensKey          = attribute.Key("gen_ai.request.max_tokens")
+	GenAIRequestTemperatureKey        = attribute.Key("gen_ai.request.temperature")
+	GenAIRequestTopPKey               = attribute.Key("gen_ai.request.top_p")
+	GenAIRequestFrequencyPenaltyKey   = attribute.Key("gen_ai.request.frequency_penalty")
+	GenAIRequestPresencePenaltyKey    = attribute.Key("gen_ai.request.presence_penalty")
+	GenAIRequestIsStreamKey           = attribute.Key("gen_ai.request.is_stream")
```
kakkoyun, line 26: gen_ai.request.is_stream and gen_ai.response.time_to_first_token (next line) are not in the GenAI registry. The spec does define a time-to-first-chunk span attribute, but it's named gen_ai.response.time_to_first_chunk (not time_to_first_token), typed double in seconds — not int64 in microseconds. Emitting custom keys under gen_ai.* also risks clashing with future official names. A vendor prefix (for example otelc.*) or aligning with the spec name and unit would address both. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+			trace.WithAttributes(spanAttrs...),
+		)
+		ctx = runtime.SuppressHTTPClientInstrumentation(ctx)
+		req = req.WithContext(ctx)
+
+		resp, err := next(req)
+
+		if err != nil {
```
kakkoyun, line 159: This branch only catches transport errors. A 4xx/5xx response with a nil Go error falls through to the success path, gets parsed as a completion, and the span ends OK with empty token values. It would also help to set error.type on the error span (conditionally required by the spec). Consider checking resp.StatusCode here. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+	"siliconflow":        "siliconflow",
+	"together":           "together",
+	"mistral":            "mistral",
+	"groq.com":           "groq",
+	"ollama":             "ollama",
+	"localhost":          "local",
+	"127.0.0.1":          "local",
+}
+
+func getProviderName(host string) string {
```
kakkoyun, line 54: Two things here. The function iterates a map[string]string, and Go map iteration order is randomized, so if a future mapping ever has two keywords that both match the same host but point to different providers the result would be non-deterministic. More immediately, strings.Contains on bare short keywords like "mistral" or "together" can match unrelated hostnames, and anything that does not match defaults to "openai", which mislabels unknown providers. An ordered slice of {suffix, provider} pairs with domain-suffix matching would make this deterministic and avoid the false positives. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+func parseChatRequest(body []byte) (string, []attribute.KeyValue) {
+	var req struct {
+		Model            string   `json:"model"`
+		MaxTokens        *int64   `json:"max_tokens,omitempty"`
+		Temperature      *float64 `json:"temperature,omitempty"`
+		TopP             *float64 `json:"top_p,omitempty"`
+		FrequencyPenalty *float64 `json:"frequency_penalty,omitempty"`
+		PresencePenalty  *float64 `json:"presence_penalty,omitempty"`
+	}
+	if err := json.Unmarshal(body, &req); err != nil {
```
kakkoyun, line 222: When Unmarshal fails the function returns an empty model and no attributes, with nothing logged, so a malformed body produces a span named "chat " (empty model) and no signal that parsing failed. A debug log or a span event here would make that visible. The same pattern repeats in the other parse* functions. Applies to v2/v3.

instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -0,0 +1,250 @@
...
+	totalTokens   int64
+	id            string
+	responseModel string
+	reasons       []string
+	span          trace.Span
+	model         string
+	opName        string
+	provider      string
+	op            operationType
+	ctx           context.Context
```
kakkoyun: This ctx field is stored but never read anywhere in the type. Storing a context in a struct is discouraged in general, and here it is also dead, so the field and the corresponding newStreamingReader parameter can be dropped. Applies to v2/v3.

kakkoyun (commented):
Nice to see OpenAI Go support landing. Good scope here: three SDK majors, streaming and non-streaming, embeddings, and a provider map. I read the diff closely, mostly around request/response handling, the streaming reader, and GenAI semantic-convention conformance. A few items look like correctness bugs worth resolving before merge; specifics are inline, summarized below.

Correctness and data integrity
- The middleware reads the request and response bodies through io.LimitReader (1 MB and 4 MB) and then puts the truncated copy back as the body that actually gets sent or returned. io.LimitReader stops at the limit without reporting an error, so any payload above the cap (inline base64 images, long context windows, large embedding batches) is truncated before it reaches the API or the caller. The metadata parsing should not change the bytes that flow through.
- v3/middleware.go is missing the `ctx = runtime.SuppressHTTPClientInstrumentation(ctx)` line that v1 and v2 have, so the net/http client hook still fires for v3 and you get two spans per call.
- The three version directories are byte-for-byte identical apart from the package name and import paths (I diffed them). If the shared logic lived in one internal package with only the per-version hook wiring kept separate, this class of drift mostly goes away.

Streaming reader
- `done` is read and written from both Read and Close with no synchronization, and finalize()/span.End() can run from either path. The current tests read then close on the same goroutine, so -race will not surface it.
- The span ends only when Read reaches EOF/error or Close() is called. A caller that abandons the body without closing it leaves the span open.
- lineBuffer keeps growing across reads until a newline arrives, so an upstream that never sends one has no upper bound.

GenAI semantic conventions
- gen_ai.system is hardcoded to "openai" while gen_ai.provider.name is detected from the host, so a DeepSeek or Azure call ends up with both, and they conflict. gen_ai.system is also deprecated in favor of gen_ai.provider.name.
- The span kind is INTERNAL; the spec recommends CLIENT for calls to a remote model.
- gen_ai.request.is_stream and gen_ai.response.time_to_first_token are not in the registry. Time-to-first-token is defined as a metric in seconds rather than a microsecond span attribute.
- error.type is not set when a span ends in error, and a non-2xx response with a nil Go error currently goes down the success path and ends OK.

Tests
- Only v1 has an end-to-end test, and the suppression behavior is not exercised anywhere, so the v3 duplicate-span regression would not be caught by CI. An e2e test per version asserting a single span, a concurrent Read/Close test under -race, and a round-trip test for bodies above the cap would cover the gaps here.

I left smaller style notes out of the inline set: missing doc comments in the semconv package, the package v1 naming, the identical parseChatResponse/parseCompletionResponse pair, the info-level init log, and the demo logging the prompt. Glad to help with any of these, and thanks for the work so far.

instrumentation/github.com/openai/openai-go/middleware.go
NameHaibinZhang: Done. Replaced the LimitReader+ReadAll+replace pattern with a TeeReader approach: we now read a bounded preview for attribute parsing while reassembling the full original body stream (via MultiReader) so that the SDK/transport always receives the complete, unmodified payload.

instrumentation/github.com/openai/openai-go/middleware.go
NameHaibinZhang: Done. Same TeeReader fix applied to the response side — we now read a bounded preview for parsing but reassemble the full body stream for callers. Applied consistently across v1/v2/v3.

instrumentation/github.com/openai/openai-go/v3/middleware.go
NameHaibinZhang: Done. Added `ctx = runtime.SuppressHTTPClientInstrumentation(ctx)` and the corresponding `pkg/runtime` import to v3/middleware.go, bringing it in line with v1 and v2.

instrumentation/github.com/openai/openai-go/streaming.go
NameHaibinZhang: Done. Replaced the `bool` field with `sync/atomic.Bool` and guarded finalize() with `CompareAndSwap(false, true)` in both Read() and Close(), preventing double-End and data races. Applied to v1/v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
NameHaibinZhang: Done. Changed `gen_ai.system` to use the detected `provider` variable instead of the hardcoded "openai" literal, so it now stays consistent with `gen_ai.provider.name` for all providers. Applied to v1/v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
NameHaibinZhang: Done. Changed from `SpanKindInternal` to `SpanKindClient` since these are outgoing network calls to a remote model service. Applied to v1/v2/v3.

instrumentation/github.com/openai/openai-go/middleware.go
NameHaibinZhang: Done. Added a check for `resp.StatusCode >= 400` — when the API returns an error HTTP status, we now set the span status to Error, record `error.type` with the status text, and end the span early instead of attempting to parse the body as a successful completion. Applied to v1/v2/v3.

instrumentation/github.com/openai/openai-go/streaming.go
NameHaibinZhang: Done. Removed the unused `ctx` field from `streamingReader` and changed the `newStreamingReader` parameter to `_ context.Context` (keeping the signature stable for callers). Applied to v1/v2/v3.

instrumentation/github.com/openai/openai-go/streaming.go
NameHaibinZhang: Thank you — this is a valid concern. With the `done` field now using `atomic.Bool`, the immediate double-End race is resolved. However, you're right that an abandoned body (no Close, no EOF) would leave the span open indefinitely. Adding a goroutine that monitors `ctx.Done()` to bound the span lifetime is the right approach, but it introduces lifecycle complexity (goroutine leak if body is consumed normally, need to coordinate with the atomic guard). I'll track this as a follow-up improvement to keep this PR focused on the correctness fixes.

instrumentation/github.com/openai/openai-go/streaming.go
NameHaibinZhang: Good catch. In practice, SSE streams always emit newlines (the protocol requires `\n\n` between events), so this is unlikely to trigger in normal operation — but a malformed or hostile upstream could exploit it. I'll add a lineBuffer cap with a drop-and-finalize strategy in a follow-up PR to keep defensive bounds without overcomplicating this change.

instrumentation/github.com/openai/openai-go/semconv/genai.go
NameHaibinZhang: Thank you for the careful spec check. You're right that `gen_ai.request.is_stream` and `gen_ai.response.time_to_first_token` are not in the official GenAI semantic conventions registry. The spec defines `gen_ai.response.time_to_first_chunk` (typed as double in seconds). I'll address the naming alignment (either adopting the spec name/unit or using a vendor prefix like `otelc.*`) in a follow-up PR, as it requires a broader discussion on our custom attribute naming strategy.

instrumentation/github.com/openai/openai-go/middleware.go
NameHaibinZhang: Agreed on both points. The random map iteration order is technically non-deterministic (though in practice the current entries don't overlap for any real host), and short bare keywords like "together" or "mistral" could match unintended hostnames. Refactoring to an ordered slice of `{suffix, provider}` pairs with `strings.HasSuffix` checks would resolve both issues cleanly. I'll address this in a follow-up PR as it touches the matching semantics (suffix vs. contains) and needs careful thought about which patterns each provider actually uses.

instrumentation/github.com/openai/openai-go/middleware.go
NameHaibinZhang: Fair point. Currently, a failed Unmarshal produces a span named "chat " (empty model) with no indication of what went wrong. Adding a debug-level span event (e.g., `span.AddEvent("parse_error", ...)`) when Unmarshal fails would make this visible without introducing a logger dependency. I'll add this in a follow-up PR along with the other parse* functions.

instrumentation/github.com/openai/openai-go/v3/middleware.go
```diff
@@ -0,0 +1,343 @@
...
+		provider := getProviderName(req.URL.Host)
+		opName := operationName(op)
+
+		bodyBytes, err := io.ReadAll(io.LimitReader(req.Body, maxRequestBodySize))
+		if err != nil {
+			return next(req)
+		}
+		req.Body.Close()
+		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Fixed in the latest commit — now using TeeReader+MultiReader to preserve the full body stream while reading a bounded preview for attribute parsing.

instrumentation/github.com/openai/openai-go/v3/middleware.go
```diff
@@ -0,0 +1,343 @@
...
+func handleNonStreamingResponse(_ context.Context, resp *http.Response, span trace.Span, _ time.Time, op operationType) {
+	defer span.End()
+
+	bodyBytes, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBodySize))
+	if err != nil {
+		return
+	}
+	resp.Body.Close()
+	resp.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Fixed — response body is now reassembled via TeeReader+MultiReader pattern.

instrumentation/github.com/openai/openai-go/v3/middleware.go
```diff
@@ -0,0 +1,343 @@
...
+			semconv.GenAISystem("openai"),
+			semconv.GenAIOperationName(opName),
+			semconv.GenAIRequestModel(model),
+			semconv.GenAIProviderName(provider),
+		}
+		spanAttrs = append(baseAttrs, spanAttrs...)
+
+		ctx := req.Context()
+		ctx, span := tracer.Start(ctx, spanName,
+			trace.WithSpanKind(trace.SpanKindInternal),
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Changed to SpanKindClient.

instrumentation/github.com/openai/openai-go/v3/middleware.go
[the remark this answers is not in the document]
NameHaibinZhang: Done. Added `ctx = runtime.SuppressHTTPClientInstrumentation(ctx)` to v3.

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+		start := time.Now()
+		provider := getProviderName(req.URL.Host)
+		opName := operationName(op)
+
+		bodyBytes, err := io.ReadAll(io.LimitReader(req.Body, maxRequestBodySize))
+		if err != nil {
+			return next(req)
+		}
+		req.Body.Close()
+		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Fixed with TeeReader+MultiReader pattern in v2.

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+
+func handleNonStreamingResponse(_ context.Context, resp *http.Response, span trace.Span, _ time.Time, op operationType) {
+	defer span.End()
+
+	bodyBytes, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBodySize))
+	if err != nil {
+		return
+	}
+	resp.Body.Close()
+	resp.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Fixed in v2 as well.

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+			semconv.GenAISystem("openai"),
+			semconv.GenAIOperationName(opName),
+			semconv.GenAIRequestModel(model),
+			semconv.GenAIProviderName(provider),
+		}
+		spanAttrs = append(baseAttrs, spanAttrs...)
+
+		ctx := req.Context()
+		ctx, span := tracer.Start(ctx, spanName,
+			trace.WithSpanKind(trace.SpanKindInternal),
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Changed to SpanKindClient in v2.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+		start := time.Now()
+		provider := getProviderName(req.URL.Host)
+		opName := operationName(op)
+
+		bodyBytes, err := io.ReadAll(io.LimitReader(req.Body, maxRequestBodySize))
+		if err != nil {
+			return next(req)
+		}
+		req.Body.Close()
+		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Fixed with TeeReader+MultiReader pattern in v1.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -0,0 +1,345 @@
...
+
+func handleNonStreamingResponse(_ context.Context, resp *http.Response, span trace.Span, _ time.Time, op operationType) {
+	defer span.End()
+
+	bodyBytes, err := io.ReadAll(io.LimitReader(resp.Body, maxResponseBodySize))
+	if err != nil {
+		return
+	}
+	resp.Body.Close()
+	resp.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```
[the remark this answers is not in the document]
NameHaibinZhang: Done. Fixed in v1 as well.

instrumentation/github.com/openai/openai-go/middleware.go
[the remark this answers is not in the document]
NameHaibinZhang: Done. Changed to SpanKindClient in v1.

kakkoyun (approved):
LGTM! Thank you so much!

PR #605, opened 2026-06-24
doc(adr): Import-Driven Instrumentation Selection

amazingakai:
Split from #584 to allow discussion of the import-driven instrumentation design independently from the implementation.


txabman42 (approved):
LGTM!

Issue #304, opened 2026-02-27
feat(instrumentation): Support log/logrus instrumentation


NameHaibinZhang:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

amazingakai:
Closing as this was implemented in #609

PR #609, opened 2026-06-25
feat(instrumentation): add log, slog, and logrus trace context injection

### LINKED ISSUE #317: feat(instrumentation): Support Logrus instrumentation

**Additional context**

NameHaibinZhang:
Add compile-time instrumentation for Go's log, log/slog, and github.com/sirupsen/logrus packages to automatically inject trace_id and span_id into log output when an active span exists in GLS.

Introduces a function registration pattern in pkg/runtime so log instrumentations can retrieve trace context without directly depending on the OTel SDK trace package. Also adds FuncType support in trampoline.go for hooking functions with function-type parameters.

Fixes #317 #304

## Motivation

Fixes #

---

## Checklist

- [ ] Tests pass: `make test`


kakkoyun (commented):
Thanks for this. The trace context injection works end to end, and the unit tests are solid now: they register real trace functions and cover the enabled, disabled, id-only, and no-context paths.

One thing to check before merge: both slog and logrus can add the trace id twice, because two hooks each inject on the same call. Details are inline. The unit tests check each hook on its own, so they pass, but at runtime both hooks run together.

Minor: Go naming usually writes ID, so traceID, spanID, GetTraceAndSpanID. Renaming the exported names is easier now than after merge.

instrumentation/log/slog/slog_hook.go
```diff
@@ -0,0 +1,89 @@
...
+
+type logEnabler struct{}
+
+func (l logEnabler) Enable() bool {
+	return runtime.Instrumented(instrumentationKey)
+}
+
+var enabler = logEnabler{}
+
+func BeforeSlogLog(
```
kakkoyun: BeforeSlogLog adds the trace id into the message text, and AfterSlogNewRecord adds it as a structured attribute. (*Logger).log calls the exported NewRecord internally, so for one log call both hooks run and the trace id can appear twice: once in the message, once as an attribute. Is that intended? If not, keeping only AfterSlogNewRecord would be enough, and you could then drop the strings.Contains(msg, traceIdKey) guard here.

instrumentation/github.com/sirupsen/logrus/logrus_hook.go
```diff
@@ -0,0 +1,136 @@
...
+	if formatterInit {
+		return
+	}
+
+	std := logrus.StandardLogger()
+	std.AddHook(&traceHook{})
+	formatterInit = true
+}
+
+func BeforeLogrusEntryLog(ictx hook.HookContext, entry *logrus.Entry, level logrus.Level, args ...interface{}) {
```
kakkoyun: For a logger from logrus.New(), AfterLogrusNew installs traceHook, and traceHook.Fire already writes trace_id and span_id into entry.Data. BeforeLogrusEntryLog then also appends them to the message args, so for that logger they appear twice: as fields and inside the message text. If Fire is the main path, this function may be redundant for New() loggers. Worth confirming which one you want to keep.

instrumentation/log/log_hook.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/runtime"
+)
+
+const (
+	instrumentationKey = "logs/log"
+	traceIdKey         = "trace_id"
+	spanIdKey          = "span_id"
+)
+
+type logEnabler struct{}
```
kakkoyun, line 20: Small one, low priority. logEnabler, enabler, and the trace_id / span_id constants are repeated in all three modules (log, slog, logrus). They all import pkg/runtime already, so a shared helper such as runtime.Enabled(key) plus shared key constants could remove the repetition.

instrumentation/log/slog/slog_hook.go
NameHaibinZhang: Done. You're right — both hooks running together would cause duplication. Removed `BeforeSlogLog` entirely and kept only `AfterSlogNewRecord`, which injects trace context as structured slog attributes (the idiomatic approach for slog). Also removed the corresponding hook registration from `slog.yaml`.

instrumentation/github.com/sirupsen/logrus/logrus_hook.go
NameHaibinZhang: Done. Agreed — `traceHook.Fire` injecting into `entry.Data` as structured fields is the proper logrus pattern. Removed `BeforeLogrusEntryLog` entirely and the corresponding hook registration from `logrus.yaml`. The hook-based approach (`AfterLogrusNew`, `AfterLogrusWithField`, `AfterLogrusSetFormatter`) ensures trace context is always added as structured fields regardless of the logging path.

instrumentation/log/log_hook.go
NameHaibinZhang: Good observation. I agree there's some repetition here. I'll keep the current structure for now since each module has its own unique `instrumentationKey` and they're in separate Go packages — extracting a shared helper would introduce coupling between otherwise independent instrumentation modules. Happy to revisit this as a follow-up if more log instrumentation modules are added and the pattern solidifies.

NameHaibinZhang:
Thanks for the thorough review! All points addressed:

1. **slog duplication**: Removed `BeforeSlogLog` — keeping only `AfterSlogNewRecord` for structured attributes.
2. **logrus duplication**: Removed `BeforeLogrusEntryLog` — the `traceHook.Fire` approach via structured fields is the correct path.
3. **Shared helper**: Acknowledged as a good follow-up item. Keeping independent for now to avoid coupling between instrumentation modules.
4. **Go naming (ID vs Id)**: Done. Renamed all occurrences: `traceIdKey` → `traceIDKey`, `spanIdKey` → `spanIDKey`, `GetTraceAndSpanId` → `GetTraceAndSpanID`, `RegisterTraceAndSpanIdFunc` → `RegisterTraceAndSpanIDFunc`, and all local variables accordingly.

kakkoyun:
Thanks for the quick turnaround. The duplicate injection is gone now (slog keeps `AfterSlogNewRecord`, logrus keeps the `Fire` hook), and the `ID` renaming looks good.

One thing before merge: the integration tests are red on all platforms. Both log tests fail because no trace id reaches the output:
- `TestLogsSlog`, `logs_test.go:46` (no `trace_id=...`)
- `TestLogsLogrus`, `logs_test.go:74` (no `"trace_id":"..."`)

slog and logrus inject in different ways, but both come out empty. That points to the shared path (`runtime.GetTraceAndSpanID` returning empty at runtime) rather than either hook.

This does not look caused by removing the duplicate injection. Integration passed on the first commit (`df9d896`) and has been failing since the unit-test commit, so the message-text path was probably the only thing still surfacing the trace id. The timing lines up with the merge from `main`, so it may be worth checking that the `//go:linkname` registration in `otel_trace_context.go` still wires `GetTraceAndSpanID` into `pkg/runtime` after that merge, and that the span is present in GLS for the logging goroutine.

Happy to help dig in if useful.

NameHaibinZhang:
Thanks for the detailed analysis! You were right that the hooks weren't firing at runtime, but the root cause turned out to be simpler than a linkname issue.

The rule files were named `slog.yaml`, `log.yaml`, and `logrus.yaml`, but the tool's rule discovery mechanism (`isRuleFile` in `tool/internal/setup/match.go`) only recognizes files matching `otelc.yaml` / `*.otelc.yaml` patterns. So the hook rules were never loaded during compilation.

Fixed by renaming all three to `otelc.yaml`. Both `TestLogsSlog` and `TestLogsLogrus` pass now.

kakkoyun (approved):
LGMT!

Issue #587, opened 2026-06-17
feat(tool): add support for `.otel.yml` instrumentation configuration


amazingakai:
## Problem

The [`ux-design.md`](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/5d80db178d06c17e8cc08cc20edc67d86976d04f/docs/ux-design.md) document describes a `.otel.yml` configuration style intended for CI/CD and other workflows where instrumentation configuration is preferred to be managed outside of `otel.instrumentation.go`.

This configuration style is not currently implemented.

## Proposal

Add support for a module-local `.otel.yml` file.

Proposed structure:

```yaml
instrumentations:
  - github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/nethttp/server
  - github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/gin/server
```

The `instrumentations` section declares instrumentation packages and is equivalent to importing them from `otel.instrumentation.go`.

This configuration style can be useful for CI/CD environments and for instrumenting codebases without maintaining an instrumentation tool file.

As described in `ux-design.md`, builds using `.otel.yml` are not expected to be fully reproducible. Unlike `otel.instrumentation.go`, instrumentation dependencies are not tracked directly by the Go toolchain and may be resolved dynamically during the build.

macayu17:
Hey @amazingakai, can I take this one? I've been poking at it locally and have something working.

Quick rundown of what I did:
- Look for `.otel.yml` in the module root (also accepting `.otel.yaml`)
- Parse the instrumentations: list
- When the file's there, only load rules from those packages instead of all the built-in ones
- Kept the existing priority: `OTELC_RULES` and --rules still win over `.otel.yml`

A couple things I wasn't sure about:
1. Should this only allow the built-in instrumentation packages for now, or also third-party ones?
2. What should happen if someone has both `.otel.yml` and `otel.instrumentation.go` once #584 is merged ?

Happy to open a PR if this direction looks good.

amazingakai:
Thank you for your interest in this @macayu17!

That sounds great, although I was thinking of taking a slightly different approach: generating an `otel.instrumentation.go` from the `.otel.yml` file during `otelc pin` (auto-pinning would remove the generated `otel.instrumentation.go` after the build completes) and then simply running `go mod tidy`.

[...]

PR #612, opened 2026-06-25
feat(tool): add support for loading rules from otel.instrumentation.go

amazingakai:
Split from #584.

Part of #567.

Adds support for import-based instrumentation configuration through `otel.instrumentation.go` (and the compatibility alias `otelc.tool.go`).

The previously proposed `.otel.yml` configuration has been deferred to 1.x (#567). The `go.mod` `tool` directive approach mentioned in `ux-design.md` was dropped because it was not clear how it would work in practice, with this PR focusing solely on the import-based configuration mechanism.


amazingakai:
This PR addresses the following review comments from #584:

> Comment https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/584#discussion_r3455084989

Fixed typo: `unqoteErr` -> `unquoteErr`.

> Comment https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/584#discussion_r3455091334

Changed this to return a plain error instead of wrapping it in `ErrNotInstrumentation`, so misconfigured instrumentation packages now fail the build instead of being pruned.

> Comment https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/584#discussion_r3455120587

Not addressed in this PR. My plan is to generate `otel.instrumentation.go` from `.otel.yml` during `otelc pin`, so we won't need to parse `.otel.yml` here. See https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/587#issuecomment-4834844313.

> Comment https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/584#discussion_r3457401895

`instrumentation_selection_test.go` now verifies:
- built-in gin instrumentation is not used
- custom gin instrumentation from `otel.instrumentation.go` is loaded
- the custom gin instrumentation recursively loads the built-in `net/http` instrumentation

> Comment https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/584#discussion_r3459667587

Loading instrumentation from non-module-root import paths is supported now.

darccio:
@amazingakai Are the Copilot comments already addressed?

amazingakai:
@darccio yes

docs/rules.md
```diff
@@ -341,6 +345,35 @@ invalid_where_file_shape:
...
+
+For example:
+
+- If `OTELC_RULES` is set, all other rule sources are ignored.
+- If `--rules` is provided, discovery through tool files and `.otel.yml`
+  is skipped.
+- If an `otel.instrumentation.go` (or `otelc.tool.go`) file exists, automatic
+  rule discovery from `.otel.yml` is not used.
+- `.otel.yml` is only considered when no higher-precedence rule source is
+  available.
```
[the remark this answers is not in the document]
amazingakai: `.otel.yml` support is tracked in #587

tool/internal/setup/config.go
```diff
@@ -0,0 +1,189 @@
...
+		toolFile, err := findToolFile(dir)
+		if err != nil {
+			return nil, err
+		}
+		if toolFile != "" {
+			toolFiles = append(toolFiles, toolFile)
+		}
+	}
+	return toolFiles, nil
+}
```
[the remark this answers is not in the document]
amazingakai: Used `slices.Sort` to have a deterministic order for tool files.

tool/internal/setup/config.go
```diff
@@ -0,0 +1,189 @@
...
+		f, parseErr := p.Parse(toolFile, parser.ImportsOnly)
+		if parseErr != nil {
+			return parseErr
+		}
+
+		for _, imp := range f.Imports {
+			if imp.Name == nil || imp.Name.Name != ast.IdentIgnore {
+				continue
+			}
+
```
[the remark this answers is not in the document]
amazingakai: Done, non-blank imports are now a hard error instead of being silently ignored.

docs/getting-started.md
```diff
@@ -37,6 +37,25 @@ No manual code changes required.
...
+```go
+//go:build tools
+
+package tools
+
+import (
+	_ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/cmd/otelc" // pin the tool itself, not an instrumentation package
+
+	_ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/instrumentation/net/http/server" // enable net/http server instrumentation
+	_ "github.com/open-telemetry/opentelemetry-go-compile-instrumentation/instrumentation/gin-gonic/gin" // enable gin instrumentation
```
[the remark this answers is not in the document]
amazingakai: Done

tool/internal/setup/config.go
```diff
@@ -0,0 +1,204 @@
...
+	if pkg.Module == nil {
+		return nil, ex.Newf("package %s is not part of a module", importPath)
+	}
+
+	modDir := pkg.Module.Dir
+	pkgDir := pkgload.GetPackageDir(pkg)
+	if pkgDir == "" {
+		return nil, ex.Newf("could not determine directory for package %s", importPath)
+	}
```
darccio: `resolveInstrumentationConfig` requires each imported package to be *compilable*, but the walk only needs the package's directory and module — neither requires a build. A pure-aggregator instrumentation package (its only `.go` file carries `//go:build tools`) has empty `GoFiles`, so `packages.Load` reports `build constraints exclude all Go files` in `pkg.Errors` and `GetPackageDir` returns `""`. Both gates fire, so importing such a package **hard-fails the build** — yet that's exactly the shape the `otel/all` aggregator example in `docs/rules.md` shows.

Repro: `go list -e` on a directory whose only file is a `//go:build tools` file returns empty `GoFiles` and the build-constraint error. The PR's own `gincustom/instrumentation` avoids this only because it also has `hook.go`.

Fix: drop the `pkg.Errors` check (genuine failures are still caught by `pkg.Module == nil` above, and by `rulesFromDir`/`ErrNotInstrumentation` below), and fall back to module-relative dir derivation when `GoFiles` is empty:

````suggestion
	pkg := pkgs[0]
	if pkg.Module == nil {
		return nil, ex.Newf("package %s is not part of a module", importPath)
	}

	modDir := pkg.Module.Dir
	// Prefer the directory of a real Go file. A tools-only instrumentation
	// package (its only .go file carries //go:build tools) has no buildable Go
	// files, so packages.Load reports "build constraints exclude all Go files"
	// and GoFiles is empty; fall back to deriving the dir from the module path +
	// import-path suffix (correct even under replace directives). The walk only
	// needs the directory, not a compilable package.
	pkgDir := pkgload.GetPackageDir(pkg)
	if pkgDir == "" {
		pkgDir = filepath.Join(modDir, filepath.FromSlash(importPath[len(pkg.Module.Path):]))
	}
````

`TestResolveInstrumentationConfig` masks this today because every fixture writes a `dummy.go` — please add a fixture whose *only* file is a `//go:build tools` tool file (no `dummy.go`) plus an end-to-end `walkInstrumentation` aggregator case.

Alternatively, if pure-aggregator packages aren't meant to be supported, this becomes a docs fix: the `otel/all` example must include a build-unconstrained `.go` file (e.g. a `package all` `doc.go`) so the package loads.

File: docs/rules.md [hunk 6]
```diff
@@ -341,6 +345,35 @@ invalid_where_file_shape:
 
 ---
 
+## Loading Rules
+
+Rules are normally distributed through instrumentation packages and enabled using `otel.instrumentation.go` (or `otelc.tool.go`) or `.otel.yml` config files.
```
darccio: The precedence table lists `.otel.yml` as tier 4, but that file type has no reader in the codebase — `loadRules`'s tier 4 is `loadDefaultRules()` which reads the embedded `instrumentation/` bundle. There's no `.otel.yml` handler anywhere in `tool/`. The concept is referenced in `ux-design.md` but isn't implemented.

Please either:
- replace tier 4 with "embedded default instrumentation rules" (the actual fallback), or
- note `.otel.yml` as a planned/future feature and keep it out of the shipped precedence table.

darccio (changes requested):

darccio:
**Docs: two issues in the updated documentation**

**1. `.otel.yml` in the precedence table (`docs/rules.md`) — wrong tier 4**

The "Rule Source Precedence" list ends with:

> 4. `.otel.yml`

But `loadRules`'s tier 4 is `loadDefaultRules()` — the embedded `instrumentation/` bundle. There is no `.otel.yml` reader anywhere in `tool/`. The correct table is:

1. `OTELC_RULES` environment variable
2. `--rules` flag
3. `otel.instrumentation.go` / `otelc.tool.go`
4. Embedded default instrumentation rules

If `.otel.yml` is a planned future tier, please note it as such rather than listing it as current behaviour — a user following the docs today will be confused when it has no effect.

**2. `go.mod` `tool` directive strategy silently removed (`docs/ux-design.md`)**

The previous version of `ux-design.md` listed three configuration strategies, with strategy 1 being the Go 1.24 `tool` directive in `go.mod`. This PR removes it from the numbered list without explanation. If the `tool` directive was previously supported (or documented as planned) and is now intentionally de-scoped, a brief note in the PR description or a changelog entry would help existing users understand the change — particularly anyone who set up their project following the old docs.

tool/internal/setup/config.go
amazingakai: Done:

* `resolveInstrumentationConfig` now loads any importable package instead of requiring a compilable package.
* Updated some `TestResolveInstrumentationConfig` cases to omit `dummy.go`, verifying we can load any valid package, even if it has no buildable Go files.
* Added `TestWalkInstrumentation_LoadsModuleWithoutBuildableFiles` to cover the end-to-end aggregator case as well.

docs/rules.md
amazingakai: Done, updated item 4 to say "embedded default rules", which matches the current implementation.

amazingakai:
Thanks, updated.

1. The precedence table now reflects the current implementation ("embedded default rules"). `.otel.yml` support was deferred to 1.x, so I've removed it from the current precedence list.
2. The `go.mod` `tool` directive strategy was also dropped as part of #567. I've updated the PR description to reflect this.

Windows E2E tests CI Failure is unrelated here.

darccio (approved):

tool/internal/setup/config.go
```diff
@@ -0,0 +1,207 @@
...
+type InstrumentationVisit struct {
+	Config *InstrumentationConfig
+	Error  error
+}
+
+type InstrumentationVisitor func(visit *InstrumentationVisit) (recurse bool, err error)
+
+func walkInstrumentation(ctx context.Context, toolFiles []string, visit InstrumentationVisitor) error {
+	queue := append([]string(nil), toolFiles...)
+	seenImports := make(map[string]bool)
```
kakkoyun, line 228: `seenToolFiles` is only populated when a discovered config's `ToolFile` gets enqueued. The initial `toolFiles` seed list passed into `walkInstrumentation` is never recorded into it. If a package graph routes back to one of the root tool files (e.g. two import paths that both resolve into the project's own module), that file gets re-enqueued and parsed a second time. `seenImports` still caps this at one extra pass, so it's not a hang, but `TestWalkInstrumentation_AvoidsCycles` only exercises a leaf-to-leaf cycle (foo/bar), not this root-routing case that the comment at line 196 describes. Seeding `seenToolFiles` with the initial `toolFiles` before the loop would close this, and it's worth a dedicated test.

kakkoyun (commented):
Went through the new instrumentation config code in detail, checking it against the existing review thread first so this does not repeat ground already covered.

Also found two other issues (a silent rule-precedence override and a logging bug in `loadRules`), but left them out here since both are in code that predates this PR; only the function signature changed. (sorry if it is already commented).

tool/internal/setup/config.go
```diff
@@ -0,0 +1,207 @@
...
+			if unquoteErr != nil {
+				return ex.Wrapf(unquoteErr, "failed to unquote import path %s in %s", imp.Path.Value, toolFile)
+			}
+			// OtelcToolCmdRoot is the tool import itself, not an instrumentation package.
+			if importPath == util.OtelcToolCmdRoot || seenImports[importPath] {
+				continue
+			}
+			seenImports[importPath] = true
+
+			cfg, resolveErr := resolveInstrumentationConfig(ctx, filepath.Dir(toolFile), importPath)
```
kakkoyun: Each newly discovered import in `walkInstrumentation` resolves via its own `packages.Load` call (this line), and `FindModuleDirs` (pkgload.go:164) does the same per package lacking a `Module`. Both spawn a `go list` subprocess per item rather than batching into one `packages.Load(patterns...)` call. For an instrumentation graph with more than a few tool-file imports this adds linear subprocess overhead to every build using this feature. Separately, no timeout is attached to the `ctx` reaching these calls, so a stalled `go list` (network hang, proxy timeout) blocks the build indefinitely, since this tool runs as a `go build -toolexec` wrapper inside every downstream user's build.

tool/internal/pkgload/pkgload.go
```diff
@@ -106,8 +110,15 @@ func ResolveExportFiles(ctx context.Context, importPath string, buildFlags ...st
 	return result, nil
 }
 
-// ResolveModuleDir returns the module directory for a given package directory.
-func ResolveModuleDir(ctx context.Context, pkgDir string) (string, error) {
+func GetPackageDir(pkg *packages.Package) string {
```
kakkoyun: `GetPackageDir` keeps the `Get`-prefixed name this project's own conventions avoid on getters. The two other new accessors added in this same PR (`resolveModuleDir`, `FindModuleDirs`) both use a Resolve/Find verb instead. Worth renaming to `PackageDir` while it's being relocated into this package.

amazingakai: Renamed to `PackageDir`

tool/internal/setup/config.go
amazingakai: Updated `walkInstrumentation` to batch package resolution into a single `packages.Load` call per tool file instead of resolving each import individually. also wrapped the packages.Load in `resolveInstrumentationConfigs` call with a timeout to prevent it from blocking indefinitely if module resolution stalls

tool/internal/setup/config.go
amazingakai: Good catch, thanks! Seeded `seenToolFiles` with the root tool files so they are treated the same as newly discovered ones. also added a dedicated test covering the case where an import resolves back to the root module's tool file

tool/internal/pkgload/pkgload.go
```diff
@@ -128,3 +139,38 @@ func ResolveModuleDir(ctx context.Context, pkgDir string) (string, error) {
...
+		var moduleDir string
+		if pkg.Module != nil {
+			moduleDir = pkg.Module.Dir
+		} else {
+			pkgDir := PackageDir(pkg)
+			if pkgDir == "" {
+				logger.WarnContext(ctx, "skipping package without Go files", "package", pkg.PkgPath)
+				continue
+			}
+
```
txabman42: Nitpick: I was thinking sth like this for cases where there is module without go files

```suggestion
		pkgDir := PackageDir(pkg)
		if pkgDir == "" {
			logger.WarnContext(ctx, "skipping package without Go files", "package", pkg.PkgPath)
			continue
		}

		var moduleDir string
		if pkg.Module != nil {
			moduleDir = pkg.Module.Dir
		} else 
```

tool/internal/setup/config.go
```diff
@@ -0,0 +1,269 @@
...
+		// Prefer the directory of a real Go file. A tools-only instrumentation
+		// package (its only .go file carries //go:build tools) has no buildable Go
+		// files, so packages.Load reports "build constraints exclude all Go files"
+		// and GoFiles is empty; fall back to deriving the dir from the module path +
+		// import-path suffix (correct even under replace directives). The walk only
+		// needs the directory, not a compilable package.
+		pkgDir := pkgload.PackageDir(pkg)
+		if pkgDir == "" {
+			pkgDir = filepath.Join(modDir, filepath.FromSlash(strings.TrimPrefix(importPath, pkg.Module.Path)))
+		}
```
txabman42, line 132:
```suggestion
		if pkgDir == "" {
			pkgDir = filepath.Join(modDir, filepath.FromSlash(strings.TrimPrefix(importPath, pkg.Module.Path)))
		} else if len(pkg.Errors) > 0 {
			// Surface directly package errors
			cfgs[importPath].Error = ex.Newf("loading package %s: %v", importPath, pkg.Errors[0])
			continue
		}

```

txabman42 (approved):
LGTM! Nice PR, solid! ❤️  Just couple of nitpicks

tool/internal/pkgload/pkgload.go
amazingakai: Done in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/compare/29b2820db2378bc059ebf35400338bd5fe14f649..32de8f9d62af8103b98dc9e5a838c034e45c9302

I dont think the check is strictly required today since `listBuildPlan` (`go build -a -x -n`) already rejects packages/modules without go files, but I'm happy to move it earlier as a sanity check.

tool/internal/setup/config.go
amazingakai: I think it's safe to ignore `pkg.Errors` here. At this point we already have the information we need to locate the tool/rule files, so I don't think package loading errors should prevent that.

txabman42: Agree on not skipping them. Lets use that pkg.Errors for adding extra context to `instrumentation package %s contains neither %s nor any rule files`. It could explain why.

amazingakai: Good idea, done in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/compare/32de8f9d62af8103b98dc9e5a838c034e45c9302..ef2d9d74aedaee861edcd10376ed6e33bae06034

kakkoyun (approved):

PR #616, opened 2026-06-25
feat(tool): support projects using vendoring

### LINKED ISSUE #195: Support projects using vendoring
Coming from https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/120#issuecomment-3545920119

Right now, it fails instrumenting an app with vendoring (`go mod vendor`), for example, https://github.com/VictoriaMetrics/VictoriaMetrics. `otel go build -a ./app/vmctl` will fail with:

<details>
<summary>Error logs</summary>

```bash
...
go: inconsistent vendoring in /Users/xabier/Documents/github/VictoriaMetrics:
        github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/grpc@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
        github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/nethttp@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
        github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg@v0.0.0-20251111000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
        github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/grpc: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
        github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg/instrumentation/nethttp: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
        github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pkg: is replaced in go.mod, but not marked as replaced in vendor/modules.txt

        To ignore the vendor directory, use -mod=readonly or -mod=mod.
        To sync the vendor directory, run:
                go mod vendor
Error:
[0] failed to run command "go" with args: [build -work -toolexec=/Users/xabier/Documents/github/VictoriaMetrics/../opentelemetry-go-compile-instrumentation/otel toolexec -a -a ./app/vmctl]: exit status 
...
```

</details>

Same error could be replicated by generating the vendor folder in `/demo/basic`

The `go.mod` is being modified resulting in an incompatibility between the `vendor.txt` and the `go.mod`.

vyagh:
## Description

otelc fails on vendored projects: setup adds the hook modules to `go.mod` but not `vendor/modules.txt`, so `go build` bails with `inconsistent vendoring`.
so when there's a vendor dir, build with `-mod=mod` so go ignores it (outer build + the `go list` toolexec runs).

Fixes #195 

### Notes

- doesn't handle go workspaces (`go work vendor`) yet.

### Testing

- unit tests for the detection helpers
---

## Checklist

- [x] Tests pass: `make test`


NameHaibinZhang:
@y1yang0

tool/internal/setup/setup.go
```diff
@@ -407,6 +407,44 @@ func extractBuildFlags(args []string) []string {
...
+			return dir
+		}
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			break
+		}
+		dir = parent
+	}
+	return ""
+}
```
amazingakai: Could we reuse `pkgload.ResolveModuleDir` here instead of introducing another way of resolving the module root?

We could extend it to first check whether `workDir` is a package directory and use `LoadPackages` to resolve the module directory. If it isn't a package, we could fall back to `go env GOMOD` to locate the active `go.mod`.

amazingakai (changes requested):
Okay, I was originally planning to solve this by tracking `vendor/` and restoring it after the build, but I actually like this approach more.

One thing to keep in mind is that this means we won't support fully offline builds yet. My thought was that once we have `otel.instrumentation.go` / `otelc pin`, we could do something like:

1. `otelc go build` (auto-pinning / clean-room usage):
   * In workspace vendor mode: run `go work vendor` and revert changes to `vendor/` after the build (using #618).
   * In regular (single-module) vendor mode: simply use `-mod=mod` to avoid rewriting `vendor/`.
2. `otelc pin`: run `go mod tidy` followed by `go mod vendor` / `go work vendor` so the instrumentation dependencies are vendored and can be used for fully offline builds.

   * Subsequent `otelc go build` would then just reuse the pinned/vendored dependencies.

What do you think? If we agree this is the direction we want to take, we could track the offline vendoring support in a follow-up issue.

cc @y1yang0 @kakkoyun

amazingakai:
Can we also have an integration test to show that we support builds using `vendor/`?

tool/internal/setup/setup.go
vyagh: I moved it to pkgload and used `go env GOMOD` instead of go list, go list trips on the half updated vendor dir at this point

vyagh:
> Okay, I was originally planning to solve this by tracking `vendor/` and restoring it after the build, but I actually like this approach more.
> 
> One thing to keep in mind is that this means we won't support fully offline builds yet. My thought was that once we have `otel.instrumentation.go` / `otelc pin`, we could do something like:
> 
> 1. `otelc go build` (auto-pinning / clean-room usage):
>    
>    * In workspace vendor mode: run `go work vendor` and revert changes to `vendor/` after the build (using #618).
>    * In regular (single-module) vendor mode: simply use `-mod=mod` to avoid rewriting `vendor/`.
> 2. `otelc pin`: run `go mod tidy` followed by `go mod vendor` / `go work vendor` so the instrumentation dependencies are vendored and can be used for fully offline builds.
>    
>    * Subsequent `otelc go build` would then just reuse the pinned/vendored dependencies.
> 
> What do you think? If we agree this is the direction we want to take, we could track the offline vendoring support in a follow-up issue.
> 
> cc @y1yang0 @kakkoyun

sounds good to me!

tool/internal/setup/setup.go
amazingakai, line 554: If the user runs `otelc go build -mod=vendor`, this will still fail. Should we also check whether `restArgs` contains a `-mod` flag and force its value to `mod`? Otherwise the CLI flag takes precedence over `GOFLAGS`, so the top-level `go build` will still use `-mod=vendor`.

amazingakai (changes requested):
Another problem: during setup we're still resolving source files from the `vendor/` directory while the compiler is compiling the copy from `$GOMODCACHE` (due to `-mod=mod`). That means we're trying to replace a different file than the one being compiled.

For example, the compile command contains:

```text
... ../../../go/pkg/mod/google.golang.org/grpc@v1.81.1/clientconn.go
```

but instrumentation later tries to replace:

```text
.../VictoriaMetrics/vendor/google.golang.org/grpc/clientconn.go
```

which then fails with:

```text
cannot replace .../vendor/google.golang.org/grpc/clientconn.go with $WORK/b469/clientconn.go
```

amazingakai:
@vyagh This would be great to have in v1. Would you like to work through the review comments? Otherwise, I'm happy to take over if you're busy.

tool/internal/setup/setup.go
vyagh: thanks for pointing this out :)

vyagh:
> Another problem: during setup we're still resolving source files from the `vendor/` directory while the compiler is compiling the copy from `$GOMODCACHE` (due to `-mod=mod`). That means we're trying to replace a different file than the one being compiled.
> 
> For example, the compile command contains:
> 
> ```
> ... ../../../go/pkg/mod/google.golang.org/grpc@v1.81.1/clientconn.go
> ```
> 
> but instrumentation later tries to replace:
> 
> ```
> .../VictoriaMetrics/vendor/google.golang.org/grpc/clientconn.go
> ```
> 
> which then fails with:
> 
> ```
> cannot replace .../vendor/google.golang.org/grpc/clientconn.go with $WORK/b469/clientconn.go
> ```

moved -mod=mod ahead of setup, runs in module mode too.
and added a vendored test to cover it.

vyagh:
thanks for the ping @amazingakai, appreciate your support.
I just pushed a fix ( will revert faster now :p)

tool/internal/setup/setup.go
```diff
@@ -320,8 +320,15 @@ func Setup(ctx context.Context, cmd *cli.Command) error {
 		return err
 	}
 
+	// A CLI -mod=vendor beats the GOFLAGS=-mod=mod that GoBuild sets, so the
+	// Phase-1 build-plan dry run would still resolve vendor/ paths. findDeps
+	// rewrites it to module mode when vendoring is active so both phases agree
+	// on where dependency source lives. Recomputed here (rather than threaded)
+	// because Setup is also a standalone command action (otelc setup).
+	vendored := vendoringActive(ctx, util.GetOtelcWorkDir())
```
kakkoyun: `GOFLAGS=-mod=mod` is only forced in `GoBuild` (line 662-670), before `Setup` runs. `otelc setup` is also registered as a standalone CLI command (`tool/cmd/otelc/cmd_setup.go`, `Action: setup.Setup`), and when invoked that way `GoBuild` never runs, so this line's `vendoringActive` result only reaches `findDeps`'s dry run. Nothing forces module mode for `getBuildPackages` at line 318, which runs first and hits the same "inconsistent vendoring" failure this PR fixes. Worth either forcing GOFLAGS at the top of `Setup` too, or documenting that standalone `otelc setup` on a vendored project isn't supported. Neither integration test exercises this path; both go through `otelc go build`.

tool/internal/setup/setup.go
```diff
@@ -491,8 +498,70 @@ func extractBuildFlags(args []string) []string {
 	return append(valueFlags, enabledBoolFlags...)
 }
 
-// BuildWithToolexec builds the project with the toolexec mode
-func BuildWithToolexec(ctx context.Context, cmd *cli.Command) error {
+// vendoringActive reports whether the main module vendors its dependencies. It
+// resolves the module via go env GOMOD rather than `go list`, which would fail
+// the vendor consistency check while go.mod is mid-edit, then checks for
+// vendor/modules.txt.
+func vendoringActive(ctx context.Context, workDir string) bool {
```
kakkoyun: `vendoringActive` doesn't check for workspace mode. I confirmed `go env GOMOD` still resolves a workspace member module's real go.mod (not empty/DevNull) even under `go.work`, so a workspace member with a leftover `vendor/modules.txt` makes this return `true`, and `GoBuild` then sets `GOFLAGS=-mod=mod`. Go rejects that outright in workspace mode: `-mod may only be set to readonly or vendor when in workspace mode`. Workspace mode normally ignores per-module vendor directories entirely, so this turns a build that worked before this PR into a hard failure. The "go.work workspaces are out of scope" comment at line 666 isn't backed by an actual `go env GOWORK` guard anywhere in this change.

tool/internal/setup/setup.go
```diff
@@ -491,8 +498,70 @@ func extractBuildFlags(args []string) []string {
...
+	root, err := pkgload.ModuleDir(ctx, workDir)
+	if err != nil || root == "" {
```
kakkoyun: Any error from `pkgload.ModuleDir` (missing `go` binary, permission error, context cancellation) is swallowed and treated the same as "not vendored," with no log line. On a real vendored project this fails open into the exact bug the PR fixes, with nothing in the logs pointing at why detection didn't fire.

tool/internal/setup/setup.go
```diff
@@ -491,8 +498,70 @@ func extractBuildFlags(args []string) []string {
...
+// (Go applies last-wins for repeated flags, so all occurrences must change);
+// -mod=mod and -mod=readonly are left as-is (both already ignore vendoring, so
+// we respect the user's intent). -mod=mod is appended only when no -mod token
+// is present at all.
+func forceModMod(goflags string) string {
+	fields := strings.Fields(goflags)
+	hasMod := false
+	for i, f := range fields {
+		switch {
+		case f == "-mod=vendor":
```
kakkoyun: `forceModMod` and `rewriteModVendor` only match single-dash `-mod`/`-mod=vendor`. Go's flag parsing treats `--mod=vendor` / `--mod vendor` as equivalent to the single-dash form, so an invocation using double dashes would skip the rewrite and hit the original vendor-consistency failure. Low real-world odds since single-dash is the near-universal convention, but it's the exact flag this PR exists to neutralize.

tool/internal/setup/setup.go
```diff
@@ -577,6 +653,22 @@ func GoBuild(ctx context.Context, cmd *cli.Command) error {
...
+	// mode so both build phases resolve dependency sources from the module cache
+	// (matching versions and paths) instead of vendor/, leaving the user's
+	// vendor directory untouched. This must run before Setup so the Phase 1
+	// build-plan dry run also ignores vendoring.
+	pwd := util.GetOtelcWorkDir()
+	vendored := vendoringActive(ctx, pwd)
+	if vendored {
+		logger.InfoContext(ctx, "vendored project detected; building with -mod=mod")
+		// go.work workspaces are out of scope (Go forbids -mod=mod in workspace mode).
+		if err := os.Setenv("GOFLAGS", forceModMod(os.Getenv("GOFLAGS"))); err != nil {
```
kakkoyun: `os.Setenv("GOFLAGS", ...)` mutates process-wide env with no restore. Not an issue under the current one-shot-CLI-per-process model, but there's nothing guarding against `GoBuild` running twice in the same process (library embedding, tests that call it repeatedly). If that ever happens, `-mod=mod` from one vendored build leaks into a later non-vendored one.

test/integration/vendor_test.go
```diff
@@ -0,0 +1,77 @@
...
+// TestVendoredBuild builds the basic demo from a vendored copy. setup edits
+// go.mod for the hook modules but not vendor/modules.txt, so the build must use
+// -mod=mod or it fails the vendor consistency check.
+func TestVendoredBuild(t *testing.T) {
+	t.Parallel()
+	buildVendoredBasic(t, "go", "build", "-a")
+}
+
+// TestVendoredBuildModVendor covers the flag-precedence path: an explicit
+// -mod=vendor on the CLI beats the GOFLAGS=-mod=mod otelc sets, so without
```
kakkoyun: Both new tests drive `otelc go build`. Nothing here exercises standalone `otelc setup` against a vendored project. A test on that path would have caught the standalone-command gap noted above.

test/integration/vendor_test.go
```diff
@@ -0,0 +1,77 @@
...
+	t.Helper()
+
+	appsDir := t.TempDir()
+	app := filepath.Join(appsDir, "basic")
+	src := filepath.Join("..", "..", "demo", "app", "basic")
+	require.NoError(t, os.CopyFS(app, os.DirFS(src)))
+
+	goModVendor(t, app)
+	modulesTxt := filepath.Join(app, "vendor", "modules.txt")
+	before, err := os.ReadFile(modulesTxt)
```
kakkoyun, line 66: gosec flags `os.ReadFile(modulesTxt)` here as G304 (potential file inclusion via variable). `modulesTxt` is built from `t.TempDir()` plus a hardcoded literal, not from any external input. Standard test-code false positive, no change needed.

tool/internal/setup/setup.go
kakkoyun: `"GOFLAGS"` is a bare string literal; the rest of this codebase centralizes env var names as constants in `tool/util` (`EnvOtelcWorkDir`, `EnvOtelcBuildFlags`, `EnvOtelcStats`). Not a functional issue since GOFLAGS is a real Go toolchain var rather than an otelc-specific one, just a consistency nit.

tool/internal/setup/setup.go
kakkoyun: `vendoringActive` (which shells out to `go env GOMOD`) runs once here and again in `GoBuild` at line 663 for the same directory: two subprocess execs per `otelc go build` invocation. Given the standalone-command gap noted above, the recompute here doesn't currently deliver the coverage it's justified by, so today it's a redundant exec with no corresponding benefit.

tool/internal/setup/setup.go
```diff
@@ -491,8 +498,70 @@ func extractBuildFlags(args []string) []string {
...
+// -mod=readonly and -mod=mod untouched. A CLI flag beats GOFLAGS, so this
+// neutralizes a vendor selection that setting GOFLAGS alone cannot override.
+func rewriteModVendor(args []string) []string {
+	out := make([]string, len(args))
+	copy(out, args)
+	for i := 0; i < len(out); i++ {
+		switch {
+		case out[i] == "-mod=vendor":
+			out[i] = modMod
+		case out[i] == "-mod" && i+1 < len(out) && out[i+1] == "vendor":
```
kakkoyun: `rewriteModVendor` scans positional args without distinguishing flags from build targets. A literal `-mod` token would only be misread if it appeared as some other flag's separate-form value immediately followed by `"vendor"`. Vanishingly unlikely, not worth changing, just noting the tradeoff of positional scanning vs. a flag-aware parser.

tool/internal/setup/setup.go
vyagh: added a go env GOWORK check for this.

tool/internal/setup/setup.go
vyagh: now handles --mod too

tool/internal/setup/setup.go
vyagh: moved it into Setup.

tool/internal/setup/setup.go
vyagh: now it logs a warning.

test/integration/vendor_test.go
vyagh: standalone is covered now, but I did not add a test for it, the harness only runs otelc go build. 
can add one if needed

tool/internal/setup/setup.go
vyagh: left it inline rn. I can move it to a util const if you prefer that.

tool/internal/setup/setup.go
vyagh: needed now for the standalone path. the GOWORK check adds another go env call though. I can combine them if you prefer

tool/internal/setup/setup.go
vyagh: so, nothing to be done for now?

kakkoyun:
@vyagh this needs a rebase

test/integration/vendor_test.go
kakkoyun: Not blocking, this is now fixed in code either way. If you want coverage without standing up a full otelc-binary integration test, a unit test calling `Setup` directly against a temp vendored module (same style as `TestVendoringActive`) would catch a regression here at much lower cost than extending the integration harness to invoke standalone `otelc setup`.

tool/internal/setup/setup.go
kakkoyun: No need. GOFLAGS is a real Go toolchain env var, not an otelc-owned one, so it doesn't need to follow the `EnvOtelc*` constant convention. Leave it inline.

tool/internal/setup/setup.go
kakkoyun: Worth it if it's a small change: `go env GOMOD GOWORK` returns both in one call, so `ModuleDir` and `WorkspaceActive` could share a subprocess instead of shelling out twice per `vendoringActive` check. Not blocking though, only do it if it stays simple.

tool/internal/setup/setup.go
kakkoyun: Correct, nothing to do for now given the current one-shot CLI-per-process model. Worth a one-line comment noting the assumption (GOFLAGS is set once per process and never restored) so it doesn't bite silently if `GoBuild` is ever invoked more than once in-process, but not blocking this PR.

amazingakai:
@vyagh The CI failure looks related, could you take a look?

Please also update the import path (`github.com/open-telemetry/opentelemetry-go-compile-instrumentation` -> `go.opentelemetry.io/otelc`)

tool/internal/setup/setup.go
```diff
@@ -498,8 +520,103 @@ func extractBuildFlags(args []string) []string {
...
+		switch {
+		case isModVendorToken(out[i]):
+			out[i] = modMod
+		case isModFlag(out[i]) && i+1 < len(out) && out[i+1] == "vendor":
+			out[i+1] = "mod"
+			i++
+		}
+	}
+	return out
+}
```
amazingakai: Lets move these helpers to a separate `vendor.go` file.

amazingakai (approved):
LGTM ❤️ My comment below is non-blocking but nice to have

amazingakai:
@vyagh You removed the `vendored` arg from `findDeps` in [`8992ee4` (this PR)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/616/changes/8992ee45a388e9d7ad692b830f923b0f4a8156e9)
Is that not needed anymore?

amazingakai:
Currently the following workflows do not work due to `getBuildPackages` attempting to load packages after `go.mod` has been modified, resulting in inconsistent vendoring:

1. `otelc pin` -> running `otelc pin` again fails with an inconsistent vendoring error. This is blocking and should be fixed.
2. `otelc pin` -> running `otelc go build -mod=vendor` also fails because `getBuildPackages` still loads packages with inconsistent vendoring before the build proceeds. Fixing this would also be nice, but it's less critical than (1).

<details>
<summary>
Example Failure:
</summary>

```
➜  VictoriaMetrics git:(master) ✗ ../otelc pin ./app/vmctl
go: found go.opentelemetry.io/otelc/instrumentation/basic in go.opentelemetry.io/otelc/instrumentation/basic v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel in go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/init in go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/init v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/trace in go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/trace v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/client in go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/client v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/server in go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/server v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/log in go.opentelemetry.io/otelc/instrumentation/log v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/log/slog in go.opentelemetry.io/otelc/instrumentation/log/slog v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/net/http/client in go.opentelemetry.io/otelc/instrumentation/net/http/client v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/net/http/server in go.opentelemetry.io/otelc/instrumentation/net/http/server v0.0.0-00010101000000-000000000000
go: found go.opentelemetry.io/otelc/instrumentation/runtime in go.opentelemetry.io/otelc/instrumentation/runtime v0.0.0-00010101000000-000000000000
➜  VictoriaMetrics git:(master) ✗ ../otelc pin ./app/vmctl
Error:
[0] loading packages [./app/vmctl]: err: exit status 1: stderr: go: inconsistent vendoring in /home/azhar/Projects/opentelemetry-go-compile-instrumentation-upstream/VictoriaMetrics:
	go.opentelemetry.io/otelc/instrumentation/basic@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/init@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/trace@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/client@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/server@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/log@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/log/slog@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/net/http/client@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/net/http/server@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/runtime@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/dave/dst@v0.27.3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/grpc-ecosystem/grpc-gateway/v2@v2.29.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/urfave/cli/v3@v3.6.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/contrib/bridges/prometheus@v0.69.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/contrib/exporters/autoexport@v0.69.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/contrib/instrumentation/runtime@v0.69.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc@v0.20.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp@v0.20.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc@v1.44.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp@v1.44.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/otlp/otlptrace@v1.44.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@v1.44.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp@v1.44.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/prometheus@v0.66.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/stdout/stdoutlog@v0.20.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/stdout/stdoutmetric@v1.44.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/exporters/stdout/stdouttrace@v1.44.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/log@v0.20.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otel/sdk/log@v0.20.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/pkg@v0.0.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/pkg/runtime@v0.0.0-00010101000000-000000000000: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/proto/otlp@v1.10.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/mod@v0.37.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/tools@v0.46.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/open-telemetry/opentelemetry-go-compile-instrumentation@v0.5.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	google.golang.org/grpc@v1.82.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	go.opentelemetry.io/otelc/pkg: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/net/http/client: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/log/slog: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/net/http/server: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/basic: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/runtime: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/pkg/runtime: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/client: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/trace: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/go.opentelemetry.io/otel/init: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/log: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	go.opentelemetry.io/otelc/instrumentation/google.golang.org/grpc/server: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	google.golang.org/grpc@v1.81.1: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod

	To ignore the vendor directory, use -mod=readonly or -mod=mod.
	To sync the vendor directory, run:
		go mod vendor

[1] failed to load packages for patterns [./app/vmctl]
[2] getting build packages

Stack:
[0] go.opentelemetry.io/otelc/tool/internal/pkgload/pkgload.go:41 tool/internal/pkgload.LoadPackages
[1] go.opentelemetry.io/otelc/tool/internal/setup/setup.go:171 tool/internal/setup.getBuildPackages
[2] go.opentelemetry.io/otelc/tool/internal/setup/pin.go:543 tool/internal/setup.Pin
[3] go.opentelemetry.io/otelc/tool/cmd/otelc/cmd_pin.go:47 main.init.func2
[4] github.com/urfave/cli/v3@v3.10.1/command_run.go:382 github.com/urfave/cli/v3.(*Command).run
[5] github.com/urfave/cli/v3@v3.10.1/command_run.go:320 github.com/urfave/cli/v3.(*Command).run
[6] github.com/urfave/cli/v3@v3.10.1/command_run.go:95 github.com/urfave/cli/v3.(*Command).Run
[7] go.opentelemetry.io/otelc/tool/cmd/otelc/main.go:104 main.main
[8] runtime/proc.go:290 runtime.main
```

</details>

kakkoyun (approved):
LGTM

kakkoyun:
I'm gonna merge this and we will send subsequent PRs. Feel free to add more comments to this PR.

cc @amazingakai @darccio @txabman42

kakkoyun:
@vyagh CI needs to be green though.

vyagh:
on it!

amazingakai:
@vyagh The windows integration test failure is timeout. Maybe our integration tests have grown so we can loosen up the timeout a little bit

vyagh:
> @vyagh You removed the `vendored` arg from `findDeps` in [`8992ee4` (this PR)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/616/changes/8992ee45a388e9d7ad692b830f923b0f4a8156e9) Is that not needed anymore?

actually i moved the rewrite up to the Setup so it will now hit both auto-pin and findDeps.

txabman42:
@amazingakai have you created issues for this or should I? https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/616#issuecomment-4923931675

btw, nice catch ❤️

amazingakai:
@txabman42 Thanks! I haven't created issues yet since I wanted to see whether we'd just fix them as part of this PR.

@vyagh I applied the following changes locally on top of this PR to address the issues mentioned in my earlier comment:

1. In `Setup()`, moved `rewriteModVendor(args)` before `getBuildPackages` inside the initial `if vendored` check.
2. In `Pin()`, inside the `if len(moduleDirs) == 0` block, added logic similar to `Setup()` so that `otelc pin` works correctly on vendored projects:

```go
vendored := vendoringActive(ctx, util.GetOtelcWorkDir())
if vendored {
	logger.InfoContext(ctx, "vendored project detected; building with -mod=mod")
	if err := os.Setenv("GOFLAGS", forceModMod(os.Getenv("GOFLAGS"))); err != nil {
		return nil, ex.Wrapf(err, "forcing module mode for vendored build")
	}

	opts.Args = rewriteModVendor(opts.Args)
}
```

These changes unblock both issues I reported, so we should be able to merge this PR afterward.

Would you prefer to make these changes yourself, or should I push them? If you'd rather track them separately, I'm also happy to open issues.

Also, if CI acts up again, it might be worth loosening the integration test timeout slightly.

txabman42:
@amazingakai TY for the context. Feel free to push them here if they are not too long, if not we can merge this and open another PR.

vyagh:
> @txabman42 Thanks! I haven't created issues yet since I wanted to see whether we'd just fix them as part of this PR.
> 
> @vyagh I applied the following changes locally on top of this PR to address the issues mentioned in my earlier comment:
> 
> 1. In `Setup()`, moved `rewriteModVendor(args)` before `getBuildPackages` inside the initial `if vendored` check.
> 2. In `Pin()`, inside the `if len(moduleDirs) == 0` block, added logic similar to `Setup()` so that `otelc pin` works correctly on vendored projects:
> 
> ```go
> vendored := vendoringActive(ctx, util.GetOtelcWorkDir())
> if vendored {
> 	logger.InfoContext(ctx, "vendored project detected; building with -mod=mod")
> 	if err := os.Setenv("GOFLAGS", forceModMod(os.Getenv("GOFLAGS"))); err != nil {
> 		return nil, ex.Wrapf(err, "forcing module mode for vendored build")
> 	}
> 
> 	opts.Args = rewriteModVendor(opts.Args)
> }
> ```
> 
> These changes unblock both issues I reported, so we should be able to merge this PR afterward.
> 
> Would you prefer to make these changes yourself, or should I push them? If you'd rather track them separately, I'm also happy to open issues.
> 
> Also, if CI acts up again, it might be worth loosening the integration test timeout slightly.

sure, would appreciate help :)

amazingakai:
Fixed both issues in [`2d7f270` (this PR)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/616/commits/2d7f2700ef42fc55139ecaa3ea344ad76d5481b3).

I also updated the `VendoredBuild` integration test to use a minimal gin http server instead of the previous `golang.org/x/time` application, since the `basic` instrumentation is no longer applied automatically (#664).

PR #617, opened 2026-06-26
feat(tool): resolve rule paths from go/packages

amazingakai:
## Description

Resolves hook and file rule paths using the Go module graph instead of assuming they exist under `.otelc-build`.

## Motivation

Required for #612. This allows `otelc` to load hook packages and injected source files from third-party instrumentation packages, not just the built-in instrumentations embedded in `.otelc-build`.


tool/internal/setup/add.go
```diff
@@ -27,14 +27,23 @@ var requiredImports = map[string]string{
 	"unsafe":        "_",           // The golinkname tag depends on unsafe
 }
 
-func genImportDecl(matched []*rule.InstFuncRule) []dst.Decl {
-	for _, m := range matched {
-		requiredImports[m.Path] = ast.IdentIgnore
+func genImportDecl(funcRules []*rule.InstFuncRule, fileRules []*rule.InstFileRule) []dst.Decl {
+	var imports map[string]string
+	if len(funcRules) > 0 {
+		imports = maps.Clone(requiredImports) // clone required imports to avoid mutating the global constant map
```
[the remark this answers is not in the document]
amazingakai: Done

tool/internal/instrument/apply_file.go
```diff
@@ -17,27 +17,14 @@ import (
...
 	return strings.ReplaceAll(content, "//go:build ignore", "")
 }
 
 // applyFileRule introduces the new file to the target package at compile time.
 func (ip *InstrumentPhase) applyFileRule(ctx context.Context, rule *rule.InstFileRule, pkgName string) error {
 	// List all files in the rule module path
-	files, err := listRuleFiles(rule.Path)
+	files, err := util.ListFiles(rule.ResolvedPath)
 	if err != nil {
 		return ex.Wrapf(err, "listing files for rule %s at path %s", rule.Name, rule.Path)
```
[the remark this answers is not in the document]
amazingakai: Done

tool/internal/setup/store.go
```diff
@@ -4,17 +4,77 @@
...
+				continue
+			}
+
+			util.Assert(len(pkgs) == 1, "expected exactly one package")
+
+			cache[goPath] = pkgs[0].Dir
+			return pkgs[0].Dir, nil
+		}
+
+		return "", ex.Newf("failed to resolve module path %s", goPath)
```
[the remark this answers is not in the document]
amazingakai: Done

tool/internal/setup/sync.go
```diff
@@ -103,27 +95,39 @@ func (sp *SetupPhase) warnVersion(goModPath string, before versionSnapshot) erro
 	// Go directives use Go toolchain syntax ("1.21"), not module semver.
 	if after.Go != nil && before.goVersion != "" {
 		if goversion.Compare("go"+after.Go.Version, "go"+before.goVersion) > 0 {
-			sp.Warn(fmt.Sprintf("Bumped go version (%s -> %s)", before.goVersion, after.Go.Version))
+			_, _ = fmt.Fprintf(os.Stdout, "Bumped go version (%s -> %s)\n", before.goVersion, after.Go.Version)
+			sp.Warn("bumped go version", "old", before.goVersion, "new", after.Go.Version)
```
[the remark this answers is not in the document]
amazingakai: sp.Warn logs to `.otelc-build/debug.log` but we also want to show the warning to user.

tool/internal/setup/sync.go
```diff
@@ -103,27 +95,39 @@ func (sp *SetupPhase) warnVersion(goModPath string, before versionSnapshot) erro
...
+					os.Stdout,
+					"Bumped dependency %s (%s -> %s)\n",
+					req.Mod.Path,
+					oldVer,
+					req.Mod.Version,
+				)
+				sp.Warn("bumped dependency",
+					"module", req.Mod.Path,
+					"old", oldVer,
+					"new", req.Mod.Version)
```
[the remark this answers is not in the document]
amazingakai: Same as https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/617#discussion_r3481273568

kakkoyun (commented):
Nice work on this. Splitting import path from module path is the right call, and a couple of the cleanups are good catches in their own right: `genImportDecl` now clones `requiredImports` instead of mutating the package-level map, and `addReplace` only logs when it actually adds a directive. My notes below are all non-blocking, mostly about how the new `module:` field fails when it's missing or wrong, plus one place a unit test would help.

tool/internal/rule/func_rule.go
```diff
@@ -73,6 +76,9 @@ func NewInstFuncRule(data []byte, name string) (*InstFuncRule, error) {
 	if r.Name == "" {
 		r.Name = name
 	}
+	if r.ModulePath == "" {
+		r.ModulePath = r.Path
```
kakkoyun, line 80: `ModulePath` defaults to `Path` here (and the same in `file_rule.go`), but nothing checks that the two are consistent. Path resolution in `store.go` uses `Path`, while `syncDeps` builds the replace directive from `ModulePath`. So if `module:` is missing or wrong on a package that isn't its own module root, resolution still succeeds while the replace points at the wrong module, and the failure surfaces later in the build with nothing pointing back to this rule or the `module:` field. Would it be worth validating that `ModulePath` is a prefix of `Path`, or at least warning when the default kicks in for a path under the otelc root?

tool/internal/setup/store.go
```diff
@@ -4,17 +4,77 @@
...
+			return dir, nil
+		}
+
+		for moduleDir := range moduleDirs {
+			pkgs, err := packages.Load(&packages.Config{
+				Mode:    packages.NeedFiles,
+				Context: ctx,
+				Dir:     moduleDir,
+			}, goPath)
+			if err != nil || len(pkgs) == 0 || len(pkgs[0].Errors) > 0 {
```
kakkoyun: Separate from the line-47 wording fix: this `continue` still discards the `packages.Load` error and `pkgs[0].Errors`, so when resolution fails in every module dir the caller only gets `failed to resolve import path <path>` with no underlying reason. A mistyped path and a package that fails to compile then look identical. Could we hold onto the last error (or the package's own errors) and fold it into the returned message? It would save a lot of guessing when a hook path doesn't resolve.

instrumentation/runtime/stub.go
```diff
@@ -0,0 +1,7 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// This file keeps the package importable. The implementation lives in files
```
kakkoyun, line 4: The stub plus `//go:build ignore` pairing is load-bearing here: without the stub, `packages.Load` finds no buildable files and the package comes back with errors, so resolution fails with the generic `failed to resolve import path` message. Right now that's only explained inside the stub files themselves. Could we add a short note next to `path:` in `docs/rules.md`, so the next person adding a build-ignored hook package knows the stub is required?

tool/internal/setup/sync.go
```diff
@@ -138,60 +142,43 @@ func (sp *SetupPhase) syncDeps(ctx context.Context, matched []*rule.InstRuleSet,
...
-		newPath = filepath.Join(util.GetBuildTempDir(), newPath)
-		replaces = append(replaces, &replaceDirective{
-			oldPath:    oldPath,
-			oldVersion: "",
-			newPath:    newPath,
-			newVersion: "",
-		})
+	replaces := make(map[string]string)
+	for _, m := range funcRules {
+		if path, isEmbedded := strings.CutPrefix(m.ModulePath, util.OtelcInstRoot+"/"); isEmbedded {
```
kakkoyun, line 147: This embedded branch does run during the integration build (runtime's `add_file` rule leaves `module:` unset, so `ModulePath` defaults to `.../instrumentation/runtime` and matches this prefix), but nothing pins it at the unit level. `TestSyncDeps_WithRules` leaves `ModulePath` empty, so the branch never fires there, and it only asserts the output contains `replace`, which the unconditional pkg/runtime/instrumentation entries already satisfy. A case that sets `ModulePath` under the instrumentation root and checks the replace target would catch a regression here, along with a rule-load test for the `module:` default.

tool/internal/setup/store.go
```diff
@@ -4,17 +4,77 @@
...
+				continue
+			}
+
+			util.Assert(len(pkgs) == 1, "expected exactly one package")
```
kakkoyun: `util.Assert` here ends up calling `os.Exit(1)` (through `ex.Fatalf`), so if it ever trips it takes the whole build down from inside resolution and skips the deferred `Cleanup` in `GoBuild`, leaving the backed-up `go.mod`/`go.sum` unrestored. A single import-path pattern should always resolve to one package, so this is more defensiveness than a live bug. Returning an error instead would let the normal cleanup path run if that assumption ever breaks.

amazingakai:
Thanks @kakkoyun 

* Added validation to ensure `module` is a prefix of `path`
* Documented the `stub.go` requirement for packages referenced by path whose implementation files use `//go:build ignore`
* Preserved the underlying resolution error instead of returning a generic failure
* Replaced the `Assert` with an error return for the multiple-package case
* Added test covering the `module` defaulting behavior
* Updated `TestSyncDeps_WithRules` to cover the embedded instrumentation replace path

tool/internal/setup/store.go
```diff
@@ -4,17 +4,88 @@
...
+
+	resolve := func(goPath string) (string, error) {
+		if dir, ok := cache[goPath]; ok {
+			return dir, nil
+		}
+
+		var lastErr error
+		for moduleDir := range moduleDirs {
+			pkgs, err := packages.Load(&packages.Config{
+				Mode:    packages.NeedFiles,
```
txabman42, line 34: Adding here `NeedModule` wouldn't give us the module directly, which would let us remove the new `module` in the rule yaml definition?

In this case, its a bitmask, so they should be combinable.

tool/internal/rule/func_rule.go
txabman42, line 80: Let's also check that `Path` can't be null.

txabman42 (changes requested):

tool/internal/setup/store.go
amazingakai: `module` is only required for built-in instrumentations where `path` is not a module path (currently only the otel sdk instrumentation).

We add `replace` directives to `go.mod` for built-in instrumentations. Using the import path in the `replace` directive does not work if it is not a module root, since `replace` expects a module path and the build fails otherwise.

Unfortunately, we can't use `packages.Load` to discover the module until the package is already required somewhere in the user's `go.mod`.

tool/internal/rule/func_rule.go
amazingakai: Added validation for `path` in func rule as well as file rule.

tool/internal/setup/store.go
txabman42: TY good point, agree with that!

txabman42 (approved):

kakkoyun (approved):

PR #618, opened 2026-06-27
feat(tool): add a state manager for tracking modified and generated files

### LINKED ISSUE #564: feat(tool/setup): track generated otelc.runtime.go files in a state file
Follow-up from review comment https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/515#discussion_r3330983185 by @txabman42.

Currently, generated `otelc.runtime.go` files are rediscovered during cleanup by recalculating packages from command arguments.

Instead, we could track generated files in a state file (similar to `backup.json`) and use that during cleanup. This would avoid recomputing the file list and make cleanup more deterministic.

amazingakai:
## Description

Adds a `StateManager` to track modified (`go.mod`, `go.sum`, `go.work.sum`) and generated (`otelc.runtime.go`, `go.sum`, `go.work.sum`) files.

## Motivation

* Track generated `otelc.runtime.go` files in `state.json` so `otelc cleanup` no longer needs the original package list to clean up correctly.
* Remove `go.sum` and `go.work.sum` if they did not exist before, instead of leaving behind empty files.
* Provide a clean API that can be reused across different stages and passed through `context.Context`.
* Can be easily extended to tracking directories in the future (for example `vendor/`).

Fixes #564 

---

## Checklist

- [x] Tests pass: `make test`


tool/internal/setup/setup.go
```diff
@@ -247,6 +247,20 @@ func Setup(ctx context.Context, cmd *cli.Command) error {
 		return ex.Wrapf(err, "matching dependencies to hook rules")
 	}
 
+	// Track generated & modified files with state manager
+	stateManager, found := StateManagerFromContext(ctx)
+	if !found {
```
txabman42, line 252: If there is a problem/interruption between `Setup` and `Cleanup`, file snapshots are copied but not the `state.json`.

tool/internal/setup/state.go
```diff
@@ -0,0 +1,196 @@
...
+
+func (s *StateManager) TrackAll(paths ...string) error {
+	var err error
+	for _, path := range paths {
+		err = ex.Join(err, s.Track(path))
+	}
+	return err
+}
+
+func (s *StateManager) Track(path string) error {
```
txabman42, line 146: Lets add doc for each of these public methods. Lets also mention that this is not thread safe, right now no problem, but always good to have it documented.

txabman42 (changes requested):
Nice and solid PR! ❤️ 

Just review what would happen if otelc is interrupted.

tool/internal/setup/setup.go
amazingakai: For `GoBuild`, this shouldn't be a problem. The `StateManager` is created in memory, and `Cleanup()` is deferred for the entire build. `SIGINT`/`SIGTERM` during the build unwind normally, so the deferred cleanup still runs. In that path, we don't rely on `state.json` at all.

For the standalone setup command, `Commit()` is deferred before any work begins, so it always runs before `Setup()` returns (unless the process terminates abnormally)

The remaining case is process termination where deferred functions are not executed (`os.Exit`, or `util.Assert`/`ex.Fatal`). That's not specific to this PR, and I think we should generally try to reduce the use of those where possible.

tool/internal/setup/state.go
amazingakai: Done

tool/internal/setup/setup.go
txabman42: Good point, agree 👍

txabman42 (approved):
LGTM! GJ! ❤️

NameHaibinZhang (approved):

PR #621, opened 2026-06-30
fix(instrumentation/grpc): prevent reentrant OTel SDK initialization in BeforeDialContext

amazingakai:
Split from #456.

Prevents a deadlock in applications that use `grpc.DialContext` together with an OTLP gRPC exporter configured via:
```
OTEL_EXPORTER_OTLP_PROTOCOL=grpc
OTEL_EXPORTER_OTLP_ENDPOINT=...
```

The fix skips instrumentation for OTLP exporter targets before gRPC instrumentation is initialized, preventing reentrant OpenTelemetry SDK initialization.

Co-authored-by: shown <yuluo08290126@gmail.com>

### DIFF
--- a/instrumentation/google.golang.org/grpc/internal/otlpfilter/otlpfilter.go
+++ b/instrumentation/google.golang.org/grpc/internal/otlpfilter/otlpfilter.go
@@ -1,0 +1,46 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package otlpfilter identifies OTLP exporter gRPC calls that must not be
+// instrumented by the gRPC instrumentation itself.
+package otlpfilter
+
+import (
+	"os"
+	"strings"
+)
+
+const (
+	exporterTracePath   = "/opentelemetry.proto.collector.trace.v1.TraceService/Export"
+	exporterMetricsPath = "/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+	exporterLogsPath    = "/opentelemetry.proto.collector.logs.v1.LogsService/Export"
+)
+
+var exporterEndpointEnvVars = []string{
+	"OTEL_EXPORTER_OTLP_ENDPOINT",
+	"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT",
+	"OTEL_EXPORTER_OTLP_METRICS_ENDPOINT",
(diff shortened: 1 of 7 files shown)


darccio (approved):
LGTM

instrumentation/google.golang.org/grpc/internal/otlpfilter/otlpfilter.go
```diff
@@ -0,0 +1,46 @@
...
+
+// IsExporterTarget returns true when target matches a configured OTLP exporter endpoint.
+func IsExporterTarget(target string) bool {
+	if target == "" {
+		return false
+	}
+
+	for _, envVar := range exporterEndpointEnvVars {
+		endpoint := os.Getenv(envVar)
+		if endpoint != "" && strings.Contains(endpoint, target) {
```
txabman42, line 41: This could not be the most robust solution.

- Endpoint could differ:
```
target   = "dns:///localhost:4317"
endpoint = "http://localhost:4317"   // OTEL_EXPORTER_OTLP_ENDPOINT
```
- There could be missmatches
```
endpoint = "http://collector-localhost:4317"
target   = "localhost:4317"
```

What about detecting if we are inside the inst initialization? In the end, we want to avoid the deadlock. Global flag for detecting it.

txabman42 (changes requested):

instrumentation/google.golang.org/grpc/internal/otlpfilter/otlpfilter.go
amazingakai: Agreed. I am thinking of moving the SDK initialization logic into a separate instrumentation module, since all instrumentations currently share the same code, so we can inject a file with an `init()` that initializes the SDK. I will put up a PR for it later today.

txabman42: I like that idea ❤️

kakkoyun: @amazingakai Shall we create an issue for this? If this is not in your immediate todo list?

amazingakai: I've created a PR (#625) but I need to go through it once again and rebase it before marking it ready for review.

PR #629, opened 2026-07-02
feat(tool): add root module rule target

### LINKED ISSUE #611: feat(tool): module-scoped (root module) glob target for first-party package selection

Instrumentation rules select packages via `target` — an exact import path or, since #549, a glob (`tool/internal/rule/target_glob.go`). There is no way to express "instrument all packages in the application's own (root) module" without either hardcoding the module path or relying on `target: main`, which matches only the `main` package and misses sibling first-party packages.

A module-scoped target that selects the root module's packages regardless of their import paths/names.

Prior art: DataDog/orchestrion's most-used package selector is `package-filter: { root: true }` — "instrument the application's own code only" — which matches the pattern against the import path *relative to the root module* (`internal/injector/aspect/join/package.go`). An equivalent here could be:

- a `root: true` qualifier on the rule, or
- a sentinel target, or
- resolving `<module-path>/**` from `go.mod` and feeding it through the existing glob engine.

- `target: <module-path>/**` — works today *if* the author knows the module path, but it is not portable across forks/renames and forces hardcoding.
- `target: main` — only the `main` package; misses first-party sub-packages.

**Additional context**

Surfaced while reviewing #561 (glob `target` matching) against orchestrion's implementation. The glob engine added in #561 is the natural place to build on. Note that ADR-0005 (import-driven instrumentation selection) reduces but does not eliminate the need. Refs #549, #346.

macayu17:
## Description

Adds `target: $root` for instrumentation rules.

During setup, `$root` is resolved from the build root module and expanded to `<module-path>/**`. Matching then goes through the existing glob target path, so this does not add a separate package-matching branch.

Fixes #611

## Motivation

Rules that target first-party code currently need either `target: main` or a hardcoded module path.

`target: main` only covers the main package, and hardcoding the module path is awkward for forks or renamed modules. `$root` gives rule authors a portable way to match the root package and its subpackages.

---

## Checklist

  - [x] Code formatted: `gofmt`
    - `make lint/go` passes locally
    - full `make lint` hits unrelated YAML formatting diffs in this Windows checkout
  - [x] Tests pass: `make test`
  - [ ] OpenTelemetry Registry updated (not applicable)

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -284,6 +284,11 @@ Rules:
 
 - `target: main` — matches the compile-time package named `main`.
 - `target: test_main` — not currently supported; reserved for future work.
+- `target: $root` — matches the root module of the build and every package
+  below it. If a build spans multiple modules, it matches the union of all
+  resolved roots. The setup phase expands this selector through the normal glob
+  target matcher, so rules do not need to hardcode the application's module
+  path.
 - An empty or whitespace-only `target` is rejected at load time: `target` is
   the sole package selector, so a rule without one can never match.
 
--- a/tool/internal/pkgload/pkgload.go
+++ b/tool/internal/pkgload/pkgload.go
@@ -7,6 +7,7 @@ package pkgload
 import (
 	"context"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/tools/go/packages"
 
@@ -23,16 +24,50 @@ func LoadPackages(
 	buildFlags []string,
 	patterns ...string,
 ) ([]*packages.Package, error) {
+	dir, buildFlags, err := loadDirFromBuildFlags(buildFlags)
+	if err != nil {
+		return nil, err
+	}
 	cfg := &packages.Config{
 		Mode:       mode,
 		Context:    ctx,
 		BuildFlags: buildFlags,
+		Dir:        dir,
 	}
 	pkgs, err := packages.Load(cfg, patterns...)
 	if err != nil {
 		return nil, ex.Wrapf(err, "loading packages %v", patterns)
 	}
 	return pkgs, nil
+}
+
+func loadDirFromBuildFlags(buildFlags []string) (string, []string, error) {
+	var dir string
+	filtered := make([]string, 0, len(buildFlags))
+	for i := 0; i < len(buildFlags); i++ {
+		flag := buildFlags[i]
+
+		switch {
+		case flag == "-C":
+			if i+1 >= len(buildFlags) {
+				return "", nil, ex.New("missing value for -C")
+			}
+			dir = buildFlags[i+1]
+			i++
+		case strings.HasPrefix(flag, "-C="):
+			dir = strings.TrimPrefix(flag, "-C=")
+		default:
+			filtered = append(filtered, flag)
+		}
+	}
+	if dir == "" {
+		return "", filtered, nil
+	}
+	absDir, err := filepath.Abs(dir)
+	if err != nil {
+		return "", nil, ex.Wrapf(err, "resolving -C directory %s", dir)
+	}
+	return absDir, filtered, nil
 }
 
 // ResolvePackageName returns the declared package name for an import path.
@@ -117,19 +152,19 @@ func PackageDir(pkg *packages.Package) string {
 	return ""
 }
 
-// resolveModuleDir returns the module directory for a given package directory.
-func resolveModuleDir(ctx context.Context, pkgDir string) (string, error) {
+// ResolveModule returns the module for a given package directory.
+func ResolveModule(ctx context.Context, pkgDir string) (*packages.Module, error) {
 	pkgs, err := LoadPackages(ctx, packages.NeedModule, nil, pkgDir)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	if len(pkgs) == 0 {
-		return "", ex.Newf("no packages found for directory: %s", pkgDir)
+		return nil, ex.Newf("no packages found for directory: %s", pkgDir)
 	}
 
 	pkg := pkgs[0]
 	if pkg.Module == nil || pkg.Module.Dir == "" || len(pkg.Errors) > 0 {
-		return "", ex.Newf(
+		return nil, ex.Newf(
 			"failed to load module information for package in directory %s: module=%v, errors=%v",
 			pkgDir,
 			pkg.Module,
@@ -137,7 +172,16 @@ func resolveModuleDir(ctx context.Context, pkgDir string) (string, error) {
 		)
 	}
 
-	return pkg.Module.Dir, nil
+	return pkg.Module, nil
+}
+
+// ResolveModuleDir returns the module directory for a given package directory.
+func ResolveModuleDir(ctx context.Context, pkgDir string) (string, error) {
+	mod, err := ResolveModule(ctx, pkgDir)
+	if err != nil {
+		return "", err
+	}
+	return mod.Dir, nil
 }
 
 func FindModuleDirs(ctx context.Context, pkgs []*packages.Package) (map[string]bool, error) {
@@ -161,7 +205,7 @@ func FindModuleDirs(ctx context.Context, pkgs []*packages.Package) (map[string]b
 		if pkg.Module != nil {
 			moduleDir = pkg.Module.Dir
 		} else {
-			modDir, err := resolveModuleDir(ctx, pkgDir)
+			modDir, err := ResolveModuleDir(ctx, pkgDir)
 			if err != nil {
 				return nil, ex.Wrapf(err, "finding module dir for package %s", pkg.PkgPath)
 			}
--- a/tool/internal/rule/target_glob.go
+++ b/tool/internal/rule/target_glob.go
@@ -16,6 +16,15 @@ import (
 // import path that keeps the fast map-lookup matching path.
 const globMeta = "*?[{"
 
+// TargetRoot selects the root module of the build. The setup phase expands it
+// to a concrete module glob before matching rules.
+const TargetRoot = "$root"
+
+// IsRootTarget reports whether target is the root-module selector.
+func IsRootTarget(target string) bool {
+	return target == TargetRoot
+}
+
 // IsGlobTarget reports whether target uses glob syntax and therefore must be
 // matched against every dependency import path rather than looked up by exact
 // key. An empty target is not a glob.
@@ -32,6 +41,9 @@ func IsGlobTarget(target string) bool {
 // before it reaches ValidateTarget; a non-glob target is a literal import path
 // and is always valid.
 func ValidateTarget(target string) error {
+	if strings.Contains(strings.ToLower(target), TargetRoot) && target != TargetRoot {
+		return ex.Newf("target %q must be exactly %q", target, TargetRoot)
+	}
 	if !IsGlobTarget(target) {
 		return nil
 	}
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -134,6 +134,43 @@ func matchVersion(dependency *Dependency, rule rule.InstRule) bool {
 	return util.VersionInRange(dependency.Version, rule.GetVersion())
 }
 
+type targetRule struct {
+	target string
+	rule   rule.InstRule
+}
+
+func (sp *SetupPhase) matchGlobRules(
+	dep *Dependency,
+	relevantRules []rule.InstRule,
+	globRules []targetRule,
+) []rule.InstRule {
+	var matched []rule.InstRule
+	var seen map[rule.InstRule]bool
+	for _, gr := range globRules {
+		if !rule.MatchGlobTarget(gr.target, dep.ImportPath) {
+			continue
+		}
+		if matched == nil {
+			matched = make([]rule.InstRule, 0, len(relevantRules)+1)
+			matched = append(matched, relevantRules...)
+			seen = make(map[rule.InstRule]bool, len(relevantRules)+1)
+			for _, r := range relevantRules {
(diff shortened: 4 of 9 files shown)


tool/internal/setup/match.go
```diff
@@ -506,11 +511,20 @@ func (sp *SetupPhase) matchDeps(ctx context.Context, deps []*Dependency) ([]*rul
...
 	exactRules := make(map[string][]rule.InstRule)
-	globRules := make([]rule.InstRule, 0)
+	globRules := make([]targetRule, 0)
 	for _, r := range allRules {
 		target := r.GetTarget()
+		if rule.IsRootTarget(target) {
+			if len(sp.rootModulePaths) == 0 {
+				return nil, ex.Newf("rule %q uses target %q, but no root module was found", r.GetName(), target)
+			}
+			for _, root := range sp.rootModulePaths {
```
kakkoyun, line 591: When `sp.rootModulePaths` has more than one entry (e.g. a monorepo with nested modules, or a `go.work` build), this loop pushes the same `rule.InstRule` into `globRules` once per root:

```go
for _, root := range sp.rootModulePaths {
    globRules = append(globRules, targetRule{target: root + "/**", rule: r})
}
```

If two resolved root paths overlap as glob prefixes (e.g. `example.com/app` and `example.com/app/plugin` as separate modules), a dependency under the nested module matches both expanded globs. `runMatch`'s glob loop has no dedup:

```go
for _, gr := range globRules {
    if !rule.MatchGlobTarget(gr.target, dep.ImportPath) {
        continue
    }
    ...
    matched = append(matched, gr.rule)
}
```

The same rule instance gets appended twice, so the dependency ends up matched/hooked twice. Worth deduping either at expansion time (skip a root path that's a sub-path of another already-collected root) or at match time (skip a rule already present in `matched` for this dependency).

tool/internal/rule/target_glob.go
```diff
@@ -16,6 +16,15 @@ import (
 // import path that keeps the fast map-lookup matching path.
 const globMeta = "*?[{"
 
+// TargetRoot selects the root module of the build. The setup phase expands it
+// to a concrete module glob before matching rules.
+const TargetRoot = "$root"
+
+// IsRootTarget reports whether target is the root-module selector.
+func IsRootTarget(target string) bool {
```
kakkoyun, line 24:
```go
func IsRootTarget(target string) bool {
    return target == TargetRoot
}
```

is exact-string equality with no companion check in `ValidateTarget`. A rule author who writes `$ROOT`, `$root ` (trailing space from copy-paste), or `$root/**` gets no error at load time. None of those trip `IsGlobTarget` in a way that errors — `$root/**` contains `*` so it's treated as an ordinary (syntactically valid) doublestar glob, and the other two fall into the exact-match path. Either way the rule silently never matches any real import path (they can't contain `$`), with no warning.

This is the same failure class the empty-target check already guards against (`ValidateTarget` rejects blank targets specifically so they don't "silently never match any real import path"). Suggest extending that same loud-failure treatment to any target containing the substring `$root` that isn't an exact match.

tool/internal/setup/setup.go
```diff
@@ -249,6 +252,32 @@ func getPackageDir(pkg *packages.Package) string {
...
+		}
+		pkgDir := getPackageDir(pkg)
+		if pkgDir == "" {
+			continue
+		}
+		moduleDir, err := pkgload.ResolveModuleDir(ctx, pkgDir)
+		if err != nil {
+			return nil, ex.Wrapf(err, "finding module dir for package %s", pkg.PkgPath)
+		}
+		mf, err := parseGoMod(filepath.Join(moduleDir, "go.mod"))
```
kakkoyun:
```go
moduleDir, err := pkgload.ResolveModuleDir(ctx, pkgDir)
...
mf, err := parseGoMod(filepath.Join(moduleDir, "go.mod"))
```

`ResolveModuleDir` already calls `packages.Load(..., packages.NeedModule, ...)` internally, and that call's `pkg.Module.Path` is exactly the value this code re-derives by hand-parsing `go.mod` with `golang.org/x/mod/modfile`. That's a second, independently-versioned go.mod parser sitting in the same code path as the toolchain's own resolution — a go.mod a locally installed newer Go accepts (new directives the vendored `x/mod` doesn't know yet) can make this call fail even though `go build` itself would succeed.

Simplest fix: have `ResolveModuleDir` (or a sibling function) return `pkg.Module.Path` alongside `pkg.Module.Dir` from the same `packages.Load` call, and drop the `parseGoMod` step entirely for this call site.

tool/internal/setup/setup.go
```diff
@@ -319,6 +348,10 @@ func Setup(ctx context.Context, cmd *cli.Command) error {
 	if err != nil {
 		return err
 	}
+	sp.rootModulePaths, err = rootModulePaths(ctx, pkgs)
```
kakkoyun:
```go
sp.rootModulePaths, err = rootModulePaths(ctx, pkgs)
if err != nil {
    return err
}
```

This runs unconditionally before any rule is loaded. Its failure mode (see the reliability comment on `rootModulePaths`) can abort a build whose rule set never references `$root` at all. Consider deferring this resolution until `matchDeps` actually finds a `$root` rule, so unrelated rule sets aren't exposed to a corner case they don't use.

tool/internal/setup/match_test.go
```diff
@@ -1247,6 +1267,67 @@ func TestMatchDeps_GlobTargetSplit(t *testing.T) {
 	require.False(t, matchedPaths["example.com/other"], "unrelated package must not match")
 }
 
+func TestMatchDeps_RootTargetExpandsToRootModuleGlob(t *testing.T) {
```
kakkoyun, line 1353: The multi-root `$root` expansion path (`for _, root := range sp.rootModulePaths` in match.go) has no end-to-end test with 2+ roots — this test and `TestRootModulePaths` each exercise the pieces separately, never the fan-out together. That gap is exactly why the duplicate-rule-application bug flagged above wasn't caught here.

docs/rules.md
```diff
@@ -249,6 +249,10 @@ Rules:
 
 - `target: main` — matches the compile-time package named `main`.
 - `target: test_main` — not currently supported; reserved for future work.
+- `target: $root` — matches the root module of the build and every package
```
kakkoyun, line 287: > `target: $root` — matches the root module of the build and every package below it.

The wording implies a single canonical root, but `rootModulePaths()` can resolve multiple root paths (deduped, sorted) and `matchDeps` OR's a `$root` rule across all of them. Worth a line clarifying that when a build spans multiple modules, `$root` matches the union of all resolved roots, not "the" root.

tool/internal/setup/match.go
```diff
@@ -165,16 +170,16 @@ func (sp *SetupPhase) runMatch(
...
 				continue
 			}
 			if matched == nil {
 				matched = make([]rule.InstRule, 0, len(relevantRules)+1)
 				matched = append(matched, relevantRules...)
 			}
-			matched = append(matched, r)
-			sp.Debug("Match glob target", "rule", r.GetName(), "target", r.GetTarget(), "dep", dep.ImportPath)
+			matched = append(matched, gr.rule)
+			sp.Debug("Match glob target", "rule", gr.rule.GetName(), "target", gr.target, "dep", dep.ImportPath)
```
kakkoyun: This is the only log line that shows the resolved glob (`gr.target`, e.g. `example.com/app/**`). Downstream match/log calls further into `matchOneRule` still log the rule's original `Target` field, which for a `$root` rule is still the literal string `"$root"`. Someone grepping logs to see why a rule fired against a specific dependency won't see the concrete module path there — a small but real debuggability gap given `$root` exists specifically to abstract that path away.

tool/internal/setup/setup.go
```diff
@@ -249,6 +252,32 @@ func getPackageDir(pkg *packages.Package) string {
...
+		}
+		moduleDir, err := pkgload.ResolveModuleDir(ctx, pkgDir)
+		if err != nil {
+			return nil, ex.Wrapf(err, "finding module dir for package %s", pkg.PkgPath)
+		}
+		mf, err := parseGoMod(filepath.Join(moduleDir, "go.mod"))
+		if err != nil {
+			return nil, err
+		}
+		if mf.Module != nil && mf.Module.Mod.Path != "" {
```
kakkoyun:
```go
if mf.Module != nil && mf.Module.Mod.Path != "" {
    roots[mf.Module.Mod.Path] = true
}
```

This takes the parsed go.mod module path with no `golang.org/x/mod/module.CheckPath` validation, unlike the normal path (a few lines up) where `go/packages` has already validated the module path's syntax before populating `pkg.Module.Path`. A go.mod with an invalid module path would likely fail earlier in the Go toolchain anyway — still, aligning the two code paths would remove the inconsistency.

tool/internal/setup/setup.go
```diff
@@ -249,6 +252,32 @@ func getPackageDir(pkg *packages.Package) string {
...
+	for _, pkg := range pkgs {
+		if pkg.Module != nil && pkg.Module.Path != "" {
+			roots[pkg.Module.Path] = true
+			continue
+		}
+		pkgDir := getPackageDir(pkg)
+		if pkgDir == "" {
+			continue
+		}
+		moduleDir, err := pkgload.ResolveModuleDir(ctx, pkgDir)
```
kakkoyun: `pkgload.ResolveModuleDir` (which spawns a `go list` subprocess) is called twice for the same package directory in the file-target case — once here in `rootModulePaths`, again later in `generateRuntimePerPackage`. Caching/threading the first result through would save a subprocess spawn per build.

tool/internal/pkgload/pkgload.go
```diff
@@ -31,6 +38,34 @@ func LoadPackages(
 	return pkgs, nil
 }
 
+func loadDirFromBuildFlags(buildFlags []string) (string, []string, error) {
```
kakkoyun, line 44: `-C`/`-C=` flag parsing (space-separated vs `=`-joined forms) is now implemented independently here and again in `setup.go`'s `flagsWithPathValues`/`extractBuildFlags`/`splitBuildTargets`. Three places now understand this flag's shape; a shared helper would prevent them drifting apart if a fourth callsite is added later.

tool/internal/setup/setup.go
```diff
@@ -402,6 +435,8 @@ func setupGoCache(ctx context.Context, env []string) ([]string, error) {
 //
 //nolint:gochecknoglobals // private lookup table
 var buildContextFlagsWithValue = map[string]bool{
```
kakkoyun, line 409: `-C`'s value is resolved with `filepath.Abs` in `loadDirFromBuildFlags`, but `-overlay`'s path is forwarded as-is with no equivalent resolution relative to a `-C` directory. If both flags are combined with a relative `-overlay` path, the base directory used to resolve it may not match what the real `go` toolchain assumes. Not verified against `go/packages`' actual overlay-resolution semantics — worth checking before relying on this.

tool/internal/pkgload/pkgload.go
```diff
@@ -31,6 +38,34 @@ func LoadPackages(
...
+	var dir string
+	filtered := make([]string, 0, len(buildFlags))
+	for i := 0; i < len(buildFlags); i++ {
+		flag := buildFlags[i]
+		switch {
```
kakkoyun, line 50:
```go
switch {
case flag == "-C":
```

`wsl` (from an unrestricted `golangci-lint run --no-config --default=all` pass, not part of this repo's own lint config): anonymous switch statement cuddled with the preceding line. No action needed unless the team wants to adopt `wsl` project-wide.

kakkoyun (commented):
Reviewed this against the stated intent (portable `$root` selector reusing the existing glob matcher). Baseline is clean: `go build`, `go vet`, `gofmt -l`, and `go test -race -count=1 ./tool/...` all pass, and all 47 CI checks on the PR are green.

Three Major issues worth fixing before merge (inline below): a duplicate-rule-application bug when root module paths overlap as glob prefixes, a silent-failure gap for `$root` typos, and a redundant/riskier go.mod re-parse in the file-target fallback. The rest are minor/advisory hardening and process notes.

One more note that doesn't have a code location to anchor to: the `-C`/`-overlay` build-flag plumbing in this PR is a legitimate prerequisite for `$root` resolution to work correctly under `-C`, but it also fixes `ResolvePackageName`/`ResolveExportFiles` (used during the toolexec phase, unrelated to `$root`) for any `-C`/`-overlay` build — a wider blast radius than the PR description discloses. Worth a line in the description calling that out.

macayu17:
Sorry for the noise around here. I should have tightened the edge cases before pushing the first pass.

I pushed `642068e` addressing the review items: deduped overlapping `$root` matches, made invalid `$root` variants fail during validation, stopped reparsing `go.mod`, deferred root resolution until needed, and added multi-root coverage plus the docs clarification.

macayu17:
I also looked into the benchmark failure. It looks like the baseline overhead check is sitting right on the edge rather than failing deterministically from this PR.

kakkoyun:
@macayu17 this needs a rebase

kakkoyun:
Thanks for the fast turnaround on `642068e`. The dedup fix, the `$root` validation, dropping the go.mod reparse, all of it checks out and the tests back it up. Resolved those threads.

A handful of the minor/advisory comments are still open: logging, the duplicate `ResolveModule` call, the `-C` parsing duplication, the `-overlay` path handling, and the PR description note. None of that blocks merging as far as I'm concerned, happy to see it as a follow-up if that's easier for you.

kakkoyun:
@macayu17 This needs one last rebase. Let's push this to the finish line.

kakkoyun (approved):

PR #649, opened 2026-07-04
refactor(tool): optimize package name discovery for file-only rules

### LINKED ISSUE #648: perf(tool): optimize package name discovery for file-only rules to avoid AST parsing on the hot path
# Issue: Optimize package name discovery for file-only rules on the hot path

## Description
Currently, in [match.go](file:///d:/opentelemetry-go-compile-instrumentation/tool/internal/setup/match.go#L144-L151), the tool chain parses source files on the hot path to discover the package name for file-only rules using `ast.ParseFileOnlyPackage`:

```go
// TODO: Optimize package name discovery for file-only rules to avoid
// parsing source files on the hot path.
tree, err := ast.ParseFileOnlyPackage(dep.Sources[0])
if err != nil {
    return nil, err
}
set.SetPackageName(tree.Name.Name)
```

Since the setup/instrumentation phase runs during compilation, performing AST parsing of source files on the hot path increases build times, especially for large projects with numerous file-level rules.

## Proposed Optimization
We should optimize package name discovery to avoid parsing source files on the hot path. Possible approaches include:
1. Caching discovered package names.
2. Fast regex/string scanning of the package declaration at the start of the source files rather than full AST parsing.
3. Pre-extracting package names during the setup phase.

SatyamPandey-07:
## PR Title
refactor(tool): optimize package name discovery for file-only rules

fixes #648 

## Description

This PR optimizes the package name discovery phase for file-only rules during the compile setup step. 

Key changes:
- Introduced `ast.ParsePackageName` in `tool/internal/ast/parser.go`, which uses the standard library `go/parser` directly to read the package name without AST/DST decoration.
- Updated `tool/internal/setup/match.go` to use `ast.ParsePackageName` instead of `ast.ParseFileOnlyPackage`.
- Added a unit test `TestParsePackageName` in `tool/internal/ast/ast_test.go` to verify the functionality.
- Added benchmarks comparing `ParseFileOnlyPackage` and `ParsePackageName` in `tool/internal/ast/ast_test.go`.

## Motivation

When executing file-only rules, the setup step previously parsed source files using `ast.ParseFileOnlyPackage`. While this only parsed the package clause, it still invoked full AST/DST decoration and comment mapping via the `decorator` package, introducing unnecessary overhead on the compilation hot path. This optimization avoids the decoration overhead.

### Benchmark Results
Below are the benchmark results comparing the original `ParseFileOnlyPackage` (with DST/AST decoration) against the optimized `ParsePackageName` (without decoration) when parsing `parser.go` (136 lines):

```text
goos: windows
goarch: amd64
BenchmarkParseFileOnlyPackage-16    	    1765	    626382 ns/op	   17141 B/op	      86 allocs/op
BenchmarkParsePackageName-16        	    6126	    204047 ns/op	   12538 B/op	      31 allocs/op
```

- **Execution Speed:** ~3x faster (~204 µs vs. ~626 µs)
- **Allocated Bytes:** ~1.37x less memory (12.5 KB vs. 17.1 KB)
- **Heap Allocations:** ~2.7x fewer allocations (31 vs. 86)

Fixes #648

---

## Checklist

- [x] Tests pass: `make test`


SatyamPandey-07:
@kakkoyun @txabman42 @darccio 
The Overhead Threshold Check failure is unrelated to this change — this PR optimizes package name parsing to be faster (avoids DST decoration). The benchmark runs on a shared ubuntu-latest runner which is susceptible to noisy-neighbour timing variance. Re-running the check should confirm no regression.

kakkoyun (approved):

kakkoyun (commented):
LGTM.

However, could we add a benchmark to show the improvement?

SatyamPandey-07:
@kakkoyun 
Here are the benchmark results comparing the original `ParseFileOnlyPackage` (which performs DST/AST decoration) against the optimized `ParsePackageName` (which skips decoration) when parsing `parser.go` (136 lines):

```text
goos: windows
goarch: amd64
BenchmarkParseFileOnlyPackage-8   	    1550	    751141 ns/op	  359560 B/op	    6556 allocs/op
BenchmarkParsePackageName-8       	   13968	     85293 ns/op	   10474 B/op	     170 allocs/op

**Comment by kakkoyun** [comment]
@SatyamPandey-07 Thanks. Let's put that into the PR description. And I'd be happy to merge this when the CI is green.

File: tool/internal/ast/parser.go [hunk 1]
```diff
@@ -106,6 +106,21 @@ func ParseFileOnlyPackage(filePath string) (*dst.File, error) {
...
+	if err != nil {
+		return "", ex.Wrapf(err, "failed to open file %s", filePath)
+	}
+	defer file.Close()
+	astFile, err := parser.ParseFile(fset, filepath.Base(filePath), file, parser.PackageClauseOnly)
+	if err != nil {
+		return "", ex.Wrapf(err, "failed to parse package name from file %s", filePath)
+	}
+	return astFile.Name.Name, nil
+}
```
y1yang0, line 116: Identical to ParseFileOnlyPackage?

y1yang0 (changes requested):

y1yang0 (commented):
It looks like this TODO has already been completed; all that's left to do is remove the TODO comment.

tool/internal/ast/parser.go
SatyamPandey-07: > Identical to ParseFileOnlyPackage?

Not quite — the key difference is in the call path:

- `ParseFileOnlyPackage` calls `NewAstParser().Parse(...)`, which constructs a `decorator.Decorator` and runs `dec.DecorateFile(astFile)`. Even with `PackageClauseOnly`, the DST/AST decoration and comment-mapping overhead still runs.
- `ParsePackageName` calls `go/parser.ParseFile` directly, skips the decorator entirely, and returns just the `string` package name.

The benchmarks added in this PR demonstrate the difference:

```text
BenchmarkParseFileOnlyPackage-16   1765   626382 ns/op   17141 B/op   86 allocs/op
BenchmarkParsePackageName-16       6126   204047 ns/op   12538 B/op   31 allocs/op
~3x faster, fewer allocations, by avoiding the decorator entirely.

SatyamPandey-07:
> It looks like this TODO has already been completed; all that's left to do is remove the TODO comment.

@y1yang0 
Yes, the TODO comment was already removed as part of this PR — the current match.go no longer contains it. The PR replaces the old ast.ParseFileOnlyPackage call (and its TODO) with ast.ParsePackageName directly.

SatyamPandey-07:
@kakkoyun The PR description has been updated with the benchmark results. All CI checks are now green. Ready for merge whenever you are!

kakkoyun:
@SatyamPandey-07 Let's wait for @y1yang0 answer. If he is satisfied then we can merge.

y1yang0:
> @kakkoyun The PR description has been updated with the benchmark results. All CI checks are now green. Ready for merge whenever you are!

But you added a new ast.ParsePackageName without removing the old ast.ParseFileOnlyPackage. On the other hand, ParsePackageName duplicates most of the code in AstParser.Parse, which is a code smell.

I suggest: Instead of adding ParsePackageName, add a `parseOnly bool` parameter to the AstParser constructor and skip the decorator phase.

SatyamPandey-07:
Good catch, thank you! `ParseFileOnlyPackage` had no remaining production callers after `match.go` was updated — it was only kept for the benchmark comparison. I've now removed it and updated the benchmark to call `NewAstParser().Parse("parser.go", parser.PackageClauseOnly)` directly, which is exactly equivalent.

Regarding the `parseOnly bool` suggestion: the challenge is that `AstParser.Parse` returns `*dst.File`, which inherently requires the DST decoration step. A `parseOnly` flag would either break the return type contract or return a misleadingly incomplete DST tree. A standalone `ParsePackageName(string) (string, error)` is the most honest API for this use case since the caller only needs the name.
@y1yang0

SatyamPandey-07:
The "Coverage Integration Tests" failure appears to be a flaky CI issue — it ran for the full 10-minute timeout and the test suite covers network-dependent tests (HTTP, gRPC, Kafka, MongoDB, etc.) that are unrelated to this change. Our change only touches package name parsing in `tool/internal/ast/` and `tool/internal/setup/match.go`.

Could a maintainer re-run the failing check? Happy to do so if I have the permissions.
@kakkoyun @y1yang0

y1yang0:
There are several approaches you can take. You could store the parseOnly field in the AST parser and then use it during parsing. Alternatively, a simpler approach would be to detect the PackageClauseOnly mode during parsing and skip decorator processing accordingly. I'd recommend the latter.

SatyamPandey-07:
Good suggestion, thanks! I've reworked the implementation: `AstParser.Parse` now detects `parser.PackageClauseOnly` mode and skips the DST decoration step, returning a minimal `*dst.File` with only `Name` populated. `ParsePackageName` is now just a thin wrapper over `NewAstParser().Parse()` — no duplicated file-open/parse logic.
@y1yang0

y1yang0 (approved):
LGTM. Let's see if @kakkoyun  is happy with the latest version.

SatyamPandey-07:
@kakkoyun pls review when u are free

amazingakai (approved):

kakkoyun (approved):
LGTM

PR #657, opened 2026-07-06
fix(cli): handle -C flag for go build and go install

### LINKED ISSUE #507: Support go build -C

HARSHVARANDANI:
## Description

This PR adds support for the `-C <dir>` flag in `otelc go build` and `otelc go install` invocations.

A new `goInvocation` struct is introduced to parse and carry all relevant information from the raw argument slice: the subcommand (`build`/`install`), the build directory extracted from `-C`, its placement relative to the subcommand (before or after), and the remaining build arguments with `-C` stripped out.

The parsed `goInvocation` is threaded through the following call chain:

- `parseGoInvocation` validates the subcommand and extracts `-C <dir>` when it appears in a supported position (immediately before or immediately after the subcommand).
- `getBuildPackages` / `splitBuildTargets` now accept a `buildDir` parameter and set `packages.Config.Dir` so that patterns like `.` and `./...` are resolved relative to `<dir>` instead of the process working directory.
- `absBuildPath`: resolves file-target paths relative to `buildDir`.
- `BuildWithToolexec` / `Cleanup`: reconstruct the `go` invocation with `-C` in the correct position and forward the build directory to all downstream steps.
- `SetupCommand` / `CleanupCommand`: thin CLI adapter wrappers that parse the invocation from `cli.Command` args before delegating to `Setup`/`Cleanup`, keeping the CLI layer decoupled from the core logic.

### Tests added

- `TestParseGoInvocation`: verifies `-C` extraction in all supported positions (before subcommand, after subcommand, and ignored when mid-args).
- `TestGetPackagesWithBuildDir`: verifies that `getBuildPackages` correctly loads packages when invoked from a directory different from the module root.
- Existing `TestGetPackages` and `TestSplitBuildTargets` call sites updated to pass the new `buildDir` parameter.

## Motivation

`otelc go build -C <dir> [pkgs]` did not respect the `-C` flag. The flag was never forwarded to the package loading step: `packages.Config` had no `Dir` set, so package patterns like `.` and `./...` were resolved against the process working directory instead of `<dir>`. This made it impossible to instrument a project located in a different directory without first `cd`-ing into it.

Fixes #507 

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -127,119 +127,6 @@ func absBuildPath(buildDir, target string) (string, error) {
 	return filepath.Abs(target)
 }
 
-func resolveDirRelativeToBuild(buildDir, dir string) string {
-	if buildDir == "" || filepath.IsAbs(dir) || strings.HasPrefix(dir, "$") {
-		return dir
-	}
-	absDir, err := absBuildPath(buildDir, dir)
-	if err != nil {
-		return dir
-	}
-	return absDir
-}
-
-type buildDirPlacement int
-
-const (
-	buildDirPlacementNone             buildDirPlacement = iota
-	buildDirPlacementBeforeSubcommand                   // go -C dir build
-	buildDirPlacementAfterSubcommand                    // go build -C dir
-)
-
-type buildInvocation struct {
-	command          string            // "build", "install" or "test"
-	buildDir         string            // the -C directory, if any
-	buildDirPosition buildDirPlacement // where -C appeared
-	args             []string          // remaining args with -C stripped
-}
-
-// appendGoCommand reconstructs the `go` command with `-C` in its original
-// position and appends it to the provided slice.
-func (inv buildInvocation) appendGoCommand(args []string) []string {
-	args = append(args, "go")
-	if inv.buildDirPosition == buildDirPlacementBeforeSubcommand && inv.buildDir != "" {
-		args = append(args, "-C", inv.buildDir)
-	}
-	args = append(args, inv.command)
-	if inv.buildDirPosition == buildDirPlacementAfterSubcommand && inv.buildDir != "" {
-		args = append(args, "-C", inv.buildDir)
-	}
-	return args
-}
-
-// minArgsForCFlag is the minimum number of arguments required when the first
-// argument is "-C": the flag itself, the directory, and the subcommand.
-const minArgsForCFlag = 3
-
-func parseGoInvocation(args []string) (buildInvocation, error) {
-	if len(args) == 0 {
-		return buildInvocation{}, ex.Newf(
-			"no command provided. Only 'go build' and 'go install' are supported",
-		)
-	}
-
-	invocation := buildInvocation{buildDirPosition: buildDirPlacementNone}
-	switch args[0] {
-	case subcmdBuild, subcmdInstall, subcmdTest:
-		invocation.command = args[0]
-		invocation.args = args[1:]
-		// Check for -C immediately after the subcommand: `build -C dir`
-		if len(invocation.args) >= 2 && invocation.args[0] == "-C" {
-			invocation.buildDir = invocation.args[1]
-			invocation.buildDirPosition = buildDirPlacementAfterSubcommand
-			invocation.args = invocation.args[2:]
-		}
-		return invocation, nil
-	case "-C":
-		// -C before the subcommand: `-C dir build`
-		if len(args) < minArgsForCFlag {
-			return buildInvocation{}, ex.Newf(
-				"no command provided. Only 'go build' and 'go install' are supported",
-			)
-		}
-		switch args[2] {
-		case subcmdBuild, subcmdInstall, subcmdTest:
-			// supported
-		default:
-			return buildInvocation{}, ex.Newf(
-				"unsupported command: %s. Only 'go build' and 'go install' are supported",
-				args[2],
-			)
-		}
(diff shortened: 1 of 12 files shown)


HARSHVARANDANI:
These are the same changes as PR#622 but in a new PR because I messed up a little bit with the merge conflicts :sweat_smile:

amazingakai (changes requested):
GJ. I have one suggestion though: wouldn't it be simpler to just `Chdir` to the path passed via `-C` instead of saving the directory and passing it around everywhere?

IMO it'd be easier to maintain, since it's easy to forget to pass the directory correctly in future changes.

amazingakai:
Please also fix merge conflicts.

tool/internal/setup/setup.go
```diff
@@ -263,17 +365,18 @@ func (sp *SetupPhase) generateRuntimePerPackage(
...
-	// we need to handle the arguments accordingly. If the command is `go build` or `go install`, we should trim the first argument
-	args := cmd.Args().Slice()
-	subcommand := subcmdBuild
-	if cmd.Name == "go" {
-		subcommand = cmd.Args().First() // build / install / test
-		args = cmd.Args().Tail()        // trim the subcommand
+// SetupCommand is the CLI entry point for `otelc setup`.
+// It parses the go invocation from the CLI args and delegates to Setup.
+func SetupCommand(ctx context.Context, cmd *cli.Command) error {
+	invocation, err := parseGoInvocation(cmd.Args().Slice())
```
txabman42: We have removed here the `cmd.Name == "go"` condition. This would make commands like `otelc setup` fail

txabman42 (changes requested):
Let's also support -C=<path>, lets see if that happen with other flags

HARSHVARANDANI:
> GJ. I have one suggestion though: wouldn't it be simpler to just `Chdir` to the path passed via `-C` instead of saving the directory and passing it around everywhere?
> 
> IMO it'd be easier to maintain, since it's easy to forget to pass the directory correctly in future changes.

Apologies for the late response. `go -C <dir>` already performs a directory change, internally the Go tool itself switches to that directory before executing the build. My reasoning for carrying the build directory through the code instead of calling o`s.Chdir` is that `os.Chdir` changes the working directory for the entire process. From that point onward, every relative path lookup and subprocess invocation would run from the new directory. That would require auditing any code that relies on the current working directory to ensure it's still operating correctly which would make this a much larger and more invasive change than explicitly passing the build directory where it's needed.

tool/internal/setup/setup.go
HARSHVARANDANI: I must've missed that, I'll fix it :+1:

txabman42:
> Apologies for the late response. go -C <dir> already performs a directory change, internally the Go tool itself switches to that directory before executing the build. My reasoning for carrying the build directory through the code instead of calling os.Chdir is that os.Chdir changes the working directory for the entire process. From that point onward, every relative path lookup and subprocess invocation would run from the new directory. That would require auditing any code that relies on the current working directory to ensure it's still operating correctly which would make this a much larger and more invasive change than explicitly passing the build directory where it's needed.

I think @amazingakai is the good direction at first sight, although we need to see if there are any limitations. Here Chdir reduces the audit surface rather than enlarging it. This change would force us to _every current and future cwd consumer must remember to accept and forward buildDir_, which looks good that idea. However, we need to make sure nothing resolves a path before the chdir happens (logger setup, -rules resolution, GetOtelcWorkDir initialization, ...)

For example, right now, the threading approach missed `findDeps`/`listBuildPlan` (the dry-run build plan still runs from the original cwd), and `vendoringActive` still checks the wrong directory.

HARSHVARANDANI:
> I think @amazingakai is the good direction at first sight, although we need to see if there are any limitations. Here Chdir reduces the audit surface rather than enlarging it. This change would force us to _every current and future cwd consumer must remember to accept and forward buildDir_, which looks good that idea. However, we need to make sure nothing resolves a path before the chdir happens (logger setup, -rules resolution, GetOtelcWorkDir initialization, ...)
> 
> For example, right now, the threading approach missed `findDeps`/`listBuildPlan` (the dry-run build plan still runs from the original cwd), and `vendoringActive` still checks the wrong directory.

I took a closer look at what would actually need to be audited before an early os.Chdir, and from what I found the main areas are global file flag resolution (e.g. --rules), GetOtelcWorkDir/logger initialization, vendoring detection, and ensuring we don't forward -C to subprocesses after changing directories.

Looking through the execution flow, this does contradict my earlier assumption that os.Chdir would have a larger audit surface. In comparison, explicitly passing buildDir requires every current and future cwd consumer to remember to accept and propagate it, and as you pointed out, that already led to a couple of misses that you pointed out (findDeps/listBuildPlan and vendoringActive).

Based on that comparison, the os.Chdir approach does seem like the simpler long-term maintenance option for the codebase. If you're happy with that direction, I'm happy to rework the implementation accordingly.

txabman42:
@HARSHVARANDANI sounds good to me to explore it

Issue #552, opened 2026-05-29
Proposal: Move file-level rule matching to toolexec phase


amazingakai:
## Background
PR #495 fixes a bug where matched instrumentation rules became stale after `syncDeps` mutated `go.mod` during setup. The fix (conditional re-run of setup when an instrumented dependency is bumped) works, but was acknowledged as a patch over a deeper design problem.

*(Update: The correctness motivation for this RFC was resolved by #655 (pin-stage go mod tidy). However, the core design change proposed here still offers major architectural and efficiency benefits, so this issue tracks the exploration of this refactor as a longer-term structural improvement.)*

## Proposed Idea

Move file-level rule matching to the `toolexec` phase. This would involve removing absolute file paths from `matched.json` at setup time -- setup would only record which packages/targets to instrument based on `target` + `version` selectors.

Eventually, we can make `setup` an internal step of `otelc go build`. Since `otelc go build` (auto-pinning) already does some matching based on target + version, the internal setup phase would do that again and generate a **filtered list of rules**. The actual file-level resolution would happen in `toolexec`: `otelc` would check if any of the rules in the filtered list match the target currently being built (by inspecting the compile commands provided to toolexec) and apply those rules to the exact files being compiled.

This idea was originally proposed by @txabman42 in [this review comment](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/495#pullrequestreview-4323561021) on [#495](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/495).

The goal of this issue is to explore whether this approach is worth pursuing, document pros/cons of both designs, and reach a consensus before any implementation work begins.

### Option A: Current design (setup-time matching)

Setup uses `go build -a -x -n` to get the full build plan, resolves file paths, and stores them in `matched.json`. `toolexec` reads from it without any additional resolution.

**Pros**
- Already implemented
- File resolution happens once and is cached

**Cons**
- Matched file paths can go stale if `go.mod` changes during setup (the bug in #494 - *now mitigated by #584*)
- Once isSetup is implemented, the all-or-nothing cache becomes a problem: if any instrumented dependency changes, isSetup returns false and the entire setup re-runs, re-matching all packages even if only one dependency was bumped.

### Option B: toolexec-time matching

[...]

PR #659, opened 2026-07-07
feat(tool): implement isSetup check in setup phase

### LINKED ISSUE #658: chore: Implement `isSetup` check in the setup phase
## Description
In `tool/internal/setup/setup.go`, the function `isSetup()` is a stub that always returns `false`:

```go
// This function can be used to check if the setup has been completed.
func isSetup() bool {
	// TODO: Implement Task
	return false
}
```

Because it always returns `false`, the compilation/setup pipeline executes the entire setup process (extracting embedded modules, matching dependencies to hook rules, generating runtime code, backing up module files, etc.) even when the setup was already successfully completed in a prior invocation.

## Proposed Solution
We can detect whether the setup has been completed by checking if a valid `state.json` file is present. In the tool's architecture, `StateManager.Commit()` is called at the end of a successful `Setup()` invocation to write the `state.json` metadata file to the build temp directory.

By invoking the existing `LoadStateManager()` helper, we can check if the state file is present and readable:
1. If `LoadStateManager()` returns a non-nil `StateManager` and no error, setup has been completed (`isSetup` should return `true`).
2. If it returns a nil `StateManager` or an error, it indicates setup is not yet complete or the state is invalid (`isSetup` should return `false`).

### Proposed Change in `tool/internal/setup/setup.go`
```go
func isSetup() bool {
	stateManager, err := LoadStateManager()
	return err == nil && stateManager != nil
}
```

## Verification Plan
1. Add a unit test `TestIsSetup` in `tool/internal/setup/setup_test.go`.
2. The test should:
   - Verify `isSetup()` returns `false` in a clean environment.
   - Write a mock `state.json` file to the temporary build directory.
   - Verify `isSetup()` returns `true` when the file is present.
   - Remove the state file and verify `isSetup()` returns `false` again.

SatyamPandey-07:
## Description

This PR implements the `isSetup` check in the compilation/setup pipeline. 

Key changes:
- Swapped the `isSetup()` stub in `tool/internal/setup/setup.go` for a functional check using `LoadStateManager()`.
- Added a unit test `TestIsSetup` in `tool/internal/setup/setup_test.go` to cover clean, valid, invalid, and deleted state scenarios.

## Motivation

Previously, `isSetup()` was an unimplemented stub returning `false`. Because of this, consecutive invocations of `otelc setup` or compile-time setup runs executed the entire setup lifecycle (extracting modules, matching rules, generating package runtime code, etc.) repeatedly. 

With this change, the pipeline checks for a successfully committed `state.json` file on disk via `LoadStateManager()`, avoiding redundant execution steps and disk I/O.

Fixes #658

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -46,8 +46,8 @@ func (sp *SetupPhase) keepForDebug(srcPath string) {
 
 // This function can be used to check if the setup has been completed.
 func isSetup() bool {
-	// TODO: Implement Task
-	return false
+	stateManager, err := LoadStateManager()
+	return err == nil && stateManager != nil
 }
 
 // flagsWithPathValues contains flags that accept a value from "go build" command.
--- a/tool/internal/setup/setup_test.go
+++ b/tool/internal/setup/setup_test.go
@@ -493,3 +493,34 @@ func TestExtractBuildFlags(t *testing.T) {
 		})
 	}
 }
+
+func TestIsSetup(t *testing.T) {
+	tmp := t.TempDir()
+	t.Chdir(tmp)
+
+	// In a clean directory, isSetup should return false
+	require.False(t, isSetup())
+
+	// Create the build temp directory first so Commit doesn't fail
+	err := os.MkdirAll(util.GetBuildTempDir(), 0o755)
+	require.NoError(t, err)
+
+	// Create stateManager and commit to create the state file
+	stateManager := NewStateManager()
+	err = stateManager.Commit()
+	require.NoError(t, err)
+
+	// Since state.json now exists, isSetup should return true
+	require.True(t, isSetup())
+
+	// If we write invalid JSON to state.json, isSetup should return false
+	stateFile := util.GetBuildTemp(stateFileName)
+	err = os.WriteFile(stateFile, []byte("{invalid json"), 0o644)
+	require.NoError(t, err)
+	require.False(t, isSetup())
+
+	// Remove the state file, isSetup should return false
+	err = os.Remove(stateFile)
+	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


SatyamPandey-07:
@kakkoyun @darccio review the pr when u are free

kakkoyun:
Leaving it to @amazingakai since he recently added the `StateManager`

SatyamPandey-07:
@amazingakai review the pr when u are free

amazingakai (changes requested):
I'm not convinced `LoadStateManager()` is the right signal for `isSetup`. `LoadStateManager()` only reflects persisted state. For `otelc go build`, the state manager can exist only in memory and never be committed to `state.json`, so this check can report "not setup" even though setup has already completed for the current invocation.

#513 used `matched.json` as the signal, but that also had drawbacks (as @txabman42 pointed out in that review).

I previously suggested another approach in #161 (https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/161#issuecomment-4389948535). However, we'd still need to detect when setup is no longer valid and rerun it, for example when:
- the dependency graph (`go.mod`) changes
- the user's source code changes
- the build environment changes (flags, env, etc.)
- the set of modules changes (e.g. in a `go.work` workspace)

Keeping track of all of those inputs correctly is non-trivial, so I'm not sure a simple `isSetup` check can be made correct.

I'm leaning towards removing the setup phase entirely (#552, by performing matching during the instrument phase), but that's still mostly theory and would require a much larger refactor (removing `otelc.runtime.go` files, moving matching, and more), so it's probably outside the scope of this PR.

darccio:
Closing in favour of #854

PR #660, opened 2026-07-07
chore(integration/kafka): split instrumentation into producer and consumer submodules

### LINKED ISSUE #635: chore(integration/kafka): improve the quality assurance
LGTM once the other review comments are addressed.

A couple of non-blocking comments from my side:

1. Should we also have an e2e test for this?
2. grpc and net/http instrumentations use `client/` and `server/` subdirectories. Could we follow the same convention here and use `producer/` and `consumer/`? It's not super important right now, but with `otel.instrumentation.go` support users could choose to enable only one of them.

Also, once this is merged, we'll need to add it to the OpenTelemetry registry (same for OpenAI instrumentation), similar to https://github.com/open-telemetry/opentelemetry.io/pull/10609. Would you like to take care of that, or should I?

_Originally posted by @amazingakai in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/302#pullrequestreview-4619075652_

macayu17:
## Description

The `segmentio/kafka-go` instrumentation lived in a single module that hooked both `(*Writer).WriteMessages` and `(*Reader).ReadMessage`. This splits it into two submodules, following the same `client/` + `server/` layout that `grpc` and `net/http` already use:

- `producer/` holds the write-side hooks (`BeforeWriteMessages` / `AfterWriteMessages`), with its own `go.mod` and `otelc.yaml`.
- `consumer/` holds the read-side hooks (`BeforeReadMessage` / `AfterReadMessage`) and the public `ExtractContext` helper, again with its own `go.mod` and `otelc.yaml`.
- The shared `semconv/` package stays under the root instrumentation module, the same way `grpc/semconv` and `net/http/semconv` do (the old `kafka-go/go.mod` is removed so `semconv` is covered by the root module).

The existing unit tests moved with their hooks into the new packages, so producer and consumer are each covered on their own.

## Motivation

Right now a user pulls in the producer and consumer instrumentation together because they share one module. With the `otel.instrumentation.go` selection mechanism, splitting them lets a service enable only the side it needs, for example a write-only publisher that never consumes. This matches how `grpc` and `net/http` already let you pick `client` or `server` independently, so kafka now follows the same convention.

This came out of the review discussion on #302.

Fixes #635

---

## Checklist

- [x] Tests pass: `make test`


NameHaibinZhang:
LGTM! Clean split that follows the established grpc/net/http convention. The semconv placement, module boundaries, replace paths, and otelc.yaml configs are all correct.

One optional follow-up: consider adding a `TestHeaderCarrier_SetGetKeys` in the consumer test file for parity with the producer tests — though the carrier is already exercised indirectly through the propagation tests.

NameHaibinZhang (approved):

Issue #665, opened 2026-07-10
bug: go get -tool workflow fails or builds an unusable otelc in v0.5.0


lrwh:
## Description

Following the documented Go tool dependency installation workflow does not produce a working `otelc` with v0.5.0.

I found two separate problems while trying to instrument a minimal Go `net/http` application.

### 1. Documented vanity import path fails

The documented workflow uses:

```sh
go get -tool go.opentelemetry.io/otelc/tool/cmd/otelc@latest
```

With v0.5.0 this fails with:

```text
go: module go.opentelemetry.io/otelc@latest found (v0.5.0), but does not contain package go.opentelemetry.io/otelc/tool/cmd/otelc
```

Trying the apparent package path without the final `/otelc` also fails because the module declares a different module path:

```sh
go get -tool go.opentelemetry.io/otelc/tool/cmd@latest
```

```text
go: go.opentelemetry.io/otelc@v0.5.0 (matching go.opentelemetry.io/otelc/tool/cmd@latest) requires go.opentelemetry.io/otelc@v0.5.0: parsing go.mod:
        module declares its path as: github.com/open-telemetry/opentelemetry-go-compile-instrumentation
                but was required as: go.opentelemetry.io/otelc
```

### 2. GitHub module path installs, but the built tool is unusable

Using the GitHub module path and the actual CLI package path succeeds:

```sh
go get -tool github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/cmd@v0.5.0
go tool github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tool/cmd go build -o bin/http-otelc-demo .
```

But running the tool fails at setup time:

```text
Error:
[0] failed to read file: open otelc-pkg.gz: file does not exist
[1] extracting embedded instrumentation pkg

Stack:
[0] .../tool/data/export.go:18 tool/data.ReadEmbedFile
[1] .../tool/internal/setup/extract.go:133 tool/internal/setup.(*SetupPhase).extract
[2] .../tool/internal/setup/setup.go:247 tool/internal/setup.Setup
[3] .../tool/internal/setup/setup.go:508 tool/internal/setup.GoBuild
```

## Expected behavior

The documented `go get -tool` / `go tool` workflow should install and run a working `otelc`, or the docs should clearly state that users must build from source with `make build` for v0.5.0.

## Actual behavior

* The documented vanity import path does not resolve to the documented CLI package.
[...]

Issue #585, opened 2026-06-17
feat(tool): decouple built-in instrumentations from `otelc-bundle.tgz`


amazingakai:
## Problem

Today, all built-in instrumentations are bundled into the `otelc` binary through `otelc-bundle.tgz`. As a result, instrumentation versions are tightly coupled to the version of the tool itself.

This makes it difficult for users to independently manage instrumentation versions. For example, using an older version of a particular instrumentation would require using an older version of `otelc`, even when the tool itself is otherwise compatible.

## Proposal

With `otel.instrumentation.go` support (#584), built-in instrumentations can be distributed as normal instrumentation packages instead of being embedded into the tool.

Users would then manage instrumentation versions through `go.mod`, just like any other Go dependency. `otelc pin` would generate imports to those instrumentation packages rather than relying on embedded rule definitions.

## Benefits

* Instrumentations can evolve independently of the tool.
* Users can pin instrumentation versions through `go.mod`.
* Reduces the amount of embedded data shipped with `otelc`.
* Unifies built-in and third-party instrumentation workflows.

## Related

#849

kakkoyun:
@amazingakai very well put and great idea. We should note that minimum required version of go stated in `go.mod` files will be binding still. There is still a change where we need to upgrade `otelc` to support newer versions of the integrations. To be able to broaden the supported instrumentation we can be conservative on bumping the version of `otelc` itself.

txabman42:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/655/changes#r3537497576

>  I'm leaning towards #585 as the solution for this rather than adding additional sync/extract logic to the update path.
>
> Updated the documentation with a small note clarifying that otelc pin-generated otel.instrumentation.go files are not intended to be committed yet, and that this workflow is still under development pending #585.

txabman42:
Reminder, let's revert this https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/666 once this issue is done

We would need to test the new workflow, but go get -tool install + defining rules repo

darccio:
I did a quick analysis and I scoped the required work to these six steps:

1. Add a small JSON manifest (`modulePath`, `target`, `versionRange`) generated from the real instrumentation/ tree: replaces the embedded bundle for discovery, unwired at first.
2. Wire `otelc pin` to load it instead of extracting the bundle; delete bundle/embed machinery + its CI.
[...]

PR #666, opened 2026-07-10
docs: mark go get -tool install as not-yet-supported

txabman42:
The documented `go get -tool go.opentelemetry.io/otelc/tool/cmd/otelc` workflow cannot produce a working binary: the instrumentation bundle (`otelc-bundle.tgz`) is generated by `make build` and gitignored, so it is absent from the published module and the embed fails / the binary cannot instrument. Flag `make build`/`make install` as the only supported path until the packages are decoupled from the executable (#585).

Refs #665


amazingakai (approved):

amazingakai:
@txabman42 It seems CLA bot isn't happy about the `Co-Authored By` part.

txabman42:
@amazingakai fixed, testing new AI config...

PR #670, opened 2026-07-10
fix(state): make StateManager crash-safe with incremental atomic persistence

amazingakai:
Based on: https://github.com/kakkoyun/opentelemetry-go-compile-instrumentation/pull/10

## Summary

Make StateManager crash-safe with incremental atomic persistence

Previously the state manifest was written only once at the end of a successful build via an explicit `Commit()`. A crash (SIGKILL, OOM, CI cancellation) before that point left no trace on disk, making it impossible for a subsequent `otelc cleanup` to restore the tree.

## Changes

- `Track()` now calls `Commit()` before returning, persisting the manifest incrementally after every tracked file. No explicit `Commit()` needed from callers.
- `Commit()` uses `WriteFileAtomic` (temp file + rename) so a crash mid-write never leaves a truncated manifest.
- Add `Discard()` to remove the consumed manifest and snapshots after a successful `Revert`.
- A failed revert in `Cleanup()` no longer deletes snapshots, they are preserved as the only recovery path.
- Increase test timeout from 10m to 15m for latestlibbuild, latestlibrun, and versionmatrix tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -548,13 +548,13 @@ test-integration: build build-demo
 test-latestlibbuild: build ## Run LatestLibBuild tests
 	@echo "Running LatestLibBuild tests..."
 	set -euo pipefail
-	go -C "test" test -json -v -shuffle=on -timeout=10m -count=1 -tags latestlibbuild ./latestlibbuild/... 2>&1 | tee ./gotest-latestlibbuild.log
+	go -C "test" test -json -v -shuffle=on -timeout=15m -count=1 -tags latestlibbuild ./latestlibbuild/... 2>&1 | tee ./gotest-latestlibbuild.log
 
 .ONESHELL:
 test-latestlibrun: build ## Run LatestLibRun tests (bump apps to @latest then run integration suite)
 	@echo "Bumping test apps to @latest..."
 	set -euo pipefail
-	go -C "test" test -json -v -shuffle=on -timeout=10m -count=1 -tags latestlibrun ./latestlibrun/... 2>&1 | tee ./gotest-latestlibrun.log
+	go -C "test" test -json -v -shuffle=on -timeout=15m -count=1 -tags latestlibrun ./latestlibrun/... 2>&1 | tee ./gotest-latestlibrun.log
 	$(MAKE) tidy/test-apps
 	@echo "Syncing test module with bumped apps..."
 	go -C "test" mod tidy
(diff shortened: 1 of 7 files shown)


tool/internal/setup/cleanup.go
```diff
@@ -28,13 +29,26 @@ func Cleanup(ctx context.Context, cleanAll bool) error {
...
 
+	reverted := true
 	if stateManager != nil {
 		if err := stateManager.Revert(); err != nil {
+			reverted = false
 			logger.WarnContext(ctx, "failed to revert state", "error", err)
+		} else if discardErr := stateManager.Discard(); discardErr != nil {
+			logger.WarnContext(ctx, "failed to discard consumed state", "error", discardErr)
 		}
 	}
```
kakkoyun, line 45: Could we rewrite this in more readable way? It's hard to follow.

tool/internal/setup/setup.go
```diff
@@ -365,18 +365,14 @@ func Setup(ctx context.Context, cmd *cli.Command) error {
 		return ex.Wrapf(findModErr, "finding module directories for build packages")
 	}
 
-	// Track generated & modified files with state manager
+	// Track generated & modified files with state manager. Track persists the
```
kakkoyun: I don't see any explicit commit here. Maybe the comment is misleading?

kakkoyun (approved):
LGTM

Makefile
```diff
@@ -548,13 +548,13 @@ test-integration: build build-demo
 test-latestlibbuild: build ## Run LatestLibBuild tests
 	@echo "Running LatestLibBuild tests..."
 	set -euo pipefail
-	go -C "test" test -json -v -shuffle=on -timeout=10m -count=1 -tags latestlibbuild ./latestlibbuild/... 2>&1 | tee ./gotest-latestlibbuild.log
```
txabman42, line 551: Why this?

darccio (approved):

txabman42 (changes requested):
I see the PR ok ❤️ 

Requested changes as I want to understand well the test timeout increase

Makefile
amazingakai: I had some CI failures in latestlibuild & integration tests due to timeout. Since our integration test suite has grown, I think this is reasonable. Not directly related to this PR, I can move this to a separate PR if you prefer.

txabman42: Yes, please, lets move it to 15 for now, but create an issue to track this. I wouldn't like to have test exec times increase every time we add extra tests. We need to keep those numbers under control, just to see if we can refactor them so exec time dont increase so much.

txabman42 (approved):

Makefile
txabman42: @amazingakai I was ok with 15min as commented. But if they are not too noisy for now, 👌 keeping them as 10

amazingakai: @txabman42 I moved it to a separate PR #675
Let's monitor the CI for a bit before merging, maybe we won't need it.

PR #672, opened 2026-07-10
fix(setup): serialize concurrent otelc invocations with a build lock

kakkoyun:
## Description

Adds a build lock so concurrent otelc invocations against the same module can no longer corrupt it. Every mutating entry point in the setup package — `Setup`, `GoBuild`, `Cleanup`, and `Pin` — acquires an OS advisory file lock (via `gofrs/flock`) around its whole body. A second invocation logs that it is waiting and blocks until the holder finishes or the wait is canceled with Ctrl-C. Because the lock is advisory, the OS releases it when the holding process dies, so a crashed build can never wedge the module.

Design points worth knowing before reviewing:

- The lock lives in the setup package's entry points, not in the CLI layer, and nested chains reuse the outer lock through a context marker (same idiom as `ContextWithStateManager`). `GoBuild -> Setup -> AutoPin -> Pin` and GoBuild's deferred `Cleanup` therefore never re-acquire, and a verb added later is serialized by construction instead of by remembering boilerplate. `toolexec` runs as a child of the locked build and never calls these entry points, so it stays outside the lock and Go's internal build parallelism is untouched.
- The lock file is a sibling of the build dir (`.otelc-build.lock`), not inside it: `cleanup` deletes `.otelc-build/` while holding the lock, and Windows refuses to delete a file the process itself has open. The Windows CI leg caught this; POSIX unlink had masked it.
- The file exists only while an invocation runs. Release removes it in an order that is safe on POSIX and Windows, every acquisition attempt reopens the path and verifies the locked handle still matches it, and unexpected stat failures surface as errors instead of silent retries. A leftover file from a killed process holds no lock, never blocks anyone, and gets reused and cleaned up by the next run — the waiting message spells out that only a live invocation can cause waiting.
- The work dir is canonicalized through `EvalSymlinks` so `/tmp/x` and `/private/tmp/x` (macOS) contend on one lock, and a work dir that does not exist degrades to a no-op acquisition so `otelc cleanup` on a never-set-up tree stays a true no-op.
- `GoBuild` validates its arguments before locking; a bad command line fails immediately instead of waiting behind a running build.
- `otelc pin` landed while this branch was in progress; it rewrites go.mod/go.sum through the same window, so it is one of the locked entry points.
- `test/go.mod`/`test/go.sum` are retidied: `gofrs/flock` raises the module graph's `golang.org/x/sys` floor to v0.46.0.

`gofrs/flock` (BSD-3, no transitive dependencies beyond `x/sys`) was picked over a hand-rolled `O_EXCL` pidfile because stale-lock handling across the three supported OSes is exactly the part that is easy to get wrong.

Known limitation, documented in code: the lock is scoped to one module's work dir, so in a `go.work` workspace two invocations for different member modules still share `go.work.sum` unserialized. That needs a workspace-scoped lock and is left for a follow-up.

## Motivation

Nothing serializes concurrent otelc runs in one module today. When a second build starts while the first is mid-setup, it snapshots the already-mutated go.mod as its "original"; whichever run restores last bakes the replace directives into go.mod for good. The two runs also race on `.otelc-build/` itself, deleting each other's import-tracking files and clobbering `matched.json`.

This is easy to hit without doing anything unusual: an IDE save-hook firing while a terminal build runs, or two CI jobs sharing a checkout. Reproduced before the fix with two parallel `otelc go build` runs: one build fails, the other "succeeds", and go.mod keeps all 8 replace directives. With the lock, both runs exit 0 and go.mod comes back pristine.

## Verification

- Unit tests: mutual exclusion between concurrent acquisitions, cancellation while waiting, lock-file removal on release, a pre-existing stale file not blocking acquisition, nested re-entrancy through the context marker, and the missing-work-dir no-op.
- `go test ./tool/...` green; golangci-lint clean; the repo-wide `go mod tidy` consistency check passes.
- End-to-end: the concurrent-build repro flips from corrupted go.mod to both builds exiting 0 with go.mod restored.
- Full CI matrix (linux/macos/windows, amd64/arm64) green on my fork first: kakkoyun#11.

---

## Checklist

- [x] Tests pass: `make test`
- [ ] Documentation updated (if applicable) — behavior is documented in code; no user-facing doc pages affected
- [ ] OpenTelemetry Registry updated (if applicable) — n/a, not a new instrumentation


tool/internal/setup/lock.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	logger.InfoContext(
+		ctx,
+		"another otelc invocation is running and holds the build lock; waiting for it to finish",
+		"path",
+		path,
+		"note",
+		"a leftover lock file from a crashed build cannot cause waiting; the OS releases its lock when the process dies",
+	)
```
amazingakai: Let's use `fmt.Fprintf` here with `os.Stderr`, this log goes to `debug.log` otherwise and user won't know.

tool/internal/setup/setup.go
```diff
@@ -613,9 +616,25 @@ func GoBuild(ctx context.Context, cmd *cli.Command) error {
 			cmd.Args().First())
 	}
 
+	ctx = ContextWithStateManager(ctx, NewStateManager())
```
amazingakai: Maybe move this to `goBuildLocked`?

tool/internal/setup/lock.go
```diff
@@ -0,0 +1,267 @@
...
+// Known limitation: the lock is scoped to one module's work dir. In a
+// go.work workspace, invocations for different member modules hold
+// different locks yet share the workspace-level go.work.sum; serializing
+// that file needs a workspace-scoped lock and is out of scope here.
+func buildLockPath() string {
+	workDir := util.GetOtelcWorkDir()
+	if resolved, err := filepath.EvalSymlinks(workDir); err == nil {
+		workDir = resolved
+	}
+	return filepath.Join(workDir, util.BuildTempDir) + ".lock"
```
amazingakai: Let's make this a constant and maybe place it in the same file as `BuildTempDir` constant.

amazingakai (changes requested):
Clean ❤️ 

Just some small changes

amazingakai (approved):

Makefile
```diff
@@ -542,13 +542,13 @@ test-integration: go-protobuf-plugins ## Run integration tests
 test-integration: build build-demo
 	@echo "Running integration tests..."
 	set -euo pipefail
-	go -C "test" test -json -v -shuffle=on -timeout=15m -count=1 -tags integration ./integration/... 2>&1 | tee ./gotest-integration.log
+	go -C "test" test -json -v -shuffle=on -timeout=20m -count=1 -tags integration ./integration/... 2>&1 | tee ./gotest-integration.log
```
txabman42, line 548: We have this issue https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/676, lets dont start increasing the test timeouts without understanding why they are starting to get so slow.

tool/internal/setup/lock.go
```diff
@@ -0,0 +1,264 @@
...
+	"go.opentelemetry.io/otelc/tool/ex"
+	"go.opentelemetry.io/otelc/tool/util"
+)
+
+// buildLockRetryInterval is how often a waiting invocation re-attempts the
+// lock. The wait itself is unbounded: it ends when the holder finishes or
+// the caller cancels (Ctrl-C), and a log line makes the waiting visible.
+const buildLockRetryInterval = 200 * time.Millisecond
+
+// buildLockPath returns the lock file path. The lock deliberately lives
```
txabman42: I suggest adding shorter comments, straight to the key points. Same for other functions.

tool/internal/setup/setup.go
```diff
@@ -613,9 +616,25 @@ func GoBuild(ctx context.Context, cmd *cli.Command) error {
...
 	}
 
+	// Serialize with other otelc invocations in this module before touching
+	// any shared state (tracking files, go.mod, .otelc-build contents).
+	return withBuildLock(ctx, func(ctx context.Context) error {
+		return goBuildLocked(ctx, cmd)
+	})
+}
+
+func goBuildLocked(ctx context.Context, cmd *cli.Command) error {
```
txabman42: `goBuildLocked` doesn't know about locks, no? This name is confusing.

The functions shouldn't know its wrapped by `withBuildLock`

tool/internal/setup/pin.go
```diff
@@ -552,6 +552,21 @@ type PinResult struct {
 // (not just those with instrumentation). The PinResult can be used by later stages (setup) instead of
 // re-parsing compile commands again.
 func Pin(ctx context.Context, opts PinOptions) (*PinResult, error) {
+	// Pin rewrites go.mod/go.sum (ensureOtelcRequire, syncDeps, go mod tidy)
```
txabman42: Let's remove this kind of comment, they dont add too much value, easier to read the code than adding comments everywhere

txabman42 (changes requested):
PR is solid, just some changes

Makefile
kakkoyun: I bumped it here just to make CI green. We can revert the changes when the fixes for #676 lands.

txabman42 (approved):
LGTM ❤️

Issue #671, opened 2026-07-10
feat: otelc toolexec drop-in via GOFLAGS fails (missing matched.json; work-dir permission denied in module cache)


kakkoyun:
## What happened

We want `otelc` to work as an Orchestrion-style drop-in, where users point the
Go toolchain at `otelc` through `GOFLAGS` and run a normal `go build`:

```sh
export GOFLAGS="${GOFLAGS} '-toolexec=otelc toolexec'"
go build ./...
```

Today this path fails, while `otelc go build` succeeds on the same module.

## Reproduction

Environment: `otelc` installed via `make install` from `main`
(`v0.0.0-20260710101053-1324a37eb0eb`), Go 1.26.5, macOS. Minimal module with a
`net/http` `main`.

**Case 1 — no prep step:**

```sh
export GOFLAGS="${GOFLAGS} '-toolexec=otelc toolexec'"
go build .
```

```
failed to read file ./.otelc-build/matched.json: open ./.otelc-build/matched.json: no such file or directory
  tool/internal/instrument/match.go:21  (*InstrumentPhase).load
  tool/internal/instrument/toolexec.go:115  interceptCompile
  tool/internal/instrument/toolexec.go:362  Toolexec
```

The compile interceptor expects `.otelc-build/matched.json`, which is only
produced by the setup phase that `otelc go build` runs.

**Case 2 — after `otelc setup`:**

`otelc setup` does generate `.otelc-build/matched.json`, `otel.instrumentation.go`,
and `otelc.runtime.go`. The subsequent drop-in build then fails differently:

```
failed to create work directory ".../go/pkg/mod/golang.org/x/sys@v0.45.0/unix/.otelc-build": mkdir ...: permission denied
  tool/cmd/otelc/main.go:129  initLogger
```

When the toolchain invokes `otelc toolexec` for a dependency package, `otelc`
tries to create its `.otelc-build` work directory inside the (read-only) module
cache directory of that package, rather than under the build root.

## Expected

The `-toolexec=otelc toolexec` drop-in (optionally after an `otelc setup` /
`otelc pin` prep step) should build successfully and produce the same
instrumented binary as `otelc go build`, so `otelc` is a true drop-in for the
Orchestrion `GOFLAGS` workflow.

## Control

`otelc go build -o app .` succeeds on the same module.

## Why now

We plan to document this drop-in usage in the v1 announcement blog post, so it
would be good to have it working before the v1 release.

[...]

PR #674, opened 2026-07-10
fix: toolexec goflags dropin

txabman42:
Contributing to https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/671

This solves `Case 2`, we'll need another PRs for fixing `Case 1`.

Main changes:
1. **Work-dir discovery (DiscoverWorkDir)**: walk up from cwd to .otelc-build, bounded by go.mod, instead of trusting cwd (which may be the read-only module cache for asm/cgo).
2. **GOFLAGS strip at startup**: remove -toolexec so spawned go children don't recurse.
3. **-V=full rewrite**: append an otelc@<version>/<rulesHash> marker to the tool ID so instrumented and plain artifacts never share build-cache entries.
4. **Nested version-only toolexec (OTELC_NESTED_TOOLEXEC)**: keep archives resolved during instrumentation aligned with the outer build's cache keys.
5. **Actionable error when matched.json is missing + docs**.

### DIFF
--- a/tool/internal/instrument/toolexec.go
+++ b/tool/internal/instrument/toolexec.go
@@ -355,15 +355,20 @@ func interceptLink(ctx context.Context, args []string) ([]string, error) {
 // toolVersionLine appends an otelc marker to a `tool -V=full` line so the tool
 // ID (and every build cache key derived from it) differs from a plain build.
 // The rules hash is included so editing the config invalidates cached
-// artifacts too. A devel toolchain's line ends in `buildID=...` and go keys
-// the cache on that field alone, so the marker goes just before it.
+// artifacts too. For devel toolchains, Go uses only the content ID in the
+// trailing `buildID=...` field, so append the marker to that field's value.
 func toolVersionLine(line, rulesHash string) string {
+	hasBuildID := strings.Contains(line, " buildID=")
 	marker := "otelc@" + util.Version
 	if rulesHash != "" {
-		marker += "/" + rulesHash
+		if hasBuildID {
+			marker += "+" + rulesHash
+		} else {
+			marker += "/" + rulesHash
+		}
 	}
-	if idx := strings.LastIndex(line, " buildID="); idx >= 0 {
-		return line[:idx] + " " + marker + line[idx:]
+	if hasBuildID {
+		return line + "+" + marker
 	}
 	return line + " " + marker
 }
--- a/tool/internal/instrument/toolexec.go
+++ b/tool/internal/instrument/toolexec.go
@@ -473,7 +473,11 @@ func EnableNestedToolexec() error {
 	if err != nil {
 		return ex.Wrapf(err, "resolving otelc executable path")
 	}
-	goflags := strings.TrimSpace(os.Getenv("GOFLAGS") + fmt.Sprintf(" '-toolexec=%s toolexec'", execPath))
+	toolexecFlag, err := util.QuoteGoflagsToken(fmt.Sprintf("-toolexec=%s toolexec", execPath))
+	if err != nil {
+		return ex.Wrapf(err, "quoting nested toolexec GOFLAGS entry")
+	}
+	goflags := strings.TrimSpace(os.Getenv("GOFLAGS") + " " + toolexecFlag)
 	if err = os.Setenv("GOFLAGS", goflags); err != nil {
 		return ex.Wrapf(err, "setting GOFLAGS for nested go commands")
 	}
--- a/tool/internal/instrument/toolexec.go
+++ b/tool/internal/instrument/toolexec.go
@@ -5,9 +5,13 @@ package instrument
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"log/slog"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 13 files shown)


txabman42:
@kakkoyun could you try to reproduce the same case from https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/671?

kakkoyun:
I validated the direct GOFLAGS='-toolexec=otelc toolexec' path against #671 and added end-to-end regression coverage in ea71812 (https://github.com/txabman42/opentelemetry-go-compile-instrumentation/commit/ea7181232a5efbdbeb30807eff8456d1224fad64).

tool/util/toolexec_dropin_test.go
```diff
@@ -0,0 +1,118 @@
...
+		{
+			name:     "bare toolexec flag",
+			goflags:  "-toolexec=otelc",
+			expected: "",
+		},
+		{
+			name:     "single-quoted toolexec flag with space",
+			goflags:  "'-toolexec=otelc toolexec'",
+			expected: "",
+		},
```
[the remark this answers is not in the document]
txabman42: Not valid, skipping...

tool/util/go.go
```diff
@@ -92,6 +92,62 @@ func IsCgoCommand(line string) bool {
...
+			unquoted[len(unquoted)-1] == unquoted[0] {
+			unquoted = unquoted[1 : len(unquoted)-1]
+		}
+		if unquoted == "-toolexec" || strings.HasPrefix(unquoted, "-toolexec=") {
+			continue
+		}
+		kept = append(kept, token)
+	}
+	return strings.Join(kept, " ")
+}
```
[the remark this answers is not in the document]
txabman42: `-toolexec ...` not valid, skipping...

kakkoyun (approved):

tool/util/go.go
```diff
@@ -92,6 +92,62 @@ func IsCgoCommand(line string) bool {
...
+// splitGoflags splits a GOFLAGS value into tokens like the go command does
+// (https://cs.opensource.google/go/go/+/master:src/cmd/internal/quoted/quoted.go)
+//
+// space-separated, but a token starting with a quote runs to the matching close quote.
+// Quotes are kept so tokens re-join verbatim.
+func splitGoflags(goflags string) []string {
+	var tokens []string
+	i := 0
+	for i < len(goflags) {
+		for i < len(goflags) && (goflags[i] == ' ' || goflags[i] == '\t') {
```
kakkoyun: `splitGoflags` only separates on space and tab, but Go parses GOFLAGS with `cmd/internal/quoted.Split`, which also treats `\n` and `\r` as separators. A newline before the toolexec entry (`-race\n'-toolexec=otelc toolexec'`) therefore stays in one token, so the `-toolexec` prefix check never matches and the flag survives the strip. The spawned `go` children then inherit `-toolexec` and recurse, which is the exact thing this strip prevents. It's an unlikely input, but the doc comment claims parity with `quoted.Split`, so splitting on `\n` and `\r` here would match it.

tool/internal/instrument/toolexec.go
```diff
@@ -390,3 +448,37 @@ func Toolexec(ctx context.Context, args []string) error {
...
+// commands this process spawns (e.g. `go list -export`) run through a
+// version-only otelc toolexec and share this build's cache keys. Any existing
+// -toolexec was stripped at startup. Must only be called from the real otelc
+// binary, since os.Executable is what nested go commands will run.
+func EnableNestedToolexec() error {
+	execPath, err := os.Executable()
+	if err != nil {
+		return ex.Wrapf(err, "resolving otelc executable path")
+	}
+	goflags := strings.TrimSpace(os.Getenv("GOFLAGS") + fmt.Sprintf(" '-toolexec=%s toolexec'", execPath))
```
kakkoyun: This wraps the executable path in single quotes by hand, so a path containing a single quote breaks the quoting. For an otelc binary under something like `/home/kakkoyun/go/bin/otelc`, the wrapped value becomes `'-toolexec=/home/kakkoyun/... toolexec'`, which Go's GOFLAGS parser rejects as an unterminated quote, and every nested `go` command otelc spawns then fails. The single quotes cover spaces but not apostrophes. `quoted.Join` switches quote style when the value already contains the quote character, so it would handle this safely.

tool/internal/instrument/toolexec_exec_test.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// versionMarkerPattern documents the exact tool-ID shape the go build cache
+// keys on, guarding against accidental format drift.
+var versionMarkerPattern = regexp.MustCompile(`otelc@[^\s/]+(/[0-9a-f]{16})?`)
+
+func TestToolVersionLineMatchesCachePattern(t *testing.T) {
+	assert.Regexp(t, versionMarkerPattern, toolVersionLine("compile version go1.26.5", ""))
+	assert.Regexp(t, versionMarkerPattern, toolVersionLine("compile version go1.26.5", "0123456789abcdef"))
+}
```
amazingakai: Move to `tool_version_test.go`?

tool/util/go.go
txabman42: 205db4dd927f3884c561d870aaa73fa1941dbd01

tool/internal/instrument/toolexec.go
txabman42: 205db4dd927f3884c561d870aaa73fa1941dbd01

amazingakai (approved):
LGTM ❤️

tool/internal/instrument/toolexec_exec_test.go
txabman42: Moved

PR #677, opened 2026-07-11
chore: commit otelc-bundle.tgz to git and add bundle verification CI

amazingakai:
## Description

Commit `tool/data/otelc-bundle.tgz` directly to the repository and remove it from `.gitignore`. Add a verification script and GitHub Actions workflow to ensure the committed bundle stays in sync with the `pkg/` and `instrumentation/` source trees on every pull request and push to main.

Revert the docs change that marked `go get -tool` as not-yet-supported, since it now works.

## Motivation

The `//go:embed` directive in `tool/data/export.go` requires `otelc-bundle.tgz`
to exist at build time. Previously this file was gitignored and only available
after running `make package`, which meant `go install` and `go get -tool` failed
for anyone cloning the repository without first building the tool. Checking the
bundle into git removes that friction and makes the standard Go toolchain work
out of the box.

The new CI workflow prevents the bundle from drifting out of sync with the source
trees by comparing their contents on every PR that touches the relevant paths.

---

## Checklist

- [ ] Tests pass: `make test`


kakkoyun (approved):

.github/scripts/verify-bundle.sh
```diff
@@ -0,0 +1,110 @@
...
+        status=1
+    fi
+
+    # Git only preserves the executable bit for regular files.
+    while IFS= read -r -d '' file; do
+        rel="${file#$src/}"
+        other="$dst/$rel"
+
+        if [[ -x "$file" && ! -x "$other" ]] || [[ ! -x "$file" && -x "$other" ]]; then
+            echo "executable bit mismatch: $rel"
```
txabman42: We need to throw in these errors a helper for how to solve the problem, e.g. run `make package`

.github/scripts/verify-bundle.sh
```diff
@@ -0,0 +1,110 @@
...
+    if ! diff --recursive --brief "$src" "$dst"; then
+        status=1
+    fi
+
+    # Git only preserves the executable bit for regular files.
+    while IFS= read -r -d '' file; do
+        rel="${file#$src/}"
+        other="$dst/$rel"
+
+        if [[ -x "$file" && ! -x "$other" ]] || [[ ! -x "$file" && -x "$other" ]]; then
```
txabman42: I've done:

```bash
rm tool/data/otelc-bundle.tgz
make package
```

and the resulting `otelc-bundle.tgz` differs from the committed one.

I'd expected it to be the same

.github/scripts/verify-bundle.sh
```diff
@@ -0,0 +1,110 @@
...
+        pkg_temp|instrumentation_temp)
+            ;;
+        *)
+            echo "unexpected top-level entry in archive: $name"
+            status=1
+            ;;
+    esac
+done < <(find "$tmpdir" -mindepth 1 -maxdepth 1 -print0)
+
+status=0
```
txabman42: Lets put this before the while, if not, `status=1` will be overridden

docs/getting-started.md
```diff
@@ -29,19 +29,13 @@ No manual code changes required.
...
-   > `make build` (or `make install`) from a clone is currently the only supported
-   > way to obtain a working `otelc` binary. Installing via
-   > `go get -tool go.opentelemetry.io/otelc/tool/cmd/otelc` is **not yet supported**:
-   > the instrumentation bundle is embedded from a file that is generated by
-   > `make build` and is not part of the published module, so a `go get`-installed
-   > binary cannot instrument builds. This will be resolved once the instrumentation
-   > packages are decoupled from the executable
-   > ([#585](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/585)).
+   # Option 2: Install as tool dependency (Go 1.24+)
+   go get -tool go.opentelemetry.io/otelc/tool/cmd/otelc
```
txabman42, line 36: We would need to test if this works

txabman42 (changes requested):

docs/getting-started.md
amazingakai: I tested with:

```bash
git clone --depth 1 --branch work/amazingakai/otelc-bundle-git https://github.com/amazingakai/opentelemetry-go-compile-instrumentation.git /tmp/otelc-amazingakai
cd /tmp/otelc-amazingakai
go build ./tool/cmd/otelc
cd demo/app/basic
../../../otelc go build
```

Which is similar to `go get -tool`/`go install` workflows.

.github/scripts/verify-bundle.sh
amazingakai: Good catch, done

.github/scripts/verify-bundle.sh
amazingakai: Added at [`7b188d0` (this PR)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/677/changes/7b188d06177a9ab77c0a86af39f390aa8466f6ea#diff-8bdf3106cfd84148a5729fa6e3e2af578873f2ba28b51b6b4acc19351951c2abR107)

.github/scripts/verify-bundle.sh
amazingakai: Thanks for checking. This is most likely due to differences between tar implementations (GNU tar vs. bsdtar), which can produce different archives even when the archived contents are identical. I'll look into making the packaging more reproducible so regenerating the bundle produces the same archive.

amazingakai: Added  `.tools/bundle`, which uses go's stdlib to produce reproducible archives across platforms. Could you PTAL?

amazingakai:
Verified across Fedora and FreeBSD that `.tools/bundle` produces reproducible output. I've also simplified `verify-bundle.sh` to run `make package` and check `git diff`, since the generated archive should now be reproducible.

txabman42 (approved):
LGTM! ❤️

.github/scripts/verify-bundle.sh
txabman42: Fixed, working fine, TY

PR #678, opened 2026-07-11
fix(gls): prevent stale parents in pooled goroutines

### LINKED ISSUE #673: bug(GLS): reused pool goroutines keep a stale parent span forever, parenting later requests into the wrong trace
GLS trace context is copied from parent to child goroutine in one place: the `runtime.newproc1` hook (`instrumentation/runtime/otelc.yaml`), the only call site of `propagateOtelContext`. Nothing removes a span from a goroutine's GLS except that goroutine ending the span itself. `ClearTraceContext` has no callers in any rule, and there is no goexit hook.

Worker pools break this model. Pools like `ants` spawn workers lazily inside `Submit`, so a worker created during one request inherits that request's span and keeps it for its entire life:

1. Request A is in flight; its span is on the handler goroutine's GLS.
2. The pool spawns a worker for the first time from inside that handler. `newproc1` clones the GLS, and the worker now holds span A (`Clone` in `otel_trace_context.go` copies the tail span).
3. Span A ends. `End` removes it from the ending goroutine's GLS only. The worker's copy survives, its `SpanContext` stays valid, and the `recordingSpan` stays pinned for as long as the worker lives.
4. Request B submits a task that starts a span from `context.Background()`. `afterSpanFromContext` (`instrumentation/go.opentelemetry.io/otel/trace/hook.go`) substitutes the stale span A, and `tracer.Start` uses it as the parent.

From then on, every span started this way on that worker lands in request A's trace. A propagation gap would produce an orphaned root; this produces a wrong parent, silently. Workers pre-spawned before any span exists get empty clones and only produce orphan roots, so whether you get contamination depends on which request happened to spawn the worker. That makes it hard to reproduce and harder to notice.

## Reproduction

Built with `otelc` at `1324a37` (`make build`), Go 1.25.10, `go.opentelemetry.io/otel` and `otel/sdk` v1.44.0. Full program below; the pool is a minimal hand-rolled version of what `ants` does (worker goroutine spawned lazily by the first `Submit`), so the mechanism is dependency-free.

```go
package main

import (
	"context"
	"fmt"
	"net/http"
	"time"

	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
	sdktrace "go.opentelemetry.io/otel/sdk/trace"
	"go.opentelemetry.io/otel/trace"
)

type pool struct {
	tasks   chan func()
	started bool
}

func (p *pool) Submit(f func()) {
	if !p.started {
		p.started = true
		go func() { // worker spawned here: newproc1 clones the submitter's GLS
			for t := range p.tasks {
				t()
			}
		}()
	}
	p.tasks <- f
}

var workers = &pool{tasks: make(chan func(), 16)}

func spanInfo(prefix string, sc trace.SpanContext) {
	fmt.Printf(">>> %s trace_id=%s span_id=%s\n", prefix, sc.TraceID(), sc.SpanID())
}

func main() {
	exp, err := stdouttrace.New()
	if err != nil {
		panic(err)
	}
	tp := sdktrace.NewTracerProvider(sdktrace.WithBatcher(exp))
	otel.SetTracerProvider(tp)
	defer func() { _ = tp.Shutdown(context.Background()) }()

	http.HandleFunc("/a", func(w http.ResponseWriter, r *http.Request) {
		spanInfo("REQUEST-A-SERVER-SPAN", trace.SpanFromContext(r.Context()).SpanContext())
		done := make(chan struct{})
		workers.Submit(func() { // first Submit: worker is born under span A
			defer close(done)
			time.Sleep(5 * time.Millisecond)
		})
		<-done
		fmt.Fprintln(w, "a done")
	})

	http.HandleFunc("/b", func(w http.ResponseWriter, r *http.Request) {
		spanInfo("REQUEST-B-SERVER-SPAN", trace.SpanFromContext(r.Context()).SpanContext())
		done := make(chan struct{})
		workers.Submit(func() {
			defer close(done)
			// correct Go: a deliberately detached root span
			_, span := otel.Tracer("repro").Start(context.Background(), "job-from-b")
			defer span.End()
			spanInfo("JOB-FROM-B", span.SpanContext())
			time.Sleep(5 * time.Millisecond)
		})
		<-done
		fmt.Fprintln(w, "b done")
	})

	fmt.Println("listening on :8123")
	_ = http.ListenAndServe(":8123", nil)
}
```

```
otelc go build -o glsrepro-instr .
./glsrepro-instr &
curl localhost:8123/a; curl localhost:8123/b; curl localhost:8123/b
```

Observed (span records from the stdouttrace exporter):

```
GET        trace: 33f812d2d3eb1917...  span: 866fcc9e92d178a8  parent: none        <- request A
GET        trace: 6040b14277bd4f2e...  span: 11b491150c138f1d  parent: none        <- request B1
job-from-b trace: 33f812d2d3eb1917...  span: c74dcd4c0e1ab6ed  parent: 866fcc9e92d178a8
GET        trace: 4a893d42d43119b5...  span: 725c84375da0eb9e  parent: none        <- request B2
job-from-b trace: 33f812d2d3eb1917...  span: 84674481151d221a  parent: 866fcc9e92d178a8
```

Both `job-from-b` spans, submitted by two different later requests, are parented under request A's
server span, in request A's trace, after span A ended. The application code is correct Go; the
wrong parent comes entirely from the injected propagation.

One precondition worth its own attention: the repro only triggers when the application imports
`go.opentelemetry.io/otel/sdk` itself (here via its own `TracerProvider` setup). In the
zero-config case, where the app imports only the API and `otelc` injects the SDK, the
`instrumentation/go.opentelemetry.io/otel/sdk/trace` module is not selected, because module
selection runs against the pre-instrumentation dependency graph. The build then contains the
`newproc1` copy and the `SpanFromContext` fallback but no GLS pushes at all, so cross-goroutine
propagation and `GetTraceAndSpanID` are silently inert. In other words: `otelc` does not instrument
the SDK it injects. That looks like a separate bug; happy to file it separately.

## Prior art

The two GLS lineages this project merged handled the problem differently.

dd-trace-go with Orchestrion never copies GLS at spawn. Its runtime aspects are the `g` field, same-goroutine accessors, and a `goexit1` hook that nils the slot (`internal/orchestrion/gls.orchestrion.yml`). `Finish` pops the entry, entries finished from another goroutine are reclaimed on the holder's next push, and an explicit context always wins over GLS (DataDog/dd-trace-go#4528). Their failure mode at goroutine boundaries is a visible trace split, and their docs call that expected behavior. Contamination requires an application bug, namely a span that never finishes.

loongsuite-go-agent copies at spawn like `otelc` does, and ships dedicated pool plugins (`pkg/rules/goants-v1`, `goants-v2`) because spawn-time copy cannot cover pooled workers.

`otelc` currently has the copy-at-spawn model with neither the pool plugins nor documentation of the limitation.

## Possible directions

In rough order of value per effort:

1. Skip ended spans in the fallback: `afterSpanFromContext` (or `tail()`) ignores spans that have ended. Contamination degrades to an orphan root, matching Datadog's failure mode, and the pinned `recordingSpan` gets released. Needs an "ended" signal readable from the hook.
2. Clean up on span end across GLS clones, or clear GLS in a goexit hook the way Orchestrion does. The goexit hook helps recycled goroutines but not long-lived pool workers.
3. Pool-aware rules for common pool libraries, following loongsuite's goants plugins.
4. Document the limitation and emit a diagnostic (debug log or metric) when the GLS fallback supplies a parent, so the behavior is at least discoverable.

Related, probably separate issues: the same fallback re-parents intentional root spans started from `context.Background()` on any goroutine with a non-empty GLS, and past `OTEL_GLS_MAX_SPANS` the tracker drops adds silently while `tail()` keeps returning a wrong parent. Neither behavior is in `docs/configuration.md`. I can file these separately.

macayu17:
## What

A worker reused by a pool could retain a span copied into its GLS after that span had ended. Later work on that worker could then be parented to the old request trace.

## Why

Each goroutine had its own GLS wrapper. Ending a span only removed the wrapper visible to the goroutine that called `End`, so copies held by pooled workers could stay valid indefinitely. This also happened when a span was ended from another goroutine.

## How

  - Track shared lifecycle state for each valid SDK span.
  - Mark that state before removing it when the span ends.
  - Ignore ended GLS entries when reading or cloning context.
  - Add integration coverage for reused workers, cross-goroutine `End`, and sampled-out spans.
  - Verify with `make test`, Go lint, and license-header checks.

  Closes #673


instrumentation/go.opentelemetry.io/otel/sdk/trace/otel_trace_context.go
```diff
@@ -42,12 +44,37 @@ type traceContext struct {
...
+}
+
+var spanStates sync.Map
+
+func stateForSpan(span trace.Span) *atomic.Bool {
+	sc := span.SpanContext()
+	if !sc.IsValid() {
+		return &atomic.Bool{}
+	}
+	state, _ := spanStates.LoadOrStore(spanKey{sc.TraceID(), sc.SpanID()}, &atomic.Bool{})
```
kakkoyun: An entry only leaves `spanStates` in `markSpanEnded`, which runs from the `End` hook. A span that is started but never ended (a leaked span, or a panic that skips `defer span.End()`) keeps its `spanKey -> *atomic.Bool` entry for the life of the process. Before this change that bookkeeping lived in the per-goroutine GLS list, which is bounded by `otelGLSMaxSpans` and collected when the goroutine dies; this map is global, uncapped, and outlives the goroutine. A healthy app stays bounded by its live-span count, but any steady rate of unended spans grows it without limit, and this file is compiled into every instrumented binary.

I checked that the normal start-then-end path drains the map cleanly, so this is only the unended case. Is "every span ends" the contract you want to lean on here, or is a cap (or at least a documented ceiling) worth adding?

instrumentation/go.opentelemetry.io/otel/sdk/trace/otel_trace_context.go
```diff
@@ -42,12 +44,37 @@ type traceContext struct {
...
 }
 
-func (tc *traceContext) size() int {
-	return tc.n
+type spanKey struct {
+	traceID trace.TraceID
+	spanID  trace.SpanID
+}
+
+var spanStates sync.Map
```
kakkoyun: If the registry stays, `sync.Map` matches this workload poorly. Every span start stores a fresh key and every end deletes it, which is the create-then-delete churn its own docs steer away from (it is tuned for read-mostly caches or disjoint stable keys). I benchmarked this exact pattern: `sync.Map` came out around 3 allocs/op where a plain `map` behind a `sync.Mutex` did roughly zero, and it was faster too. The overhead check passes, so this is not urgent, but a mutex-guarded map or a small shard array would suit the churn better.

instrumentation/go.opentelemetry.io/otel/sdk/trace/otel_trace_context.go
```diff
@@ -67,11 +94,14 @@ func (tc *traceContext) add(span trace.Span) bool {
 
 //go:norace
 func (tc *traceContext) tail() trace.Span {
-	if tc.n == 0 {
+	for tc.sw != nil && tc.sw.ended.Load() {
```
kakkoyun, line 140: `tail()` reclaims ended wrappers only from the head, and a cross-goroutine `End` marks a wrapper ended through `markSpanEnded` without touching that goroutine's list or `n`. So an ended span sitting below a still-live span stays linked and counted. The pooled-worker path drains after each task, so it is fine there. A long-lived goroutine that ends its locally-started spans on other goroutines, out of LIFO order, is the problem case: `n` stays inflated until those wrappers reach the head, and since `add()` rejects on `n >= otelGLSMaxSpans`, it can start silently dropping spans while most of the stack is already dead.

Sweeping ended wrappers across the whole list before rejecting at the cap closes this. I verified a compaction pass on the cap path admits the span again.

instrumentation/go.opentelemetry.io/otel/sdk/trace/otel_trace_context.go
```diff
@@ -83,6 +113,7 @@ func (tc *traceContext) localRootSpan() trace.Span {
 }
 
 func (tc *traceContext) del(span trace.Span) {
+	markSpanEnded(span)
```
kakkoyun: With `End` now expected to run on a goroutine that never held the span, `traceContextDelSpan` -> `getOrInitTraceContext()` allocates and installs an empty `traceContext` into that ender's GLS only to hit `n == 0` and return. Calling `markSpanEnded` first and then operating on the context only when one already exists keeps the cross-goroutine signal and drops the wasted allocation on the end hook. It is small and predates this PR, but this change turns the cross-goroutine end into a normal path, so it is worth tidying here.

instrumentation/go.opentelemetry.io/otel/sdk/trace/otel_trace_context.go
```diff
@@ -67,11 +94,14 @@ func (tc *traceContext) add(span trace.Span) bool {
 
 //go:norace
```
kakkoyun, line 138: `tail()` now mutates the context by popping ended wrappers, but it is still called as a getter from `GetTraceAndSpanID`, `GetTraceContext`, and `spanFromGLS`. That is safe because every caller runs on the owning goroutine and the list fields are only ever touched by that goroutine, which is also what makes `//go:norace` sound. Documenting that invariant keeps a later helper from calling `tail()` on another goroutine's context and corrupting the list with no race-detector warning to catch it.

```suggestion
// tail must be called only on the current goroutine's own context: it mutates the
// stack (popping ended wrappers) and the list fields are unsynchronized. The
// shared ended flags are the only cross-goroutine state.
//
//go:norace
```

kakkoyun:
@macayu17 Also CI is red.

macayu17:
Addressed the review comments in cf8e3b3. The failing TestAddDeps/non_main_package_name test is also red on current main: the golden file expects Printf, while the current code generates Print. The bundle, lint, consistency, and license checks are green.

test/apps/otelsdk/main.go
```diff
@@ -46,6 +105,13 @@ func main() {
 	addr := fmt.Sprintf(":%s", *port)
 
 	http.HandleFunc("/otel", otelHandler)
+	http.HandleFunc("/worker", workerHandler)
```
txabman42: Lets add some comments on why and what each of them does.

e.g. why _[server, lower, top] stack and why MAX_SPANS=3_ etc.

txabman42 (approved):
LGTM! Complete solution ❤️

NameHaibinZhang (approved):

test/apps/otelsdk/main.go
txabman42: I added those comments myself just to unblock this

txabman42:
@kakkoyun could you give this PR another 👀 since you reviewed it before? If its ok from your side, this is ready to be merged and v1-ready

PR #688, opened 2026-07-14
chore: migrate constants to new otelc root

amazingakai:


tool/internal/setup/pin_test.go
```diff
@@ -113,7 +113,7 @@ func TestGenerateDirective(t *testing.T) {
 				Generate: &trueValue,
 			},
 			want: "//go:generate go tool " +
-				util.OtelcOldToolExe +
+				util.OtelcToolExe +
```
kakkoyun: I know this is an existing issue. However, let's just use literals in the tests instead of the constant. That would make test more useful to catch any regressions.

kakkoyun (changes requested):
One things we need to change.

tool/internal/setup/pin_test.go
amazingakai: Good idea. Agreed

amazingakai: [`23dc4d7` (this PR)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/688/commits/23dc4d7f530bf6c66a9dbc7de80f9d104cd0efa6)

kakkoyun (approved):

txabman42 (approved):
LGTM! ❤️

Issue #679, opened 2026-07-13
Add compile-time instrumentation for anthropic-sdk-go (gen_ai.* spans)


pradhyum6144:
## Summary

Add zero-code, compile-time instrumentation for `github.com/anthropics/anthropic-sdk-go`
producing GenAI client spans, mirroring the merged OpenAI instrumentation
(`instrumentation/github.com/openai/openai-go/`, #604).

## Motivation

- OpenAI Go SDK instrumentation landed in #604 (v1/v2/v3). `anthropic-sdk-go` is the other
  major Go LLM SDK with no otelc coverage.
- Go still trails Python here: `opentelemetry-python-contrib` ships dedicated GenAI
  instrumentations, while Go users get at most an `http.client` span with no `gen_ai.*`
  attributes.
- The cost is low: `anthropic-sdk-go` is Stainless-generated like `openai-go`, so the
  established pattern transfers almost 1:1 — `anthropic.NewClient(opts ...option.RequestOption)`
  accepts `option.WithMiddleware(...)` with the exact same middleware signature the openai
  hook injects (`instrumentation/github.com/openai/openai-go/hook.go`).

## Scope — Phase 1 (traces)

New module `instrumentation/github.com/anthropics/anthropic-sdk-go/` following the openai-go
layout (`otelc.yaml`, `hook.go`, `middleware.go`, `streaming.go`, `semconv/`), per
`docs/instrument-guide.md`.

**Hook point** — one `inject_hooks` rule on `anthropic.NewClient` that prepends
`option.WithMiddleware(OtelMiddleware())`, gated by `runtime.Instrumented("ANTHROPIC")`,
with `runtime.SuppressHTTPClientInstrumentation` to avoid a duplicate `net/http` client span
(same as `instrumentation/github.com/openai/openai-go/hook.go:46` / `middleware.go:154`).

**Instrumented calls** (both hit `POST /v1/messages`; classified by URL path suffix, so
`/v1/messages/count_tokens` and `/v1/messages/batches` are excluded for free):

- Non-streaming: `client.Messages.New(ctx, MessageNewParams)` (sdk `message.go:56`)
- Streaming: `client.Messages.NewStreaming(ctx, MessageNewParams)` returning
  `ssestream.Stream[MessageStreamEventUnion]` (sdk `message.go:86`) — detected via
  `Content-Type: text/event-stream`, same as openai (`middleware.go:172-177`)

[...]

PR #689, opened 2026-07-14
feat(instrumentation): add Anthropic Go SDK instrumentation (non-streaming)

pradhyum6144:
Adds compile-time instrumentation for github.com/anthropics/anthropic-sdk-go, producing GenAI client spans for the non-streaming Messages API. This is the first phase of #679 and follows the pattern established by the openai-go instrumentation.

The rule hooks anthropic.NewClient and injects an HTTP middleware via option.WithMiddleware, gated by the ANTHROPIC instrumentation key. The middleware classifies POST /v1/messages calls (the path suffix keeps count_tokens and batches out of scope), starts a client span, and suppresses the underlying net/http client span so there is no duplicate.

Attributes follow the gen_ai semantic conventions: request model, max_tokens, temperature, top_p, and top_k (top_k is Anthropic specific, with no openai equivalent), provider name, response id and model, finish reasons, and token usage. Anthropic returns a single stop_reason rather than per-choice finish_reason values, so it is wrapped into a one element slice to keep the shared RequireGenAIClientSemconv helper reusable. The API has no total_tokens field, so it is computed from input plus output. Prompt-cache usage (cache_read and cache_creation input tokens) is recorded when present.

Per the design discussion in #679, this covers only the non-streaming path. The streaming reader is left as a documented stub and will land in a follow-up once the accumulation approach is confirmed. Content capture is deferred to a later phase.

The rule is pinned to a v1.0.0 lower bound so the single module covers a broad version range.

Tested with unit tests covering the hook, operation classification, provider mapping, request parsing, and the full span shape including the cache-token cases. An integration test app and mock-server test mirror the openai-go setup and run under the existing integration CI.

Part of #679


NameHaibinZhang (commented):
## Code Review Summary

Thanks for this PR! The implementation closely follows the OpenAI instrumentation pattern and covers the non-streaming Messages API well. The unit tests are thorough for the main paths.

### Findings

**Must fix before merge:**
- `total_tokens` calculation is incorrect when prompt caching is used — Anthropic's `input_tokens` excludes cache tokens (unlike OpenAI)
- Streaming requests emit incomplete spans missing all `gen_ai.response.*` and `gen_ai.usage.*` attributes

**Should fix:**
- Request body is partially consumed on `ReadAll` error, passing a truncated body to the SDK
- Missing nil guard on `resp.Body` in `handleNonStreamingResponse`
- `streamingReader` edge cases (close-before-read, double-close, empty stream) are untested

**Consider:**
- `getProviderName` uses map iteration with non-deterministic order
- Integration test doesn't verify Anthropic-specific cache attributes end-to-end
- Consider adding a `demo/app/anthropic/` for consistency with OpenAI

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
```diff
@@ -0,0 +1,239 @@
...
+
+	span.SetAttributes(
+		semconv.GenAIResponseID(resp.ID),
+		semconv.GenAIResponseModel(resp.Model),
+		semconv.GenAIResponseFinishReasons(reasons),
+		semconv.GenAIUsageInputTokens(resp.Usage.InputTokens),
+		semconv.GenAIUsageOutputTokens(resp.Usage.OutputTokens),
+		// The Messages API reports no total_tokens field; derive it so the
+		// span shape matches the other GenAI instrumentations.
+		semconv.GenAIUsageTotalTokens(resp.Usage.InputTokens+resp.Usage.OutputTokens),
```
NameHaibinZhang: **[Major] `total_tokens` and `input_tokens` are incorrect when prompt caching is used**

Anthropic's `input_tokens` field excludes cache tokens — cache reads and creations are reported separately in `cache_read_input_tokens` and `cache_creation_input_tokens`. This differs from OpenAI, where `prompt_tokens` already includes cached tokens, so the `input + output` formula works there but not here.

Example with caching:
```
input_tokens: 500, cache_creation_input_tokens: 1000, output_tokens: 200
Expected total = 500 + 1000 + 200 = 1700
Actual (code)   = 500 + 200             = 700  ← underreports by 1000
```

Similarly, `gen_ai.usage.input_tokens` should include cache tokens per the semconv definition ("number of tokens used in the prompt").

Suggested fix:
```go
totalInput := resp.Usage.InputTokens +
    resp.Usage.CacheReadInputTokens +
    resp.Usage.CacheCreationInputTokens

span.SetAttributes(
    semconv.GenAIUsageInputTokens(totalInput),
    semconv.GenAIUsageOutputTokens(resp.Usage.OutputTokens),
    semconv.GenAIUsageTotalTokens(totalInput + resp.Usage.OutputTokens),
)
```

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
```diff
@@ -0,0 +1,239 @@
...
+			span.End()
+			return resp, nil
+		}
+
+		contentType := resp.Header.Get("Content-Type")
+		isStreaming := strings.HasPrefix(contentType, "text/event-stream")
+
+		if isStreaming {
+			span.SetAttributes(semconv.GenAIRequestIsStream(true))
+			resp.Body = newStreamingReader(resp.Body, span, start)
```
NameHaibinZhang: **[Major] Streaming requests emit incomplete spans missing all response/usage attributes**

The middleware creates a span and sets `gen_ai.request.is_stream=true` for streaming responses, but `streamingReader.finalize()` only calls `span.End()` without setting any `gen_ai.response.*` or `gen_ai.usage.*` attributes. This means production users of the streaming Messages API will export gen_ai spans that are missing token usage data.

Downstream dashboards/alerts that expect `gen_ai.usage.input_tokens` on every gen_ai span (as validated by `RequireGenAIClientSemconv`) will show data gaps for all streaming requests.

Since the PR title says "non-streaming", consider one of:
1. **Pass-through** (skip span creation for streaming) — treat it like `opUnknown` until accumulation is implemented:
```go
if isStreaming {
    span.End()
    return resp, nil
}
```
2. **Keep the span but mark it** — add an attribute like `gen_ai.response.incomplete=true` so consumers can filter them.

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
```diff
@@ -0,0 +1,239 @@
...
+		provider := getProviderName(req.URL.Host)
+		opName := operationName(op)
+
+		// Read a bounded copy for attribute parsing, but preserve the full body for the SDK.
+		var buf bytes.Buffer
+		tee := io.TeeReader(req.Body, &buf)
+		bodyBytes, err := io.ReadAll(io.LimitReader(tee, maxRequestBodySize))
+		if err != nil {
+			return next(req)
+		}
```
NameHaibinZhang: **[Minor] Request body is partially consumed when `ReadAll` returns an error**

If `io.ReadAll` fails, `req.Body` has already been partially drained by the TeeReader, but the body is not reassembled before calling `next(req)`. This passes a truncated request body to the SDK, which will send a malformed request to the API.

While the body is typically a `bytes.Reader` that won't error in practice, a defensive fix is straightforward:

```go
if err != nil {
    req.Body = struct {
        io.Reader
        io.Closer
    }{io.MultiReader(&buf, req.Body), req.Body}
    return next(req)
}
```

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
```diff
@@ -0,0 +1,239 @@
...
+	_ context.Context,
+	resp *http.Response,
+	span trace.Span,
+	_ time.Time,
+) {
+	defer span.End()
+
+	// Read a bounded preview for parsing, but reassemble the full body for callers.
+	var buf bytes.Buffer
+	tee := io.TeeReader(resp.Body, &buf)
```
NameHaibinZhang, line 172: **[Minor] Missing nil check on `resp.Body`**

If `resp.Body` is nil, `io.TeeReader(resp.Body, &buf)` will create a TeeReader with a nil source reader, causing a nil pointer dereference panic when `ReadAll` calls `Read`.

In practice, the Anthropic SDK's HTTP transport always returns a non-nil body for successful responses, but adding a guard is cheap:

```go
if resp.Body == nil {
    return
}
```

instrumentation/github.com/anthropics/anthropic-sdk-go/streaming.go
```diff
@@ -0,0 +1,62 @@
...
+	if r.done.CompareAndSwap(false, true) {
+		r.finalize()
+	}
+	if r.reader != nil {
+		return r.reader.Close()
+	}
+	return nil
+}
+
+func (r *streamingReader) finalize() {
```
NameHaibinZhang: **[Minor] `streamingReader` edge cases are untested**

The `streamingReader` contains real logic (atomic CAS guard for idempotent `finalize`, span lifecycle management), but the only test coverage is `TestOtelMiddleware_Streaming` in `middleware_test.go`, which only verifies the happy path. The following edge cases from the comparable OpenAI `streaming_test.go` are missing:

- **Close-before-Read**: caller closes `resp.Body` without reading (common in error paths) — span must still end exactly once
- **Multiple Close**: defensive double-close must not double-end the span
- **Empty stream**: empty body reaching EOF — span must end exactly once
- **Body integrity**: the test discards the read body content (`_, err = io.ReadAll(resp.Body)`); consider asserting the content matches the original SSE data

Consider adding a dedicated `streaming_test.go` mirroring the OpenAI test suite.

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func getProviderName(host string) string {
+	for keyword, provider := range providerMapping {
+		if strings.Contains(host, keyword) {
+			return provider
+		}
+	}
+	return "anthropic"
+}
```
NameHaibinZhang, line 43: **[Suggestion] `getProviderName` uses map iteration with non-deterministic order**

Go map iteration order is randomized. If a host matches multiple keywords (e.g., a hypothetical `anthropic.com.localhost`), the returned provider would be non-deterministic. While the current three keywords are effectively mutually exclusive for valid hosts, using a map for ordered matching is a code smell.

Consider using an ordered slice for deterministic priority:

```go
var providerEntries = []struct{ keyword, provider string }{
    {"anthropic.com", "anthropic"},
    {"localhost", "local"},
    {"127.0.0.1", "local"},
}

func getProviderName(host string) string {
    for _, e := range providerEntries {
        if strings.Contains(host, e.keyword) {
            return e.provider
        }
    }
    return "anthropic"
}
```

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
NameHaibinZhang: Anthropic's `input_tokens` field excludes cache tokens — cache reads and creations are reported separately in `cache_read_input_tokens` and `cache_creation_input_tokens`. This differs from OpenAI, where `prompt_tokens` already includes cached tokens, so the `input + output` formula works there but not here.

Example with caching:
```
input_tokens: 500, cache_creation_input_tokens: 1000, output_tokens: 200
Expected total = 500 + 1000 + 200 = 1700
Actual (code)   = 500 + 200             = 700  ← underreports by 1000
```

Similarly, `gen_ai.usage.input_tokens` should include cache tokens per the semconv definition ("number of tokens used in the prompt").

Suggested fix:
```go
totalInput := resp.Usage.InputTokens +
    resp.Usage.CacheReadInputTokens +
    resp.Usage.CacheCreationInputTokens

span.SetAttributes(
    semconv.GenAIUsageInputTokens(totalInput),
    semconv.GenAIUsageOutputTokens(resp.Usage.OutputTokens),
    semconv.GenAIUsageTotalTokens(totalInput + resp.Usage.OutputTokens),
)
```

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
NameHaibinZhang: The middleware creates a span and sets `gen_ai.request.is_stream=true` for streaming responses, but `streamingReader.finalize()` only calls `span.End()` without setting any `gen_ai.response.*` or `gen_ai.usage.*` attributes. This means production users of the streaming Messages API will export gen_ai spans that are missing token usage data.

Downstream dashboards/alerts that expect `gen_ai.usage.input_tokens` on every gen_ai span (as validated by `RequireGenAIClientSemconv`) will show data gaps for all streaming requests.

Since the PR title says "non-streaming", consider one of:
1. **Pass-through** (skip span creation for streaming) — treat it like `opUnknown` until accumulation is implemented:
```go
if isStreaming {
    span.End()
    return resp, nil
}
```
2. **Keep the span but mark it** — add an attribute like `gen_ai.response.incomplete=true` so consumers can filter them.

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
NameHaibinZhang: If `io.ReadAll` fails, `req.Body` has already been partially drained by the TeeReader, but the body is not reassembled before calling `next(req)`. This passes a truncated request body to the SDK, which will send a malformed request to the API.

While the body is typically a `bytes.Reader` that won't error in practice, a defensive fix is straightforward:

```go
if err != nil {
    req.Body = struct {
        io.Reader
        io.Closer
    }{io.MultiReader(&buf, req.Body), req.Body}
    return next(req)
}
```

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
NameHaibinZhang, line 172: If `resp.Body` is nil, `io.TeeReader(resp.Body, &buf)` will create a TeeReader with a nil source reader, causing a nil pointer dereference panic when `ReadAll` calls `Read`.

In practice, the Anthropic SDK's HTTP transport always returns a non-nil body for successful responses, but adding a guard is cheap:

```go
if resp.Body == nil {
    return
}
```

instrumentation/github.com/anthropics/anthropic-sdk-go/streaming.go
NameHaibinZhang: The `streamingReader` contains real logic (atomic CAS guard for idempotent `finalize`, span lifecycle management), but the only test coverage is `TestOtelMiddleware_Streaming` in `middleware_test.go`, which only verifies the happy path. The following edge cases from the comparable OpenAI `streaming_test.go` are missing:

- **Close-before-Read**: caller closes `resp.Body` without reading (common in error paths) — span must still end exactly once
- **Multiple Close**: defensive double-close must not double-end the span
- **Empty stream**: empty body reaching EOF — span must end exactly once
- **Body integrity**: the test discards the read body content (`_, err = io.ReadAll(resp.Body)`); consider asserting the content matches the original SSE data

Consider adding a dedicated `streaming_test.go` mirroring the OpenAI test suite.

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
NameHaibinZhang, line 43: Go map iteration order is randomized. If a host matches multiple keywords (e.g., a hypothetical `anthropic.com.localhost`), the returned provider would be non-deterministic. While the current three keywords are effectively mutually exclusive for valid hosts, using a map for ordered matching is a code smell.

Consider using an ordered slice for deterministic priority:

```go
var providerEntries = []struct{ keyword, provider string }{
    {"anthropic.com", "anthropic"},
    {"localhost", "local"},
    {"127.0.0.1", "local"},
}

func getProviderName(host string) string {
    for _, e := range providerEntries {
        if strings.Contains(host, e.keyword) {
            return e.provider
        }
    }
    return "anthropic"
}
```

test/integration/anthropic_client_test.go
```diff
@@ -0,0 +1,104 @@
...
+					"text": "Hello!",
+				},
+			},
+			"stop_reason":   "end_turn",
+			"stop_sequence": nil,
+			"usage": map[string]any{
+				"input_tokens":  10,
+				"output_tokens": 20,
+			},
+		}
```
NameHaibinZhang, line 127: The mock server returns `input_tokens: 10, output_tokens: 20` but omits `cache_read_input_tokens` and `cache_creation_input_tokens`. The `RequireGenAIClientSemconv` helper only validates standard attributes. The Anthropic-specific cache token attributes (`gen_ai.usage.cache_read.input_tokens`, `gen_ai.usage.cache_creation.input_tokens`) are unit-tested but not verified end-to-end through the instrumented binary.

Consider adding a second test case whose mock server returns non-zero cache values and asserting those attributes appear on the span.

tool/data/otelc-bundle.tgz
NameHaibinZhang, line 1: this is not use

pradhyum6144:
Thanks for the thorough review! All findings addressed in 782b79c. Module coverage is now 100% with -race clean.

On tool/data/otelc-bundle.tgz  could you clarify? The Verify Bundle check regenerates the archive via make package and fails if the committed file differs, which is why it's included. Happy to drop it if there's a different workflow for new instrumentations.

NameHaibinZhang (approved):

NameHaibinZhang:
A conversation must be resolved before this pull request can be merged.

pradhyum6144:
Thanks @NameHaibinZhang! All conversations resolved, CI is green, and it's approved ready to merge whenever you are.

NameHaibinZhang (approved):

amazingakai:
@pradhyum6144 Please also add a corresponding entry to the OTel registry (see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation)).

pradhyum6144:
> @pradhyum6144 Please also add a corresponding entry to the OTel registry (see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation)).

Thanks @amazingakai opened open-telemetry/opentelemetry.io#10909 with the registry entry. I also added Anthropic to the compile-time supported-libraries page, since OpenAI was listed there but the Anthropic row was missing

PR #696, opened 2026-07-15
chore(semconv): add local OTel Weaver registry for the emitted telemetry contract

NameHaibinZhang:
## What

Adds a project-owned [OTel Weaver](https://github.com/open-telemetry/weaver) schema registry under `schemas/otelc/` that declares the telemetry this project's instrumentations emit, and validates it in CI against the pinned upstream semantic conventions.

## Why

Today `make lint/semantic-conventions` runs `weaver registry check` against a clone of the upstream `semantic-conventions` repo. Upstream is always valid, so the check gives little signal, and there is no machine-readable description of the metrics/spans/attributes this project actually produces.

This follows the pattern used by [opentelemetry-ebpf-instrumentation](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation) (OBI), which maintains its emission contract as a Weaver registry under `schemas/obi/`.

## What's included

- `schemas/otelc/registry_manifest.yaml` — registry metadata + a dependency on the upstream semconv version pinned in `.semconv-version`, pre-fetched under `.deps/` (git-ignored).
- `schemas/otelc/groups/http.yaml` — the first instrumentation contract (`net/http` client & server metrics). Upstream attributes are referenced with `ref:`; library-specific telemetry would be declared with `id:`. Other instrumentations can follow this template.
- `scripts/semconv/` — `fetch-upstream-semconv.sh`, `lint-schema.sh`, and `lint-schema-filter.jq`. Weaver runs from the `otel/weaver` OCI image (no host install needed).
- `Makefile` — new `fetch-upstream-semconv` and `lint-schema` targets; `lint/semantic-conventions` now validates the local registry.
- CI `check-registry-diff.yaml` — runs the Docker-based check and asserts the manifest version matches `.semconv-version`.

### Note on `DuplicateMetricName`

Instrumentations mostly emit standard upstream metrics (e.g. `http.client.request.duration`). Weaver has no way to *reference* a metric defined in a dependency, so we re-declare the metrics we emit to pin the contract. Weaver then reports a `DuplicateMetricName` between our registry and the upstream dependency; this specific, expected duplicate is allowlisted in `lint-schema-filter.jq`. Any other diagnostic — including a metric declared twice within our own registry, or an unresolved `ref:` — still fails the lint.

## Scope

Scaffold + `net/http` as the worked example. The remaining instrumentations (grpc, database/sql, kafka, redis, k8s, openai, mongo) can be added incrementally using the same pattern. No runtime Go code is changed.

## Testing

- `make lint-schema` passes locally (Weaver via Docker).
- Reverse-checked: injecting an unresolved `ref:` makes the lint fail as expected.
- `yamlfmt` and `checkmake` are clean.

Requires Docker (or `OCI_BIN=podman`) and `jq`; both are preinstalled on the CI runner.


Makefile
```diff
@@ -28,6 +29,13 @@ TOOLS_DIR = .tools
 GO_VERSION = 1.25
 INTEGRATION_TEST_RUN ?= .
 
+# OTel Weaver execution for the local semantic-convention registry under
+# schemas/otelc/. Weaver runs from an OCI image (no host install required);
+# override OCI_BIN=podman or WEAVER_IMAGE=... to use a different runtime/version.
+OCI_BIN ?= docker
+WEAVER_IMAGE ?= otel/weaver:v0.19.0
```
kakkoyun, line 36: Can we configure renovate to auto-update this?

kakkoyun (approved):
LGTM.

Only one nit.

Makefile
NameHaibinZhang: Done. Added a `customManagers` regex rule to `.github/renovate.json5` that tracks the `WEAVER_IMAGE` variable in the Makefile—Renovate will now propose PRs when new `otel/weaver` tags are published.

scripts/semconv/fetch-upstream-semconv.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Pre-fetch the upstream OpenTelemetry semantic-conventions registry into
+# `schemas/otelc/.deps/` so weaver doesn't have to clone it over the network
+# on every container start. The pinned upstream version is parsed out of
+# `schemas/otelc/manifest.yaml` so there's a single source of truth.
```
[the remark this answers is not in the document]
NameHaibinZhang: Good catch, fixed. Updated the comment to reference the correct filename `schemas/otelc/registry_manifest.yaml`.

schemas/otelc/groups/http.yaml
```diff
@@ -0,0 +1,171 @@
+groups:
```
txabman42, line 1: @NameHaibinZhang, I have two quick questions on this:

- Schemas sync --> How is it ensured that our semconv go code is in sync with these semconv schemas?

- Adoption Scope --> The issue scope mentions adding this to other instrumentation packages. Should all packages be forced to be weave-compliant?
    - If YES: We should open a follow-up issue to enforce this compliance across the board.
    - If NO: When exactly should a package use it? We will need: clear, documented guidelines explaining when to apply it. Follow-up issues to add it to any existing instrumentation packages that meet those guidelines.

cc @kakkoyun

NameHaibinZhang: **Schemas sync:**

The sync is enforced at multiple levels:

1. **Version consistency** — The CI workflow (`check-registry-diff.yaml`) validates that the semconv version declared in `.semconv-version`, the upstream dependency path in `registry_manifest.yaml`, and the Go import paths in `instrumentation/**/semconv/*.go` all reference the same version.

2. **Structural validity** — `make lint-schema` runs `weaver registry check` against the local registry with the cached upstream dependency. This ensures all `ref:` attributes resolve correctly against the pinned upstream version. Any unresolved reference or internal duplication fails the check.

3. **Manual co-maintenance** — The YAML files include source-of-truth comments pointing to the corresponding Go files (e.g., `client.go`, `server.go`). When a developer modifies the Go semconv code, they are expected to update the corresponding YAML declaration in the same PR.

What we do **not** yet have is an automated tool that cross-checks metric/span names between Go code and YAML declarations. This could be a follow-up enhancement (e.g., a `weaver codegen` template or a custom linter that parses both sources). I'd be happy to open a tracking issue for that.

**Adoption Scope:**

I'd suggest a **gradual, opt-in approach** rather than forcing all packages to be weave-compliant immediately:

- **New instrumentations** should include a registry YAML declaration from day one (as part of the contribution checklist).
- **Existing instrumentations** that already have Go-level semconv code (gRPC, database/sql, kafka, redis, k8s, openai — 6 remaining) should be migrated incrementally via follow-up PRs. I can open a tracking issue for this.
- **Instrumentations without semconv code** (mongo, log, runtime, OTel core) can be deferred until they mature enough to emit well-defined telemetry.

The guideline for "when to add it" would be: **any instrumentation that emits metrics or spans with attributes referencing the OpenTelemetry semantic conventions should have a corresponding registry declaration.** This naturally excludes pure lifecycle hooks (like `runtime` or `otel/init`) that don't emit user-facing telemetry.

I'll open a follow-up issue to track the migration of the remaining 6 packages and document this guideline in `docs/semantic-conventions.md`.

txabman42: > Schemas sync

All three CI checks only verify that everyone agrees on the version string v1.37.0. Nothing verifies that the metrics/attributes the Go code actually emits match what the YAML declares. A dev can add `http.client.connection.duration` in Go forget the YAML and CI stays green.

> Manual co-maintenance — The YAML files include source-of-truth comments pointing to the corresponding Go files (e.g., client.go, server.go). When a developer modifies the Go semconv code, they are expected to update the corresponding YAML declaration in the same PR.

We expect that, but how we ensure its being done in that way? Now we have the context, but it some months? I would expect having a defined contract that avoids diverging from reality (go code). 

> What we do not yet have is an automated tool that cross-checks metric/span names between Go code and YAML declarations. This could be a follow-up enhancement (e.g., a weaver codegen template or a custom linter that parses both sources). I'd be happy to open a tracking issue for that.

IMO, this is a must and the key of the initiative. This is what generates the contract. This would generate the Go semconv constants from the registry, so the registry becomes the SSOT. The other idea would be to capture emited signals and compare them with the registry in order to verify if they are in sync. However, I prefer the first idea by using directly [weaver registry generate](https://github.com/open-telemetry/weaver/blob/main/docs/codegen.md). So the registry becomes the SSOT and we avoid any divergence. We can also check how other repos are using it. Lets push it!

txabman42: Idea: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/728

PR #704, opened 2026-07-17
refactor(tool): dispatch rule type on the do modifier name

### LINKED ISSUE #541: feat(tool/rule): consume structured where/do schema in rule constructors (drop normalize.go)
Follow-up from #377 (review threads from @txabman42 and @amazingakai).

## Context

Branch 0 of the OP4 stack landed the structured `where`/`do` YAML schema together with `tool/internal/rule/normalize.go`, which hoists selectors out of `where` and modifiers out of `do` back into the legacy flat-field shape that `createRuleFromFields` consumes. This double translation was a deliberate scope-limiter for branch 0.

## Problem

OP4.1–OP4.6 will continue to touch (a) the YAML schema, (b) `WhereDef`/`FilterDef`, (c) the flat-rule structs, and (d) the docs in lockstep. The flat representation has no semantic value of its own — it only exists because rule constructors were written before the structured schema existed. Maintaining both shapes doubles the surface area where selectors and modifiers can drift.

## Proposed work

- Add `DoDef` (modifier name + payload) symmetric with `WhereDef`.
- Change `parseRuleFromYaml` to dispatch directly on the `do` modifier name (no flat re-marshal round-trip).
- Refactor `createRuleFromFields` callers to accept the structured payload directly. The function likely splits per rule type.
- Delete `tool/internal/rule/normalize.go` and its tests.
- Verify by re-running `tool/internal/instrument/testdata/golden/...` with no output diff.

## Out of scope

Schema changes — this is purely an internal refactor.

## References

- #377 inline thread on `tool/internal/rule/normalize.go:12`
- #377 inline thread on `tool/internal/rule/base.go:52`
- @amazingakai's review-summary point about `WhereDef` containing rule-specific fields

kakkoyun:
Follow-up from #377; closes #541.

## Problem

Branch 0 of the `where`/`do` schema (ADR-0003) parses rules through a normalization layer (`tool/internal/rule/normalize.go`) that translates the structured shape back into the legacy flat field bag, drops the `do` modifier name, and re-infers the rule type from field presence. Each rule makes two YAML round-trips (map → normalize → re-marshal → struct), and the flat and structured shapes both carry the same selectors and modifiers, so they can drift apart.

## Change

Parse structured YAML directly and dispatch on the `do` modifier name:

- a typed payload per modifier (`HookAction`, `CodeAction`, …) and a `DoDef` union, symmetric with `WhereDef`; the modifier name selects the rule type with no re-marshal;
- `rule.ParseRules` is the single parser, shared by the setup phase and the golden-test harness (which previously kept its own copy);
- `normalize.go` and its tests are removed.

The concrete rule structs stay as the execution representation; the typed payloads are the structured layer a future JSON Schema generator (editor validation of rule files) can reflect over. Rationale in ADR-0006.

## Scope

Internal refactor only — the `where`/`do` YAML schema is unchanged. The five `func-sig-*` golden fixtures were the last flat-form rule files in the repo; they are migrated to the structured shape and their generated output is unchanged.

## Verification

- `go build`/`vet`/`gofmt`, repo `golangci-lint`, `markdownlint`: clean.
- All `tool/...` unit tests pass; `rule`/`setup` pass with `-race`.
- Golden instrumentation output is byte-identical (only input `rules.yml` fixtures changed).
- Validation parity with the removed normalizer is covered by new `parse_test.go` tests, including duplicate-key rejection.

Opening as draft to run CI.


rudraymehra:
Heyy @kakkoyun  I've been studying the rule-parsing pipeline in depth and independently prototyped this same refactor to understand it, with equivalence checks against main: golden test data regenerates with zero diff, and matched.json for demo/app/basic and test/apps/gincustom comes out semantically identical before/aftr. A few findings from that exercise that might be worth pinning as tests here:

- Legacy flat dispatch: a rule carrying both `func` and `raw` must keep `raw` winning, or flat inject_code rules silently become func rules.
- Stray top-level flat fields on a structured rule currently seed every expanded rule (precedence: common fields < where selectors < do payload, via maps.Copy in normalize.go). Worth deciding/pinning either way.
- `otelc pin --validate` prunes instrumentation imports on any parse error (pin.go:424) with strictr modifier-name validation, a typo'd third-party rule file becomes silently disabled instrumentation. A loud error there might be safer.
- Rule order in matched.json is map-iteration-random today (two consecutive setup runs differ); sorting entry names makes it deterministic cheaply.

This is conflicting with main after #846 tag changes  happy to help with the rebase or the test coverage if useful.

PR #722, opened 2026-07-20
fix(tool): reject malformed version ranges in instrumentation rules

### LINKED ISSUE #717: fix(tool): reject malformed version ranges in instrumentation rules
### Describe the bug

Malformed `version` ranges in instrumentation rules are **accepted at load time** but produce **incorrect matching behavior** with no error or warning.

The docs specify the format as `start_inclusive,end_exclusive` (e.g. `v0.11.0,v0.12.0`), and empty/invalid `target` values are rejected loudly at parse time - but **`version` is never validated**. Typos cause instrumentation to silently not apply, or to apply far too broadly.

**Verified behavior** (`tool/util/shared.go` → `VersionInRange`):

| Rule `version` | Test version | Result | Problem |
|----------------|--------------|--------|---------|
| `v1.0.0,v2.0.0` | `v1.5.0` | ✅ match | Correct |
| `v1.0.0,` (trailing comma) | `v1.5.0` | ❌ no match | Silent miss - instrumentation never applies |
| `,v2.0.0` (missing lower bound) | `v0.5.0` | ✅ match | Overly broad -matches versions that should be excluded |
| `v1.0.0,v2.0.0,v3.0.0` (extra comma) | `v1.5.0` | ❌ no match | Silent miss - user likely meant a valid range |

Rules with these malformed values load without error because `tool/internal/setup/match.go` validates `target` but passes `version` straight to `VersionInRange()`.

### Environment

- OS: Linux
- Go Version: 1.25.0
- Version: `otelc` v1.0.1 / main branch

### To Reproduce

1. Run this small program using `VersionInRange` from `tool/util`:

   ```go
   package main

   import (
       "fmt"
       "go.opentelemetry.io/otelc/tool/util"
   )

   func main() {
       cases := []struct{ version, versionRange string }{
           {"v1.5.0", "v1.0.0,v2.0.0"},
           {"v1.5.0", "v1.0.0,"},
           {"v0.5.0", ",v2.0.0"},
           {"v1.5.0", "v1.0.0,v2.0.0,v3.0.0"},
       }
       for _, c := range cases {
           fmt.Printf("VersionInRange(%q, %q) = %v\n", c.version, c.versionRange, util.VersionInRange(c.version, c.versionRange))
       }
   }
   ```

2. Observe output:
   ```
   VersionInRange("v1.5.0", "v1.0.0,v2.0.0") = true
   VersionInRange("v1.5.0", "v1.0.0,") = false
   VersionInRange("v0.5.0", ",v2.0.0") = true
   VersionInRange("v1.5.0", "v1.0.0,v2.0.0,v3.0.0") = false
   ```

3. Create a rule YAML with `version: "v1.0.0,"` or `version: ",v2.0.0"` - it loads without error.

4. Compare with empty `target`, which **is** rejected at load time in `tool/internal/setup/match.go` (lines 84–86).

### Expected behavior

Invalid `version` ranges should be **rejected at rule load time** with a clear error (consistent with how invalid `target` values are handled), for example:

- Trailing comma: `v1.0.0,`
- Missing lower bound: `,v2.0.0`
- More than one comma: `v1.0.0,v2.0.0,v3.0.0`
- Empty version components

At minimum, malformed ranges should not silently cause instrumentation to never match or to match all versions.

### Additional context

- **Docs:** `docs/rules.md` line 74 defines format as `start_inclusive,end_exclusive`
- **Code:** `tool/util/shared.go` lines 127–146 (`VersionInRange`)
- **Load path:** `tool/internal/setup/match.go` line 109 (`matchVersion`) - no validation before use
- **Existing tests:** `tool/util/shared_test.go` covers valid ranges only; no malformed-range cases
- **Suggested fix:** Add `ValidateVersionRange()` at rule load time (alongside existing `ValidateTarget()`), plus unit tests for malformed inputs

**Impact:** Version gating is a safety mechanism for instrumentation rules. Silent failures make typos very hard to debug - especially in CI/CD where rules are applied automatically.

ADITYA-CODE-SOURCE:
## Summary
- validate rule `version` values at load time instead of silently accepting malformed ranges
- preserve the repo's existing single-version lower-bound syntax like `version: "v1.44.0"`
- reject malformed comma-based ranges such as trailing/missing bounds and extra commas
- add focused tests for both the validator and the rule-loading path

Fixes #717.

## Test plan
- `go test ./tool/util ./tool/internal/setup`


kakkoyun (changes requested):
We should be more resilient here instead of rejecting. Like removing  the trailing and preceding commas.

Until we have the linter in-place. cc @amazingakai

ADITYA-CODE-SOURCE:
> We should be more resilient here instead of rejecting. Like removing the trailing and preceding commas.
> 
> Until we have the linter in-place. cc @amazingakai

Thanks understood.

I see the preferred direction is to normalize/rescue malformed ranges for now  and `#724` is already covering the same issue. I’m happy to pause this PR unless you’d prefer me to rework it that way.

amazingakai:
@kakkoyun imo, trying to guess or silently fix malformed version ranges would likely create more confusion than rejecting them. I think we should keep the validation strict here and reject malformed input.

The linter would a be separate layer and not called internally by otelc, so this validation should be kept.

kakkoyun (approved):

PR #729, opened 2026-07-21
feat: populate more template variables for directive rule

### LINKED ISSUE #702: feat: template variables for `signature_contains`
## Problem Context
The `signature` and `signature_contains` rules allow us to filter for functions that match certain parameters and return values, but does not allow us to access or use these parameters. In the `do` section of a rule, we do not currently have a way to update the values of a parameter or change the return value of the matched function.

Orchestrion currently supports `{{ .Function.ArgumentOfType }}`, `{{ .Function.Argument n }}` , and several other template variables as documented [here](https://datadoghq.dev/orchestrion/contributing/aspects/code-templates/#the-function-method). A rule that exists in Orchestrion but cannot be replicated yet in `otelc` is the following in `net/http`:

```
{{- $ctx := .Function.ArgumentOfType "context.Context" -}}
{{- $req := .Function.ArgumentOfType "*net/http.Request" }}
{{- if $ctx -}}
	__ddtrace_client.{{ .AST.Fun.Name }}(
		{{ $ctx }},
		{{ range .AST.Args }}{{ . }},
		{{ end }}
	)
{{- else if $req -}}
	__ddtrace_client.{{ .AST.Fun.Name }}(
		{{ $req }}.Context(),
		{{ range .AST.Args }}{{ . }},
		{{ end }}
	)
{{- else -}}
	{{ . }}
{{- end -}}
```

([full net/http advices here](https://datadoghq.dev/orchestrion/docs/dd-trace-go/v2/contrib-net-http/#getheadpostpostform))

## Request
Similar to the template variable `{{ FuncName }}` already supported by the directive rule, we should have something like `{{ FuncArgument 0 }}` to grab the first (0-indexed) argument of the matched function, `{{ FuncReturn 0 }}` to get the first (0-indexed) return value of the function, etc. 

### Alternatives
We could also consider `{{ Func.ArgumentOfType type }}` and `{{ Func.ReturnOfType type }}` as possible template variables. This could help solve a different variety of problems, but can also be more restrictive if there are multiple parameters/return values of the same type.

**NOTE**
It would also be extremely useful to have access to template variables like `{{ FuncName }}` from all rules, not just `directive`. 

hannahkm:
**NOTE:** This is part of a stack of PRs meant to bring more function template variables to different rules in otelc. Full stack is as follows:

**NOTE 2:** This PR adds a `CHANGELOG.md` file due to a breaking change in the FuncName template variable.

1. [This PR]
2. [feat: make template variables work for inject_code](https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/pull/1)
3. [feat: apply template variables to wrap_call](https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/pull/2)
4. [feat: ArgumentOfType and CallArgument on wrap_call](https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/pull/4)

The following PRs will be opened after this one has been merged.

## Description

1. Migrates work to `text/template`.
2. This allows easier support for new function template variables, namely `FuncArgument N`, `FuncArgumentCount`, `FuncArgumentOfType`, `FuncReturnOfType`,  `FuncReturn N`, `FuncReturnCount`, and `DirectiveArgs`. 
3. ONLY does the above for the directive rule. Subsequent PRs expand to other rules.

## Motivation

Fixes #702 

---

## Checklist

- [ ] Tests pass: `make test`


amazingakai:
@hannahkm Needs a rebase

docs/rules.md
```diff
@@ -946,19 +946,33 @@ This rule instruments functions annotated with a magic comment (a "directive") b
...
 Top-level `imports` (map[string]string, optional): Additional imports needed by the injected code. Same format as [Top-level fields](#top-level-fields).
 
 **Template Placeholders:**
 
-| Placeholder    | Replaced with                      |
-| -------------- | ---------------------------------- |
-| `{{FuncName}}` | The name of the annotated function |
+| Placeholder                     | Replaced with                                                               |
+| ------------------------------- | --------------------------------------------------------------------------- |
+| `{{.FuncName}}`                 | The name of the annotated function                                          |
```
kakkoyun, line 961: This is a breaking change. Even minor so we need to document this on the `CHANGELOG.md`

And we don't have it yet, so you have to create it.

kakkoyun (commented):
Looking good.

We should remove `fasttemplate` as dependency and make sure we use standard library template everywhere.

And as I suggested we need a changelog. And we need to make sure all the existing integrations work (I guess CI would tell us otherwise).

tool/internal/instrument/func_template.go
```diff
@@ -0,0 +1,155 @@
...
+
+// FuncArgumentOfType returns the identifier of the first parameter (excluding
+// the receiver) whose type matches typeStr, or "" if none match. Template
+// usage: {{ .FuncArgumentOfType "context.Context" }}
+func (d *funcTemplateData) FuncArgumentOfType(typeStr string) (string, error) {
+	args := d.arguments() // ensures synthetic names are assigned first
+	idx := 0
+	for _, field := range d.funcDecl.Type.Params.List {
+		for range field.Names {
+			matched, err := ast.MatchesTypeName(field.Type, typeStr)
```
amazingakai, line 101: Why not pass the import alias map here? `MatchesTypeName` passes `nil` for the imports, which means it falls back to path-tail matching. That works for single-segment packages like `context` but breaks for aliased imports (e.g. `import althttp "net/http"`) or multi-segment paths that share a tail (e.g. `text/template` vs `html/template`).

tool/internal/instrument/func_template.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// FuncArgumentOfType returns the identifier of the first parameter (excluding
+// the receiver) whose type matches typeStr, or "" if none match. Template
+// usage: {{ .FuncArgumentOfType "context.Context" }}
+func (d *funcTemplateData) FuncArgumentOfType(typeStr string) (string, error) {
+	args := d.arguments() // ensures synthetic names are assigned first
+	idx := 0
+	for _, field := range d.funcDecl.Type.Params.List {
+		for range field.Names {
```
amazingakai: Instead of putting this in a loop, we can do `idx += len(field.Names)`

tool/internal/instrument/apply_directive.go
amazingakai, line 46: Is this needed? If the template references return values (e.g. `{{.FuncReturn 0}}`), `funcTemplateData.returns()` already calls `collectReturnValues` which names them before we get here. and if the template doesn't reference returns, the injected code doesn't use return names either, so naming them here does nothing useful.

tool/internal/instrument/apply_func.go
```diff
@@ -49,29 +49,56 @@ func findJumpPoint(jumpIf *dst.IfStmt) *dst.BlockStmt {
...
+			name := fmt.Sprintf("%s%d", prefix, *idx)
+			*idx++
+			if !taken[name] {
+				taken[name] = true
+				return name
+			}
+		}
+	}
+}
+
```
amazingakai, line 79: This same fix was discussed in https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/799#discussion_r3689788318

I'd suggest dropping this from this PR and doing it properly in a followup. The fix handles signature-level collisions, but doesn't account for global identifiers or locals in the function body. A synthetic name like `_ignoredParam1` could still shadow a package-level variable.

See the issue I opened for this: #1014, we should probably use CRC32 hashes in the identifier names instead to make collisions nearly impossible.

tool/internal/instrument/func_template.go
```diff
@@ -0,0 +1,155 @@
...
+			matched, err := ast.MatchesTypeName(field.Type, typeStr)
+			if err != nil {
+				return "", err
+			}
+			if matched {
+				return args[idx], nil
+			}
+			idx++
+		}
```
amazingakai, line 109: All names in a single `dst.Field` share the same type, so looping over `field.Names` just to call `MatchesTypeName` repeatedly is unnecessary. Can replace the inner loop with a single match check and `idx += len(field.Names)` when it doesn't match:

```suggestion
		matched, err := ast.MatchesTypeName(field.Type, typeStr)
		if err != nil {
			return "", err
		}
		if matched {
			return args[idx], nil
		}
		idx += len(field.Names)
```

tool/internal/instrument/func_template.go
```diff
@@ -0,0 +1,155 @@
...
+		for range field.Names {
+			matched, err := ast.MatchesTypeName(field.Type, typeStr)
+			if err != nil {
+				return "", err
+			}
+			if matched {
+				return rets[idx], nil
+			}
+			idx++
+		}
```
amazingakai, line 133: Same fix applies here as well:

```suggestion
		matched, err := ast.MatchesTypeName(field.Type, typeStr)
		if err != nil {
			return "", err
		}
		if matched {
			return rets[idx], nil
		}
		idx += len(field.Names)
```

amazingakai (changes requested):
Solid PR! I have left some review comments.

This way of referencing function args/returns by template variable is nice. I see in the PR description that there's a planned followup ([hannahkm/opentelemetry-go-compile-instrumentation#1](https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/pull/1)) to expand this to `inject_code` (raw rules) as well, which is great. Currently those rules have to hardcode `_unnamedRetVal0` (e.g. in `instrumentation/runtime/otelc.yaml`).

Once raw rules use template variables too, we'll be able to fix the synthetic name shadowing bug properly. I've opened an issue for that here: #1014

tool/internal/instrument/func_template.go
hannahkm: Good catch! I think this was a leftover from before I added `ArgumentOfType` -- fixed in [29b1952](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/729/commits/29b195278d845a623b0e835cf0422719a3accee6)

tool/internal/instrument/apply_directive.go
hannahkm: True, removed in [c124fc8](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/729/commits/c124fc83a56c837fc216e1366d4498ac06e4ccd5)

tool/internal/instrument/apply_func.go
hannahkm: Good idea, I'll revert the commit and open a new [PR](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1035)! I'll open it up for review once I'm done there.

tool/internal/ast/typename.go
```diff
@@ -108,6 +108,28 @@ func fieldListContainsType(fields *dst.FieldList, typeStr string, imports map[st
...
+}
+
+// ImportAliasMap builds a map from the local identifier used to reference an
+// imported package within file to that package's real import path. See imports on
+// MatchesTypeName for how to use the result, and importAliasMap for the
+// resolution rules.
+// Returns nil when file is nil.
+func ImportAliasMap(file *dst.File) map[string]string {
+	return importAliasMap(file)
+}
```
amazingakai: Why not just make `importAliasMap` public?

tool/internal/instrument/func_template.go
amazingakai: Non-blocking but this is still unaddressed?

amazingakai (approved):
LGTM.

tool/internal/instrument/func_template.go
hannahkm: Ah sorry, I thought I had applied the suggestion 💀, I probably overwrote it by accident. Let me fix these now!

amazingakai:
@hannahkm This needs an update after #1035 I think

hannahkm:
@amazingakai Merging in the PR required a few extra changes to make it work with this branch (adding hash identity to directive rules to pass into the newly added arguments) -- do you want to take another look at it? Thanks!

amazingakai (approved):
I did a quick review, looks good.

PR #730, opened 2026-07-21
docs: update external configuration guide for shipped otelc pin flow

ADITYA-CODE-SOURCE:
## Summary

Updates `docs/external-configuration.md` to reflect that `otelc pin` is now available.

The page still described `otelc pin` as planned, but the command has since landed and can create or update `otel.instrumentation.go` and run `go mod tidy`.

This also avoids duplicating the longer note about current limitations around committed generated instrumentation files by linking to the existing Getting Started guidance.

## Validation

Docs-only change. No code behavior changed.


txabman42 (approved):
TY ❤️

PR #731, opened 2026-07-21
fix(instrument): write nil through the pointer in SetParam/SetReturnVal

### LINKED ISSUE #726: HookContext.SetParam/SetReturnVal(idx, nil) silently drops the write and can panic on later reads
### Describe the bug

I was writing a hook that clears a returned `error` in some cases (`ictx.SetReturnVal(errIdx, nil)`), and noticed the error was still coming back non-nil afterward. Dug into the generated `HookContextImpl` code and found `SetParam`/`SetReturnVal` basically no-op when you pass `nil` - they don't touch the real value, and if you read the same slot again with `GetParam`/`GetReturnVal` it panics.

Here's the generated `SetReturnVal` for a function returning `(float32, error)` (from `tool/internal/instrument/impl.tmpl`):

```go
func (c *HookContextImpl) SetReturnVal(idx int, val interface{}) {
	if val == nil {
		c.returnVals[idx] = nil
		return
	}
	switch idx {
	case 1:
		*(c.returnVals[1].(*error)) = val.(error)
	}
}
```

`c.returnVals[1]` holds a `*error` pointing at the real return value. The non-nil branch dereferences it and assigns through, which is correct. But the `val == nil` branch just stomps `c.returnVals[1]` itself with a bare `nil` so it never touches what the pointer was pointing at, and now the slot isn't a `*error` anymore, so the next `GetReturnVal(1)` (which does `c.returnVals[1].(*error)`) panics with `interface conversion: interface {} is nil, not *error`.

So two problems from one line: the actual error value never gets cleared, and the context object is left in a state where reading it back blows up. The panic gets swallowed by the trampoline's recover, but the hook just silently fails.

Nothing in the bundled instrumentations currently calls `SetParam`/`SetReturnVal` with `nil`, so this hasn't shown up in any of the shipped packages. I hit it writing a custom hook against the public `HookContext` API.

### To Reproduce

Trimmed it down to just the part that matters (same shape as the generated code, minus the switch):

```go
package main

import "fmt"

type Ctx struct{ returnVals []interface{} }

func (c *Ctx) SetReturnVal(idx int, val interface{}) {
	if val == nil {
		c.returnVals[idx] = nil
		return
	}
	*(c.returnVals[0].(*error)) = val.(error)
}
func (c *Ctx) GetReturnVal(idx int) interface{} { return *(c.returnVals[0].(*error)) }

func main() {
	ret := fmt.Errorf("original error")
	c := &Ctx{returnVals: []interface{}{&ret}}

	c.SetReturnVal(0, nil)
	fmt.Println(ret)
	fmt.Println(c.GetReturnVal(0))
}
```

```
$ go run main.go
original error
panic: interface conversion: interface {} is nil, not *error
```

Or in an actual instrumented setup: hook a function that returns an `error`, in the After hook call `ictx.SetReturnVal(errIdx, nil)`, then check the real return value afterward — still whatever it was before.

### Expected behavior

Calling `SetReturnVal(idx, nil)` (or `SetParam`) should actually clear the real value, same as the non-nil path does but writing the zero value through the pointer instead of replacing the slot. A later `GetReturnVal`/`GetParam` on that index should return `nil`, not panic.

Anshumancanrock:
## Summary

`SetParam`/`SetReturnVal` on the generated `HookContextImpl` no-op on `nil`: instead of writing through the stored pointer, they replace the slot itself, so the real value never changes and the next `GetParam`/`GetReturnVal` on that index panics on the type assertion.

Fixes #726

Move the nil check into each generated switch case, where the concrete type is known, and zero through the pointer instead of replacing the slot:

```go
case 1:
    if val == nil {
        *(c.returnVals[1].(*error)) = *new(error)
    } else {
        *(c.returnVals[1].(*error)) = val.(error)
    }
```


tool/internal/instrument/trampoline.go
```diff
@@ -682,23 +682,29 @@ func (ip *InstrumentPhase) implementHookContext(t *rule.InstFuncRule) {
...
-	te := ast.TypeAssertExpr(ie, ast.DereferenceOf(t))
-	pe := ast.ParenExpr(te)
-	de := ast.DereferenceOf(pe)
-	val := ast.Ident(trampolineValIdentifier)
-	assign := ast.AssignStmt(de, ast.TypeAssertExpr(val, t))
 	if ast.IsInterfaceType(t) {
-		assign = ast.AssignStmt(ie, val)
+		ie := ast.IndexExpr(se, ast.IntLit(idx))
+		assign := ast.AssignStmt(ie, ast.Ident(trampolineValIdentifier))
+		return ast.SwitchCase(ast.Exprs(ast.IntLit(idx)), ast.Stmts(assign))
```
amazingakai, line 690: Unless I missed it, we do not have a golden test which covers this path. It'd be nice to have (maybe you can have it in some existing golden test?)

tool/internal/instrument/trampoline.go
```diff
@@ -682,23 +682,29 @@ func (ip *InstrumentPhase) implementHookContext(t *rule.InstFuncRule) {
...
+	}
+	nonNilAssign := ast.AssignStmt(target(), ast.TypeAssertExpr(ast.Ident(trampolineValIdentifier), t))
+	tClone := util.AssertType[dst.Expr](dst.Clone(t))
+	nilAssign := ast.AssignStmt(target(), ast.DereferenceOf(ast.CallTo("new", nil, ast.Exprs(tClone))))
+
+	ifStmt := &dst.IfStmt{
+		Cond: &dst.BinaryExpr{X: ast.Ident(trampolineValIdentifier), Op: token.EQL, Y: ast.Nil()},
+		Body: ast.Block(nilAssign),
+		Else: ast.Block(nonNilAssign),
+	}
```
amazingakai, line 708: Now we're setting the value to its zero value when the provided value is nil and the underlying type is not nilable. This makes sense, but I think we should document this behavior somewhere.

amazingakai (approved):
Good catch ❤️ 

I left some suggestions

Please also work on fixing the CI

Anshumancanrock:
hii @amazingakai , thanks for the review. I addressed both of your comments.

darccio (changes requested):
Leaving a change request.

tool/internal/instrument/trampoline.go
```diff
@@ -682,23 +682,31 @@ func (ip *InstrumentPhase) implementHookContext(t *rule.InstFuncRule) {
 }
 
 func setValue(field string, idx int, t dst.Expr) *dst.CaseClause {
```
darccio, line 684: This fix is verified only by "generated source compiles + matches golden text"
(TestInstrumentation_Integration in instrument_test.go). Nothing exercises the
generated code at runtime — no test calls the generated SetParam(idx, nil)
followed by GetParam(idx) to confirm the actual symptom from #726 (panic on
the type assertion, and the real value never changing) is fixed.

None of the real instrumentation hooks in this repo pass nil to SetParam/
SetReturnVal either, so the nil path is currently exercised by neither the
golden suite nor any instrumentation call site.

Suggest adding a small runtime test — e.g. instantiate a generated
HookContextImpl (or a minimal hand-written stand-in with the same shape) for
a case with a concrete-typed param/return, call SetParam(idx, nil), and assert
GetParam(idx) returns the zero value without panicking. This would have
caught the original bug directly and guards against regressing it.

Anshumancanrock:
@darccio you're right, the golden suite only compiles the code, so the `nil` path was never actually run. Added `TestHookContextSetNilWritesZeroValue`: stands in a `HookContextImpl` of the same shape (`*int` param, `*error` return), calls `Set*(idx, nil)`, and checks `Get*` returns the zero value without panicking. Went with the stand-in since the real impl is generated per-function and can't be imported. Verified it reproduces the original `interface conversion: interface {} is nil, not *error` panic with the old code and passes with the fix.

PLease take a looK now !

darccio (approved):

PR #733, opened 2026-07-21
fix(instrument): no-op call rules on files without matching calls

### LINKED ISSUE #732: wrap_call rules fatal on multi-file packages when only some files contain matching calls
### Describe the bug

`wrap_call` instrumentation rules crash the build when the target package has **multiple `.go` files** and only some of those files contain the matching function call.

During setup, call rules are attached to **every** source file in the package. During instrumentation, `applyCallRule` fatals if a file has no matching call - even though `match.go` documents that files without matches should be a no-op.

**Error:**
```
Assertion failed: call rule did not match any call
```

This breaks real-world packages where call sites live in one file (e.g. `handler.go`) and helper code lives in another (e.g. `utils.go`).

---

### Environment

- **OS:** Linux (also affects any OS; not platform-specific)
- **Go Version:** 1.25.0
- **Version:** `main` (latest as of Jul 2026)

---

### To Reproduce

1. Create a Go package with two files:

   **`calls.go`**
   ```go
   package main

   import "unsafe"

   func CallSizeof() {
       _ = unsafe.Sizeof(42)
   }
   ```

   **`other.go`**
   ```go
   package main

   func Helper() {
       println("no matching calls here")
   }
   ```

2. Add an instrumentation rule targeting `unsafe.Sizeof`:

   **`rules.yml`**
   ```yaml
   wrap_sizeof:
     target: main
     where:
       function_call: unsafe.Sizeof
     do:
       - wrap_call:
           replace: "Wrapper({{ . }})"
   ```

3. Build with `otelc`:
   ```bash
   otelc go build .
   ```

4. See error  build fatals with:
   ```
   Assertion failed: call rule did not match any call
   ```

**Minimal unit-level reproduction** (same root cause):

```go
// File with fmt.Println — no unsafe.Sizeof
err := applyCallRule(ctx, callRule, file)
// → Fatal: "call rule did not match any call"
```

**Why existing tests miss this:** all `call-rule-*` golden fixtures use a single `source.go` file. Setup adds call rules to all files in a package, but the test harness only models one file.

---

### Expected behavior

- Build should **succeed**
- Files containing matching calls (e.g. `calls.go`) should be instrumented
- Files with no matching calls (e.g. `other.go`) should be **skipped silently** (no-op), as documented in `match.go`:

  > *"Files without matching calls are a no-op in applyCallRule."*

---

### Additional context

**Contradiction in the codebase:**

| Location | Behavior |
|----------|----------|
| `tool/internal/setup/match.go:370–376` | Adds call rules to **all** source files in the package |
| `tool/internal/instrument/apply_call.go:33` | `util.Assert(...)` fatals if **no** call matched in a file |

**Verified locally:**
- `preciseMatching` with 2 files → call rule added to **both** files
- `applyCallRule` on a file without matching calls → **fatal** (not no-op)

Jay2006sawant:
## Description

When a package has multiple `.go` files, the setup phase attaches `wrap_call` rules to every source file in the package. The instrument phase previously fatalled on files that contained no matching call sites:

```
Assertion failed: call rule did not match any call
```

This change makes `applyCallRule` return early (no-op) when neither `append_args` nor `replace` modified the file, matching the behavior documented in `match.go`.

Tests added:
- `TestApplyCallRule_NoMatchIsNoOp` - unit test for the no-op path
- `TestPreciseMatching_CallRuleAddedToAllFiles` - setup attaches call rules to all files
- `TestInstrument_CallRuleMultiFilePackage` - end-to-end multi-file package build

## Motivation

Real-world packages often split call sites across files (e.g. `handler.go` has the instrumented call, `utils.go` does not). Call rules must not crash the build on files without matching calls; those files should be skipped silently while files with matches are still instrumented.

Fixes #732 

---

## Checklist

- [x] Tests pass: `make test` (instrument + setup packages verified locally)

### DIFF
--- a/tool/internal/instrument/instrument_test.go
+++ b/tool/internal/instrument/instrument_test.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"log/slog"
 	"os"
 	"os/exec"
@@ -54,6 +53,7 @@ const (
 	importPathFileName = "importpath"
 	mainGoFileName     = "main.go"
 	mainTestFileName   = "main_test.go"
+	otherGoFileName    = "other.go"
 	mainPackage        = "main"
 	buildID            = "foo/bar"
 	compiledOutput     = "_pkg_.a"
@@ -100,14 +100,22 @@ func runTest(t *testing.T, testName string) {
 	require.NoError(t, util.CopyFile(testSpecificSource, sourceFile),
 		"missing %s for test %q at %s", srcName, testName, testSpecificSource)
 
+	packageFiles := copyExtraPackageFiles(t, testName, tempDir, sourceFile)
+
 	importPath := testImportPath(t, testName)
-	ruleSet := loadRulesYAML(t, testName, sourceFile, importPath, isTest)
+	ruleSet := loadRulesYAML(t, loadRulesParams{
+		testName:     testName,
+		sourceFile:   sourceFile,
+		packageFiles: packageFiles,
+		importPath:   importPath,
+		isTest:       isTest,
+	})
 	writeMatchedJSON(ruleSet)
 
 	testcaseDir := filepath.Join(testdataDir, goldenDir, testName)
 	helpers := buildTestcaseHelpers(ctx, t, testcaseDir)
 
-	args := compileArgs(tempDir, helpers, importPath, sourceFile)
+	args := compileArgs(tempDir, helpers, importPath, packageFiles...)
 	err := Toolexec(ctx, args, false)
 
 	if testName == invalidReceiver {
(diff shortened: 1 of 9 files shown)


tool/internal/instrument/instrument_test.go
```diff
@@ -119,6 +120,82 @@ func runTest(t *testing.T, testName string) {
...
+
+	mainContent, err := os.ReadFile(mainFile)
+	require.NoError(t, err)
+	otherContent, err := os.ReadFile(otherFile)
+	require.NoError(t, err)
+
+	assert.Contains(t, string(mainContent), "Wrapper(unsafe.Sizeof")
+	assert.NotContains(t, string(otherContent), "Wrapper(")
+}
+
```
amazingakai: Maybe a simple golden test instead to show no match is a no-op?

amazingakai (commented):
@Jay2006sawant good catch, I only have one comment.

tool/internal/instrument/instrument_test.go
Jay2006sawant: @amazingakai Thanks for the review! Done in acce063. Replaced the inline test with a call-rule-multi-file golden fixture that shows main.go is instrumented and other.go stays unchanged when the call rule is attached to both files.

Jay2006sawant: Let me know if anything else should change.

amazingakai (approved):

PR #738, opened 2026-07-22
fix(instrument): avoid _ignoredParam name collision in generated sign…

### LINKED ISSUE #736: Instrumenting a func with a blank named return + unnamed param/receiver produces uncompilable code
## Description

If the tool instruments a function that has **both**:

- an unnamed or blank (`_`) parameter, or an unnamed receiver, **and**
- a blank (`_`) *named* return value

…the generated signature ends up with `_ignoredParam0` in two binding
positions, and the instrumented package won't compile:

```
_ignoredParam0 redeclared in this block
```

Worse, the error points at generated code the rule author never wrote.

### Reason:

In `tool/internal/instrument/apply_func.go`, `insertTJump` runs both collectors
on the same `funcDecl`:

```go
retVals := collectReturnValues(funcDecl) // blank named return -> _ignoredParam0
args    := collectArguments(funcDecl)    // unnamed param/recv  -> _ignoredParam0
```

Both rename blank/unnamed bindings using the same `ignoredParam` ("_ignoredParam")
prefix, and each keeps its **own** `idx` that starts at 0. Since the receiver,
params, and named results all share one scope in Go, the two `_ignoredParam0`
bindings collide. With multiple blanks on each side you get `_ignoredParam1`,
`_ignoredParam2`, … colliding too.

Unnamed *returns* dodge this because they use a separate prefix
(`_unnamedRetVal`) — so only a blank `_` **named** return trips it.

## Reproduce

All three of these turn a green `go build` into a compile error once instrumented:

```go
func F(_ int) (_ error)          { return nil }
func F(int) (_ error)            { return nil }
func (T) M() (_ error)           { return nil }
```

Rendering the rewritten AST and feeding it to `go/types` (or plain `go build`)
confirms it:

```
func F(_ignoredParam0 int) (_ignoredParam0 error)   // _ignoredParam0 redeclared
func (_ignoredParam0 T) M() (_ignoredParam0 error)  // _ignoredParam0 redeclared
```

Control: `func F(int) (int)` renders as
`func F(_ignoredParam0 int) (_unnamedRetVal0 int)` and compiles fine. That's the
boundary that shows the two prefixes are what save the common case.

Anshumancanrock:
## Description

Instrumenting a function with both a blank/unnamed param (or unnamed receiver)
and a blank (`_`) named return generated uncompilable code:

```
_ignoredParam0 redeclared in this block
```

`collectArguments` and `collectReturnValues` both renamed blank bindings with the
same `_ignoredParam` prefix, each counting from 0. Since a receiver, params, and
named results share one Go scope, the two `_ignoredParam0` bindings collided.

### Fix

Fixes #736

Blank named returns now use their own `_ignoredRetVal` prefix, mirroring the
existing `_unnamedRetVal` handling for unnamed returns. Params and returns can no
longer produce the same name.

## Tests

- Unit regression that type-checks the rewritten signature via `go/types`.
- New golden fixture that compiles a blank-param + blank-named-return function
  through the real toolchain (fails without the fix).

### DIFF
--- a/tool/internal/instrument/apply_func_test.go
+++ b/tool/internal/instrument/apply_func_test.go
@@ -5,8 +5,16 @@ package instrument
 
 import (
 	"context"
+	goast "go/ast"
+	"go/importer"
+	"go/parser"
+	"go/token"
+	"go/types"
+	"strings"
 	"testing"
 
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -121,7 +129,7 @@ func TestCollectReturnValues(t *testing.T) {
 		{
 			name:     "underscore return values",
 			src:      "package main\nfunc F() (_ int, _ string) { return }",
-			expected: []string{"_ignoredParam0", "_ignoredParam1"},
+			expected: []string{"_ignoredRetVal0", "_ignoredRetVal1"},
 		},
 	}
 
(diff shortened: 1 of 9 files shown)


tool/internal/instrument/apply_func_test.go
```diff
@@ -133,3 +141,88 @@ func TestCollectReturnValues(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			file, funcDecl := parseFileFunc(t, tt.src)
+
+			// Mirror insertTJump: returns are collected first, then arguments.
+			retVals := collectReturnValues(funcDecl)
+			args := collectArguments(funcDecl)
+
+			seen := make(map[string]struct{})
+			for _, name := range append(append([]string{}, retVals...), args...) {
```
amazingakai, line 183: This is slightly unreadable, can we use `slices.Concat` here?

amazingakai (approved):
Good catch, no major comments from me.

tool/internal/instrument/apply_func_test.go
kakkoyun: We should run https://golangci-lint.run/docs/linters/configuration/#modernize as part of our CI.

amazingakai: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/747

Issue #709, opened 2026-07-20
LFX Mentorship 2026 Term 3 — Zero-code AI Agent observability for otelc


NameHaibinZhang:
## Summary

This is the upstream project-idea issue for a **CNCF LFX Mentorship 2026 Term 3** program under OpenTelemetry, targeting `opentelemetry-go-compile-instrumentation` (`otelc`) — OpenTelemetry's compile-time, zero-code instrumentation for Go.

The goal is to design and build an **AI provider adapter interface** in `otelc` that emits the `gen_ai.*` semantic conventions (spans, metrics, events) through a single shared abstraction, so AI providers and libraries plug into a common contract instead of being instrumented ad hoc. The already-merged OpenAI instrumentation (#604) is migrated onto this interface as the reference, then extended to `anthropic-sdk-go`, **Gemini, MCP, and LangChain**. MCP/LangChain require non-HTTP hooks (following the Kafka/Redis instrumentation patterns) rather than the HTTP-based path. This design-first framing gives the mentee real architecture and problem-solving work rather than repetitive per-library instrumentation.

## Motivation

Community demand is concrete and vendor-neutral: the OpenAI instrumentation (#604) closed community issue #327, and the request pipeline is active (#706 for aws-sdk-go-v2, #553 MongoDB shipped). AI Agent frameworks are a fast-growing gap for zero-code Go observability.

## Deliverables (week-by-week)

A per-week plan where each milestone ships a **runnable demo + green CI** so progress is visible at a glance: adapter interface + OpenAI migration → `anthropic-sdk-go` → Gemini → MCP → LangChain, each with instrumentation rules, tests under the paired `test/` directory, and an example app emitting `gen_ai.*` spans.

## Privacy posture

Message bodies are **off by default** and strictly opt-in via `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` — nothing is captured unless the operator turns it on.

## Mentors

- Haibin Zhang (@NameHaibinZhang)
- Xabier Martinez (@txabman42)

PRAteek-singHWY:
Hi @NameHaibinZhang @txabman42 , really glad to see this one. I have been reading through otelc a bit before commenting and wanted to share where I landed, in case it is useful.

[...]

PR #740, opened 2026-07-22
feat(openai): implement opt-in message content capture

Aryanbhargava18:
## Proposed changes

This PR implements opt-in message content capture natively within the `openai-go` instrumentation (covering v1, v2, and v3). 

Following @NameHaibinZhang's guidance that content capture is a cross-cutting capability requiring tight scoping and zero-overhead defaults, this implementation:
- Strictly gates capture logic behind the `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` environment variable (cached via `sync.OnceValue` with a testable comparison helper).
- Safely handles truncation bounds (`ContentCaptureLimit` = 16 KiB) to prevent OOM panics on massive payloads (e.g., multimodal base64 image strings), joining text parts for multimodal content.
- Bounds streaming choice accumulation: caps distinct stream choice indices to 16 to prevent memory exhaustion from server-controlled indices.
- Unifies the `gen_ai.content.completion` event format across both standard responses and SSE chunked streaming.
- Preserves zero-overhead telemetry execution when the flag is disabled.

### Scope & Design Notes for #709
- **Embedding capture**: Embedding operations are deliberately excluded from content capture in this PR. Vector representations and embedding inputs are high-dimensional arrays that are ill-suited for span event attributes; their capture mechanics and policy will be designed as part of the broader cross-cutting abstraction in #709.
- **Span Events vs Log Records**: Captured prompt and completion payloads are currently emitted as Span Events per GenAI semantic conventions. As noted in review discussion, requests with extensive message history can record multiple large events; whether GenAI content should transition to Log Records or remain Span Events is flagged for decision during the generalization in #709.

## Types of changes
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist
- [x] I have read the [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/CONTRIBUTING.md) doc
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)


instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -34,7 +36,8 @@ type streamingReader struct {
 	opName        string
 	provider      string
```
NameHaibinZhang, line 37: **OOM risk**: `capturedStream` accumulates all streaming chunks without any memory bound. For long responses (e.g. code generation, translations), this can grow to tens of MBs. The PR description claims "safely handles truncation bounds (maxRequestBodySize) to prevent OOM panics" but that constant only limits the *request* body read (pre-existing logic) — there is no truncation applied to the streaming accumulation here.

Please add a size limit (e.g. `maxResponseBodySize`) and stop accumulating once exceeded, similar to how `logBuffer` is bounded elsewhere.

instrumentation/github.com/openai/openai-go/middleware_test.go
```diff
@@ -61,14 +61,14 @@ func TestOperationName(t *testing.T) {
 
 func TestParseChatRequest(t *testing.T) {
 	body := []byte(`{"model":"gpt-4","max_tokens":100,"temperature":0.7}`)
-	model, attrs := parseChatRequest(body)
+	model, attrs, _ := parseChatRequest(body)
```
NameHaibinZhang, line 80: The only test changes here are adapting existing tests to the new `parseChatRequest` return signature. There are no new tests that verify:
- Content is captured when the env var is enabled
- Content is NOT captured when the env var is disabled
- Streaming content accumulation works correctly
- Memory bounds are respected (once added)

Every new feature must have tests that validate the new behavior — this is a project requirement (see CONTRIBUTING.md).

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -20,6 +21,8 @@ import (
 	"go.opentelemetry.io/otelc/pkg/runtime"
 )
```
NameHaibinZhang, line 27: suggestion: `os.Getenv` evaluated at package init time makes this feature untestable in unit tests — you cannot toggle the behavior within a test without process-level env manipulation, which races under `t.Parallel()`.

Consider using a `sync.OnceValue` pattern or passing the config through a function parameter so tests can exercise both code paths.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -151,6 +155,14 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 			trace.WithSpanKind(trace.SpanKindClient),
 			trace.WithAttributes(spanAttrs...),
 		)
+
+		if captureContent && len(promptMessages) > 0 {
```
NameHaibinZhang: suggestion: `string(msg)` dumps the entire raw JSON message object (including `role`, `tool_calls`, `name`, etc.) as a single opaque string attribute. This produces low-quality telemetry that is hard to consume.

Consider extracting at least the `content` field from each message, or following the OTel GenAI semconv event body structure more closely.

NameHaibinZhang (dismissed):
Thanks for working on content capture! The feature direction is useful, but there are several issues that need to be addressed before this can be merged.

om7057 (approved):
@Aryanbhargava18 I've verified the changes and added a few comments.

om7057 (commented):
I was working on a same issue, and it perfectly resolves this @Aryanbhargava18.

Aryanbhargava18:
Just pushed a final hardening pass to cover a few extra edge cases I caught. I unified the logic to support the legacy /v1/completions API, refactored the extraction to safely preserve multimodal vision arrays, and ensured the 16KiB truncation strictly respects UTF-8 boundaries to prevent any malformed string panics.

Everything is synced across v1/v2/v3 and CI is perfectly green.

Aryanbhargava18:
Rebased on `main` and addressed the review feedback:
- Cached env flag lookup with `sync.OnceValue`
- Bounded stream capture to 16 KiB with safe UTF-8 boundary handling
- Added comprehensive unit tests for completion/chat streaming

@NameHaibinZhang let me know if anything else needs adjusting!

instrumentation/github.com/openai/openai-go/middleware_test.go
```diff
@@ -401,9 +432,200 @@ func TestParseCompletionRequest_Penalties(t *testing.T) {
...
+		Body: io.NopCloser(bytes.NewBufferString(`{}`)),
+	}
+
+	next := func(req *http.Request) (*http.Response, error) {
+		return resp, nil
+	}
+
+	middleware := otelMiddleware(func() bool { return true })
+	_, _ = middleware(req, next)
+}
```
darccio, line 591: This test has no assertions. The request body has no `model` field, so
`parseChatRequest` returns an empty model and the middleware returns before
it creates a span. A test without assertions cannot fail. The
`setupTestTracer` helper in `middleware_integration_test.go` registers a
span recorder as the active tracer, so the test can assert on the emitted
span.

```suggestion
func TestOtelMiddleware_ContentCapture_Enabled(t *testing.T) {
	sr := setupTestTracer(t)

	reqBody := `{"model":"gpt-4","messages":[{"role":"user","content":"hello"}]}`
	req, err := http.NewRequestWithContext(context.Background(), "POST", "https://api.openai.com/v1/chat/completions", bytes.NewBufferString(reqBody))
	require.NoError(t, err)

	resp := &http.Response{
		StatusCode: 200, Header: make(http.Header),
		Body: io.NopCloser(bytes.NewBufferString(`{}`)),
	}

	next := func(req *http.Request) (*http.Response, error) {
		return resp, nil
	}

	middleware := otelMiddleware(func() bool { return true })
	_, err = middleware(req, next)
	require.NoError(t, err)

	spans := sr.Ended()
	require.Len(t, spans, 1)
	require.Len(t, spans[0].Events(), 1)
	assert.Equal(t, "gen_ai.content.prompt", spans[0].Events()[0].Name)
	assert.Equal(t, "hello", spans[0].Events()[0].Attributes[0].Value.AsString())
}
```

instrumentation/github.com/openai/openai-go/middleware_test.go
```diff
@@ -401,9 +432,200 @@ func TestParseCompletionRequest_Penalties(t *testing.T) {
...
+		for _, event := range spans[0].Events() {
+			if event.Name == "gen_ai.content.completion" {
+				for _, attr := range event.Attributes {
+					if attr.Key == "gen_ai.completion" {
+						assert.Equal(t, "This is extracted content", attr.Value.AsString())
+					}
+				}
+			}
+		}
+	}
```
darccio: All assertions sit inside `if len(spans) > 0`. The test passes when capture
produces no span and no event, so it does not detect the regression that it
exists to catch. Fail the test when the span or the event is missing.

```suggestion
	require.Len(t, spans, 1)

	var found bool
	for _, event := range spans[0].Events() {
		if event.Name == "gen_ai.content.completion" {
			found = true
			for _, attr := range event.Attributes {
				if attr.Key == "gen_ai.completion" {
					assert.Equal(t, "This is extracted content", attr.Value.AsString())
				}
			}
		}
	}
	require.True(t, found, "expected a gen_ai.content.completion event")
```

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -324,7 +347,76 @@ func parseCompletionRequest(body []byte) (string, []attribute.KeyValue) {
...
+	var text string
+	if err := json.Unmarshal(content, &text); err == nil {
+		return text
+	}
+
+	marshaled, err := json.Marshal(content)
+	if err != nil {
+		return ""
+	}
+	return string(marshaled)
```
darccio, line 432: For multimodal requests, `content` is an array of typed parts. The fallback
emits the full compacted JSON, and the JSON includes the base64 image data.
The 16 KiB truncation then removes the text that follows the image data, so
the event loses the textual content. Joining the `text` parts keeps the
event readable. Keep the raw-JSON marshal for shapes that match no earlier
branch.

```suggestion
	// Array content (e.g. multimodal parts): join the text parts so the event
	// stays readable instead of embedding base64 payloads in raw JSON.
	var parts []struct {
		Text string `json:"text"`
	}
	if err := json.Unmarshal(content, &parts); err == nil {
		texts := make([]string, 0, len(parts))
		for _, part := range parts {
			if part.Text != "" {
				texts = append(texts, part.Text)
			}
		}
		return strings.Join(texts, "\n")
	}

	// Last resort for other non-string content: compacted raw JSON.
	marshaled, err := json.Marshal(content)
	if err != nil {
		return ""
	}
	return string(marshaled)
```

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -324,7 +347,76 @@ func parseCompletionRequest(body []byte) (string, []attribute.KeyValue) {
...
+		if content == "" {
+			continue
+		}
+		span.AddEvent(name, trace.WithAttributes(
+			attribute.String(key, streaming.TruncateContent(content)),
+		))
+	}
 }
 
 func parseEmbeddingRequest(body []byte) (string, []attribute.KeyValue) {
```
darccio, line 446: Embedding requests and responses are the only operations without a content
capture path. Under the GenAI semantic conventions, `input` is message
content, and the middleware already holds the parsed body when the capture
flag is on. Confirm whether the omission is deliberate or an oversight. If
it is deliberate, state the decision in the PR description. This PR is the
template for the capture adapter in #709, so other instrumentations will
inherit the scope decision that lands here.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -324,7 +347,76 @@ func parseCompletionRequest(body []byte) (string, []attribute.KeyValue) {
...
+
+	marshaled, err := json.Marshal(content)
+	if err != nil {
+		return ""
+	}
+	return string(marshaled)
+}
+
+func recordContentEvents(span trace.Span, name, key string, contents []string) {
+	for _, content := range contents {
```
darccio, line 436: The middleware emits one span event per captured message, and each event can
reach 16 KiB. A single request with a long `messages` history can therefore
add hundreds of events to one span. That cost is acceptable for an opt-in
feature. Confirm the intended consumption pattern (span events or log
records) with the maintainers, because #709 will generalize the format that
this PR chooses.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
```diff
@@ -58,18 +139,22 @@ func NewStreamingReader(
 	span trace.Span,
 	start time.Time,
 	op OperationType,
+	captureContent bool,
+	maxBodySize int,
```
darccio: The bridge always passes `streaming.ContentCaptureLimit` as `maxBodySize`, and
`contentLimit` clamps larger values back to 16 KiB. The parameter therefore
cannot change production behavior. It works as a test seam, but the name says
body size while the value bounds event content. Rename the parameter and the
struct field to `contentLimit`, and add a doc comment on `NewStreamingReader`
that states the clamp. The rename touches the bridge call and the test call
sites in the same change.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
```diff
@@ -155,12 +239,42 @@ func (r *StreamingReader) finalize(flush bool, err error) {
...
 }
 
+func streamError(err error) error {
+	if err == io.EOF {
+		return nil
+	}
+	return err
+}
+
+func (r *StreamingReader) recordCapturedContent() {
```
darccio, line 263: A truncated stream ends with `... [truncated]`. An aborted stream ends with
clean text and no marker. The span status carries the error, but at the event
level a consumer cannot tell partial content from complete content. Consider
appending a marker to the accumulated content when the stream did not
complete, so each event is self-describing.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
```diff
@@ -327,9 +447,29 @@ func (r *StreamingReader) processCompletionChunk(payload []byte) {
...
+
+func (r *StreamingReader) captureChunkContent(index int, content string) {
+	if r.capturedStreams == nil {
+		r.capturedStreams = make(map[int]*contentAccumulator)
+	}
+	if r.capturedStreams[index] == nil {
+		r.capturedStreams[index] = &contentAccumulator{limit: r.maxBodySize}
 	}
+	r.capturedStreams[index].Append(content)
 }
```
darccio, line 486: `contentAccumulator` bounds the size per choice, but not the number of
choices. The server controls the `index` values, so a stream that sends many
distinct indexes allocates one accumulator per index, and no cap exists on
that count. A normal response has a few choices, so the practical risk is
low, but the PR claims a memory bound and the count is the one unbounded
dimension. Cap the number of tracked indexes.

```suggestion
func (r *StreamingReader) captureChunkContent(index int, content string) {
	if r.capturedStreams == nil {
		r.capturedStreams = make(map[int]*contentAccumulator)
	}
	if r.capturedStreams[index] == nil {
		// The server controls the index values. Cap the map so that a stream
		// with many distinct indexes cannot grow memory without bound.
		if len(r.capturedStreams) >= 16 {
			return
		}
		r.capturedStreams[index] = &contentAccumulator{limit: r.maxBodySize}
	}
	r.capturedStreams[index].Append(content)
}
```

instrumentation/github.com/openai/openai-go/internal/streaming/streaming_test.go
```diff
@@ -541,3 +543,150 @@ func assertSliceAttribute(t *testing.T, attrs []attribute.KeyValue, key string,
...
+	reader := NewStreamingReader(io.NopCloser(strings.NewReader(streamData)), span, time.Now(), OpChat, false, ContentCaptureLimit)
+	_, err := io.ReadAll(reader)
+	require.NoError(t, err)
+	require.NoError(t, reader.Close())
+
+	spans := sr.Ended()
+	require.Len(t, spans, 1)
+	assert.Equal(t, codes.Error, spans[0].Status().Code)
+	assertSliceAttribute(t, spans[0].Attributes(), "gen_ai.response.finish_reasons", []string{"error"})
+}
```
darccio, line 692: The code emits the accumulated content when a stream aborts, and no test
covers that path. The missing test also pins the B3 behavior: the partial
event carries no marker that distinguishes it from complete content.

```suggestion
}

func TestStreamingReader_AbortedStreamRecordsContent(t *testing.T) {
	sr := tracetest.NewSpanRecorder()
	tp := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))
	_, span := tp.Tracer("test").Start(t.Context(), "aborted-stream-content")

	streamData := "data: {\"choices\":[{\"delta\":{\"content\":\"partial\"}}]}\n\n"
	reader := NewStreamingReader(io.NopCloser(strings.NewReader(streamData)), span, time.Now(), OpChat, true, ContentCaptureLimit)
	_, err := io.ReadAll(reader)
	require.NoError(t, err)
	require.NoError(t, reader.Close())

	spans := sr.Ended()
	require.Len(t, spans, 1)
	assert.Equal(t, codes.Error, spans[0].Status().Code)

	var found bool
	for _, event := range spans[0].Events() {
		if event.Name == "gen_ai.content.completion" {
			found = true
			assertAttribute(t, event.Attributes, "gen_ai.completion", "partial")
		}
	}
	require.True(t, found, "expected the partial content event on an aborted stream")
}
```

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -213,15 +232,15 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 
 		if isStreaming {
 			span.SetAttributes(semconv.GenAIRequestIsStream(true))
-			resp.Body = newStreamingReader(resp.Body, span, start, op, func() {
+			resp.Body = newStreamingReader(resp.Body, span, start, op, captureContent, func() {
```
darccio, line 242: No test drives `otelMiddleware` with a streaming response body, so the
streaming branch is only covered piecewise. The uncovered wiring includes the
`newStreamingReader` call, the propagation of `captureContent`, and the prompt
events that the middleware records before the request. The `setupTestTracer`
helper in `middleware_integration_test.go` makes this path assertable: feed a
`text/event-stream` body through the middleware and assert on both the prompt
event and the completion events that the stream emits at finalize.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -21,10 +23,15 @@ import (
...
 
+	"go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go/internal/streaming"
 	"go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go/semconv"
 	"go.opentelemetry.io/otelc/pkg/runtime"
 )
 
+var captureContentEnabled = sync.OnceValue(func() bool {
+	return os.Getenv("OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT") == "true"
+})
+
```
darccio, line 41: The `"true"` comparison against
`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` has no test. The
`sync.OnceValue` cache is process-wide, so a test cannot toggle the flag after
the first read, and the injection seam covers everything downstream. That is
an acceptable split. If coverage here matters, extract the comparison into a
small function such as `contentCaptureFromEnv(value string) bool` and test
that function directly.

darccio (commented):
Please make sure to validate these comments against the v2 and v3 variants too.

instrumentation/github.com/openai/openai-go/middleware_test.go
Aryanbhargava18: yeah good catch, added the model field and wired setupTestTracer, asserts on the prompt event now

instrumentation/github.com/openai/openai-go/middleware_test.go
Aryanbhargava18: swapped to require.Len and require.True here. also called span.End() since parseChatResponse doesn't end it

instrumentation/github.com/openai/openai-go/middleware.go
Aryanbhargava18: joins the text parts together now and only falls back to raw json if needed, updated the test too

instrumentation/github.com/openai/openai-go/middleware.go
Aryanbhargava18: deliberate yeah — embedding vectors get huge for span events. leaving a note in the PR desc so we can decide on it in #709

instrumentation/github.com/openai/openai-go/middleware.go
Aryanbhargava18: fair point, gonna call this out in the PR desc so #709 can decide on events vs log records

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
Aryanbhargava18: renamed to contentLimit across the board, and named the helper clampContentLimit so it doesn't clash

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
Aryanbhargava18: makes sense. i'll open a follow-up issue for the marker, added a test to lock in the current behavior for now

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
Aryanbhargava18: capped it at 16 entries — only blocks new index allocations, existing streams still process fine

instrumentation/github.com/openai/openai-go/internal/streaming/streaming_test.go
Aryanbhargava18: added the test — captureContent=true, cuts off before DONE, verifies codes.Error and the partial completion event

instrumentation/github.com/openai/openai-go/middleware.go
Aryanbhargava18: added TestOtelMiddleware_StreamingContentCapture to drive the full middleware with SSE, checks both prompt and finalized completion events

instrumentation/github.com/openai/openai-go/middleware.go
Aryanbhargava18: pulled it out into contentCaptureFromEnv(val string) bool so it's directly testable, OnceValue just calls into it

Aryanbhargava18:
@darccio addressed all review comments inline and ported the changes across v2 and v3 variants as well. Re-generated and verified the bundle archive with `make package`, and updated the PR description with the notes on embedding scope and event format for #709.

darccio (approved):
LGTM

PR #752, opened 2026-07-22
test: reproduce logrus AfterLogrusNew init-order panic

mtanda:
## Description

Building and running an otelc-instrumented app that has `github.com/sirupsen/logrus` anywhere in its dependency graph -- even if the application never calls a single logrus function -- currently prints this before any application code runs:

```
failed to exec After hook AfterLogrusNew
assignment to entry in nil map
```

## Cause

`AfterLogrusNew` (the compile-time hook on `logrus.New()`, wired via `//go:linkname`) fires while logrus's own `var std = New()` package-level initializer runs. `go:linkname` doesn't create a normal Go import edge, so Go's init-order guarantees don't apply here: the hook implementation package's own state (`hookInitMap`) is not guaranteed to be initialized yet when the hook fires. The resulting nil-map write panics, and the panic is silently recovered by the generated hook trampoline (`tool/internal/instrument/impl.tmpl`), so the process keeps running.

## Minimal reproduction

This is not specific to any particular usage pattern -- it reproduces unconditionally, even when the application code never references `github.com/sirupsen/logrus` at all:

```go
package main

import (
	"fmt"

	_ "github.com/sirupsen/logrus"
)

func main() {
	fmt.Println("app never touches logrus directly")
}
```

Built and run with `otelc go build -a`:

```
$ /tmp/logrusblankimport-repro
failed to exec After hook AfterLogrusNew
assignment to entry in nil map
{"time":"...","level":"INFO","msg":"trace provider initialized with auto-export"}
{"time":"...","level":"INFO","msg":"meter provider initialized with auto-export"}
{"time":"...","level":"INFO","msg":"logger provider initialized with auto-export"}
{"time":"...","level":"INFO","msg":"OpenTelemetry initialized","instrumentation_name":"go.opentelemetry.io/otelc","instrumentation_version":"dev"}
{"time":"...","level":"INFO","msg":"runtime metrics enabled"}
app never touches logrus directly
```

The panic fires before any application code runs, from logrus's own package-level `var std = New()` initializer -- confirming the trigger is logrus being present in the build graph at all, not anything the application itself does with it.

## User impact

Every user who builds with `otelc` and has logrus anywhere in their dependency graph sees this panic-looking message on every startup, unconditionally, regardless of how (or whether) their own code uses logrus. There is nothing an application can do about it -- the trigger is logrus's own package init, which runs before `main`.

This PR adds a regression test documenting the panic, skipped for now so it doesn't block CI on an existing, unfixed issue.

---

## Checklist

- [x] Linters pass: `make lint/go`
- [x] Tests pass: `make test`
- [ ] Tests follow testing guidelines


amazingakai (changes requested):
@mtanda I think we should instead create an issue to track this problem

mtanda:
@amazingakai Sorry for late reply. I made a issue for this problem.
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/1028
So, I close this PR.

PR #757, opened 2026-07-22
fix(ast): handle composite types in signature filters

### LINKED ISSUE #756: Signature filters fatal on functions with slice/chan/map params instead of no-match
### Describe the bug

Signature sub-filters (`signature_contains`, `param`, `result`, `last_result`, and exact `signature`) **fatal during `otelc setup`** when they scan a function whose parameters or return types use composite Go types (`[]T`, `chan T`, `map[K]V`, `func(...)`, etc.).

The tool calls `util.Unimplemented` → `ex.Fatalf` instead of returning “no match” or recursing into element types. This aborts the entire build.

**Docs vs behavior mismatch:** `docs/rules.md` states that if a parameter or return uses a composite type, “the filter will never match it.” The actual behavior is a **hard crash**, not a non-match.

**Affected code path:**

- `tool/internal/ast/typename.go` - `parsedTypeName.matches()` hits the `default` branch for `*dst.ChanType`, `*dst.ArrayType`, `*dst.MapType`, `*dst.FuncType`, etc. and calls `util.Unimplemented(...)`.
- Triggered via `fieldListContainsType()` → `matchesSignatureContains()` / `funcDeclMatchesFilters()` in `tool/internal/ast/shared.go`.

**Example crash (verified locally):**

Target function: `func(ch chan int) error`  
Rule filter: `signature_contains: {args: [int]}`  

```
Unimplemented: signature filter: unsupported type node *dst.ChanType
```

Same crash occurs for common stdlib-style signatures such as `func Read(p []byte) (n int, err error)` when using `signature_contains: {args: [byte]}` - the filter looks for `byte` but hits `[]byte` and fatals.

---

### Environment

- **OS:** Linux (any OS; not platform-specific)
- **Go Version:** 1.25.0
- **Version:** `main` / latest `otelc`

---

### To Reproduce

**Option A - unit-level (fastest):**

1. In `tool/internal/ast/funcmatch_test.go`, add a test for a function with a composite param:

   ```go
   // func(ch chan int) error
   decl := makeFuncDecl(
       []*dst.Field{field(&dst.ChanType{Value: ident("int")})},
       []*dst.Field{field(ident("error"))},
   )
   sig := rule.FuncSignature{Args: []string{"int"}}
   r := &rule.InstFuncRule{SignatureContains: &sig}
   _, _ = funcDeclMatchesFilters(decl, r) // fatals
   ```

2. Run:

   ```bash
   go test ./tool/internal/ast/ -run TestFuncDeclMatchesFilters -v
   ```

3. See fatal:

   ```
   Unimplemented: signature filter: unsupported type node *dst.ChanType
   ```

**Option B - rule + setup (real-world):**

1. Add a target package with:

   ```go
   package main

   func Read(p []byte) (n int, err error) { return len(p), nil }
   ```

2. Create a rule:

   ```yaml
   read_hook:
     target: main
     func: Read
     signature_contains:
       args: [byte]
     before: BeforeRead
     path: example.com/hooks
   ```

3. Run `otelc setup` / `otelc go build` against that package.

4. Setup fatals while evaluating signature filters instead of matching `Read` or skipping it gracefully.

**Also affects:** `param: string` on `func(p []string)`, `signature` exact match on functions with slice/map/chan params, etc.

---

### Expected behavior

When a signature filter encounters a composite type in a function signature:

1. **No fatal error** -  setup/instrumentation must not abort.
2. For `signature_contains` / `param` / `result`: either return `false` (no match), or recurse into element types (e.g. `[]byte` contains `byte`) per documented “never match” / partial-match semantics.
3. For `signature` exact match: return `false` when composite types are present in the signature.
4. Behavior should align with `docs/rules.md`:

   > If a parameter or return value uses one of these forms, the filter will never match it.

   “Never match” must mean **skip / no match**, not **crash the tool**.

---

### Additional context

**Root cause:** `parsedTypeName.matches()` in `tool/internal/ast/typename.go:69–73` calls `util.Unimplemented(...)` for unsupported AST nodes instead of returning `false` or handling composite types.

**Impact:**

- Rule authors cannot safely use `signature_contains` to disambiguate overloaded function names when any candidate has `[]byte`, `chan`, `map`, or func-typed parameters - common in stdlib and networking code.
- Any instrumentation rule targeting packages with such signatures can crash `otelc setup` at load time.
- Existing golden tests only cover simple types (`string`, `int`, `error`); see `tool/internal/instrument/testdata/golden/func-sig-contains/`.

Jay2006sawant:
## Description

Signature sub-filters (`signature_contains`, `param`, `result`, `last_result`) previously called `util.Unimplemented` and fatally aborted `otelc setup` when scanning function signatures that use composite Go types (`[]T`, `chan T`, `map[K]V`, `func(...)`, etc.).

This PR fixes that crash by:

- Adding `matchesContainedIn()` in `tool/internal/ast/typename.go` to traverse composite types when evaluating contains-style filters
- Keeping `matches()` as exact-only matching for the `signature` exact-match path
- Replacing the fatal `Unimplemented` path with safe match/no-match behavior
- Adding regression tests for `chan`, `[]byte`, and exact vs contains semantics
- Updating `docs/rules.md` to document composite traversal behavior

## Motivation

Rule authors use signature filters to disambiguate overloaded function names (e.g. `signature_contains: {args: [byte]}` on `func Read(p []byte) (n int, err error)`). Before this fix, any such rule against a function with slice/chan/map params caused:

```
Unimplemented: signature filter: unsupported type node *dst.ChanType
```

That aborted the entire build instead of matching or skipping gracefully, contradicting the documented behavior that filters should not match composite types fatally.

Fixes #756 

---

## Checklist

- [ ] Linters pass: `make lint` *(pre-existing `nolintlint` failure in `tool/internal/instrument/trampoline.go` on main; unchanged by this PR)*
- [x] Tests pass: `make test` *(verified: `go test ./tool/internal/ast/...`, `./tool/internal/setup/...`, `./tool/internal/instrument/...`)*
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](docs/instrument-guide.md#4-register-the-instrumentation)) *(N/A - tool-only change)*

### DIFF
--- a/tool/internal/ast/typename.go
+++ b/tool/internal/ast/typename.go
@@ -4,13 +4,11 @@
 package ast
 
 import (
-	"fmt"
 	"regexp"
 
 	"github.com/dave/dst"
 
 	"go.opentelemetry.io/otelc/tool/ex"
-	"go.opentelemetry.io/otelc/tool/util"
 )
 
 // typeNameRe parses type-name strings of the form [*][pkg.]Name.
@@ -37,7 +35,7 @@ func parseTypeName(s string) (parsedTypeName, error) {
 	return parsedTypeName{pointer: m[1] == "*", importPath: m[2], name: m[3]}, nil
 }
 
-// matches reports whether the dst.Expr node represents this type.
+// matches reports whether node is exactly this type (no composite traversal).
 func (t parsedTypeName) matches(node dst.Expr) bool {
 	switch n := node.(type) {
 	case *dst.Ident:
@@ -63,15 +61,44 @@ func (t parsedTypeName) matches(node dst.Expr) bool {
 		return !t.pointer && t.matches(n.X)
 
 	case *dst.InterfaceType:
-		// Only the empty interface matches "any".
+		// Only the empty interface matches "any"; named interfaces never match.
 		return len(n.Methods.List) == 0 && t.importPath == "" && t.name == "any"
 
 	default:
-		// Unsupported AST node types (chan, func, map, slice, array, interface
-		// literals) cannot be matched by type-name filters.
-		util.Unimplemented(fmt.Sprintf("signature filter: unsupported type node %T", node))
 		return false
 	}
+}
+
+// matchesContainedIn reports whether this type appears in node, traversing
+// composite types such as slices, channels, maps, and function signatures.
+func (t parsedTypeName) matchesContainedIn(node dst.Expr) bool {
+	if t.matches(node) {
+		return true
+	}
+	switch n := node.(type) {
+	case *dst.ArrayType:
+		return t.matchesContainedIn(n.Elt)
+	case *dst.ChanType:
+		return t.matchesContainedIn(n.Value)
+	case *dst.MapType:
+		return t.matchesContainedIn(n.Key) || t.matchesContainedIn(n.Value)
+	case *dst.FuncType:
+		return typeExprListContainsMatch(t, n.Params) || typeExprListContainsMatch(t, n.Results)
+	default:
+		return false
+	}
+}
+
+func typeExprListContainsMatch(t parsedTypeName, fields *dst.FieldList) bool {
+	if fields == nil {
+		return false
+	}
+	for _, field := range fields.List {
+		if t.matchesContainedIn(field.Type) {
+			return true
+		}
+	}
+	return false
 }
 
 // fieldListContainsType reports whether any field in fields has a type that
@@ -85,7 +112,7 @@ func fieldListContainsType(fields *dst.FieldList, typeStr string) (bool, error) 
 		return false, err
 	}
 	for _, field := range fields.List {
-		if tn.matches(field.Type) {
+		if tn.matchesContainedIn(field.Type) {
 			return true, nil
 		}
 	}
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -483,7 +483,9 @@ Type names follow the form `[*][pkg.]Name`, for example `error`, `context.Contex
 
 > **Note on scalar filter semantics:** Because there is no type checker at instrumentation time, `result`, `last_result`, and `param` perform an exact type-name match. For example, `result: error` matches functions that literally return the `error` type, but not functions that return a concrete type (e.g. `*MyError`) that happens to implement `error`.
 >
-> **Unsupported type expressions:** Complex type expressions — `chan`, `func`, `map`, slice (`[]T`), and non-empty interface literals — cannot be matched by type-name filters. If a parameter or return value uses one of these forms, the filter will never match it.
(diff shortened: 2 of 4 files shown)


tool/internal/ast/typename.go
```diff
@@ -63,17 +61,46 @@ func (t parsedTypeName) matches(node dst.Expr) bool {
...
+	case *dst.ArrayType:
+		return t.matchesContainedIn(n.Elt)
+	case *dst.ChanType:
+		return t.matchesContainedIn(n.Value)
+	case *dst.MapType:
+		return t.matchesContainedIn(n.Key) || t.matchesContainedIn(n.Value)
+	case *dst.FuncType:
+		return typeExprListContainsMatch(t, n.Params) || typeExprListContainsMatch(t, n.Results)
+	default:
+		return false
```
Aryanbhargava18, line 88: Great catch on removing `util.Unimplemented` from `matches()` to fix the fatal error.

However, your `matchesContainedIn` traversal misses three critical AST nodes that wrap types in Go signatures:
1. `*dst.Ellipsis`: Variadic parameters like `...string`. A filter for `string` will fail to match because `matchesContainedIn` drops to `default: return false`.
2. `*dst.StarExpr`: Pointers to composite types like `*[]byte` or `*map[string]int`. While `t.matches` handles `StarExpr` for named types, if the pointer wraps a composite type, `t.matches` returns `false` and `matchesContainedIn` skips it because there's no `case *dst.StarExpr:` in the switch block.
3. `*dst.ParenExpr`: Parenthesized types like `([]byte)`.

You should add them to the switch block in `matchesContainedIn`:
```go
	case *dst.Ellipsis:
		if n.Elt != nil {
			return t.matchesContainedIn(n.Elt)
		}
	case *dst.StarExpr:
		return t.matchesContainedIn(n.X)
	case *dst.ParenExpr:
		return t.matchesContainedIn(n.X)
```

Without these, `signature_contains: {args: [string]}` will silently fail to match `func Log(msgs ...string)` or `func Decode(buf *[]byte)`. Add those cases and a couple of test cases in `TestFuncDeclMatchesFilters_CompositeParamTypes` for variadic and pointer-to-slice, and this will be rock solid.

Jay2006sawant:
@kakkoyun @NameHaibinZhang PTAL

amazingakai (changes requested):
Just to make sure I understand the semantics:

```yaml
signature_contains:
  args: [string]
```

Would this also match `[]string`?

If so, I think that could be a bit confusing. My expectation would be that `signature_contains` checks whether the signature contains a parameter of the specified type, with the type matching exactly. In other words, "contains" would apply to the list of arguments, not to partial matches within the argument types.

CC @darccio, since you originally implemented `signature_contains`, I'd be interested in hearing your thoughts on the intended behavior here.

PR #762, opened 2026-07-23
fix(tool): implement where.file.has_directive filter

### LINKED ISSUE #739: `Implement where.file.has_directive filter (documented but unsupported)`
### Describe the bug

The `where.file.has_directive` predicate is **documented** in the rules schema and [docs/rules.md](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/rules.md), but **not implemented** at build time. Any rule using it fails with:

```
where.file.has_directive is not yet supported
```

This blocks rule authors from targeting source files that carry specific `//go:` directives (for example `//otelc:span`), even though the underlying AST helper already exists and is used elsewhere in the tool.

**Gap between docs and code:**

| Location | Behavior |
|----------|----------|
| `docs/rules.md` | Lists `has_directive` as a valid `where.file` predicate |
| `tool/internal/rule/base.go` | Field defined: `HasDirective string` with comment "validated, not yet executed" |
| `tool/internal/setup/filter.go:308-309` | Returns error: `where.file.has_directive is not yet supported` |
| `tool/internal/ast/directive.go` | `FileHasDirective()` already implemented and tested |
| `tool/internal/setup/match.go` | Uses `FileHasDirective()` for **directive rules**, but not for **file filters** |

Rule authors expect to write rules like "only instrument files that contain `//otelc:external-log`" using `where.file.has_directive`, but the build rejects them.

---

### Environment

- **OS:** Linux (any OS; not platform-specific)
- **Go Version:** 1.25.0
- **Version:** `main` / latest `otelc`

---

### To Reproduce

1. Create an instrumentation rule using `where.file.has_directive`:

   **`otelc.yaml`**
   ```yaml
   trace_directive_files:
     target: example.com/app
     where:
       file:
         has_directive: otelc:span
       func: Handler
     do:
       - inject_hooks:
           before: BeforeHandler
           path: example.com/instrumentation/app
   ```

2. Build with `otelc`:
   ```bash
   otelc go build .
   ```

3. See error at rule load time:
   ```
   where.file.has_directive is not yet supported
   ```

4. Compare with existing test fixture that expects this rejection:

   **`tool/internal/setup/testdata/where/err_has_directive.yml`**
   ```yaml
   has_directive: otelc:span
   ```

   **`tool/internal/setup/filter_test.go`** confirms `Build()` returns an error for this predicate today.

**Existing infrastructure (ready to wire up):**

```go
// tool/internal/ast/directive.go
func FileHasDirective(file *dst.File, directive string) bool { ... }
```

Directive rules in `match.go` already use this for matching; only the `where.file` filter path is missing.

---

### Expected behavior

- Rules with `where.file.has_directive` should load and apply successfully
- Files containing the specified directive comment should match the filter
- Files without the directive should be skipped (same semantics as `has_func`, `has_struct`, etc.)
- Behavior should match what is documented in `docs/rules.md`

**Example that should work after the fix:**

```yaml
trace_directive_files:
  target: example.com/app
  where:
    file:
      has_directive: otelc:span
    func: Handler
  do:
    - inject_hooks:
        before: BeforeHandler
        path: example.com/instrumentation/app
```

---

### Additional context

**Root cause:** `buildFile()` in `tool/internal/setup/filter.go` explicitly rejects `has_directive` instead of compiling a filter:

```go
case def.HasDirective != "":
    return nil, ex.Newf("where.file.has_directive is not yet supported")
```

**Impact:** Rule authors cannot use a documented predicate to scope instrumentation to files marked with otelc directives. This limits fine-grained control that the schema already promises.

Aryanbhargava18:
## Description

This PR implements the `where.file.has_directive` filter in the `setup` package, which was previously documented but unsupported. It delegates to `ast.FileHasLeadingDirective` to scan the AST for the requested compiler directive.

Added positive and negative tests to `filter_test.go` and `testdata/where`.

## Motivation

This resolves an inconsistency between the documented schema and the actual build-time behavior, allowing rule authors to target specific `//go:directive` tags in source files.

Fixes #739

---

## Checklist

- [x] Tests pass: `make test`


tool/internal/setup/filter.go
amazingakai, line 72: Let's add the new filter here as well?

tool/internal/setup/filter.go
```diff
@@ -110,6 +110,15 @@ func (f *PackageNameFilter) Match(ctx *MatchContext) bool {
 	return ctx.AST.Name.Name == f.Name
 }
 
+// DirectiveFilter matches source files that contain the specified directive.
+type DirectiveFilter struct {
+	Directive string
+}
+
+func (f *DirectiveFilter) Match(ctx *MatchContext) bool {
+	return ast.FileHasDirective(ctx.AST, f.Directive)
```
amazingakai: I think we should match go's notion of a file-level directive here. Go only recognizes `//go:build` in the file's leading run of `//` comments before the `package` clause, rather than anywhere in the file:
https://github.com/golang/go/blob/master/src/go/build/build.go#L1554-L1557

I think this should follow the same semantics instead of searching the entire file. That matches the behavior of go itself and avoids directives attached to declarations or other nodes being treated as file-level directives.

`ast.FileHasDirective` searches anywhere in the file, which I don't think is the right behavior for this particular use case.

WDYT? If we follow this behavior, we should also document this in `docs/rules.md`

amazingakai (changes requested):

Aryanbhargava18:
@amazingakai Good catch regarding the Go compiler semantics.

I've pushed a fix that addresses both points:
1. Restricts `FileHasDirective` to only scan `file.Decs.Start` (the file's leading run of comments before the `package` clause).
2. Adds the missing `_ Filter = (*DirectiveFilter)(nil)` assertion in `filter.go`.
3. Updates `docs/rules.md` to clarify that directives attached to declarations are not treated as file-level directives.

tool/internal/ast/directive.go
```diff
@@ -109,30 +109,12 @@ func scanArgs(input string) ([]DirectiveArg, error) {
...
-				found = true
-				return false
-			}
-		}
-		return true
-	})
-	return found
+	}
+	return false
 }
```
amazingakai, line 147: Could we add a new function instead of modifying this one? This function is also used during the setup phase for directive rule matching, so changing its behavior would break that logic.

amazingakai (changes requested):
It looks like your AI agent committed a number of unintended files as part of this PR. Could you clean those up?

Also, please review the project's AI policy before opening AI-generated PRs: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md.

Aryanbhargava18:
Ah, apologies for that! I carelessly ran `git add .`, which picked up some untracked scratchpads and JSON logs from my workspace. I've force-pushed to completely clean up the branch.

I have also reviewed the AI Policy. I want to assure you that I completely own, understand, and stand behind every line of Go code and AST logic in this PR. Thanks for your patience as I get my workspace workflow dialed in!

Regarding the `FileHasDirective` feedback:
1. I restored the original `FileHasDirective` function so it doesn't break the setup phase's directive matching logic.
2. I introduced a new `FileHasLeadingDirective` function in `ast/directive.go` specifically for this rule, which only scans `file.Decs.Start` to match the Go compiler semantics.
3. Updated the tests and `filter.go` to use the new function.

Let me know if this looks good to go!

amazingakai (approved):
LGTM, Good Job!

Aryanbhargava18:
i have Rebased against main to resolve the out-of-date status

tool/internal/ast/directive.go
```diff
@@ -135,6 +135,17 @@ func FileHasDirective(file *dst.File, directive string) bool {
 	return found
 }
 
+// FileHasLeadingDirective returns true if the file contains a leading run of comments
+// before the package clause that matches the given directive.
+func FileHasLeadingDirective(file *dst.File, directive string) bool {
```
darccio, line 140: The PR description says this "delegates to `ast.FileHasDirective`" — the
actual implementation delegates to the new `ast.FileHasLeadingDirective`,
which is deliberately narrower (leading file comments only, not all AST
nodes). Worth updating the description so readers aren't misled about which
helper backs the new filter.

darccio (approved):
LGTM, just a minor change request before merging.

Aryanbhargava18:
@darccio i have updated the description.

PR #765, opened 2026-07-23
docs(rfcs): propose genai.Observer architecture for unified telemetry

Aryanbhargava18:
## Description
Following up on the LFX Term 3 architecture discussions in #709, this RFC formally proposes the `genai.Observer` abstraction for the `pkg/genai` namespace.

Currently, SDK instrumentations like `openai-go` manually parse JSON and handle chunked Server-Sent Events (SSE) to map their idiosyncratic payloads to OpenTelemetry `gen_ai.*` semantic conventions. As we expand to support Gemini, LangChain, and MCP, duplicating this stateful accumulation and truncation logic across every SDK creates an unsustainable maintenance burden.

This RFC standardizes the flow by moving all Semantic Convention mapping, token usage aggregation, and memory-safe content accumulation out of the SDK hooks and into a central `genai.Observer`. The SDKs are reduced to simple, stateless Extractors.

This lays the exact architectural foundation required for Milestone 1 of the LFX GenAI initiative.

### DIFF
--- a/docs/genai_observer_rfc.md
+++ b/docs/genai_observer_rfc.md
@@ -1,0 +1,95 @@
+# RFC: `genai.Observer` — A Unified Adapter for GenAI Telemetry
+
+## Background
+Currently, AI SDK instrumentations (e.g., `openai-go`, `anthropic-sdk-go`) in `otelc` manually parse JSON, handle chunked Server-Sent Events (SSE), and independently map these idiosyncratic payloads to OpenTelemetry `gen_ai.*` semantic conventions. 
+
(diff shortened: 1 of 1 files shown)


NameHaibinZhang (changes requested):
Thanks for taking the initiative on this! The direction of centralizing GenAI telemetry logic is worth exploring, but this PR is premature for the following reasons:

1. **No maintainer consensus on direction**: The architecture discussion in #709 hasn't reached agreement from maintainers yet. Please get buy-in on the approach there before submitting an RFC.

2. **Wrong format/location**: This project uses ADRs (`docs/adr/`) with a specific template (Status, Context, Decision, Consequences). An architectural proposal should follow that process rather than introducing a new `docs/genai_observer_rfc.md` file.

3. **Missing critical analysis**: The RFC doesn't discuss tradeoffs, migration path from the existing `streaming.go` implementation, backward compatibility, or alternatives considered.

Suggested next steps:
- Discuss the high-level approach in #709 and get maintainer feedback first
- Once direction is agreed, submit as a properly numbered ADR following the template in `docs/adr/template.md`

PR #768, opened 2026-07-23
feat: implement declarative configuration pipeline

### LINKED ISSUE #705: LFX Term 3: Per-instrumentation configuration: let users control how individual instrumentations behave
## Problem

Users can only turn instrumentations on or off (`OTEL_GO_ENABLED/DISABLED_INSTRUMENTATIONS`); there is no way to configure how one behaves: capture specific HTTP headers or gRPC metadata, redact URL query parameters, opt into a semconv migration gradually, or set any library-specific option. Configuration granularity was the top post-v1 community ask, and upcoming work (messaging span-links migration, GenAI content capture) needs an opt-in mechanism; ad-hoc env vars per feature do not scale.

## Proposed solution

Adopt the OTel [declarative configuration](https://github.com/open-telemetry/opentelemetry-configuration) model instead of inventing our own format. Its `instrumentation/development` node defines exactly this surface:

```yaml
instrumentation/development:
  general:                        # cross-language, semconv-domain keyed
    http:
      server:
        request_captured_headers: [X-Tenant-Id]
    messaging:
      semconv:
        dual_emit: true           # migration gate
  go:                             # language-specific, owned by us
    kafka-go:
      enabled: ${OTEL_GO_KAFKA:-true}
```

otelc reads the file at build time and exposes the values to hooks at runtime; existing env vars keep working and take precedence. Mechanics go in a follow-up ADR.

Similar approach to what Java agent does.

Context: the SDK half of that model is stable (Go has `otelconf`); the instrumentation node is experimental, consumed by the Java agent, and implemented by nothing in Go yet. The closed `general` node gives us cross-language semantics (captured headers, URL sanitization, per-domain semconv migration); the open `go:` node is a free-form `<instrumentation-name>: <options>` map we fully own, so upstream coverage never limits us. Gaps we hit get proposed upstream.

## Goals

1. otelc accepts the config file; hooks read their configuration at runtime with documented precedence (env vars > file > defaults).
2. First consumers on existing instrumentations: HTTP captured headers (net/http, gin), URL sanitization, gRPC metadata capture (under `go:`), enable/disable expressed as `go: <name>: enabled`.
3. A maintainable process, not just options: each instrumentation declares its configurable options in a manifest (as the Java agent does with `metadata.yaml`); the `go:` validation schema, a generated catalog of every configurable option, and defaults all derive from it, and CI forbids hooks from reading undeclared configuration. This makes "what can I configure?" a build artifact that cannot drift from the code.
4. ADR settling the model: the split with `otel.instrumentation.go` (tool files decide what gets woven, this file decides how it behaves), the decision tree for which parameters become configurable and where they live (semconv opt-in must use upstream `general` keys; cross-library gaps ship under `go:` with an upstream proposal; library-specific stays under `go:`), and a development-to-stable lifecycle for options.

## Concerns to discuss

- `instrumentation/development` is experimental upstream; Java ships it anyway. Comfortable building on it now (pinning a schema version)?
- otelc is a build tool: file must exist at build time; deploy-time changes go through `${ENV}` substitution or env vars, unlike the Java agent.
- Configurable options are permanent API surface: `general`/shared additions need maintainer sign-off; library-local options ride normal review. Is that the right review gate?
- `go:` key naming/ownership, incl. third-party instrumentation packages.
- Keep `OTEL_GO_*` env vars indefinitely or deprecate once config lands?

Aryanbhargava18:
Fixes #705 

## Description
This PR introduces the foundational pipeline for declarative instrumentation configuration via `.otel.yml`. It parses user configuration and embeds it into the compiled application using an AST-injected `otelc.runtime.go` registry, making the config available to instrumentation hooks at runtime.

This fulfills the core architecture proposed in #705, providing the registry and pipeline needed to transition away from environment variables to a structured `genai.Observer` pattern and other library-specific toggles.

## Implementation Details
- **Schema**: Added `tool/internal/config/schema.go` containing `InstrumentConfig` and `OtelConfig`.
- **Parsing**: Added `.otel.yml` loading in `tool/internal/setup/setup.go`.
- **Runtime Registry**: Created `pkg/runtime/config.go` containing the global configuration registry (`Registry()`).
- **AST Injection**: Updated `tool/internal/setup/add.go` to inject the user's parsed config directly into the generated `otelc.runtime.go` file using an `init()` block.

## Testing
Tested locally. Verified that `otelc go build` correctly reads `.otel.yml` and that the generated `otelc.runtime.go` correctly registers the data structures in the `runtime` package without any import cycle issues.


Aryanbhargava18:
Closing this draft implementation. 

After diving deep into the architecture, I've realized my primary passion and focus for the upcoming LFX term lies entirely in the GenAI Observer adapter (#709), which builds on the OpenAI content capture prototype I recently built in #740. 

I'm stepping back from this declarative config implementation to focus 100% of my energy on the GenAI space. I've left my architectural notes in #705 for whoever takes up the RFC. Looking forward to seeing the final design!

Issue #546, opened 2026-05-27
feat(tool/rule): derive rule type from do modifier name (not field presence)


kakkoyun:
Follow-up from #377 (review thread from @amazingakai on `docs/rules.md:100`).

## Context

ADR-0003 and `docs/rules.md` state that the rule type (FuncHook, StructFieldInject, CallWrap, etc.) is derived from the modifier key in `do`. The current implementation still falls back to field-presence inference because the flat-rule normalization shape erases the modifier name (see `tool/internal/setup/match.go` `createRuleFromFields`).

## Problem

The doc and the implementation disagree. The doc has been softened in #377 to mark this as planned, but the gap should be closed.

## Proposed work

Likely subsumed by the structured-form constructor refactor (see related follow-up). When `createRuleFromFields` accepts a `DoDef`, the modifier name *is* the dispatch key, and field-presence inference goes away.

If the structured-form refactor lands first, this can be closed as a duplicate. If this ships standalone, it adds a `modifier` argument to `createRuleFromFields` while keeping the flat fields in place.

## References

- #377 inline thread on `docs/rules.md:100`

dobbydobap:
Hi, I'd like to take a shot at this one if it's still open.

From reading the issue and ADR-0003, my understanding is the fix is to have
createRuleFromFields work off the do-modifier name as the dispatch key instead
of inferring the rule type from which fields are present, the structured-form
approach mentioned in the issue (passing a DoDef through).

Before I start: should this be done as a standalone change, or is it meant to
land as part of the bigger #541 refactor? I want to make sure I build it in the
direction you'd want rather than duplicate effort. Happy to open a PR once the
approach is confirmed.

Harsh63870:
@kakkoyun #854 landed registry + unknown-modifier rejection (#542). #546 remaining: normalizeDo still drops modifier name; createRuleFromFields still uses field presence. Happy to do a minimal standalone fix per issue text, or help rebase #704, which do maintainers prefer?

PR #770, opened 2026-07-23
refactor(tool/rule): add selector and modifier registry

### LINKED ISSUE #542: feat(tool/rule): central registry of valid selectors and modifiers
Follow-up from #377 (review thread from @txabman42 on `tool/internal/rule/normalize.go:116`).

## Context

The set of valid `where` selectors (`func`, `recv`, `struct`, `function_call`, `directive`, `kind`, `identifier`) and `do` modifiers (`inject_hooks`, `replace`, `wrap`, `add_field`, `raw_inject`, `template`, …) is currently encoded in three places:

1. `tool/internal/rule/normalize.go` — string switches in `normalizeWhere` and `normalizeDo`.
2. `tool/internal/setup/match.go` — `createRuleFromFields` field-presence switch.
3. `tool/internal/rule/filter.go` — selector composition rejection list.

Branch 0 of the OP4 stack consolidated these into unexported constants in `normalize.go` as a stepping stone, but they are not yet a public registry.

## Problem

Adding or renaming a selector requires touching all three sites manually. Mistakes silently fall through (an unknown selector simply does not match anything).

## Proposed work

- New file `tool/internal/rule/schema.go` declaring `Selector` and `Modifier` as named string types with package-level constants.
- Exposed predicates: `IsValidSelector(string) bool`, `IsValidModifier(string) bool`, plus iteration helpers `AllSelectors() []Selector` / `AllModifiers() []Modifier` for tooling.
- All three call sites switch to the registry; adding a selector becomes a single-file change.
- This is a prerequisite for the rule-schema linter (see related follow-up).

## References

- #377 inline thread on `tool/internal/rule/normalize.go:116`

ADITYA-CODE-SOURCE:
## Summary

This PR adds a central registry for structured rule selectors and modifiers.

It introduces `tool/internal/rule/schema.go` as the shared place for valid `where` selector keys and `do` modifier names, and updates normalization/filter validation to use that registry.

Unknown structured `do` modifiers now fail early instead of being accepted silently.

This PR does not change rule dispatch semantics; that remains separate work under `#546`.

Fixes #542.

## Testing

Ran:
- `go test ./tool/internal/rule ./tool/internal/setup`
- `make test-unit/tool`
- `make lint/go`

### DIFF
--- a/tool/internal/rule/normalize.go
+++ b/tool/internal/rule/normalize.go
@@ -10,48 +10,6 @@ import (
 	"go.opentelemetry.io/otelc/tool/util"
 )
 
-// Structured top-level keys.
-const (
-	KeyWhere = "where"
-	KeyDo    = "do"
-)
-
-// where selectors (hoisted to flat by normalizeWhere).
-const (
-	SelTarget       = "target"
-	SelVersion      = "version"
-	SelFunc         = "func"
-	SelRecv         = "recv"
-	SelStruct       = "struct"
-	SelFunctionCall = "function_call"
-	SelDirective    = "directive"
-	SelKind         = "kind"
-	SelIdentifier   = "identifier"
-
-	// Signature match-narrowing selectors for func rules (see InstFuncRule).
-	SelSignature         = "signature"
-	SelSignatureContains = "signature_contains"
-	SelResult            = "result"
-	SelLastResult        = "last_result"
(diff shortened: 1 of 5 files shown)


ADITYA-CODE-SOURCE:
Closing because #854 merged the selector/modifier registry and resolved #542.

PR #772, opened 2026-07-23
fix(instrumentation): emit HTTP status code for anthropic error.type

malladinagarjuna2:
The Anthropic middleware set `error.type` to the full HTTP status line
(e.g. `"429 Too Many Requests"`). Per the error semantic conventions it
should be the numeric status code (`"429"`). This routes the value through
a new `semconv.ErrorType` helper instead of a raw `attribute.String`, and
updates the `TestOtelMiddleware_HTTPError` assertion accordingly.

## Motivation

`error.type` is meant to be a low-cardinality error class — for HTTP that's
the status code. Emitting the full status line ("429 Too Many Requests")
breaks metric/trace grouping in backends that key on `error.type` and
diverges from the error semantic conventions.

---

## Checklist

- [ ] Tests pass: `make test`


instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
```diff
@@ -145,7 +146,9 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 
 		if resp.StatusCode >= 400 {
 			span.SetStatus(codes.Error, resp.Status)
-			span.SetAttributes(attribute.String("error.type", resp.Status))
+			// error.type carries the numeric status code per the error semantic
+			// conventions (e.g. "429"), not the full status line.
+			span.SetAttributes(semconv.ErrorType(strconv.Itoa(resp.StatusCode)))
```
darccio, line 151: This introduces a second, locally-defined `ErrorTypeKey`/`ErrorType` for an
attribute that's already defined in the official
`go.opentelemetry.io/otel/semconv/v1.37.0` package, which this module already
depends on directly (go.mod requires `go.opentelemetry.io/otel v1.44.0`).
Two other instrumentations in this repo already emit `error.type` via the
official key — `net/http/semconv/server.go:270` and
`linode/.../semconv/client.go:129` — so this diverges from the established
pattern for this specific attribute. (The rest of this file's local keys are
fine to keep local — they're GenAI-specific and not yet stable upstream;
`error.type` isn't a GenAI attribute.)

Suggest importing the official package under an alias (it collides with the
local `semconv` import already in this file) and dropping the local
`ErrorTypeKey`/`ErrorType` additions from `semconv/genai.go`.

```suggestion
			span.SetAttributes(otelsemconv.ErrorTypeKey.String(strconv.Itoa(resp.StatusCode)))
```

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go
```diff
@@ -505,7 +505,7 @@ func TestOtelMiddleware_HTTPError(t *testing.T) {
 
 	spans := sr.Ended()
 	require.Len(t, spans, 1)
-	assertAttribute(t, spans[0].Attributes(), "error.type", "429 Too Many Requests")
+	assertAttribute(t, spans[0].Attributes(), "error.type", "429")
 	assert.Equal(t, codes.Error, spans[0].Status().Code)
 }
```
darccio, line 510: The PR description frames this as a cardinality fix ("error.type is meant to
be a low-cardinality error class... emitting the full status line breaks
metric/trace grouping"), but "429" and "429 Too Many Requests" have identical
cardinality — one status line per code, 1:1. The actual risk is format
conformance: backends group on `error.type` expecting the OTel-specified
numeric string, and status *reason phrases* aren't guaranteed to be identical
across servers/proxies for the same code — two different phrases for the same
status would fragment one group into several under the old behavior.

That framing is also the reason a single hardcoded case (429 with a fixed
reason phrase) doesn't fully cover the fix: it can't distinguish "we always
strip the code correctly" from "we happened to match this one status line."
Converting this into a table-driven test across a couple of codes (e.g. a
4xx and a 5xx, with distinct reason phrases) demonstrates the transformation
generalizes rather than just matching the one fixture.

```suggestion
func TestOtelMiddleware_HTTPError(t *testing.T) {
	tests := []struct {
		name       string
		statusCode int
		status     string
	}{
		{"client error", 429, "429 Too Many Requests"},
		{"server error", 500, "500 Internal Server Error"},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			sr := setupTestTracer(t)

			middleware := OtelMiddleware()

			req, _ := http.NewRequest(
				"POST",
				"http://api.anthropic.com/v1/messages",
				io.NopCloser(bytes.NewReader([]byte(`{"model":"claude-sonnet-4-5","max_tokens":10}`))),
			)

			next := func(r *http.Request) (*http.Response, error) {
				return &http.Response{
					StatusCode: tt.statusCode,
					Status:     tt.status,
					Header:     http.Header{},
					Body:       io.NopCloser(strings.NewReader("")),
				}, nil
			}

			resp, err := middleware(req, next)
			require.NoError(t, err)
			require.NotNil(t, resp)

			spans := sr.Ended()
			require.Len(t, spans, 1)
			assertAttribute(t, spans[0].Attributes(), "error.type", strconv.Itoa(tt.statusCode))
			assert.Equal(t, codes.Error, spans[0].Status().Code)
		})
	}
}
```

darccio (changes requested):

PR #786, opened 2026-07-23

### LINKED ISSUE #766: fix(ast): stripGenericTypes fatals on qualified pointer receivers
### Describe the bug

**Example crash:**
```
```

### Environment

 - Go Version: 1.25.0

### To Reproduce

**Option A (unit test):**

```go
```

2. Look it up with a rule:

```yaml
func: Write
recv: "*Writer"
```

4. See fatal:

```
```

```go
decl := &dst.FuncDecl{
    Type: &dst.FuncType{},
        Type: &dst.StarExpr{
            X: &dst.SelectorExpr{
            },
        },
    }}},
    Body: &dst.BlockStmt{},
}

    Func: "Write",
    Recv: "*Writer",
})
// fatals before returning
```

### Expected behavior

The tool should either:
- return no match gracefully

### Additional context

Adii-45:
## Description

This PR fixes receiver matching for qualified pointer receivers in `stripGenericTypes()`.

Previously, receiver rules such as `recv: "*Writer"` would fail when the actual receiver was a qualified pointer type (for example, `*bufio.Writer` or `*http.ResponseWriter`). `stripGenericTypes()` did not handle `*dst.SelectorExpr`, which caused `findFuncDecl()` to reach the fatal path instead of gracefully treating it as a non-match.

This change extends receiver type extraction to support qualified identifiers while preserving the existing behavior for regular and generic receivers.

### Changes

- Add support for qualified pointer and value receivers represented by `*dst.SelectorExpr`.
- Introduce a shared helper to resolve unqualified type names from receiver expressions.
- Replace the fatal path for unsupported receiver shapes with graceful "no match" handling instead of aborting instrumentation.
- Add regression tests covering:
  - qualified pointer receivers (`*bufio.Writer`, `*http.ResponseWriter`)
  - qualified value receivers
  - existing generic receiver behavior
  - unmatched receiver rules to ensure graceful handling

## Motivation

Rules targeting qualified pointer receivers could terminate instrumentation with:

```text
```

This happened because qualified receiver expressions were not recognized when extracting receiver type names. Instead of treating these as normal non-matches, the lookup aborted the entire instrumentation process.

This change ensures qualified receiver types are handled correctly while preserving existing behavior for all previously supported receiver forms and preventing unexpected crashes.

Fixes #766

---

## Checklist

- [x] Tests pass: `make test`
- [x] Tests follow Testing guidelines


Adii-45:
Hi @NameHaibinZhang, @kakkoyun, @amazingakai ,
All CI checks have passed, and I believe this PR is ready for review. I'd really appreciate any feedback whenever you have a chance. Thanks!

amazingakai:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/766#issuecomment-5141503703

tool/internal/ast/shared.go
```diff
@@ -47,41 +46,66 @@ func FindFuncDeclWithoutRecv(root *dst.File, funcName string) *dst.FuncDecl {
...
+// base of a generic receiver's type-parameter list.
+func baseTypeName(expr dst.Expr) (string, bool) {
+	switch x := expr.(type) {
+	case *dst.Ident:
+		return x.Name, true
+	case *dst.SelectorExpr:
+		return x.Sel.Name, true
+	}
+	return "", false
+}
```
darccio, line 107: Once the SelectorExpr branches above are removed, `baseTypeName` only needs to resolve `*dst.Ident` — its SelectorExpr case exists solely to support the now-removed qualified-generic-receiver case (`*pkg.Gen[T]`), which is equally illegal Go and unreachable.

```suggestion
// baseTypeName returns the unqualified type name of expr when it is an
// Ident (MyStruct), used to resolve the base of a generic receiver's
// type-parameter list.
func baseTypeName(expr dst.Expr) (string, bool) {
	if ident, ok := expr.(*dst.Ident); ok {
		return ident.Name, true
	}
	return "", false
}
```

tool/internal/ast/shared_test.go
```diff
@@ -109,6 +109,162 @@ func TestFindFuncDeclForRule(t *testing.T) {
...
+			Func: "Foo",
+			Recv: "*ResponseWriter",
+		}
+
+		fn, ok, err := FindFuncDecl(file, r)
+		require.NoError(t, err)
+		require.True(t, ok)
+		require.NotNil(t, fn)
+		assert.Equal(t, "Foo", fn.Name.Name)
+	})
```
darccio, line 181: These subtests assert `ok == true` for qualified pointer receivers. Once the `SelectorExpr` matchers are removed from `stripGenericTypes` (per the Area A discussion), `*bufio.Writer` and `*http.ResponseWriter` become unrecognized shapes and fall through to a no-match, not a match — this needs to assert the opposite.

```suggestion
	t.Run("qualified pointer receiver *bufio.Writer is not matched (illegal Go)", func(t *testing.T) {
		r := &rule.InstFuncRule{
			Func: "Flush",
			Recv: "*Writer",
		}

		fn, ok, err := FindFuncDecl(file, r)
		require.NoError(t, err)
		assert.False(t, ok)
		assert.Nil(t, fn)
	})

	t.Run("qualified pointer receiver *http.ResponseWriter is not matched (illegal Go)", func(t *testing.T) {
		r := &rule.InstFuncRule{
			Func: "Foo",
			Recv: "*ResponseWriter",
		}

		fn, ok, err := FindFuncDecl(file, r)
		require.NoError(t, err)
		assert.False(t, ok)
		assert.Nil(t, fn)
	})
```

tool/internal/ast/shared_test.go
```diff
@@ -109,6 +109,162 @@ func TestFindFuncDeclForRule(t *testing.T) {
...
+			name: "qualified pointer receiver *http.ResponseWriter",
+			src:  "package main\nimport \"net/http\"\nfunc (w *http.ResponseWriter) M() {}\n",
+			want: "*ResponseWriter",
+		},
+		{
+			name: "bare qualified value receiver",
+			src:  "package main\nimport \"bufio\"\nfunc (w bufio.Writer) M() {}\n",
+			want: "Writer",
+		},
+	}
```
darccio, line 258: Two things here: these three cases assert a positive match on illegal-Go qualified receivers, which should flip to `""` once the SelectorExpr matchers are removed. More importantly, this table has no coverage for `ParenExpr` receivers (`func (r (T)) M()`, `func (r (*T)) M()`) — the shape that actually compiles, actually appears in unformatted third-party source, and is the real crash this PR's `findFuncDecl` fix (return no-match instead of fataling) protects against. Verified both forms build with `go build` and parse to `*dst.ParenExpr` via this package's own parser, with `stripGenericTypes` already returning `""` for them.

```suggestion
		{
			name: "qualified pointer receiver is not matched (illegal Go)",
			src:  "package main\nimport \"bufio\"\nfunc (w *bufio.Writer) M() {}\n",
			want: "",
		},
		{
			name: "qualified value receiver is not matched (illegal Go)",
			src:  "package main\nimport \"bufio\"\nfunc (w bufio.Writer) M() {}\n",
			want: "",
		},
		{
			name: "parenthesized value receiver falls through to no match",
			src:  "package main\ntype T struct{}\nfunc (r (T)) M() {}\n",
			want: "",
		},
		{
			name: "parenthesized pointer receiver falls through to no match",
			src:  "package main\ntype T struct{}\nfunc (r (*T)) M() {}\n",
			want: "",
		},
	}
```

darccio (changes requested):

darccio:
Closed in favour of #1257

PR #789, opened 2026-07-23
fix(tool): resolve canFlattenTJump panic on unnamed HookContext parameter

### LINKED ISSUE #788: bug(optimize): canFlattenTJump panics when Before hook declares HookContext as an unnamed parameter
### Describe the bug

`canFlattenTJump` in `tool/internal/instrument/optimize.go` (line 229) panics with a
**runtime index-out-of-range** error when a Before hook function declares its first
parameter (the `HookContext`) without an explicit name — i.e. `func(HookContext, int)`
instead of `func(ctx HookContext, int)` or `func(_ HookContext, int)`.

In Go's AST (and in the `dst` library used here), a parameter written without a name
produces a `*dst.Field` whose `Names` slice is `nil`. The code unconditionally indexes
`Names[0]`, which panics on a nil/empty slice:

```go
// optimize.go:229  ← panics when Names is nil
hookContextParam := hookFunc.Type.Params.List[0].Names[0].Name
```

The existing `_ HookContext` (blank-identifier) path is safe because `_` produces a
non-nil `Names` slice. Only a truly unnamed first parameter triggers the bug.

**Inconsistency:** `collectArguments` in `apply_func.go` (lines 103–109) already guards
`field.Names == nil` before indexing. `canFlattenTJump` does not, making the two
functions inconsistent in their handling of unnamed parameters.

### Environment

 - OS: Linux / macOS / Windows (all platforms)
 - Go Version: 1.21+
 - Version: current `main` (commit where `canFlattenTJump` was introduced)

### To Reproduce

1. Create a Before hook whose **first parameter (`HookContext`) is unnamed** — valid Go,
   but `Names == nil` in the AST:

```go
// hooks/before.go
package hooks

import "go.opentelemetry.io/otelc/pkg/hook"

// Unnamed HookContext — valid Go, less common but legal
func BeforeMyFunc(hook.HookContext, *string) {
    // HookContext intentionally unused
}
```

2. Write a rule file with `before: BeforeMyFunc` and **no** `after` hook:

```yaml
# hooks/otelc.yaml
my_rule:
  target: example.com/mylib
  func:   MyFunc
  before: BeforeMyFunc
  path:   example.com/hooks
```

3. Run `otelc build ./...` against a project that depends on `example.com/mylib`.

4. The build crashes immediately with the panic shown below.

### Expected behavior

`canFlattenTJump` should treat a hook function whose first parameter has `Names == nil`
(fully unnamed) the same way it treats `_ HookContext`: the parameter is definitionally
unreferenced in the function body, so the function should return `true` (safe to flatten)
without panicking.

A minimal fix, consistent with the guard already present in `collectArguments`:

```go
// Proposed fix in canFlattenTJump (optimize.go ~L227)
params := hookFunc.Type.Params.List[0]
if len(params.Names) == 0 {
    // Fully unnamed first param — cannot be referenced in the body
    return true
}
hookContextParam := params.Names[0].Name
```

### Additional context

**Panic stack trace:**

```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
go.opentelemetry.io/otelc/tool/internal/instrument.canFlattenTJump(0xc0004e8000)
        tool/internal/instrument/optimize.go:229 +0x...
go.opentelemetry.io/otelc/tool/internal/instrument.(*InstrumentPhase).optimizeTJumps(...)
        tool/internal/instrument/optimize.go:386 +0x...
go.opentelemetry.io/otelc/tool/internal/instrument.(*InstrumentPhase).instrument(...)
        tool/internal/instrument/instrument.go +0x...
```

**Test gap:** The existing tests in `optimize_test.go` exercise only `_ HookContext`
(`Names[0].Name == "_"`) and `ctx HookContext` (named). No test covers the
unnamed-parameter (`Names == nil`) case.

**Duplicate search:** Checked all open and closed issues. Issue #777 (local/replace path
silent skip), #766 (`stripGenericTypes` fatal on pointer receivers), and #756 (signature
filter fatal on slice/chan/map params) are unrelated panics in different functions. No
existing issue matches `canFlattenTJump`, `Names[0]`, or unnamed HookContext parameters.

vishalmore90:
## Description

This PR fixes a runtime panic in `canFlattenTJump` (`tool/internal/instrument/optimize.go`) when a `Before` hook function declares its `HookContext` parameter without an explicit name (e.g., `func(HookContext, string)`).

In Go's `dst` AST representation, unnamed parameters result in `len(field.Names) == 0`. We now check for empty `Names` before attempting to access `Names[0]`. Unnamed parameters cannot be referenced in the function body and are guaranteed not to escape, so `canFlattenTJump` safely returns `true`.

## Motivation

Prevents `otelc build` from crashing with `panic: runtime error: index out of range [0] with length 0` when encountering valid Go hook functions declared with unnamed parameters.

Fixes #788

---

## Checklist

- [x] Tests pass: `make test`
- [x] OpenTelemetry Registry updated (if applicable, see [registry guide](docs/instrument-guide.md#4-register-the-instrumentation))

### DIFF
--- a/tool/internal/instrument/optimize.go
+++ b/tool/internal/instrument/optimize.go
@@ -203,8 +203,18 @@ func removeAfterTrampolineDecl(targetFile *dst.File, tjump *TJump) error {
 // 1. SetSkipCall is never called (so skip is always false)
 // 2. The HookContext parameter is only used as a receiver for method calls
 func canFlattenTJump(hookFunc *dst.FuncDecl) bool {
+	if hookFunc.Type == nil || hookFunc.Type.Params == nil || len(hookFunc.Type.Params.List) == 0 {
+		return true
+	}
+	firstParam := hookFunc.Type.Params.List[0]
+	if len(firstParam.Names) == 0 {
+		// If the parameter is unnamed, it cannot be referenced in the body
+		// Thus it doesn't escape and SetSkipCall cannot be called.
+		return true
+	}
+
 	// The hook context parameter is always the first parameter of the hook function.
-	hookContextParam := hookFunc.Type.Params.List[0].Names[0].Name
+	hookContextParam := firstParam.Names[0].Name
 
 	// Check if the hook function references SetSkipCall on the hook context.
 	// If found, the trampoline-jump-if cannot be flattened. Match the bare
--- a/tool/internal/instrument/optimize_test.go
+++ b/tool/internal/instrument/optimize_test.go
@@ -504,6 +504,24 @@ func TestFlattenTJump(t *testing.T) {
 			validate:      nil,
 		},
 		{
+			name: "can flatten with unnamed HookContext param",
+			hookSrc: `package main
+			func hookFunc(HookContext, string) {
+			}`,
+			canFlatten:    true,
+			removedOnExit: false,
+			validate:      nil,
+		},
+		{
+			name: "can flatten with no parameters",
+			hookSrc: `package main
+			func hookFunc() {
+			}`,
+			canFlatten:    true,
+			removedOnExit: false,
+			validate:      nil,
+		},
+		{
 			name: "flatten despite unrelated identifier substring",
 			hookSrc: `package main
 			func hookFunc(ctx HookContext, arg1 string) {


AkshitBhandariCodes:
I pulled this PR locally and tested the affected optimization path.

Validation:
- `go test ./tool/internal/instrument -run TestFlattenTJump -count=1` passed
- `go test ./tool/internal/instrument -count=1` passed

I also reviewed the changed files:
- `tool/internal/instrument/optimize.go`
- `tool/internal/instrument/optimize_test.go`

The added test case looks useful because it specifically covers a hook signature with an unnamed `HookContext` parameter, for example `func(HookContext, string)`. In that case, the AST parameter field can have an empty `Names` slice, so the previous logic could reach an out-of-range access while checking whether the `HookContext` escapes.

The change now handles that unnamed-parameter case without treating it as escaping, which matches the expected behavior for a valid hook signature. I also checked that the existing cases around named parameters, blank identifiers, escaping behavior, and `SetSkipCall` remain covered by the test table.

The PR looks scoped to #788, and I didn’t notice regressions in the affected optimization behavior.

amazingakai (approved):
Good catch, LGTM.

Thanks for the PR ❤️

darccio (approved):

PR #795, opened 2026-07-23
fix(net/http/server): satisfy http.Pusher so HTTP/2 server push works

### LINKED ISSUE #794: net/http server instrumentation breaks HTTP/2 server push (writerWrapper does not satisfy http.Pusher)
### Description

The `net/http` server instrumentation wraps the `http.ResponseWriter` in an internal `writerWrapper` (`instrumentation/net/http/server/response_writer.go`). To preserve HTTP/2 server push, the wrapper is meant to re-expose the `http.Pusher` interface. However, it defines the method as:

```go
func (w *writerWrapper) Pusher() http.Pusher { ... }
```

The `http.Pusher` interface requires:

```go
Push(target string, opts *http.PushOptions) error
```

The method name (`Pusher`) and signature do not match `http.Pusher`, so `writerWrapper` **never actually satisfies `http.Pusher`**. Because the embedded `http.ResponseWriter` only promotes the base interface's methods, the optional interface is silently dropped.

### Impact

Handlers detect server-push support with a type assertion on the writer they are handed:

```go
if p, ok := w.(http.Pusher); ok {
    p.Push("/style.css", nil)
}
```

Once this instrumentation is enabled, that assertion silently returns `ok == false`, so **HTTP/2 server push is disabled** for every instrumented handler. There is no error or log — behavior just quietly changes relative to the uninstrumented server.

### Expected behavior

The wrapper should transparently preserve the optional interfaces of the underlying `http.ResponseWriter` (`http.Pusher`, `http.Flusher`, `http.Hijacker`), so instrumentation does not change the capabilities visible to handlers.

### Proposed fix

- Implement `Push(target string, opts *http.PushOptions) error` on `writerWrapper`, forwarding to the underlying writer when it is an `http.Pusher` and returning `http.ErrNotSupported` otherwise.
- Add compile-time interface assertions (`var _ http.Pusher = (*writerWrapper)(nil)`, etc.) so future signature drift fails the build instead of silently dropping an interface.

### Environment

- Package: `instrumentation/net/http/server`

PRAteek-singHWY:
Fixes #794

### Problem

`writerWrapper` (`instrumentation/net/http/server/response_writer.go`) exposed a method named `Pusher() http.Pusher`. That name and signature do not match the `http.Pusher` interface, which requires `Push(target string, opts *http.PushOptions) error`. Consequently the wrapper never satisfied `http.Pusher`, and since the embedded `http.ResponseWriter` only promotes the base interface's methods, the optional interface was silently dropped.

Handlers detect push support with `if p, ok := w.(http.Pusher); ok`. Once this instrumentation was enabled, that assertion silently returned `false`, disabling HTTP/2 server push for every instrumented handler — with no error or log.

### Fix

- Replace `Pusher()` with a correct `Push(target string, opts *http.PushOptions) error` that forwards to the underlying writer when it supports push and returns `http.ErrNotSupported` otherwise (the sentinel handlers already expect on a non-push connection).
- Add compile-time interface assertions (`http.ResponseWriter`, `http.Hijacker`, `http.Flusher`, `http.Pusher`) so future signature drift fails the build instead of silently dropping an optional interface.

### Tests

- `TestWriterWrapper_ImplementsPusher` — asserts the wrapper actually satisfies `http.Pusher` via the type assertion real `net/http` code performs (the regression this guards against).
- `TestWriterWrapper_Push` — forwards to the underlying pusher.
- `TestWriterWrapper_Push_NotSupported` — returns `http.ErrNotSupported` for a non-push writer.

All tests pass (`go test ./...`) and `go vet ./...` is clean in the `net/http/server` module.


PRAteek-singHWY:
Hi @NameHaibinZhang
Quick update: signed the CLA, and EasyCLA is now green ✅. The branch is also up to date with `main`, so it should be ready for workflow approval / review whenever you have a moment.

Separately, I've been enjoying digging into the GenAI instrumentation in otelc, and if it's ever useful I'm happy to help out where I can, including on weekends. The Anthropic streaming follow-up under #679 is one piece I'd be comfortable picking up, since I've already traced the non-streaming path.

Thank you.

PRAteek-singHWY:
Thanks @kakkoyun, I ran `make package` locally and pushed the regenerated `tool/data/otelc-bundle.tgz` (e2f8813). It's built on top of the current branch head (latest `main` merged in) and includes the `http.Pusher` fix.

instrumentation/net/http/server/response_writer_test.go
```diff
@@ -162,30 +162,43 @@ func (m *mockPusher) Push(target string, opts *http.PushOptions) error {
...
+// http.Pusher. This is the assertion real net/http code performs.
+func TestWriterWrapper_ImplementsPusher(t *testing.T) {
+	var rw http.ResponseWriter = &writerWrapper{
+		ResponseWriter: &mockPusher{ResponseWriter: httptest.NewRecorder()},
+		statusCode:     http.StatusOK,
+	}
+
+	_, ok := rw.(http.Pusher)
+	assert.True(t, ok, "writerWrapper must satisfy http.Pusher so handlers can use HTTP/2 server push")
+}
```
amazingakai: We already have compile-time assertion, do we need this?

instrumentation/net/http/server/response_writer.go
```diff
@@ -10,6 +10,18 @@ import (
 	"net/http"
 )
 
+// Compile-time assertions that writerWrapper preserves the optional interfaces
+// a http.ResponseWriter may satisfy. Without these, a wrong method name or
+// signature silently drops an interface (see the Push regression these guard
+// against), because the embedded http.ResponseWriter only promotes the base
+// interface methods.
```
amazingakai: The compile-time assertions are useful, but I think the comment can be shortened to just explain their purpose.

instrumentation/net/http/server/response_writer.go
```diff
@@ -52,10 +64,19 @@ func (w *writerWrapper) Flush() {
...
+// Push implements the http.Pusher interface, forwarding to the underlying
+// ResponseWriter when it supports HTTP/2 server push. It returns
+// http.ErrNotSupported when the underlying writer is not an http.Pusher, which
+// is the sentinel handlers already expect on a non-push connection.
+//
+// The previous implementation exposed a method named Pusher() http.Pusher,
+// which does not match the http.Pusher interface (Push(string, *PushOptions)
+// error). As a result the wrapper did not satisfy http.Pusher at all, so a
+// handler's `if p, ok := w.(http.Pusher); ok` check silently failed once this
+// instrumentation was enabled, disabling HTTP/2 server push.
```
amazingakai: I don't think we need to document the previous `Pusher()` bug in the code itself. That's already captured by the PR and commit history, and the implementation is straightforward enough without the extra explanation.

amazingakai (commented):
I think the comments here are more detailed than necessary, but otherwise LGTM.

PRAteek-singHWY:
Thanks for the review @amazingakai! Addressed all three points in ab01469:

- Shortened the compile-time assertion comment to just state its purpose.
- Dropped the historical-bug narrative from the `Push` doc comment (it's captured in the PR/commit history).
- Removed `TestWriterWrapper_ImplementsPusher`, since the compile-time assertion already guarantees the wrapper satisfies `http.Pusher`.

Regenerated the bundle to match. Happy to trim further if anything still reads as too verbose.

amazingakai (approved):
@PRAteek-singHWY Good catch, LGTM

rbroggi:
Thanks for fixing `http.Pusher`. There is a related missing compatibility method: `writerWrapper` should implement `Unwrap() http.ResponseWriter`. Prometheus `promhttp` relies on this for `http.ResponseController` to access optional underlying writer capabilities. Without it, an instrumented controller-runtime Prometheus endpoint can return 404 despite `/metrics` being registered. I will open a follow-up PR with `Unwrap` and a regression test. Basically without the mentioned method we rule out any Kubernetes controller which is a sufficiently large audience to exclude from the functionality of otelc.

PRAteek-singHWY:
Thanks for the part @rbroggi. Agreed that ruling out controller-runtime based projects is too big a gap to leave open, so I've added `Unwrap` to this PR in 8414506 rather than deferring it to a follow-up.

`writerWrapper` now implements `Unwrap() http.ResponseWriter`, returning the embedded writer so `http.ResponseController` can walk the decorator chain and reach optional capabilities on the real writer.

Tests added:

- `TestWriterWrapper_Unwrap` checks that `Unwrap` returns the underlying writer.
- `TestWriterWrapper_ResponseControllerReachesUnderlyingWriter` drives a real `http.ResponseController` through `SetReadDeadline`. That capability is deliberately one the wrapper does not implement itself, so the test only passes if `Unwrap` is working. A `Flush` based test would pass either way, since `writerWrapper` has its own `Flush` method that `ResponseController` finds first.

No compile-time assertion for this one, since `net/http` does not declare an interface for `Unwrap`.

I also merged the latest `main` and regenerated `tool/data/otelc-bundle.tgz` from the merged tree, so the bundle carries both the new linodego instrumentation from #710 and this fix. That clears the conflict on the bundle.

@rbroggi since your #843 also re-applies the `Pusher` to `Push` change that already landed here, this PR should now cover everything in it. Feel free to close #843, or let me know if you'd rather I drop the `Unwrap` commit and we land yours instead.

Thank you.

rbroggi:
Closed the other MR. Looking forward to see this PR land a release so we can safely use otelc in k8s controllers.

PRAteek-singHWY:
Thanks @rbroggi Glad the Unwrap addition covers the controller-runtime case. Hopefully this lands in a release soon so k8s controllers can use otelc safely.


@kakkoyun @amazingakai, with Unwrap added and the bundle regenerated, this should be ready for a final look/merge whenever you have a moment.

This has been great learning.
Thank you.

NameHaibinZhang (approved):

PR #797, opened 2026-07-23
fix(tool/rule): validate required target field during rule loading

### LINKED ISSUE #796: fix(tool/rule): validate required target field during rule loading
### Describe the bug
When loading instrumentation rules from YAML, `validate()` methods in `tool/internal/rule` check specific fields like `func`, `path`, `raw`, and `struct`, but fail to validate that `target` is non-empty.
For example, in `tool/internal/rule/func_rule.go`:
```go
func (r *InstFuncRule) validate() error {
	if strings.TrimSpace(r.Func) == "" {
		return ex.Newf("func cannot be empty")
	}
	if strings.TrimSpace(r.Before) == "" && strings.TrimSpace(r.After) == "" {
		return ex.Newf("before or after must be set")
	}
	if strings.TrimSpace(r.Path) == "" {
		return ex.Newf("path cannot be empty")
	}
	return nil
}
If a YAML rule configuration omits the target: field, constructor functions like NewInstFuncRule succeed without returning an error. The rule passes setup validation, but fails silently later at package matching time without emitting a clear rule loading error to the developer.

Affected Files & Locations
[tool/internal/rule/base.go](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/tool/internal/rule/base.go#L92-L98): InstBaseRule struct defines Target string but lacks a default validateBase() method.
[tool/internal/rule/func_rule.go](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/tool/internal/rule/func_rule.go#L84-L95): validate() checks Func, Before/After, Path — skips Target.
[tool/internal/rule/raw_rule.go](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/tool/internal/rule/raw_rule.go#L51-L61): validate() checks Raw, Pattern, Placement — skips Target.
[tool/internal/rule/struct_rule.go](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/tool/internal/rule/struct_rule.go#L50-L55): validate() checks Struct — skips Target.
tool/internal/rule/call_rule.go: validate() checks Func, Path — skips Target.
Minimal Reproduction
yaml


# malformed_rule.yaml - missing required target field
rule:
  name: "example_rule"
  func: "MyFunc"
  before: "HookFunc"
  path: "example.com/hooks"
Running rule parsing on the snippet above currently succeeds instead of rejecting the rule:

go


rule, err := rule.NewInstFuncRule(data, "example_rule")
// err is currently nil, but should return: "invalid func rule \"example_rule\": target cannot be empty"
Expected Behavior
Rule constructors (NewInstFuncRule, NewInstRawRule, NewInstStructRule, NewInstCallRule) should return a validation error if target is empty or whitespace:

invalid func rule "example": target cannot be empty
Proposed Fix
Add a validateBase() method in tool/internal/rule/base.go:
go


func (r *InstBaseRule) validateBase() error {
	if strings.TrimSpace(r.Target) == "" {
		return ex.Newf("target cannot be empty")
	}
	return nil
}
Invoke validateBase() from rule validate() methods across func_rule.go, raw_rule.go, struct_rule.go, call_rule.go, etc.:
go


func (r *InstFuncRule) validate() error {
	if err := r.validateBase(); err != nil {
		return err
	}
	if strings.TrimSpace(r.Func) == "" {
		return ex.Newf("func cannot be empty")
	}
	// ...
}
Add unit test coverage in tool/internal/rule/*_test.go verifying that rule constructors return validation errors when target is missing or whitespace.

### PR DESCRIPTION
**Pull request by bhuvan-somisetty**
## Description

When loading instrumentation rules from YAML, `validate()` methods in `tool/internal/rule` validated specific fields like `func`, `path`, `raw`, `struct`, `directive`, `file`, etc., but failed to validate that `target` is non-empty. If a YAML rule configuration omitted the `target` field, constructors like `NewInstFuncRule` or `NewInstCallRule` succeeded, passing setup validation and failing silently later at package matching time.

This PR adds a `validateBase()` method in `tool/internal/rule/base.go` to validate that `r.Target` is non-empty and non-whitespace, and invokes `validateBase()` from the `validate()` method of all rule implementations (`InstFuncRule`, `InstRawRule`, `InstStructRule`, `InstCallRule`, `InstDeclRule`, `InstDirectiveRule`, `InstFileRule`).

Additionally, unit tests were updated and added across `tool/internal/rule/*_test.go` (including creating `raw_rule_test.go` and `struct_rule_test.go`) to ensure all rule constructors return validation errors when `target` is missing or whitespace.

## Motivation

Ensures malformed rules missing a required `target` field fail early with a clear validation error during rule loading instead of failing silently later during package matching.

Fixes #796

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -73,22 +73,13 @@ func parseRuleFromYaml(content []byte) ([]rule.InstRule, error) {
 				return nil, ex.Wrap(err1)
 			}
 
+			// target is the sole package selector and is required (docs/rules.md).
+			// Each rule type validates it (empty check plus glob syntax) via
+			// InstBaseRule.validateBase during construction below, so no further
+			// target validation is needed here.
 			r, err2 := createRuleFromFields(raw, name, flatFields)
 			if err2 != nil {
 				return nil, err2
-			}
-			// target is the sole package selector and is required (docs/rules.md).
-			// An empty or whitespace-only target would land under exactRules[""]
-			// and silently never match any real import path, so reject it loudly
-			// at load time instead.
-			if strings.TrimSpace(r.GetTarget()) == "" {
-				return nil, ex.Newf("rule %q has an empty target; target is required", name)
-			}
-			// Reject ambiguous/invalid glob targets at load time so a bad rule
-			// fails loudly during parsing rather than silently matching nothing
-			// during the setup phase.
-			if err3 := rule.ValidateTarget(r.GetTarget()); err3 != nil {
-				return nil, ex.Wrapf(err3, "rule %q", name)
 			}
 			if err3 := util.ValidateVersionRange(r.GetVersion()); err3 != nil {
 				return nil, ex.Wrapf(err3, "rule %q", name)
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -76,22 +76,13 @@ func parseRuleFromYaml(content []byte) ([]rule.InstRule, error) {
 				return nil, ex.Wrap(err1)
 			}
 
+			// target is the sole package selector and is required (docs/rules.md).
+			// Each rule type validates it (empty check plus glob syntax) via
+			// InstBaseRule.validateBase during construction below, so no further
+			// target validation is needed here.
 			r, err2 := createRuleFromFields(raw, name, flatFields)
 			if err2 != nil {
 				return nil, err2
-			}
-			// target is the sole package selector and is required (docs/rules.md).
-			// An empty or whitespace-only target would land under exactRules[""]
-			// and silently never match any real import path, so reject it loudly
-			// at load time instead.
-			if strings.TrimSpace(r.GetTarget()) == "" {
-				return nil, ex.Newf("rule %q has an empty target; target is required", name)
-			}
-			// Reject ambiguous/invalid glob targets at load time so a bad rule
-			// fails loudly during parsing rather than silently matching nothing
-			// during the setup phase.
-			if err3 := rule.ValidateTarget(r.GetTarget()); err3 != nil {
-				return nil, ex.Wrapf(err3, "rule %q", name)
 			}
 			if err3 := util.ValidateVersionRange(r.GetVersion()); err3 != nil {
 				return nil, ex.Wrapf(err3, "rule %q", name)
--- a/tool/internal/rule/base.go
+++ b/tool/internal/rule/base.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"go.opentelemetry.io/otelc/tool/ex"
 	"go.opentelemetry.io/otelc/tool/util"
 )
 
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

**Comment by bhuvan-somisetty** [comment]
Hi @NameHaibinZhang @kakkoyun @amazingakai — PR #797 is ready for review. It addresses #796 by ensuring that the required target field is validated when rules are loaded, catching missing targets early during setup. All unit tests and CI checks pass. Thanks!

**Review by amazingakai — CHANGES_REQUESTED** [review]
This repeats the same check everywhere, there is already a `rule.ValidateTarget` used by `parseRuleFromYaml`. Maybe we can update that to disallow blank targets?

**Comment by bhuvan-somisetty** [comment]
Good catch, thanks @amazingakai! You're right, that was unnecessary duplication.

I moved the empty-target check into `rule.ValidateTarget` itself, so it's now the single place that rejects blank/whitespace-only targets (in addition to the existing glob-pattern checks). `InstBaseRule.validateBase()` just delegates to it now instead of re-implementing the same check, and I dropped the now-redundant manual check in `parseRuleFromYaml` since every rule type already runs `ValidateTarget` via `validateBase()` during construction.

Also rebased onto main to pick up the `raw_rule_test.go` / `struct_rule_test.go` files that landed there in the meantime, and merged in the new "empty target" / "whitespace-only target" test cases so nothing was lost. All tests pass locally — should be good for another look once CI kicks off.

File: tool/internal/setup/match.go [hunk 1]
```diff
@@ -73,23 +73,14 @@ func parseRuleFromYaml(content []byte) ([]rule.InstRule, error) {
...
-			// at load time instead.
-			if strings.TrimSpace(r.GetTarget()) == "" {
-				return nil, ex.Newf("rule %q has an empty target; target is required", name)
-			}
-			// Reject ambiguous/invalid glob targets at load time so a bad rule
-			// fails loudly during parsing rather than silently matching nothing
-			// during the setup phase.
-			if err3 := rule.ValidateTarget(r.GetTarget()); err3 != nil {
-				return nil, ex.Wrapf(err3, "rule %q", name)
-			}
```
amazingakai, line 92: I'm confused, we already had validation for target, why are we moving it?

bhuvan-somisetty: Good question! Sorry, that comment in the diff was worded confusingly, let me clarify.

We're not really *moving* validation to a new place, more like removing a duplicate. Before this PR, `parseRuleFromYaml` had its own manual "is target empty" check plus a separate call to `rule.ValidateTarget`. Then this PR added `validateBase()`, which every rule type now calls from its own `validate()` during construction, and that already covers the empty-target case. So by the time `parseRuleFromYaml` got to its manual check, the target had already been validated twice: once implicitly through rule construction, and again explicitly right after. I just deleted the now-redundant second check, since `validateBase()` (which calls `ValidateTarget`) is the single source of truth now, whether a rule is built via `parseRuleFromYaml` or directly via `NewInstXRule` (like in the unit tests). Let me know if that makes sense or if you'd rather keep the explicit check in `parseRuleFromYaml` too, as a defensive belt-and-suspenders thing.

bhuvan-somisetty:
Hey @amazingakai, done — moved the empty-target rejection into `rule.ValidateTarget` itself so it's the single canonical check now, and dropped the duplicate checks in `parseRuleFromYaml` and `validateBase()`. Updated the tests accordingly and everything's green. Ready for another look whenever you get a chance, thanks for the feedback!

tool/internal/setup/match.go
amazingakai: This PR seems to be adding more duplication (validateBase call in every rule), there was no duplication in the original code?

bhuvan-somisetty:
@amazingakai Rebased onto latest `main` and resolved the merge conflicts. All unit, integration, E2E, and linter checks are passing green.

PR #799, opened 2026-07-23
fix(tool/instrument): remap blank identifier receiver in collectArguments

### LINKED ISSUE #798: bug(tool/instrument): collectArguments fails to remap blank identifier receiver (_)
## Summary

While looking through `tool/internal/instrument/apply_func.go`, I noticed that blank identifier receivers aren't handled consistently with blank parameters and named return values.

`collectArguments` already remaps unnamed parameters (and blank parameters such as `_`) to generated names like `_ignoredParam<N>` so their addresses can be taken safely during trampoline generation.

However, for a method with a blank receiver, for example:

```go
type MyService struct{}

func (_ MyService) Execute() {}
```

the receiver name is currently preserved as `"_"`.

Later, `createTrampArgs` assumes all argument names are addressable:

```go
func createTrampArgs(names []string) []dst.Expr {
    for _, name := range names {
        util.Assert(name != ast.IdentIgnore, "must be processed before")
        exprs = append(exprs, ast.AddressOf(name))
    }
}
```

As a result, instrumentation fails because the blank identifier was never remapped.

## Steps to reproduce

1. Define a method with a blank receiver:

   ```go
   type MyService struct{}

   func (_ MyService) Execute() {}
   ```
2. Add an `otelc` instrumentation rule targeting `Execute`.
3. Build with `otelc`.

## Expected behavior

Blank receivers should be handled the same way as blank parameters: remapped to an internal generated identifier before trampoline generation, allowing instrumentation to proceed normally.

## Additional context

This appears to be related to the work done for #736, which added support for blank parameters and blank named return values. The blank receiver case seems to have been missed, resulting in this remaining edge case.

om7057:
Fixes #798

## Summary

`collectArguments` already remaps unnamed parameters (`func(int)`) and blank parameters (`func(_ int)`) to generated names such as `_ignoredParam0`, so their addresses can be taken safely during trampoline generation.

Blank receivers were missing from that treatment. For `func (_ MyType) Method()`, `recv.Names` is non-nil and holds `"_"`, so `collectArguments` returned the blank identifier unchanged. `createTrampArgs` then tries to take its address, which trips `util.Assert(name != ast.IdentIgnore)` in debug builds and produces invalid Go (`&_`) otherwise.

This PR extends the existing `_ignoredParam<N>` remapping to cover blank receivers, consistent with how unnamed and blank parameters are already handled.

The broader name-collision-avoidance generalization raised in review (checking generated names against identifiers the target already declares) is intentionally out of scope here and lives on a separate branch (`fix-799-name-collision-avoidance`) for a follow-up PR.

## Testing

`TestCollectArguments`:

- `underscore receiver`

`TestCollectNamesNoCollision`, which type-checks the rendered file and so fails on a signature that would not compile:

- `blank receiver and blank named return.`

Both cases fail without the source change and pass with it. `./tool/...` passes; `gofmt` and `go vet` are clean.

---

## Checklist

- [x] Tests pass: `make test`
- [ ] Documentation updated (if applicable) (n/a)
- [ ] OpenTelemetry Registry updated (n/a, no new instrumentation)

[^I-know-my-stuff]:


om7057:
Hi @kakkoyun, @amazingakai ! Kindly go through this PR. All unit tests pass locally. Whenever you have some bandwidth, I'd really appreciate your feedback. 
Thank you!

Aryanbhargava18:
Hi @om7057, thanks for addressing the blank receiver case! 

I was looking at this and noticed that while injecting a generated name into the AST node (`recv.Names[0].Name = receiver`) fixes the immediate extraction, hardcoding `fmt.Sprintf("%s%d", ignoredParam, idx)` might accidentally introduce a parameter collision bug during bytecode compilation if the user's code already uses `_ignoredParam0`. 

We actually recently added a collision resolution helper (`collectNamesNoCollision`) for a very similar edge case! It cross-references the generated name against the symbol table to ensure no compiler panics. I'd be happy to point you to where the AST scope resolver lives in the toolchain if it helps you integrate it here? Let me know!

om7057:
> Hi @om7057, thanks for addressing the blank receiver case!
> 
> I was looking at this and noticed that while injecting a generated name into the AST node (`recv.Names[0].Name = receiver`) fixes the immediate extraction, hardcoding `fmt.Sprintf("%s%d", ignoredParam, idx)` might accidentally introduce a parameter collision bug during bytecode compilation if the user's code already uses `_ignoredParam0`.
> 
> We actually recently added a collision resolution helper (`collectNamesNoCollision`) for a very similar edge case! It cross-references the generated name against the symbol table to ensure no compiler panics. I'd be happy to point you to where the AST scope resolver lives in the toolchain if it helps you integrate it here? Let me know!


You're right that `collectNamesNoCollision` is intended to handle name collisions. My understanding is that this particular case is a bit different. A blank receiver (`func (_ R) Method()`) belongs to the target method, so there isn't a user-defined identifier for it to collide with in the instrumented source.

The `_ignoredParam<N>` naming is also already the pattern used by `collectArguments` for unnamed parameters (`field.Names == nil`) and blank parameters (`func(_ int)`), so this change simply extends the existing behavior to blank receivers rather than introducing a new naming scheme.

From what I could see, `collectNamesNoCollision` is currently used for hook function parameter naming, whereas `collectArguments` has its own logic for collecting target function arguments.

That said, if you'd prefer that this path also goes through `collectNamesNoCollision` for consistency, I'm happy to update it. I mainly wanted to explain why I followed the existing `collectArguments` pattern here.

om7057:
@kakkoyun @amazingakai: wanted to surface this briefly. The blank receiver case applies the existing collectArguments remapping logic consistently with how unnamed and blank parameters are handled. All tests pass. Please review it.

om7057:
Thanks @Aryanbhargava18, you were right about the collision and I was wrong to wave it off.

I said a blank receiver has no user identifier to collide with. That isn't true. The collision is with another parameter, not with the receiver. If the target already declares `_ignoredParam0`, `collectArguments` hands out the same name twice:

```
func (_ T) F(_ignoredParam0 int)   ->  [_ignoredParam0 _ignoredParam0]
```

The emitted signature then declares one binding twice, which doesn't compile.

One clarification on the pointer: I couldn't find `collectNamesNoCollision` as a function anywhere in the tree. `TestCollectNamesNoCollision` is a test, the regression added for #736, and that case was solved by giving blank named returns their own `_ignoredRetVal` prefix rather than by a symbol table lookup. If you had something else in mind, point me at it and I'll use that instead of what I wrote here.

The problem also isn't specific to blank receivers. It reproduces on `main` today:

```
func (T) F(_ignoredParam0 int)         ->  [_ignoredParam0 _ignoredParam0]
func F(_ int, _ignoredParam0 string)   ->  [_ignoredParam0 _ignoredParam0]
```

So I've fixed the class rather than just the blank receiver path. `declaredNames` collects the identifiers a target already declares across receiver, params and results, and generated names skip any that are taken. Numbering stays dense when nothing collides, so the existing expectations are unchanged.

Six new cases cover receiver, param and return collisions. Two of them go through `TestCollectNamesNoCollision`, which type-checks the rendered file, so they fail on a signature that wouldn't compile. All six fail without the source change and pass with it.

Aryanbhargava18 (approved):
Ah yeah, you're exactly right—`collectNamesNoCollision` was the regression test we added in #736, not the helper itself. My wires got crossed there! But I'm really glad you dug in and caught the underlying compiler panic vulnerability. The new `declaredNames` sweep is a much more robust way to solve the entire class of AST collisions anyway, rather than just patching the blank receiver. \n\nCode looks rock solid now. Approving!

om7057:
> Ah yeah, you're exactly right—`collectNamesNoCollision` was the regression test we added in #736, not the helper itself. My wires got crossed there! But I'm really glad you dug in and caught the underlying compiler panic vulnerability. The new `declaredNames` sweep is a much more robust way to solve the entire class of AST collisions anyway, rather than just patching the blank receiver. \n\nCode looks rock solid now. Approving!

Thanks for the review & approval, and for working through this with me.

tool/internal/instrument/apply_func.go
```diff
@@ -82,18 +82,57 @@ func collectReturnValues(funcDecl *dst.FuncDecl) []string {
...
+					names[name.Name] = struct{}{}
+				}
+			}
+		}
+	}
+	collect(funcDecl.Recv)
+	collect(funcDecl.Type.Params)
+	collect(funcDecl.Type.Results)
+	return names
+}
```
amazingakai: This does not account for other identifiers declared globally. 

We can maybe use `InstFuncRule.Identify()` to make the collision rarer.

amazingakai (changes requested):
Let's keep both of these patches in separate PRs @om7057

om7057:
> Let's keep both of these patches in separate PRs @om7057

@amazingakai  I've scoped this PR back down to just the blank receiver fix for #798 and kept the existing counter-based approach.

The collision-avoidance generalization has been moved onto a separate branch (`fix-799-name-collision-avoidance`) and pushed to my fork. Since there was a note from @NameHaibinZhang in Slack about keeping the review queue manageable, I'm holding off on opening that as a separate PR for now.

If you'd find it helpful to have it up for reference anyway, Can do it now as well if it makes sense to open now, or else I'll wait until there's a bit more bandwidth in the review queue.

tool/internal/instrument/apply_func.go
om7057: I traced both `collectArguments` call sites, and the `*rule.InstFuncRule` is already available at each, so I folded `Identity()` into all generated names rather than only the collision check.

That keeps it consistent with `makeName` for trampoline names (`_ignoredParam0_<identity>`), while `declaredNames` still handles collisions within the signature itself.

om7057:
@amazingakai just a small reminder on this one!

amazingakai:
The fix looks correct to me, please update the PR title/description @om7057

om7057:
@amazingakai 
Thanks for the review.
I've updated the title and description to match the current diff; this PR is now scoped to just the blank-receiver fix for #798 (the `_ignoredParam<N>` remapping extended to blank receivers). The name-collision-avoidance generalization is split onto `fix-799-name-collision-avoidance` on my fork.
Please let me know if I can open the follow up PR.

amazingakai (approved):
LGTM, thanks for this fix ❤️

om7057:
@amazingakai, the Copilot review has completed. Could you please take a look at it again?

amazingakai:
Already approved, please wait for a maintainer.

om7057:
> Already approved, please wait for a maintainer.

Thanks!
@darccio @txabman42 just a reminder on this one!

om7057:
@darccio just a small reminder on this one!

PR #802, opened 2026-07-24
chore: stop tracking otelc-bundle.tgz and generate it at build time

NameHaibinZhang:
The bundle is a binary archive derived from pkg/ and instrumentation/, so it conflicted on nearly every rebase. It is now a build artifact: make package regenerates it (and all build/install/test targets run package first), and it is gitignored rather than committed.

Removes the now-obsolete infra: the otelc-bundle merge driver (.gitattributes, merge-bundle.sh, make setup-git), the verify-bundle and regenerate-bundle workflows, and the related CONTRIBUTING docs.

Note: go install ...@latest is no longer supported since the proxy module won't contain the generated bundle; clone and run make install.

## Motivation

Fixes #

---

## Checklist

- [ ] Tests pass: `make test`


kakkoyun (changes requested):
We agreed on this temporary solution. This will be solved when Azhar implemented his RFC. Until then we need to live with this. User experience is more important than the developer experience here.

NameHaibinZhang:
Thanks for the quick feedback, @Azhar-Momin and @kakkoyun — that makes sense. Preserving the `go install ...@latest` / `go -get tool` user experience is the right priority, and I agree the proper fix is decoupling instrumentations (#585) rather than dropping the committed bundle. Closing this PR.

For the interim rebase-conflict pain, the GitHub App idea to replace the `/regenerate-bundle` action (so it also works on fork PRs) sounds good — happy to help if that's useful.

PR #805, opened 2026-07-24
fix(instrumentation): skip anthropic span creation for unparseable model

malladinagarjuna2:
## Description

`parseMessagesRequest` returns an empty `model` when the request body fails
to parse — either truncated JSON (the middleware caps request parsing at 1MB
via `maxRequestBodySize`, e.g. on a large multimodal payload) or a
well-formed body that simply omits the `model` field. Previously this still
produced a span named `"chat "` (trailing space) with
`gen_ai.request.model=""`.

This adds a guard: when `model == ""`, the request is passed through
uninstrumented, consistent with the other pass-through cases already in the
middleware (unknown operation, streaming, nil body, request read error).

## Motivation

A span with an empty model attribute is malformed telemetry — it can't be
attributed to any model in a backend, and the `"chat "` name (note the
trailing space) is visibly broken. Skipping span creation when there's
nothing meaningful to attach is better than emitting a garbled one.

## Testing

Added two cases: an unparseable request body and a valid body missing the
`model` field, both asserting no span is created and the request still
reaches the underlying transport.

`go test -race ./...` passes.


Satya900:
@malladinagarjuna2 Both new test cases are good - nice that you covered invalid JSON and valid-JSON-missing-model as two separate scenarios rather than lumping them together.

One thing worth flagging: I checked the sibling OpenAI instrumentation (`instrumentation/github.com/openai/openai-go{,/v2,/v3}/middleware.go`), and it has the identical gap this PR just fixed for Anthropic. Sending a request with a body that omits `model` still produces a span there, named `"chat "` (trailing space) with `gen_ai.request.model = ""` and every other `gen_ai.*` attribute empty.

Not something to fix in this PR, just flagging since it's the same failure mode. Might be worth a small follow-up applying the same `if model == "" { return next(req) }` guard to the OpenAI middleware for consistency across the GenAI instrumentations. Happy to open that as a separate issue/PR if useful.

Satya900 (approved):
Anthropic fix itself is correct and well-tested, approving. Restating for visibility: the identical gap (empty model still creates a span) exists unfixed in the sibling OpenAI instrumentation, not blocking here, just flagging again since it didn't get a reply last time.

malladinagarjuna2:
hey @Satya900  sorry for the late will implement it now

malladinagarjuna2:
hey i see that the pr has been  made for thechanges so i am not making those changes, really sorry for the mess

malladinagarjuna2:
@amazingakai  @kakkoyun @NameHaibinZhang  coul you please review

darccio (approved):
LGTM. @malladinagarjuna2, please regenerate the bundle and this will be good to merge.

Satya900:
@malladinagarjuna2  please resolve the conflicts

malladinagarjuna2:
hey i am sorry will do it now

malladinagarjuna2:
yeah done with the rebase @darccio

darccio:
@malladinagarjuna2 You'll need to regenerate the bundle.

darccio (approved):

darccio:
@malladinagarjuna2 Bundle again. Sorry. We are working on avoiding this on #585.

om7057:
Noticed this is marked stale and about to close. The blocker in the thread looks like it's purely the bundle regeneration, same friction I hit twice myself this week on #990 and #1288. Happy to help get this across the line if useful, either regenerating the bundle on your existing branch, or if you'd rather hand it off given the time gap, I can pick it up from here happily!
cc: @darccio

darccio:
Thanks @om7057!

om7057:
Opened #1310 with the same fix, same two test cases, credited to @malladinagarjuna2's original diagnosis here. 
cc: @darccio

darccio:
Closed in favour of #1310

PR #806, opened 2026-07-24
chore(tool): addition of `govulncheck` scanning in ci workflows

### LINKED ISSUE #793: Add govulncheck vulnerability scanning to CI

There is currently no vulnerability scanning of Go dependencies (or the Go standard library) anywhere in CI. The `.github/workflows/` directory has CodeQL (`codeql.yml`), FOSSA (`fossa.yml`), and OSSF Scorecard (`ossf-scorecard.yml`), but a grep for `govulncheck` across all workflows returns **0 hits**.

Those existing tools cover different things CodeQL finds bugs in our own code, FOSSA covers licensing, OSSF scores repo hygiene, but none of them flag a dependency (or a stdlib version) with a **published Go vulnerability (CVE /GO-XXXX advisory)** that our code actually reaches. 
For a security-sensitive, multi-module auto-instrumentation project this is a real blind spot: today we'd learn about a vulnerable dependency from an incident rather than from CI.

Add a new workflow that runs [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) across the repository:

- **New file:** `.github/workflows/govulncheck.yaml`
- Runs `govulncheck ./...`. `govulncheck` is low-noise by design — it reports only vulnerabilities in code paths that are actually called, not ever advisory in the dependency tree.
- **Must iterate over every module.** This repo is multi-module and there are **65 `go.mod` files** (`pkg/`, `pkg/runtime/`, `tool/`, and every `instrumentation/*/...`) — so a single top-level `govulncheck ./...` will not cover them all. The workflow should loop over module directories, mirroring how the existing Makefile / workflows already fan out across modules.
- **Triggers:** on `pull_request` **and** on a `schedule` (cron). The scheduled run is important, it catches newly-disclosed CVEs in dependencies that haven't changed since the last PR.
- **Pin the action to a commit SHA** per the repo's ratchet policy (see the `ratchet`-managed pins in existing workflows, e.g. `actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1` in `.github/workflows/test-unit.yaml`). Either the official `golang/govulncheck-action` (SHA-pinned) or a plain `go run golang.org/x/vuln/cmd/govulncheck@<version>` step is acceptable.

- **Rely on existing scanners (CodeQL / FOSSA / OSSF).** Rejected, none of them detect known Go CVEs in reachable dependency/stdlib code, which is exactly the gap `govulncheck` fills.
- **Dependabot/Renovate alerts only.** These flag advisories at the manifest level but don't do reachability analysis, so they're noisier and don't confirm whether the vulnerable code is actually called. `govulncheck` complements rather than replaces them.
- **A single top-level `govulncheck ./...`.** Rejected, it silently skips the other 64 modules in this multi-module repo.

**Additional context**

Refer other go repo inside Otel ecosystem and how they do `govulncheck`

atharvamhaske:
## Description

Adds reachability-based Go vulnerability scanning to CI via `govulncheck`, split into two jobs with different blocking behaviour.

**Workflow** (`.github/workflows/govulncheck.yaml`). Triggers on `pull_request` (Go sources, `go.mod`/`go.sum`, `Makefile`, and the workflow itself), a weekly `schedule`, and `workflow_dispatch`. Read-only default permissions, `persist-credentials: false` on checkout, SHA-pinned actions.

Both jobs install Go with `go-version: stable` rather than the `go.mod` version. govulncheck reports stdlib advisories against the toolchain running it, so pinning to `go.mod` would flag stdlib vulnerabilities the current release has already fixed.

`govulncheck (core)` runs `make govulncheck`, a source-mode scan of the root module (which covers `tool/`) plus every `pkg/**` module. It is soft on pull requests via `continue-on-error: ${{ github.event_name == 'pull_request' }}`, since a newly disclosed CVE with no fix released yet should not red unrelated work. It blocks on the weekly and manual runs: job-level `continue-on-error` concludes the whole run as success, which would leave a red job buried inside a green scheduled run with nothing to act on.

`govulncheck (instrumentation)` runs `make govulncheck/instrumentation` and blocks on every trigger. It builds each top-level app under `test/apps/` with `otelc go build`, then scans the resulting binary with `govulncheck -mode=binary`. Binary mode is necessary here because a source scan of `instrumentation/` skips `//go:build ignore` hook sources and cannot typecheck modules that rely on compile-time field injection, such as `database/sql`. Demo apps are excluded, since their example dependencies are pinned deliberately.

**Tool pinning.** `govulncheck` is pinned in `.tools/go.mod` and `.tools/tools.go` and built into `.bin/` by the same rule as every other repo tool. Renovate's existing `gomod` handling keeps it current, so no dedicated version variable or customManager is needed.

**Renovate** (`.github/renovate.json5`). Adds `minimumReleaseAge: '5 days'` and `internalChecksFilter: 'strict'` as a supply-chain cooldown.

**Dependency fix.** Bumps `google.golang.org/grpc` to `v1.82.1` in every module outside `demo/`. That release fixes [GO-2026-6061](https://pkg.go.dev/vuln/GO-2026-6061) (xDS RBAC authorization bypass and HTTP/2 transport server DoS), which the new instrumentation job reports as reachable through `grpc.Server.Serve`.

## Motivation

There is no reachability-based Go vulnerability scanning in CI today. CodeQL, FOSSA, and OSSF cover different gaps, and none flag a dependency or stdlib version with a published Go advisory that our code actually reaches. `govulncheck` closes that gap and is low noise because it only reports vulnerabilities in code paths that are actually called.

The weekly scheduled run catches newly disclosed advisories in dependencies that have not changed since the last PR. The Renovate cooldown reduces the risk of pulling in a freshly published, possibly compromised release before it has been vetted. It is scoped to Renovate on purpose, since Dependabot security updates (enabled via repo settings,
for example PR #803 bumping `golang.org/x/net`) should land fast and not be delayed. In this repo Renovate owns routine version bumps; Dependabot only fires for known vulnerabilities.

## Checklist

- [x] Tests pass: `make test`


atharvamhaske:
PTAL @amazingakai @kakkoyun !

Makefile
```diff
@@ -397,6 +403,20 @@ check-golden-files: package
...
+	@echo "Running govulncheck across $(words $(GOVULNCHECK_MODULES)) modules..."
+	@set -uo pipefail
+	@status=0; \
+	for moddir in $(GOVULNCHECK_MODULES); do \
+		echo "==> govulncheck $$moddir"; \
+		(cd "$$moddir" && go run golang.org/x/vuln/cmd/govulncheck@$(GOVULNCHECK_VERSION) ./...) || status=1; \
+	done; \
+	if [ "$$status" -ne 0 ]; then echo "govulncheck: vulnerabilities found"; exit 1; fi; \
+	echo "govulncheck: no reachable vulnerabilities found"
+
```
amazingakai, line 437: Should `govulncheck` be managed via the `.tools` module instead? I think renovate would update `.tools/go.mod` automatically, which would also let us drop the custom regex manager for `GOVULNCHECK_VERSION`.

Makefile
```diff
@@ -29,6 +29,12 @@ TOOLS_DIR = .tools
 GO_VERSION = 1.25
 INTEGRATION_TEST_RUN ?= .
 
+# govulncheck version (Renovate-managed). Scans shipped modules for known Go CVEs.
+GOVULNCHECK_VERSION = v1.6.0
+# Shipped modules: root module (covers tool/), plus every pkg/ and instrumentation/
+# module. Test apps and demos are intentionally excluded (pinned example deps).
+GOVULNCHECK_MODULES := . $(shell find pkg instrumentation -type f -name 'go.mod' -exec dirname {} \; | sort)
```
amazingakai: We will probably need to skip some modules that rely on compile-time transformations (e.g, database/sql instrumentation)

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          persist-credentials: false
+      - name: Install Go
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c  # v6.1.0
+        with:
+          go-version-file: go.mod
```
amazingakai: `govulncheck` is reporting several vulnerabilities in go stdlib library. Should we use the latest stable Go toolchain for this workflow instead of the version from `go.mod`?

For reference, this is what `dd-trace-go` is using:

https://github.com/DataDog/dd-trace-go/blob/cf24b817c453c18c65250da5a0acbb029f834617/.github/workflows/govulncheck.yml#L65

It might be worth taking a look at that workflow to see if there are ideas we could adopt to improve this check as well. I'm not very knowledgeable in this area, though.

Makefile
kakkoyun: Instrumentation vulnerabilities are more critical. We should even split their checks. This is what we do in `dd-trace-go` as well. (we use contrib name in that repo).

kakkoyun (dismissed):

Makefile
atharvamhaske: So do we have to do same as you did in dd repo? i even referred those pr but i thought maybe it was out of scope, fair i will do that

.github/workflows/govulncheck.yaml
atharvamhaske: Yeah makes sense will check this out, Thanks

atharvamhaske:
@kakkoyun what i did now is created two separate `golvuncheck` scan jobs as **core** and **instrumentation** vulns(blocking).
I kept instru one blocking purposefully, so vulns which are reachable from injected code and higher impact gets their own **hard-blocking** check unlike core(root, pkg) runs **non-blocking** so a stdlib or any dependency CVE with no fix released yet does not red unrelated PRs on tooling code.

used the latest stable go toolchain rather than go.mod version as suggested by @amazingakai and added `govulncheck-instrumentation` target in Makefile.

atharvamhaske:
Moved the pinned `govulncheck` version from a hand-rolled Makefile variable into the `.tools module`, alongside every other CI tool. 
Renovate already tracks `.tools/go.mod `as a Go module, so this drops the custom regexmanager that existed only to bump `GOVULNCHECK_VERSION`.

The golang.org/x/mod, x/sync, x/sys, x/tools and x/telemetry bumps are **govulncheck v1.6.0's** declared minimum requirements, pulled in by module version selection.

Makefile
amazingakai: > We will probably need to skip some modules that rely on compile-time transformations (e.g, database/sql instrumentation)

This is still not addressed. I'm thinking since `govulncheck` supports `-mode=binary`, can we run it against our integration test apps instead?

Currently, `govulncheck` would ignore files behind `//go:build ignore` if we scan the instrumentation modules directly. We'd also need to skip modules like `database/sql` since they rely on compile-time transformations.

If we scan the final instrumented binary instead, neither of those would be a problem.

atharvamhaske: > > We will probably need to skip some modules that rely on compile-time transformations (e.g, database/sql instrumentation)
> 
> This is still not addressed. I'm thinking since `govulncheck` supports `-mode=binary`, can we run it against our integration test apps instead?
> 
> Currently, `govulncheck` would ignore files behind `//go:build ignore` if we scan the instrumentation modules directly. We'd also need to skip modules like `database/sql` since they rely on compile-time transformations.
> 
> If we scan the final instrumented binary instead, neither of those would be a problem.

hey i have made the changes, suggested with latest commit.

Makefile
```diff
@@ -406,6 +421,55 @@ check-golden-files: package
...
+govulncheck-instrumentation: $(GOVULNCHECK) build ## Scan instrumented test apps (binary mode) for known Go vulnerabilities
+	@echo "Running govulncheck -mode=binary across $(words $(GOVULNCHECK_TEST_APPS)) instrumented test apps..."
+	@set -uo pipefail
+	@status=0
+	@app_bin=app$(EXT)
+	@cleanup() { \
+		find test/apps -mindepth 2 -maxdepth 2 \( -name "$$app_bin" -o -name '.otelc-build.lock' \) -delete 2>/dev/null || true; \
+		find test/apps -type d -name '.otelc-build' -exec rm -rf {} + 2>/dev/null || true; \
+	}; \
+	trap cleanup EXIT; \
```
amazingakai: otelc should handle the cleanup

Makefile
```diff
@@ -406,6 +421,55 @@ check-golden-files: package
...
+define run_govulncheck
+	@set -uo pipefail
+	@status=0; \
+	for moddir in $(1); do \
+		echo "==> govulncheck $$moddir"; \
+		(cd "$$moddir" && $(GOVULNCHECK) ./...) || status=1; \
+	done; \
+	if [ "$$status" -ne 0 ]; then echo "govulncheck: vulnerabilities found"; exit 1; fi; \
+	echo "govulncheck: no reachable vulnerabilities found"
+endef
```
amazingakai: This can be inlined now, since its used only in once place

amazingakai (approved):
Looks good to me, both my comments are non-blocking.

Let's wait for @kakkoyun to review.


The failing deps would need to be bumped I think.

Makefile
atharvamhaske: yeah makes sense :)

Makefile
atharvamhaske: done

Makefile
```diff
@@ -406,6 +421,49 @@ check-golden-files: package
...
+		(cd "$$moddir" && "$(GOVULNCHECK)" ./...) || status=1; \
+	done; \
+	if [ "$$status" -ne 0 ]; then echo "govulncheck: vulnerabilities found"; exit 1; fi; \
+	echo "govulncheck: no reachable vulnerabilities found"
+
+# Build each integration test app with otelc, then scan the resulting binary.
+# Binary mode sees injected instrumentation (including //go:build ignore sources
+# and field-injection modules like database/sql) that source mode cannot analyze.
+.ONESHELL:
+govulncheck-instrumentation: $(GOVULNCHECK) build ## Scan instrumented test apps (binary mode) for known Go vulnerabilities
```
```suggestion
govulncheck/instrumentation: $(GOVULNCHECK) build ## Scan instrumented test apps (binary mode) for known Go vulnerabilities
```

This should be like this to follow the other command conventions.

kakkoyun (commented):
Nearly there. Thanks.

The CI is also not happy.

Makefile
atharvamhaske: oh yes thanks for flagging!

atharvamhaske:
> Nearly there. Thanks.
> 
> The CI is also not happy.

what else can be added? you can put your thoughts or what i might be missing :)

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          persist-credentials: false
+      - name: Install Go
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c  # v6.1.0
+        with:
+          go-version: stable
+          cache-dependency-path: "**/go.mod"
+      - name: Build otelc and scan instrumented test apps
+        run: make govulncheck-instrumentation
```
kaizakin: hey @atharvamhaske Makefile target is renamed but the caller isn't updated yet.

kaizakin (dismissed):

kaizakin:
@atharvamhaske bumping `google.golang.org/grpc@v1.81.1` in `pkg/runtime/go.mod` to `google.golang.org/grpc@v1.82.1` made the CI check PASS locally

as mentioned by `govulncheck` in the logs.

amazingakai:
CI needs to be fixed.

atharvamhaske:
> CI needs to be fixed.

idk whether they are related to this pr cuz there was discussion on slack few days ago that on each pr we get some ci errors like this? im not sure still will check

amazingakai:
I think you just need to run `make go-mod-tidy` + `make package`

atharvamhaske:
@amazingakai can you ptal now?

Makefile
```diff
@@ -578,7 +636,9 @@ test-integration: build build-demo
 test-latestlibbuild: build ## Run LatestLibBuild tests
 	@echo "Running LatestLibBuild tests..."
 	set -euo pipefail
-	go -C "test" test -json -v -shuffle=on -timeout=20m -count=1 -tags latestlibbuild ./latestlibbuild/... 2>&1 | tee ./gotest-latestlibbuild.log
+	# 40m: instrumented builds run sequentially, and linodego alone can take
+	# several minutes on Windows.
+	go -C "test" test -json -v -shuffle=on -timeout=40m -count=1 -tags latestlibbuild ./latestlibbuild/... 2>&1 | tee ./gotest-latestlibbuild.log
```
amazingakai, line 641: I'm not sure if we should raise this (cc @txabman42)

atharvamhaske:
@darccio can you ptal

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,62 @@
...
+permissions: read-all
+
+jobs:
+  # Core modules (root, pkg). Non-blocking: a newly-disclosed stdlib/dep CVE with
+  # no fix released yet shouldn't red unrelated PRs on tooling code. Still visible
+  # as a failed check. Instrumentation (higher impact) stays hard-blocking below.
+  govulncheck:
+    name: govulncheck (core)
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
darccio: The core job (source scan of root + pkg, which includes tool/ — the
instrumentation compiler itself) is soft-fail via continue-on-error, while
the instrumentation job (scanning already-built test app binaries) is hard-
blocking. A reachable vulnerability in the compiler that emits instrumented
code arguably deserves at least the same blocking treatment as one found in
an instrumented test app. Is the asymmetry intentional, and if so, what
surfaces a core-job failure for someone to act on it?

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,62 @@
...
+  pull_request:
+    paths:
+      - '**/*.go'
+      - '**/go.mod'
+      - '**/go.sum'
+      - 'Makefile'
+      - '.github/workflows/govulncheck.yaml'
+  schedule:
+    # Weekly, so newly-disclosed CVEs in unchanged dependencies are still caught.
+    - cron: '0 6 * * 1'
```
darccio, line 13: On the weekly schedule (and workflow_dispatch) there's no PR to turn red.
Given the core job stays non-blocking and instrumentation stays blocking,
what surfaces a scheduled-run failure — a follow-up notification/issue-filing
step, or does this rely on someone checking the Actions tab periodically?

Makefile
```diff
@@ -397,6 +403,20 @@ check-golden-files: package
...
+.ONESHELL:
+govulncheck: ## Scan shipped modules (root, pkg, instrumentation) for known Go vulnerabilities
+	@echo "Running govulncheck across $(words $(GOVULNCHECK_MODULES)) modules..."
+	@set -uo pipefail
+	@status=0; \
+	for moddir in $(GOVULNCHECK_MODULES); do \
+		echo "==> govulncheck $$moddir"; \
+		(cd "$$moddir" && go run golang.org/x/vuln/cmd/govulncheck@$(GOVULNCHECK_VERSION) ./...) || status=1; \
+	done; \
+	if [ "$$status" -ne 0 ]; then echo "govulncheck: vulnerabilities found"; exit 1; fi; \
```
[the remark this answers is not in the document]
darccio: Echoing on Copilot's comment, govulncheck ./... exits non-zero both when it finds a reachable vulnerability and when the module simply fails to build/load (bad import, syntax error, unresolved dependency). This line always prints "vulnerabilities found"
regardless of which happened, which is misleading for anyone debugging a red
build. The sibling govulncheck/instrumentation target below already hedges
correctly ("vulnerabilities found or build failed") — worth doing the same
here, e.g.:

```suggestion
	if [ "$$status" -ne 0 ]; then echo "govulncheck: vulnerabilities found or a module failed to load"; exit 1; fi; \
```

.github/renovate.json5
```diff
@@ -8,6 +8,11 @@
     'before 8am on the 1st and 15th day of the month',
   ],
   updateNotScheduled: false,
+  // Cooldown: don't raise a bump PR until a release has aged, so yanked or
+  // compromised releases are caught before they enter the tree. Renovate's
+  // equivalent of Dependabot's `cooldown`.
+  minimumReleaseAge: '5 days',
```
darccio, line 14: minimumReleaseAge is set at the top level, so it applies to every Renovate-
managed dependency, not just the ones this PR is worried about. The PR body
says this is deliberate (Dependabot handles fast security bumps separately),
which is coherent — but does Renovate's *own* vulnerability-alert flow (as
opposed to GitHub/Dependabot security alerts) bypass this cooldown, or would
a CVE that Renovate itself surfaces also sit in cooldown for 5 days? Worth
confirming against Renovate's docs/behavior for this instance before merge.

Makefile
darccio: This timeout bump (20m → 40m for Windows latestlibbuild flakiness) is
unrelated to govulncheck — it looks like a legitimate fix, but bundling it
into a "chore(ci): govulncheck" PR makes the diff harder to review/bisect
and buries an unrelated CI-reliability fix in an unrelated changelog entry.
Worth splitting into its own PR, or at minimum calling it out explicitly in
the PR description as an unrelated fix picked up along the way.

Makefile
```diff
@@ -29,6 +29,18 @@ TOOLS_DIR = .tools
 GO_VERSION = 1.25
 INTEGRATION_TEST_RUN ?= .
 
+# Modules/apps scanned by govulncheck for known Go CVEs (tool version pinned in
+# .tools/go.mod, Renovate-managed like every other .tools binary).
+# Core modules: root module (covers tool/) plus every pkg/ module.
+# Instrumentation is scanned separately via instrumented integration-test binaries
```
darccio, line 37: The comment here explains why instrumentation/ is scanned via test-app
binaries instead of source scan, but doesn't mention that coverage is only
as good as the test apps' dependency versions. Concretely,
instrumentation/github.com/openai/openai-go/v2 and /v3 have no exercising
test app anywhere in test/apps/ — only v1 is used (by openaiclient). Those
two modules get zero govulncheck coverage: not caught by the core scan
(instrumentation/ is excluded by design) and not caught by the binary scan
(nothing builds against them). Worth a one-line addition to this comment
documenting the gap, so it's a known, visible limitation rather than a
silent one.

darccio (changes requested):
The PR description doesn't match what was merged. It describes a
GOVULNCHECK_VERSION variable, a single loop over "27 shipped modules", and
a Renovate customManager to keep that variable current — none of these
exist in the final code. What actually shipped is cleaner: govulncheck is
pinned via .tools/go.mod like every other repo tool (tracked by the
existing generic Renovate gomod rule, no dedicated customManager needed),
and core/instrumentation are two differentiated jobs rather than one
27-module loop. Please update the PR description to match the implementation
that's actually being reviewed/merged.

The grpc bump (v1.81.1 → v1.82.1) is incomplete. v1.82.1 isn't a routine
bump — it's the fix release for GO-2026-6061 (xDS RBAC authorization bypass
+ HTTP/2 transport server DoS in grpc-go, CVSS 8.2/7.5), almost certainly
the vulnerability this PR's govulncheck integration was meant to close.

Four modules under test/apps/ were left on v1.82.0, one patch behind, even
though every other module (including instrumentation/google.golang.org/grpc/{client,server}
and test/apps/grpcclient) was correctly bumped to v1.82.1:

- test/apps/grpcserver/go.mod:9
- test/apps/grpcserverkafkaclient/go.mod:10
- test/apps/grpcserverdbclient/go.mod:12
- test/apps/grpcservergrpcclient/go.mod:9

This isn't theoretical — I ran govulncheck directly against test/apps/grpcserver
in this state and it reports the vulnerability live, with a concrete call
trace through grpc.Server.Serve:

  Vulnerability #2: GO-2026-6061
    Module: google.golang.org/grpc
    Found in: google.golang.org/grpc@v1.82.0
    Fixed in: google.golang.org/grpc@v1.82.1
    Example traces found:
      #1: main.go:56:2: grpcserver.main calls net.UnixListener.Close, which
          eventually calls transport.NewHTTP2Client
      #2: main.go:62:19: grpcserver.main calls grpc.Server.Serve, which
          eventually calls transport.NewServerTransport

All four of these modules are top-level dirs under test/apps/, which means
they're in GOVULNCHECK_TEST_APPS and get scanned by the new
govulncheck-instrumentation CI job — the one this PR makes hard-blocking.
As merged, that job will fail on these apps on the very grpc-server code
paths the bump was meant to fix.

Fix: in each of the four go.mod files above, bump to
`google.golang.org/grpc v1.82.1` and run `go mod tidy`, matching the rest
of the PR's bump.

(Also left on older grpc versions but outside the scanned set, so lower
priority: test/go.mod and test/shared/grpcpb/go.mod on v1.80.0, and
demo/app/grpc/{client,server}/go.mod on v1.82.0 — demos are excluded from
scanning by design per the Makefile comment.)

atharvamhaske:
> The PR description doesn't match what was merged. It describes a
> GOVULNCHECK_VERSION variable, a single loop over "27 shipped modules", and
> a Renovate customManager to keep that variable current — none of these
> exist in the final code. What actually shipped is cleaner: govulncheck is
> pinned via .tools/go.mod like every other repo tool (tracked by the
> existing generic Renovate gomod rule, no dedicated customManager needed),
> and core/instrumentation are two differentiated jobs rather than one
> 27-module loop. Please update the PR description to match the implementation
> that's actually being reviewed/merged.
> 
> The grpc bump (v1.81.1 → v1.82.1) is incomplete. v1.82.1 isn't a routine
> bump — it's the fix release for GO-2026-6061 (xDS RBAC authorization bypass
> + HTTP/2 transport server DoS in grpc-go, CVSS 8.2/7.5), almost certainly
> the vulnerability this PR's govulncheck integration was meant to close.
> 
> Four modules under test/apps/ were left on v1.82.0, one patch behind, even
> though every other module (including instrumentation/google.golang.org/grpc/{client,server}
> and test/apps/grpcclient) was correctly bumped to v1.82.1:
> 
> - test/apps/grpcserver/go.mod:9
> - test/apps/grpcserverkafkaclient/go.mod:10
> - test/apps/grpcserverdbclient/go.mod:12
> - test/apps/grpcservergrpcclient/go.mod:9
> 
> This isn't theoretical — I ran govulncheck directly against test/apps/grpcserver
> in this state and it reports the vulnerability live, with a concrete call
> trace through grpc.Server.Serve:
> 
>   Vulnerability #2: GO-2026-6061
>     Module: google.golang.org/grpc
>     Found in: google.golang.org/grpc@v1.82.0
>     Fixed in: google.golang.org/grpc@v1.82.1
>     Example traces found:
>       #1: main.go:56:2: grpcserver.main calls net.UnixListener.Close, which
>           eventually calls transport.NewHTTP2Client
>       #2: main.go:62:19: grpcserver.main calls grpc.Server.Serve, which
>           eventually calls transport.NewServerTransport
> 
> All four of these modules are top-level dirs under test/apps/, which means
> they're in GOVULNCHECK_TEST_APPS and get scanned by the new
> govulncheck-instrumentation CI job — the one this PR makes hard-blocking.
> As merged, that job will fail on these apps on the very grpc-server code
> paths the bump was meant to fix.
> 
> Fix: in each of the four go.mod files above, bump to
> `google.golang.org/grpc v1.82.1` and run `go mod tidy`, matching the rest
> of the PR's bump.
> 
> (Also left on older grpc versions but outside the scanned set, so lower
> priority: test/go.mod and test/shared/grpcpb/go.mod on v1.80.0, and
> demo/app/grpc/{client,server}/go.mod on v1.82.0 — demos are excluded from
> scanning by design per the Makefile comment.)

hey thanks, will check all this and make changes

atharvamhaske:
Thanks for the detailed review, both points were right.

## gRPC bump

Completed. All four `test/apps` modules are on `v1.82.1`, and I found two more beyond your list while sweeping:

- `instrumentation/go.mod` was also still on `v1.82.0`. That one matters more than the test apps: it owns `instrumentation/google.golang.org/grpc/semconv`, which the gRPC client and server hooks import, and it ships inside the embedded bundle. Bumped and regenerated the bundle.
- `test/go.mod` and `test/shared/grpcpb` (both `v1.80.0`), plus `test/bench/scenarios/multi` (`v1.82.0`).

Every module outside `demo/` now resolves to `v1.82.1`. Demos stay pinned, matching the Makefile comment that excludes them from scanning.

Verified with the same binary-mode check CI runs, not just a source scan:

| App | grpc in binary | otelc modules linked | GO-2026-6061 |
|---|---|---|---|
| `grpcserver` | v1.82.1 | 11 | not reported |
| `grpcserverkafkaclient` | v1.82.1 | 10 | not reported |
| `grpcserverdbclient` | v1.82.1 | 9 | not reported |
| `grpcservergrpcclient` | v1.82.1 | 8 | not reported |

`go version -m` confirms the gRPC hooks and the umbrella module are genuinely linked into those binaries, so this exercises the injected code paths rather than a bare build.

## Core job asymmetry

Fair challenge, and looking into it my reasoning in that comment was wrong. Rather than argue from the docs (which are ambiguous about job-level `continue-on-error`) I checked what GitHub actually recorded for this workflow's own runs. In run 30823112799, `govulncheck (core)` concluded `failure` while the run concluded `success`:

| Run | core | instrumentation | run conclusion |
|---|---|---|---|
| 30823112799 | **failure** | success | **success** |
| 31167757265 | success | failure | failure |

So the job does go red, but the workflow stays green and nothing gates on it.

That trade-off is defensible on a pull request, since a newly disclosed CVE with no fix released shouldn't red unrelated work. It is not defensible on the weekly run, which exists to catch exactly those: the run concludes green, and GitHub's scheduled-failure email only fires on run failure and only reaches whoever created the workflow or last
edited the cron. So to answer your second question directly: as it stood, nothing surfaced a scheduled core-only failure.

**If you have anything other than this in your mind please lmk, i am open for discussion.**

Fixed by making the softness conditional on the trigger rather than unconditional:

```yaml
continue-on-error: ${{ github.event_name == 'pull_request' }}
```

cc: @darccio

atharvamhaske:
> This timeout bump (20m → 40m for Windows latestlibbuild flakiness) is unrelated
> to govulncheck... Worth splitting into its own PR

Agreed it shouldn't be here. Dropping it rather than splitting it, because it's also obsolete now.

main shards the suite across two jobs per platform via `SHARD_INDEX`/`SHARD_TOTAL`, and `shardConfig` in latestlibbuild_test.go says why: "Each app is a full `go build -a`, so a single job building every app exceeds the slower runners' test timeout; sharding keeps each job bounded." That fixes the root cause my bump was papering over, and main correctly kept `timeout=20m`. Leaving 40m in would mask the bound sharding is there to enforce.

Reverted in 9c20ed3, so the target now matches main exactly and the Makefile diff in this PR is govulncheck-only.

atharvamhaske:
> instrumentation/github.com/openai/openai-go/v2 and /v3 have no exercising test
> app anywhere in test/apps/ ... Worth a one-line addition to this comment

Good catch, and the limitation is now documented in 9e432b1.

One correction on the specifics: v2 is covered. #990 added `test/apps/openaiclientv2`, which requires `github.com/openai/openai-go/v2 v2.7.1`, so it landed after your review.
I audited every `otelc.yaml` target against what the apps actually require (taking care that a `/vN` suffix is a distinct module, not a subpackage) and v3 is the only instrumentation module with no exercising app.

I wrote the comment as a general rule with v3 as the current instance, rather than a list of module names, so it doesn't go stale the way the "27 modules" count in my original PR description did:

    # Known gap: that makes instrumentation coverage only as good as the test apps.
    # An instrumentation module that no app under test/apps builds against is scanned
    # by neither job, so adding a module without an app leaves it uncovered (currently
    # instrumentation/github.com/openai/openai-go/v3).

Happy to open a follow-up issue for an `openaiclientv3` app if you want the gap closed rather than just recorded.

.github/renovate.json5
atharvamhaske: > minimumReleaseAge is set at the top level... does Renovate's own
> vulnerability-alert flow bypass this cooldown?

Checked the docs and the answer is yes, with a caveat worth knowing about.

Renovate's [Minimum Release Age](https://docs.renovatebot.com/key-concepts/minimum-release-age/) page is explicit: "Security updates bypass any `minimumReleaseAge` checks, and so will be raised as soon as Renovate detects them."

The caveat is that the bypass isn't airtight in combination with `internalChecksFilter: 'strict'`, which I also set here. Discussion [#44065](https://github.com/renovatebot/renovate/discussions/44065) reports that strict filtering discards the patched version as "pending" during datasource lookup, before the `vulnerabilityAlerts` override is merged in. The result is a PR labelled `[SECURITY]` that only bumps to a version still inside the vulnerable range, and setting `vulnerabilityAlerts.minimumReleaseAge: null` doesn't fix it because the override is applied too late.

For this repo that's not on the critical path: `renovate.json5` configures neither `vulnerabilityAlerts` nor `osvVulnerabilityAlerts`, and there's no `.github/dependabot.yml`, so security bumps come from GitHub/Dependabot alerts via repo settings, which `renovate.json5` doesn't affect. That's the split the PR body
describes. Happy to add an explicit `vulnerabilityAlerts` block if you'd rather the intent be stated in config rather than implied, though per #44065 it wouldn't actually change behaviour today.

darccio:
@atharvamhaske Regenerate the bundle, please.

darccio (approved):
LGTM

atharvamhaske:
@darccio regenerated the bundle, ptal

darccio:
@atharvamhaske Regenerate the bundle, please.

atharvamhaske:
> @atharvamhaske Regenerate the bundle, please.

on it

atharvamhaske:
@darccio done

PR #809, opened 2026-07-24
feat(instrumentation): add auto-instrumentation for gocql Cassandra driver

### LINKED ISSUE #780: New Instrumentation Request for cassandra-gocql-driver

### Package

[cassandra-gocql-driver](https://github.com/apache/cassandra-gocql-driver)

### Version

v1.7.0 with plans to migrate to v2 soon

### Environment

As a Deployment on Kubernetes

### Use-case

Web application servers mostly

### Telemetry

Spans for query and batch execution (via `QueryObserver` / `BatchObserver`), including attributes like statement, keyspace, consistency level, target host/peer, latency and error status. Connection-level events (via `ConnectObserver`) would be useful too, for visibility into host connect/disconnect and connection errors.

### Anything else?

- Old `otelgocql` package in opentelemetry-go-contrib, which used gocql's `QueryObserver`/`BatchObserver`/`ConnectObserver` interfaces rather than wrapping query methods directly: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gocql/gocql/otelgocql
- Deprecation and removal of `otelgocql` from opentelemetry-go-contrib: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4164, https://github.com/open-telemetry/opentelemetry-go-contrib/pull/4295. This action makes me a little uncertain whether there were some discussions around removing this package, (maybe due to some performance issues?) and whether we should add compile time instrumentation for it or not.
- The driver moved from `github.com/gocql/gocql` to `github.com/apache/cassandra-gocql-driver/v2` after being donated to Apache Cassandra, with some breaking API changes. Should we target the legacy module, the new one or both with separate version ranges?

PS: If this is a good fit I'd be more than happy to take a crack at it!

### Tip

_No response_

### LINKED ISSUE #820: gocql instrumentation: ObserveConnect never records connection telemetry (span/metric)

The `gocql` instrumentation (`instrumentation/github.com/gocql/gocql`) wraps a session's `QueryObserver`, `BatchObserver`, and `ConnectObserver` so it can emit telemetry for query and batch execution. However, `otelObserver.ObserveConnect` never records any telemetry of its own — it only forwards the call to the user-supplied `ConnectObserver`, if one was configured:

```go
func (o *otelObserver) ObserveConnect(oc gocql.ObservedConnect) {
	if o.userConnect != nil {
		o.userConnect.ObserveConnect(oc)
	}
}
```

Compare this to `ObserveQuery`/`ObserveBatch`, which both check `enabler.Enable()` and call a `record*Span` helper. `ObserveConnect` has neither the enabler check nor any span/event recording, so host connect/disconnect activity and connection-level errors are completely invisible in traces today.

This gap is called out explicitly in the original instrumentation request (#780):

> Connection-level events (via `ConnectObserver`) would be useful too, for visibility into host connect/disconnect and connection errors.

`gocql.ObservedConnect` (from `github.com/gocql/gocql`) already carries everything needed for this: `Host`, `Start`, `End`, and `Err`, so no new driver hooks are required — only the existing `ObserveConnect` method needs to actually produce telemetry.

Record a span (or span event) for each observed connection attempt, gated the same way query/batch spans are:

- Add an `enabler.Enable()` guard consistent with `recordQuerySpan`/`recordBatchSpan`.
- Add a `recordConnectSpan` (or similar) helper that:
  - Starts/ends a client-kind span using `oc.Start`/`oc.End` as timestamps (falling back to `time.Now()` when zero, matching the existing pattern).
  - Sets `server.address`/`server.port` from `oc.Host` when available.
  - Sets `db.system.name = cassandra`.
  - Records `oc.Err` via `span.RecordError` + `codes.Error` on failure, `codes.Ok` otherwise — same convention used for query/batch errors.
- Since `ObserveConnect` has no incoming `context.Context` (unlike `ObserveQuery`/`ObserveBatch`), the span should be started from `context.Background()` as a root span, or attributed as a span event on the current cluster-level context if one is threaded through in the future — happy to discuss the best approach in the PR.
- Add unit test coverage analogous to the existing `TestOtelObserver_ObserveQuery`/`TestOtelObserver_ObserveBatch` tests, verifying span attributes and error status for `ObserveConnect`, including the case where `oc.Err != nil`.

- Leaving connection events unobserved and only documenting the limitation — rejected, since the original instrumentation request explicitly asked for this and the hook already exists in the observer, just unused.
- Emitting connection events as OTel logs instead of spans — worth a short discussion, since connect/disconnect is arguably more "event" than "span," but spans keep this consistent with how query/batch telemetry is already modeled in this package.

**Additional context**

- Relevant file: `instrumentation/github.com/gocql/gocql/hook.go` (`otelObserver.ObserveConnect`, `recordQuerySpan`, `recordBatchSpan` for the existing pattern to follow).
- Existing (currently connect-telemetry-free) test: `instrumentation/github.com/gocql/gocql/hook_test.go`, `TestOtelObserver_ObserveConnect`.
- Parent tracking issue: #780.

bhuvan-somisetty:
## Description

This PR adds compile-time auto-instrumentation for the `github.com/gocql/gocql` Cassandra driver.

The instrumentation intercepts `gocql.NewSession` to wrap cluster configuration with an OpenTelemetry observer (`otelObserver`). The observer implements `gocql.QueryObserver`, `gocql.BatchObserver`, and `gocql.ConnectObserver`:
- **Queries & Batches**: Creates client spans for Cassandra query and batch executions with OpenTelemetry semantic conventions (`db.system.name: cassandra`, `db.operation.name`, `db.namespace`, `db.query.text`, `server.address`, `server.port`).
- **Connections**: Creates a `CONNECT` client span per host dial, timed from `ObservedConnect.Start`/`End` and tagged with `db.system.name`, `db.operation.name` and the host's `server.address`/`server.port`.
- **Error Handling**: Records errors and sets span status to `Error` on failed query, batch, or connect attempts.
- **Observer Chaining**: Delegates to any pre-existing user observers (`userQuery`, `userBatch`, `userConnect`) registered on `ClusterConfig`.

Unit test coverage in `hook_test.go` validates span creation, attribute population, error handling, observer delegation, and re-entrancy.

Integration coverage lives in `test/apps/cassandraclient` and `test/integration/cassandra_client_test.go`: a minimal gocql app is built with `otelc` and run against a Cassandra testcontainer, and the test asserts on the exported spans (`CONNECT`, `testks.INSERT`, `testks.SELECT`, `testks.BATCH`) and their semantic-convention attributes.

## Motivation

Provides automatic distributed tracing and observability for applications using the `gocql` Cassandra driver without requiring manual session wrapping or code modifications.

Fixes #780
Fixes #820

---

## Checklist

- [x] Tests pass: `make test`
- [x] OpenTelemetry Registry updated (if applicable, see [registry guide](docs/instrument-guide.md#4-register-the-instrumentation)) — open-telemetry/opentelemetry.io#11056


bhuvan-somisetty:
Hi @NameHaibinZhang @kakkoyun @amazingakai — PR #809 is ready for review! Following up on #780, this PR adds compile-time auto-instrumentation for the gocql Cassandra driver using query and batch observers. All unit tests and CI checks are green. Looking forward to your feedback!

instrumentation/github.com/gocql/gocql/hook.go
```diff
@@ -0,0 +1,225 @@
...
+	_, span := o.tracer.Start(ctx, spanName,
+		trace.WithTimestamp(startTime),
+		trace.WithSpanKind(trace.SpanKindClient),
+	)
+	defer span.End(trace.WithTimestamp(endTime))
+
+	stmtText := strings.Join(b.Statements, "; ")
+
+	attrs := []attribute.KeyValue{
+		attribute.String("db.system", "cassandra"),
```
rashmi-tondare: As per the [docs](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/database/cassandra.md#spans) for `semconv/1.37.0`, the attribute name should be `db.system.name` instead of just `db.system`

tool/internal/rule/base.go
```diff
@@ -103,6 +104,13 @@ func (ibr *InstBaseRule) GetTarget() string   { return ibr.Target }
 func (ibr *InstBaseRule) GetVersion() string  { return ibr.Version }
 func (ibr *InstBaseRule) GetWhere() *WhereDef { return ibr.Where }
 
+func (ibr *InstBaseRule) validateBase() error {
```
rashmi-tondare: This change seems to be unrelated to instrumentation support for gocql. Maybe this can be moved to another PR?

instrumentation/github.com/gocql/gocql/hook.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	fields := strings.Fields(stmt)
+	if len(fields) > 0 {
+		return strings.ToUpper(fields[0])
+	}
+	return "SELECT"
+}
+
+// AfterNewCluster is invoked after gocql.NewCluster returns a *ClusterConfig.
+func AfterNewCluster(ictx hook.HookContext, cluster *gocql.ClusterConfig) {
```
rashmi-tondare: Is `AfterNewCluster` needed alongside `BeforeNewSession`? `ClusterConfig.CreateSession()` just calls `NewSession(*cfg)` internally ([cluster.go#L308-L310](https://github.com/gocql/gocql/blob/v1.7.0/cluster.go#L308-L310)), which `BeforeNewSession` (line 206) already patches. So the observer gets wrapped correctly at session-creation time regardless of whether `AfterNewCluster` ran. 
The only case I can see where it matters is code that reads `cluster.QueryObserver` before ever creating a session (e.g. config-inspection tooling, or middleware that conditionally chains onto an existing observer). Is that the intent here or would relying solely on `BeforeNewSession` suffice?

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	spans := sr.Ended()
+	require.Len(t, spans, 1)
+	span := spans[0]
+
+	assert.Equal(t, codes.Error, span.Status().Code)
+	assert.Equal(t, "cassandra connection error", span.Status().Description)
+}
+
+func TestOtelObserver_ObserveBatch(t *testing.T) {
```
rashmi-tondare, line 183: This test case only covers the happy path. Would be good to have test cases for the error path too, similar to `TestOtelObserver_ObserveQueryWithError`.

rashmi-tondare (commented):
Note: gocql was donated to Apache and now has a v2 under a **different module path**: `github.com/apache/cassandra-gocql-driver/v2` ([repo](https://github.com/apache/cassandra-gocql-driver)) vs. the v1 path (`github.com/gocql/gocql`) used here. Since instrumentation is matched by exact import path, v2 support would need its own module, not just a version bump. Worth a separate follow-up PR rather than scope for this one.

bhuvan-somisetty:
Hi @rashmi-tondare, thanks for the thorough review.

I have updated the PR with all the requested fixes:
1. Updated attribute name from db.system to db.system.name using semconv.DBSystemNameCassandra per semconv/v1.37.0 spec.
2. Removed unrelated tool/internal/rule changes from this branch so the PR focuses exclusively on gocql instrumentation.
3. Added test coverage for batch error handling in TestOtelObserver_ObserveBatchWithError.
4. Added doc comments to AfterNewCluster and BeforeNewSession explaining the purpose of both hook entry points.

All unit tests and CI checks pass. Thanks again for the feedback!

tool/internal/rule/raw_rule_test.go
rashmi-tondare, line 1: Both test files under this directory would belong in the same separate PR which introduces the `validateBase()` check removed from the scope of this PR.

instrumentation/github.com/gocql/gocql/hook.go
```diff
@@ -0,0 +1,227 @@
...
+	defer span.End(trace.WithTimestamp(endTime))
+
+	attrs := []attribute.KeyValue{
+		semconv.DBSystemNameCassandra,
+		semconv.DBOperationName(opName),
+		semconv.DBQueryText(q.Statement),
+	}
+	if q.Keyspace != "" {
+		attrs = append(attrs, semconv.DBNamespace(q.Keyspace))
+		attrs = append(attrs, attribute.String("db.cassandra.keyspace", q.Keyspace))
```
rashmi-tondare: `db.cassandra.keyspace` is set here alongside `semconv.DBNamespace(...)` (`db.namespace`). That attribute isn't part of the `semconv v1.37.0` registry already imported in this file. Looks like a legacy version of `db.namespace`. Can this be dropped?

bhuvan-somisetty:
Thanks @rashmi-tondare, both fixed 

Dropped `db.cassandra.keyspace` — you're right, `db.namespace` already covers it. Also moved the two `tool/internal/rule` test files out of this branch, they'll go with the `validateBase()` PR.

And on your `AfterNewCluster` question — `BeforeNewSession` does suffice, so I removed `AfterNewCluster` and its rule. Hooking only at session creation is actually more correct since it sees the caller's final observer config. Added tests for user-observer chaining and re-entrancy.

Rebased on main and squashed to one commit. PTAL!

rashmi-tondare:
Looking good overall! Just a few more things to close out:

**1. Registry entry (docs/instrument-guide.md, step 4)**
As per the guide, new user-facing instrumentations need a companion PR against the `opentelemetry.io` repo to add gocql to the [OpenTelemetry registry](https://opentelemetry.io/ecosystem/registry/adding/). The checklist box (in this PR's description) just needs that companion PR to be *opened*, not merged (as per step 5 in the instrument-guide).

**2. Integration tests (docs/instrument-guide.md, step 3)**
The guide also calls for integration tests under `test/integration/` that builds a binary with `otelc` and asserts on exported spans, in addition to the unit tests in `hook_test.go`. Would be a good idea to add these tests also as part of this PR

**3. Relationship to #822**
#822 (`ObserveConnect` connection telemetry) is stacked on top of this branch and can't merge until #809 lands. Given the PR description here already claims `ConnectObserver` support, might be cleaner to fold that 
fix into #809 directly rather than ship it as a dependent follow-up.

bhuvan-somisetty:
Thanks @rashmi-tondare, all three done 

Registry PR is up — open-telemetry/opentelemetry.io#11056, followed the same shape as the existing compile-time entries. Checkbox ticked here.

Added the integration test too: `test/apps/cassandraclient` does a keyspace setup then INSERT / SELECT / batch, and `test/integration/cassandra_client_test.go` builds it with `otelc` against a Cassandra testcontainer and checks the exported spans and their semconv attrs. Same pattern as the kafka/redis ones.

And you're right about #822 — folded it in here so `ObserveConnect` actually records its own span instead of just forwarding. Closed #822 as superseded. Also refreshed the description, it was still mentioning the `NewCluster` hook I'd removed.

No Docker on my machine so I couldn't run the integration test locally, counting on CI for that one. PTAL!

test/apps/cassandraclient/go.mod
```diff
@@ -0,0 +1,11 @@
+module go.opentelemetry.io/otelc/test/apps/cassandraclient
+
+go 1.26.2
```
rashmi-tondare: This repo targets Go 1.25, so these tests should also adhere to the same

test/apps/cassandraclient/cassandraclient.exe
rashmi-tondare, line 1: This .exe file should be removed. Build artifacts shouldn't be committed to the repo.

rashmi-tondare (changes requested):

test/apps/cassandraclient/go.mod
bhuvan-somisetty: Good catch, fixed - go.mod now targets go 1.25.0 to match the rest of the repo.

test/apps/cassandraclient/cassandraclient.exe
bhuvan-somisetty: My bad, removed the exe and added a .gitignore rule (test/apps/*/*.exe) so this doesn't sneak in again.

bhuvan-somisetty:
Thanks for the review! Pushed a fix for both points - go.mod is back to 1.25.0 and the stray .exe is gone (added a .gitignore rule so it can't happen again). CI is showing action_required on the workflows, so it should kick off once approved - will keep an eye on it.

rashmi-tondare (approved):
@NameHaibinZhang @kakkoyun @amazingakai these changes look good to me and I'm approving this PR. However since I'm not a repo approver, it would be great if one of you can take it from here.

rashmi-tondare:
> No Docker on my machine so I couldn't run the integration test locally

@bhuvan-somisetty looks like you'll need to install docker to locally test & fix this after all :)

tool/data/otelc-bundle.tgz
rashmi-tondare, line 1: This binary can be deleted. It should not be pushed to the repo.

rashmi-tondare (changes requested):
@bhuvan-somisetty I would strongly recommend that you install docker on your system and run the integration tests you've written to ensure that everything works as expected. If you rely on CI only the debug-fix cycles will take very long since it will always require a maintainer to approve the CI workflows.

amazingakai:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/780#issuecomment-5144048223

amazingakai:
@bhuvan-somisetty Please rebase & resolve merge conflicts

instrumentation/github.com/gocql/gocql/hook.go
bhuvan-somisetty: Good catch, this is already fixed, the span now uses `semconv.DBSystemNameCassandra` (i.e. `db.system.name`) everywhere, not the old `db.system`. Marking this resolved.

instrumentation/github.com/gocql/gocql/hook.go
bhuvan-somisetty: You're right, relying solely on `BeforeNewSession` is enough since `CreateSession()` just calls `NewSession(*cfg)` internally. `AfterNewCluster` isn't in the code anymore, this was dropped in an earlier revision of the PR.

instrumentation/github.com/gocql/gocql/hook.go
bhuvan-somisetty: Agreed, that's gone now too, only `semconv.DBNamespace` (`db.namespace`) is set, no legacy `db.cassandra.keyspace` attribute anymore.

tool/internal/rule/base.go
bhuvan-somisetty: This is no longer in scope for this PR, tool/internal/rule/base.go isn't touched here anymore, that validateBase() work moved out to its own PR (#797).

tool/internal/rule/raw_rule_test.go
bhuvan-somisetty: Same here, tool/internal/rule/raw_rule_test.go isn't part of this PR's diff anymore, that's all in #797 now.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook_test.go
bhuvan-somisetty: This is already covered, there's a `TestOtelObserver_ObserveConnectWithError` right below `TestOtelObserver_ObserveConnect` that asserts the span status/description when `ObservedConnect.Err` is set, mirroring `TestOtelObserver_ObserveQueryWithError`. Let me know if you had something else in mind.

tool/data/otelc-bundle.tgz
bhuvan-somisetty: This one's actually intentional, tool/data/otelc-bundle.tgz is tracked on main and every instrumentation-adding PR updates it (it's the packaged pkg/+instrumentation bundle the CLI embeds). There's a dedicated "Verify Bundle" CI check that runs `make package` and fails if this file doesn't match, so it can't be deleted, it just needs to stay in sync, which it is here. Different situation from the cassandraclient.exe build artifact caught earlier in this PR.

bhuvan-somisetty:
Took your advice @rashmi-tondare, got Docker Desktop + WSL2 set up and ran TestCassandraClient locally against a real cassandra:4.1.3 container instead of relying on CI:

```
=== RUN   TestCassandraClient
--- PASS: TestCassandraClient (280.03s)
PASS
ok      go.opentelemetry.io/otelc/test/integration     280.059s
```

Confirms the connect span server.address fix and the apache/cassandra-gocql-driver/v2 migration both work end-to-end (CONNECT, INSERT, SELECT, and BATCH spans all coming through correctly). Should make future debug cycles here faster since I don't need to wait on CI approval anymore.

bhuvan-somisetty:
Done @amazingakai, rebased onto main. Only real conflict was the binary tool/data/otelc-bundle.tgz (expected, since it just needs regenerating whenever main moves), ran make package fresh after the rebase to make sure it's actually in sync rather than just picking a side. Build and full tool test suite pass locally.

rashmi-tondare:
@NameHaibinZhang @amazingakai linking this issue comment here as well so we can take a call whether to support gocql v1, v2 or both - [#780 comment](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/780#issuecomment-5145467078)

darccio:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/780#issuecomment-5194482350

bhuvan-somisetty:
@darccio fair point, makes sense. Dropped the gocql/gocql v1 module entirely and kept only apache/cassandra-gocql-driver/v2 — it's a clean drop-in (same package name, same observer interfaces) so the test app just needed its import switched over. Rebased, regenerated the bundle, everything's green now.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,278 @@
...
+// so guard it and fall back to no address rather than crashing the app.
+func hostAddress(host *gocql.HostInfo) (addr net.IP) {
+	if peer := host.Peer(); len(peer) > 0 {
+		return peer
+	}
+	defer func() {
+		if recover() != nil {
+			addr = nil
+		}
+	}()
```
amazingakai: This shouldn't be required, the panicky code seems to have been removed:
https://github.com/apache/cassandra-gocql-driver/commit/65e2cafa8c46534a1aef83d5483302d23f7cb792

This also defeats the internal ordering of gocql:
```go
func (h *HostInfo) connectAddressLocked() (net.IP, string) {
	if validIpAddr(h.connectAddress) {
		return h.connectAddress, "connect_address"
	} else if validIpAddr(h.rpcAddress) {
		return h.rpcAddress, "rpc_adress"
	} else if validIpAddr(h.preferredIP) {
		return h.preferredIP, "preferred_ip"
	} else if validIpAddr(h.broadcastAddress) {
		return h.broadcastAddress, "broadcast_address"
	}
	return h.peer, "peer"
}
```

Is this intentional?

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,278 @@
...
+		trace.WithTimestamp(startTime),
+		trace.WithSpanKind(trace.SpanKindClient),
+	)
+	defer span.End(trace.WithTimestamp(endTime))
+
+	stmtText := strings.Join(b.Statements, "; ")
+
+	attrs := []attribute.KeyValue{
+		semconv.DBSystemNameCassandra,
+		semconv.DBOperationName("BATCH"),
```
amazingakai: Let's also add `semconv.DBOperationBatchSize` here.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,278 @@
...
+func parseOpName(stmt string) string {
+	stmt = strings.TrimSpace(stmt)
+	if stmt == "" {
+		return "SELECT"
+	}
+	fields := strings.Fields(stmt)
+	if len(fields) > 0 {
+		return strings.ToUpper(fields[0])
+	}
+	return "SELECT"
```
rashmi-tondare: `parseOpName` returns `"SELECT"` when the statement is empty which will mislabel empty/unparseable queries as `SELECT` in `db.operation.name`. The equivalent `calOp()` in `database/sql` returns `""` in the same case ([client.go#L653-L660](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/instrumentation/database/sql/client.go#L653-L660)). I think we can match that instead of defaulting to a real operation name here.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,278 @@
+// Copyright The OpenTelemetry Authors
```
amazingakai, line 1: Please also run `make format`, there are some changes when I run it locally on this PR.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	span.SetAttributes(attrs...)
+
+	if q.Err != nil {
+		span.RecordError(q.Err)
+		span.SetStatus(codes.Error, q.Err.Error())
+	} else {
+		span.SetStatus(codes.Ok, "")
+	}
```
amazingakai: The official OTel spec says:
Generally, Instrumentation Libraries SHOULD NOT set the status code to `Ok`, unless explicitly configured to do so. Instrumentation Libraries SHOULD leave the status code as `Unset` unless there is an error...

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+
+	span.SetAttributes(attrs...)
+
+	if b.Err != nil {
+		span.RecordError(b.Err)
+		span.SetStatus(codes.Error, b.Err.Error())
+	} else {
+		span.SetStatus(codes.Ok, "")
```
amazingakai: Same here^

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+
+	span.SetAttributes(attrs...)
+
+	if oc.Err != nil {
+		span.RecordError(oc.Err)
+		span.SetStatus(codes.Error, oc.Err.Error())
+	} else {
+		span.SetStatus(codes.Ok, "")
```
amazingakai: Same here^

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
amazingakai, line 1: Nonblocking Nit: We usually split semconv-related code into a separate internal semconv package to make the code easier to read. Let's do the same here.

rashmi-tondare (changes requested):
`recordQuerySpan`, `recordBatchSpan`, and `recordConnectSpan` all call `span.SetStatus(codes.Ok, "")` on the success path. But as per the [OTel trace API spec](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status), instrumentation libraries SHOULD NOT set `Ok`, only `Error` on failure, leaving status `Unset` otherwise. 
`database/sql`'s `instrumentEnd` already follows this: it calls `SetStatus(codes.Error, ...)` only inside `if err != nil`, with no corresponding call on the success path at all ([client.go#L633-L651](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/instrumentation/database/sql/client.go#L633-L651)). Can we drop the three `else` branches here and leave status `Unset` on success, same as that module?

amazingakai (changes requested):
Nice PR! I have some initial suggestions

amazingakai:
Let's also have a schema file for this under `schemas/otelc/groups`

bhuvan-somisetty:
Thanks both for the thorough pass, addressed everything in 9274c42:

- @rashmi-tondare `parseOpName` now returns `""` for an empty/unparseable statement instead of defaulting to `"SELECT"`, matching `calOp()` in `database/sql`.
- @amazingakai the three `else { span.SetStatus(codes.Ok, "")}` branches are gone, status is left `Unset` on success per the trace API spec.
- `hostAddress` is simplified to just `host.ConnectAddress()`. You're right that the `Peer()`-first check was overriding gocql's own priority order, and since the panic was removed upstream there's no need to guard it anymore.
- Added `semconv.DBOperationBatchSize` to the batch span, with a test asserting it.
- Split the attribute building into its own `semconv` package (`instrumentation/.../v2/semconv/cassandra.go`), same layout as the redis and database/sql instrumentations.
- Added `schemas/otelc/groups/cassandra.yaml` for the three spans.
- Ran `make format` and regenerated the bundle.

Also merged main to pick up the latest conflicts and updated the docs to list the new instrumentation. Let me know if anything else is off.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (o *otelObserver) recordQuerySpan(ctx context.Context, q gocql.ObservedQuery) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	opName := parseOpName(q.Statement)
+	spanName := opName
+	if q.Keyspace != "" {
+		spanName = q.Keyspace + "." + opName
+	}
```
amazingakai: This formatting (`{keyspace}.{operation}`) doesn't match the [otel db semconv](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/database/database-spans.md#name). Spec says it should be `{db.operation.name} {target}` (so `"SELECT testks"` instead of `"testks.SELECT"`). If neither `db.operation.name` nor `target` are available, we can fallback to `"cassandra"`.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,226 @@
...
+	stmt = strings.TrimSpace(stmt)
+	if stmt == "" {
+		return ""
+	}
+	fields := strings.Fields(stmt)
+	if len(fields) > 0 {
+		return strings.ToUpper(fields[0])
+	}
+	return ""
+}
```
amazingakai, line 234: If the query starts with a comment (e.g. `/* comment */ SELECT...`), this will just return `/*`. maybe add a simple check to strip leading comments or at least note it as a limitation somewhere.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/semconv/cassandra.go
```diff
@@ -0,0 +1,89 @@
...
+	Keyspace  string
+	Host      net.IP
+	Port      int
+}
+
+// QueryClientTraceAttrs returns trace attributes for a single CQL query.
+func QueryClientTraceAttrs(req QueryRequest) []attribute.KeyValue {
+	attrs := []attribute.KeyValue{
+		semconv.DBSystemNameCassandra,
+		semconv.DBOperationName(req.OpName),
```
amazingakai: If `req.OpName` is empty, this emits `"db.operation.name": ""`. Per the [otel AnyValue spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#anyvalue):

> AnyValues expressing an empty value, a numerical value of zero, an empty string, or an empty array are considered meaningful and MUST be stored and passed on to processors / exporters.

While it makes sense to always emit `db.query.text` (since an empty query string is literally what was executed), `db.operation.name` represents a category. If no category is available, the attribute should be omitted rather than emitted as `""`.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (o *otelObserver) recordBatchSpan(ctx context.Context, b gocql.ObservedBatch) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	spanName := "BATCH"
+	if b.Keyspace != "" {
+		spanName = b.Keyspace + ".BATCH"
+	}
```
amazingakai: Same as https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/809#discussion_r3735430101. This should be `"BATCH {keyspace}"` and just `"BATCH"` if there's no keyspace.

instrumentation/github.com/apache/cassandra-gocql-driver/v2/hook.go
```diff
@@ -0,0 +1,226 @@
...
+	startTime := oc.Start
+	if startTime.IsZero() {
+		startTime = time.Now()
+	}
+	endTime := oc.End
+	if endTime.IsZero() {
+		endTime = time.Now()
+	}
+
+	_, span := o.tracer.Start(context.Background(), "CONNECT",
```
amazingakai: I think the same spec applies here too https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/809#discussion_r3735430101.

Should be `CONNECT {target}`. Since connection dial doesn't target a specific keyspace, the spec says we should fallback to `server.address:server.port`

amazingakai (changes requested):

amazingakai (approved):
LGTM

darccio:
@bhuvan-somisetty Bundle again. Regenerate it, please.

NameHaibinZhang:
please rebase main, i will merge

bhuvan-somisetty:
Done - rebased onto main and regenerated the bundle. Should be good to go now, thanks for the reviews!

amazingakai:
instrumentation-manifest needs to be regenerated.

bhuvan-somisetty:
Hi @amazingakai,

I've rebased onto the latest `main`, regenerated the instrumentation manifest to include `cassandra-gocql-driver/v2`, and repackaged `otelc-bundle.tgz`. All CI checks and manifest verifications are passing.

Ready for another look whenever you get a chance. Thanks!

rashmi-tondare:
@bhuvan-somisetty looks like there are still some conflicts that need to be manually resolved before this can be merged. Can you please check your branch for unresolved conflicts and push the fixes?

bhuvan-somisetty:
Sorry for the delay — merged main in and resolved the conflicts, all builds/tests are green now. Should be good to merge.

PR #812, opened 2026-07-24
feat(tool): support YAML instrumentation selection

### LINKED ISSUE #587: feat(tool): add support for `.otel.yml` instrumentation configuration

ADITYA-CODE-SOURCE:
This adds module-local `otel.instrumentation.yml` / `otel.instrumentation.yaml` support for selecting instrumentation packages.

The YAML file is a frontend to the existing import-driven pin flow and supports third-party instrumentation packages. Selection is discovered independently per module. When a module has both a tool file and instrumentation YAML, their imports are merged and deduplicated; selections from all participating modules are then combined into the build-wide rule set.

Behavior differs by invocation:

- `otelc pin` updates persistent configuration. It updates an existing tool file, or validates and optionally prunes YAML without leaving a generated tool file. If neither exists, it generates `otel.instrumentation.go`.
- AutoPin uses temporary tool-file state, prunes invalid imports there, and restores generated files plus dependency changes after the build.

The precedence order is:

`OTELC_RULES` > `--rules` > merged import-driven selection > embedded defaults

This also updates the configuration and UX documentation for the new filename and merge behavior.

Tests:

- `go test ./internal/setup -count=1`
- `make lint/markdown`
- `make lint/go` (CI)

Fixes #587


ADITYA-CODE-SOURCE:
Hi @amazingakai  While adding test coverage I had a few questions.

In a multi-module workspace, if one module has a tool file and another has `.otel.yml`, should the tool file suppress YAML globally or only for its own module?

Also, if `.otel.yml` changes and `otelc pin` is run again, should the tool file remain authoritative?

Finally, if an import is a valid Go package but not an instrumentation package, should it fail or be removed by `--prune`?

I'll add tests based on the confirmed behavior.

om7057:
I went looking at your third question and ended up somewhere that might matter more than the answer itself.

On question 3, the existing flow already decides this, and it doesn't fail. In `updatePinnedProjects`, `ErrNotInstrumentation` goes through `pruneImport`, which warns to the log and to stderr, drops the import from the tool file only when `--prune` is set and the file is a root tool file, then returns `(false, nil)` and carries on. Only environment and module resolution problems return a real error. So a valid Go package that isn't an instrumentation package is a warning today, not a build failure.

`validateGeneratedToolFiles` here takes the other branch. It wraps any `v.Config.Error`, `ErrNotInstrumentation` included, and returns it. So the same bad entry hard-fails coming from `.otel.yml` but only warns coming from a hand-written `otel.instrumentation.go`. Given the description frames the YAML as a thin frontend to the same flow, that difference is probably worth a deliberate decision either way.

Question 2 looks sharper to me. `generatePinnedProjectsFromOtelYAML` writes `ToolFileCanonical` into each module dir, and `pinLocked` checks for tool files before it looks for the YAML:

```go
toolFiles, findToolErr := findToolFiles(moduleDirs)
...
if len(toolFiles) > 0 {
    return updatePinnedProjects(ctx, toolFiles, opts)
}
// the .otel.yml branch sits below this
```

The first run reads `.otel.yml` and generates `otel.instrumentation.go`. Every run after that finds the generated file, takes the earlier branch, and never reads the YAML again. Editing `.otel.yml` stops having any effect, with nothing to tell the user why. "If a tool file already exists it takes precedence and the YAML is ignored" is doing more work than it looks, because this path is what creates that file.

It also makes `--prune` strange for the YAML case, since pruning writes back to a file that gets regenerated from the YAML, so the invalid entry returns.

That same early return answers question 1, incidentally: `toolFiles` is the aggregate across all module dirs, so one module having a tool file suppresses the YAML for every module rather than just its own.

Marking the generated file so `findToolFiles` can distinguish it from a hand-written one would sort out most of this, or regenerating when the YAML is newer. Seems worth settling before tests get written around the current behaviour.

amazingakai:
I haven't looked at the code yet, but here's what I'd expect from a behavior/design perspective.

If a multi-module workspace has one module with a tool file and another with a `.otel.yml`, I'd expect the module with the yaml to generate its own tool file, and then have the rules merged from both tool files. Following similar behavior if both modules already contained tool files.

For second, tool file should always take precedence over `.otel.yml`, however for `otelc pin` vs `autopin`, my expectation would be:

* `otelc pin`: updates `.otel.yml` and prunes invalid imports (or warns about them), but does **not** generate a tool file.
* `autopin`: generates a tool file and performs pruning there.

One simple implementation could be to prune imports from `.otel.yml` during the process, and for `autopin` restore the original `.otel.yml` after the build (following our clean room usage pattern).

This also answers the third question. `autopin` shouldn't modify `.otel.yml`, instead invalid imports should just be omitted from the generated tool file, whereas `otelc pin` should update the `.otel.yml` itself.

`.otel.yml` can be a thin frontend internally, but from a user's perspective I'd expect it to behave the same as `otel.instrumentation.go`, with the only real difference being that the instrumentation dependencies don't end up tracked in `go.mod`.

I only mentioned the thin-frontend behavior in #587 because it seemed like the simplest implementation to me, but if you think treating tool files and `.otel.yml` more independently would simplify the implementation then go ahead.

ADITYA-CODE-SOURCE:
> I haven't looked at the code yet, but here's what I'd expect from a behavior/design perspective.
> 
> If a multi-module workspace has one module with a tool file and another with a `.otel.yml`, I'd expect the module with the yaml to generate its own tool file, and then have the rules merged from both tool files. Following similar behavior if both modules already contained tool files.
> 
> For second, tool file should always take precedence over `.otel.yml`, however for `otelc pin` vs `autopin`, my expectation would be:
> 
> * `otelc pin`: updates `.otel.yml` and prunes invalid imports (or warns about them), but does **not** generate a tool file.
> * `autopin`: generates a tool file and performs pruning there.
> 
> One simple implementation could be to prune imports from `.otel.yml` during the process, and for `autopin` restore the original `.otel.yml` after the build (following our clean room usage pattern).
> 
> This also answers the third question. `autopin` shouldn't modify `.otel.yml`, instead invalid imports should just be omitted from the generated tool file, whereas `otelc pin` should update the `.otel.yml` itself.
> 
> `.otel.yml` can be a thin frontend internally, but from a user's perspective I'd expect it to behave the same as `otel.instrumentation.go`, with the only real difference being that the instrumentation dependencies don't end up tracked in `go.mod`.
> 
> I only mentioned the thin-frontend behavior in #587 because it seemed like the simplest implementation to me, but if you think treating tool files and `.otel.yml` more independently would simplify the implementation then go ahead.

Thanks, that clears things up. I’ll update the implementation and add tests for multi-module workspaces, pruning, and cleanup.

amazingakai:
@ADITYA-CODE-SOURCE Is this PR ready for review? If not, please mark as draft until this is ready.

tool/internal/setup/config.go
```diff
@@ -77,6 +79,29 @@ func findToolFiles(moduleDirs map[string]bool) ([]string, error) {
...
+			InstrumentationYAMLAlias,
+		)
+	case canonicalExists:
+		return canonical, nil
+	case aliasExists:
+		return alias, nil
+	default:
+		return "", nil
+	}
+}
```
darccio: `findInstrumentationYAMLFile` duplicates `findToolFile` (lines 43-64) almost verbatim — same canonical/alias existence checks, same both-exist error shape, differing only in the two path constants and one word in the error string. Given there are now two of these (and potentially more filename pairs down the line), consider factoring out the shared logic, e.g.:

```suggestion
func findOneOf(moduleDir, canonicalName, aliasName, label string) (string, error) {
	canonical := filepath.Join(moduleDir, canonicalName)
	alias := filepath.Join(moduleDir, aliasName)

	canonicalExists := util.PathExists(canonical)
	aliasExists := util.PathExists(alias)

	switch {
	case canonicalExists && aliasExists:
		return "", ex.Newf(
			"both %q and %q exist; only one %s file is allowed",
			canonicalName,
			aliasName,
			label,
		)
	case canonicalExists:
		return canonical, nil
	case aliasExists:
		return alias, nil
	default:
		return "", nil
	}
}

func findInstrumentationYAMLFile(moduleDir string) (string, error) {
	return findOneOf(moduleDir, InstrumentationYAMLCanonical, InstrumentationYAMLAlias, "instrumentation YAML")
}
```

`findToolFile` would then become a one-line wrapper around the same helper.

tool/internal/setup/pin.go
```diff
@@ -166,6 +163,140 @@ func generateOtelInstrumentationGo(imports map[string]bool, opts PinOptions) *ds
...
+	if err != nil {
+		return nil, ex.Wrapf(err, "reading %s", path)
+	}
+
+	var cfg otelYAMLConfig
+	decoder := yaml.NewDecoder(strings.NewReader(string(data)))
+	decoder.KnownFields(true)
+	if decodeErr := decoder.Decode(&cfg); decodeErr != nil {
+		return nil, ex.Wrapf(decodeErr, "parsing %s", path)
+	}
```
darccio: A genuinely empty (0-byte) `otel.instrumentation.yml` — e.g. from `touch otel.instrumentation.yml` as a starting point — makes `decoder.Decode(&cfg)` return `io.EOF` on the very first call, which gets wrapped into `"parsing <path>: EOF"`. That reads as a YAML syntax error rather than "no instrumentations selected," and it's inconsistent with the already-tested `instrumentations: []` case, which is treated as empty and succeeds. Treating a first-decode `io.EOF` the same way as the second-decode `io.EOF` check just below (empty input, not an error) would make an empty file behave like an empty list:

```suggestion
	if decodeErr := decoder.Decode(&cfg); decodeErr != nil && !errors.Is(decodeErr, io.EOF) {
		return nil, ex.Wrapf(decodeErr, "parsing %s", path)
	}
```

`errors` and `io` are already imported. With this change, an empty file leaves `cfg` at its zero value and the function returns an empty `imports` map, matching the "empty list" test case's expectation.

tool/internal/setup/pin.go
```diff
@@ -433,16 +564,246 @@ func updatePinnedProjects(
...
+		} else if opts.Prune {
+			if writeErr := writeOtelYAMLImports(
+				otelYAMLFiles[moduleDir], yamlImportsByModule[moduleDir],
+			); writeErr != nil {
+				return writeErr
+			}
+		}
+	}
+	return nil
+}
```
darccio: For a module with both a tool file and an `otel.instrumentation.yml`, `applyOtelYAMLValidation` only prunes invalid entries out of `yamlImportsByModule` (written back to the `.yml`). The physical tool file is validated too (its temporarily-merged copy goes through `invalidInstrumentationImports`), but it's then restored verbatim from `backupOtelYAMLValidationFiles` regardless of what was found — so under plain `otelc pin --prune`, an invalid import that lives in the *tool file* is only logged to stderr and never actually removed, even though a tool-file-only module (no YAML present) does get pruned via `updatePinnedProjects`.

Is this intentional — i.e. is the tool file meant to be authoritative and untouched by `otelc pin` once a YAML file is also present in the same module — or should `--prune` also remove invalid entries from the tool file in this combined case? `TestPinLocked_PrunePreservesToolAndYAMLOwnership` confirms valid entries stay in their original file, but doesn't cover an invalid entry originating from the tool file side while YAML is also present, so this behavior isn't locked in by a test either way.

Whichever direction is correct, it'd be worth a test asserting it explicitly and (if intentional) a doc note next to the "each file retains ownership of its entries" line in `docs/rules.md`.

tool/internal/setup/pin.go
```diff
@@ -433,16 +564,246 @@ func updatePinnedProjects(
...
+			if writeErr := ast.WriteFileAtomic(toolFile, generated); writeErr != nil {
+				return ex.Wrapf(writeErr, "writing %s", toolFile)
+			}
+		} else if opts.Prune {
+			if writeErr := writeOtelYAMLImports(
+				otelYAMLFiles[moduleDir], yamlImportsByModule[moduleDir],
+			); writeErr != nil {
+				return writeErr
+			}
+		}
```
darccio: The `opts.AutoPin` branch here can't execute in production. `applyOtelYAMLValidation` is only reachable through `processOtelYAMLFiles`, whose one production call site (`processPinConfigs`, guarded by `if !opts.AutoPin`) never invokes it when `AutoPin` is true — when AutoPin actually runs, YAML-bearing configs go through `materializeModuleConfigs` + `updatePinnedProjects` instead, not through this function.

The only thing that exercises this branch is `TestProcessOtelYAMLFiles_AutoPin`, which calls `processOtelYAMLFiles` directly with `PinOptions{AutoPin: true}` — bypassing the orchestrator that would otherwise never produce that combination. That gives false confidence: the test passes, but nothing in `Pin`/`AutoPin`/`pinLocked` can reach this code path.

Worth deciding: either this was meant to be wired up as an alternate AutoPin path and the routing in `processPinConfigs` should call it, or it's leftover from an earlier design and should go. If the latter, removing the branch simplifies this to a single unconditional prune-on-write:

```suggestion
		if opts.Prune {
			if writeErr := writeOtelYAMLImports(
				otelYAMLFiles[moduleDir], yamlImportsByModule[moduleDir],
			); writeErr != nil {
				return writeErr
			}
		}
```

`TestProcessOtelYAMLFiles_AutoPin` would need to go or be rewritten to call the function the way production code actually does, so it stops asserting behavior no caller can trigger.

tool/internal/setup/pin.go
```diff
@@ -433,16 +564,246 @@ func updatePinnedProjects(
...
+			generated := generateOtelInstrumentationGo(imports, opts)
+			if writeErr := ast.WriteFileAtomic(toolFile, generated); writeErr != nil {
+				return nil, ex.Wrapf(writeErr, "writing %s", toolFile)
+			}
+		} else if addErr := addToolFileImports(toolFile, yamlImports); addErr != nil {
+			return nil, addErr
+		}
+		if _, ensureErr := ensureOtelcRequire(config.moduleDir, util.Version); ensureErr != nil {
+			return nil, ex.Wrapf(ensureErr, "ensuring otelc require in go.mod in %s", config.moduleDir)
+		}
```
darccio: This call is new in this PR (confirmed via diff — `ensureOtelcRequire` itself is byte-identical to `main`, but this call site inside `materializeModuleConfigs` wasn't there before). For a tool-file-only module under plain `otelc pin`, `processPinConfigs` runs this function and then immediately calls `updatePinnedProjects`, which calls `updateToolFile` → `ensureOtelcRequire` again on the same `moduleDir` (`pin.go:493`). Previously these two call sites were mutually exclusive (tool-file path vs. embedded-defaults path), so every module hit `ensureOtelcRequire` once per `Pin` run; now tool-file-only modules hit it twice — an extra `go.mod` read + parse + re-check per module per run.

It's harmless (the function is idempotent — it no-ops once the tool directive/require are already present), so this is a nitpick rather than a bug. Worth a look only if `otelc pin` performance on large multi-module workspaces matters; the two call sites would need to coordinate (e.g. `materializeModuleConfigs` skipping the ensure step for configs it knows will flow into `updatePinnedProjects` next) since neither can be dropped outright — `processOtelYAMLFiles` relies on `materializeModuleConfigs`'s call as its *only* one, since that path never reaches `updateToolFile`.

tool/internal/setup/pin.go
```diff
@@ -166,6 +163,140 @@ func generateOtelInstrumentationGo(imports map[string]bool, opts PinOptions) *ds
...
+}
+
+func sortedImportPaths(imports map[string]bool) []string {
+	importPaths := make([]string, 0, len(imports))
+	for path := range imports {
+		importPaths = append(importPaths, path)
+	}
+	slices.Sort(importPaths)
+	return importPaths
+}
```
darccio: `sortedImportPaths` duplicates `sortedStringsMapKeys` one-for-one — `sortedImportPaths(imports)` is exactly `sortedStringsMapKeys[bool](imports)` since `map[string]bool` is the only concrete type it's ever called with. Three call sites use this one (`pin.go:126`, `pin.go:227`, `pin.go:671`), two use the generic (`pin.go:743`, `pin.go:950`).

```suggestion
func sortedImportPaths(imports map[string]bool) []string {
	return sortedStringsMapKeys(imports)
}
```

Or, to remove the duplication entirely rather than just delegate, replace the three `sortedImportPaths(...)` call sites with `sortedStringsMapKeys(...)` and delete this function.

darccio (changes requested):

ADITYA-CODE-SOURCE:
Thanks for the detailed review  @darccio. I addressed all comments in 7dd5c16  including the shared file finder, empty YAML handling, source precedence, unreachable AutoPin path, and duplicate work.
I also added production-path tests and rebased onto the latest main. 
Could you please take another look?

darccio (approved):

darccio:
@ADITYA-CODE-SOURCE There are some linting issues.

PR #819, opened 2026-07-25
test(kafka/consumer): cover headerCarrier Set and Keys

Anshumancanrock:
## Description

Adds unit tests for `headerCarrier`, the `TextMapCarrier` adapter over
`[]kafka.Header` in the kafka-go consumer instrumentation. `Set` was at 40% and
`Keys` at 0%, since no existing test constructed a carrier directly.

The new `carrier_test.go` covers both `Set` branches (append an absent key,
overwrite one already present), `Keys` enumeration, `Get`'s first-match and
byte-for-byte key comparison, a message with nil `Headers`, and an
inject/extract round trip through `propagation.TraceContext`.

No production code changed. Package coverage 74.5% -> 89.4%, with `Get`, `Set`
and `Keys` all at 100%.

## Motivation

`AfterReadMessage` and `ExtractContext` both wrap `msg.Headers` in a
`headerCarrier` and pass it to `propagator.Extract`, so a bug in here won't
throw. It just quietly mis-parents consumer spans.

`Set` is the interesting one. `Get` returns the first match for a repeated key,
so if `Set` ever appended instead of overwriting in place, a stale `traceparent`
would keep winning. The test drives that the way it would actually happen, by
injecting twice into the same message rather than calling `Set` twice.

To be clear about scope: `TraceContext` only calls `Get`, so `Set` and `Keys`
aren't on the consumer's hot path today. But the propagator is whatever
`otel.GetTextMapPropagator()` returns, and B3 or custom propagators do call
them, so the carrier still has to hold up as a full `TextMapCarrier`. There's a
compile-time assertion of that in the test file.

---

## Checklist

- [x] Tests pass: `make test`


Anshumancanrock:
hii @Aryanbhargava18 , could you please take a look at this PR when you get chance? Thanks !

instrumentation/github.com/segmentio/kafka-go/consumer/carrier_test.go
```diff
@@ -0,0 +1,137 @@
...
+	// returning the stale "tp-1", since it resolves to the first match.
+	c.Set("traceparent", "tp-2")
+
+	assert.Len(t, headers, 2)
+	assert.Equal(t, "tp-2", c.Get("traceparent"))
+	assert.Equal(t, "traceparent", headers[0].Key)
+	assert.Equal(t, "bg-1", c.Get("baggage"))
+}
+
+func TestHeaderCarrier_Keys(t *testing.T) {
```
Aryanbhargava18, line 69: since you explicitly noted in `Get` that Kafka allows duplicate keys, we should probably test the behavior of `Keys()` when duplicates are present in the headers. does it return the duplicates or deduplicate them? let's add an assertion for that edge case just so the expected behavior is locked in.

Aryanbhargava18 (changes requested):
hey, thanks for adding this! the overwrite vs append tests for Set/Get look solid. left a couple notes — mainly just an accidental binary commit that's breaking your codecov check, and a small edge case in Keys().

**Regarding `tool/data/otelc-bundle.tgz`:**
looks like this snuck in by accident. this is why your `codecov/project/tool` check is failing (the tool module got touched but has no new tests). just revert this file from the branch so it stays scoped to the kafka tests and CI will go green.

Anshumancanrock:
@Aryanbhargava18 done ! I reverted the bundle, branch is just the kafka test file now.

also added the `Keys()` dupe assertion: it returns duplicates, doesn't
dedupe. threw in a check that Set only rewrites the first one so the
key count stays stable.

heads up : verify-bundle will probably go red now. it runs make package
and diffs against the committed archive, and adding a file under
instrumentation/ changes it. could you please run /regenerate-bundle?

Anshumancanrock:
@Aryanbhargava18 /regenerate-bundle won't work on fork PRs.

I have to push the bundle back.

Anshumancanrock:
Pushed the bundle back.

codecov/project/tool will still show red at ~64.75%, but afaict it's
not a required check so it shouldn't block.

instrumentation/github.com/segmentio/kafka-go/consumer/carrier_test.go
Anshumancanrock: good call, added TestHeaderCarrier_KeysReportsDuplicates. answer,
it returns them, no dedup: Keys() just appends every header key.

also covered the Set interaction since it seemed related. Set only
rewrites the first match, so after Set the key count is unchanged and
the trailing duplicate keeps its old value. that locks in Get/Set
both resolving to the first one while Keys() reports both.

Anshumancanrock:
> hey, thanks for adding this! the overwrite vs append tests for Set/Get look solid. left a couple notes — mainly just an accidental binary commit that's breaking your codecov check, and a small edge case in Keys().
> 
> **Regarding `tool/data/otelc-bundle.tgz`:** looks like this snuck in by accident. this is why your `codecov/project/tool` check is failing (the tool module got touched but has no new tests). just revert this file from the branch so it stays scoped to the kafka tests and CI will go green.

@Aryanbhargava18 removing  `tool/data/otelc-bundle.tgz` will fail bundle checks so i pushed it back.

cc @amazingakai

Aryanbhargava18 (approved):
ah right, always forget the `/regenerate-bundle` bot is broken for forks. my bad on that one. the `Keys()` dedupe tests look perfect now. good stuff, approving.

om7057:
Worth coordinating with #847, since these two collide.

#847 extracts `headerCarrier` out of the consumer package into `internal/propagation`, deleting the type and its `Get`/`Set`/`Keys` methods from `consumer_hook.go`. The tests here construct `headerCarrier{...}` in the consumer package, so if #847 lands first this file stops compiling.

#847 does add its own `internal/propagation/carrier_test.go`, so coverage of the carrier isn't lost either way. The cases here look broader though, duplicate keys, a message with no headers, and the propagator round trip. Might be worth porting those to the new location rather than having the two collide.

Anshumancanrock:
hii @NameHaibinZhang , could you please review this PR when you get a chance ? Thanks !

darccio (approved):
LGTM

PR #822, opened 2026-07-25
fix(instrumentation/gocql): record connection telemetry in ObserveConnect

### LINKED ISSUE #820: gocql instrumentation: ObserveConnect never records connection telemetry (span/metric)

bhuvan-somisetty:
> **Note:** This branch is stacked on top of #809 (the gocql instrumentation itself hasn't landed on `main` yet), so the diff below includes #809's commits in addition to this fix until that one merges. The change specific to this PR is the last commit only.

## Description

`otelObserver.ObserveConnect` in the gocql instrumentation currently only forwards the callback to a user-supplied `ConnectObserver`, if one is configured, and never records any telemetry of its own:

```go
func (o *otelObserver) ObserveConnect(oc gocql.ObservedConnect) {
	if o.userConnect != nil {
		o.userConnect.ObserveConnect(oc)
	}
}
```

This PR brings `ObserveConnect` in line with `ObserveQuery`/`ObserveBatch` by adding a `recordConnectSpan` helper, gated by the same `enabler.Enable()` check, that starts a `CONNECT` client span timed from `ObservedConnect.Start`/`End`, tagged with `db.system.name`, `db.operation.name`, and the target host's `server.address`/`server.port` when available, and marked errored via `RecordError`/`SetStatus` when `ObservedConnect.Err` is set.

Unit tests are added covering both the success path (span emitted, status OK, user observer still invoked) and the error path (span status set to error with the underlying dial error message), mirroring the existing query/batch test coverage.

## Motivation

Fixes #820

The original instrumentation request (#780) explicitly asked for connection-level visibility:

> Connection-level events (via `ConnectObserver`) would be useful too, for visibility into host connect/disconnect and connection errors.

Without this, host connect/disconnect activity and connection failures against a Cassandra cluster are completely invisible in traces, which is a real gap for a client that could be silently failing to establish connections to some hosts while others succeed.

This branches off `feat/gocql-instrumentation` (#809) since that's where the `gocql` package and its observer wrapping currently live; it can be rebased/retargeted once #809 lands on `main`.

---

## Checklist

- [x] Tests pass: `make test`


bhuvan-somisetty:
Hey folks, this is a small follow-up on top of the gocql work in #809 — it fills in the connection-level telemetry gap that #780 originally called out (ObserveConnect wasn't recording anything).

@kakkoyun @NameHaibinZhang @y1yang0 @amazingakai @txabman42 whenever you get a chance, would appreciate a look. Happy to make changes if anything's off.

bhuvan-somisetty:
Folding this into #809 as @rashmi-tondare suggested — the `ObserveConnect` telemetry now lives there, so this doesn't need to be a stacked follow-up. Closing in favour of #809.

PR #825, opened 2026-07-25
fix(instrumentation/openai): make provider name detection deterministic

### LINKED ISSUE #824: fix(genai): getProviderName returns inconsistent provider names for the same host
### Describe the bug

`getProviderName()` in the OpenAI GenAI instrumentation sets the `gen_ai.provider.name` span attribute by looping over a `map[string]string` of hostname keywords and returning the first substring match. Go does not guarantee map iteration order and randomizes it on every `range`. When a request host contains more than one keyword from the map (realistic behind API gateways or proxies that route to multiple providers), the function can return a different provider name on different calls for the exact same input.

The same code is duplicated in three places:
- `instrumentation/github.com/openai/openai-go/middleware.go`
- `instrumentation/github.com/openai/openai-go/v2/middleware.go`
- `instrumentation/github.com/openai/openai-go/v3/middleware.go`

### Environment

### To Reproduce

Take the `providerMapping` map and `getProviderName` function as they exist today (around lines 28-61 in any of the three `middleware.go` files) and call it repeatedly with a host that matches two different keywords:

```go
host := "litellm-gateway.mistral-together-proxy.internal" // matches both "mistral" and "together"
results := map[string]int{}
for i := 0; i < 20; i++ {
    results[getProviderName(host)]++
}
fmt.Println(results)
```

On my machine this printed `map[mistral:14 together:6]` across 20 calls with no code path change in between - same input, two different outputs.

### Expected behavior

`getProviderName` should return the same provider name for the same host on every call. It's doing "first matching keyword wins" style lookup, which needs a data structure with a fixed, predictable check order. A map doesn't give you that in Go.

### Additional context

Root cause: `providerMapping` is a `map[string]string`, iterated with `for keyword, provider := range providerMapping`. The Go spec leaves map iteration order unspecified, and the runtime deliberately randomizes the starting point on each `range`. Returning on the first match inside a map loop is non-deterministic any time more than one key can match.

A fix would replace `providerMapping` with an ordered slice of `{keyword, provider}` pairs and iterate the slice instead of the map, so the check order comes from how the list is written rather than from the runtime.

This affects the `gen_ai.provider.name` attribute on every OpenAI-instrumented span, so anyone running behind a gateway or proxy whose hostname matches more than one entry in the table gets inconsistent telemetry.

Satya900:
## Description

`getProviderName()` in the OpenAI GenAI instrumentation picked the `gen_ai.provider.name` span attribute by ranging over a `map[string]string` of hostname keywords and returning the first substring match. Go randomizes map iteration order on every `range`, so a host matching more than one keyword (common behind API gateways or proxies) could get a different provider name on different calls for the same input.

This replaces `providerMapping` with an ordered slice of `{keyword, provider}` pairs in all three affected packages (`openai-go`, `openai-go/v2`, `openai-go/v3` - the logic was duplicated across all three), so the check order is fixed by declaration order instead of left to the runtime. Also adds a regression test to each package that pins the expected result across 50 repeated calls against a host matching two different provider keywords.

## Motivation

I reproduced the bug with the existing map/function copied verbatim: calling `getProviderName` 20 times against one fixed host (`litellm-gateway.mistral-together-proxy.internal`, which contains both `"mistral"` and `"together"`) returned `mistral` 14 times and `together` 6 times, with no code path change between calls. The existing test suite didn't catch this because every case in `TestGetProviderName` uses a host that matches exactly one keyword.

Fixes #824

---

## Checklist

- [x] Tests pass: `make test` (`go test ./...` in all three affected modules)


bhumikadangayach:
Hi @Satya900 — the Codex bot's flag about CRLF contamination in the bundle matches something I hit myself this weekend: cloning on Windows with Git's default core.autocrlf setting introduces CRLF into checked-out source files, which then propagates into anything generated from them (in my case, it broke golden-file test comparisons in tool/internal/instrument; here it looks like it's affecting the embedded bundle archive). Re-cloning with git config --global core.autocrlf false and regenerating from a clean LF checkout should fix it — looks like you've already pushed a fix commit (3ad0096) for this. Nice catch from Codex, and good fix.

NameHaibinZhang (approved):

NameHaibinZhang:
LGTM!

NameHaibinZhang:
please resolve conflicts

Satya900:
@NameHaibinZhang i have resolved the conflicts.

Satya900:
@amazingakai please review this PR

PR #826, opened 2026-07-25
fix(instrumentation): make openai getProviderName deterministic

om7057:
## Description

`getProviderName()` in the OpenAI GenAI instrumentation picks a provider by scanning `providerMapping`, a `map[string]string` of hostname keywords, and returning the first substring match it finds. Go doesn't guarantee map iteration order, so a host that happens to contain more than one keyword can get tagged with a different `gen_ai.provider.name` on different calls for the exact same request.

## Motivation

This logic is duplicated identically across the v1, v2, and v3 middleware files, so the same nondeterminism exists in all three. Fixed it by replacing the map with an ordered slice of keyword/provider pairs, so "first match wins" is actually well-defined instead of depending on map iteration order.

Added a test (`TestGetProviderNameDeterministic`) that repeats the lookup 50 times against a host matching two keywords (`mistral` and `together`) and checks the result stays constant, in addition to the existing `TestGetProviderName` cases.

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/instrumentation/github.com/openai/openai-go/middleware.go
+++ b/instrumentation/github.com/openai/openai-go/middleware.go
@@ -25,36 +25,43 @@ const (
 	maxResponseBodySize = 4 << 20 // 4 MB
 )
 
-var providerMapping = map[string]string{
-	"openai.com":         "openai",
-	"azure.com":          "azure",
-	"anthropic.com":      "anthropic",
-	"dashscope.aliyuncs": "qwen",
-	"volces.com":         "ark",
-	"ark.cn":             "ark",
-	"hunyuan":            "tencent",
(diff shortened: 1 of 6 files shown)


om7057:
Hi @amazingakai @NameHaibinZhang @txabman42, could you take a look at this one? It's a small, self-contained fix (map iteration was making the OpenAI provider lookup nondeterministic), so hopefully quick to review. Open to feedback if you'd want this handled differently. Thanks!

om7057:
Closing: @Satya900 already had a PR (#825) open for #824 before mine went up. Didn't mean to duplicate; will check open PRs more carefully before starting next time.

Satya900:
Thank you @om7057 for your observation.

PR #828, opened 2026-07-25
fix(instrumentation): flush final SSE chunk on stream close without trailing newline

### LINKED ISSUE #827: bug(instrumentation): openai streaming reader drops final SSE chunk without trailing newline
### Describe the bug

`streamingReader.processSSELines()` in the OpenAI GenAI streaming instrumentation only flushes and parses lines that are terminated by `\n`. It buffers everything after the last `\n` as a possibly-incomplete line and holds onto it for the next `Read()` call.

If the upstream SSE stream ends without a trailing newline after the final `data: ...` line, say the connection closes right after the last chunk with no following newline before EOF, that leftover data just sits in `lineBuffer` and is never parsed. `Read()` calls `finalize()` on the same call where it sees the EOF, so the final chunk's `finish_reason`, response `id`, and usage/token counts never make it onto the span.

Same code is duplicated in `openai-go`, `openai-go/v2`, and `openai-go/v3` (`streaming.go`), so it affects all three.

### Environment

- Version: main / latest `otelc`

### To Reproduce

I wrote a quick reproduction with a fake `io.ReadCloser` that returns the final SSE chunk and `io.EOF` in the same `Read()` call, which mirrors how a lot of real readers behave once the connection is closed:

```go
body := []byte(`data: {"id":"chatcmpl-final","model":"gpt-4","choices":[{"finish_reason":"stop","delta":{}}],"usage":{"prompt_tokens":10,"completion_tokens":5,"total_tokens":15}}`)

sr := newStreamingReader(reader, span, start, "gpt-4", "chat", "openai", opChat, ctx)
buf := make([]byte, 4096)
for {
    _, err := sr.Read(buf)
    if err != nil {
        break
    }
}
```

`sr.id` stays empty and `sr.outputTokens` stays `0` even though the chunk was fully delivered to `Read()`, it's just sitting unparsed in `lineBuffer` when `finalize()` runs.

In practice OpenAI's real streaming responses are `data: ...\n\n` terminated so this may be rare in the wild, but any proxy/gateway or SDK version that trims the trailing newline, or any connection that gets cut right after the last full chunk, would silently lose the final usage/finish_reason attributes with no error.

### Expected behavior

`processSSELines()` (or the `Read()` caller) should also parse whatever's left in `lineBuffer` when the underlying reader reports an error (EOF or otherwise), not just on a following `\n`. The final chunk's data shouldn't depend on whether the stream happens to end with a trailing newline.

### Additional context

- `instrumentation/github.com/openai/openai-go/streaming.go`: `processSSELines()`, `Read()`
- Duplicated in `v2/streaming.go` and `v3/streaming.go`
- Not related to open PR #825/#826 (`getProviderName` determinism), different bug, different file.

om7057:
## Description

`streamingReader.processSSELines()` only flushes and parses lines that end in `\n`, buffering everything after the last newline as a possibly incomplete line for the next `Read()` call. If the upstream SSE stream ends right after the final `data: ...` line with no trailing newline before EOF, that leftover line never gets parsed, since `Read()` calls `finalize()` on the same call where it sees the error. The final chunk's `finish_reason`, response `id`, and usage/token counts silently never make it onto the span.

## Motivation

I ran into this while looking through the OpenAI streaming reader, wrote a small reproduction with a fake reader that returns the last chunk and `io.EOF` together with no trailing newline, and confirmed the id and token attributes stayed empty even though the chunk was fully delivered to `Read()`.

Fixed it by adding `flushRemaining()`, called right before `finalize()` on any read error, so whatever's left in `lineBuffer` gets parsed as a final line instead of silently dropped. Same code is duplicated across v1, v2, v3, so applied the fix to all three.

Added a regression test with a reader that returns the final chunk and `io.EOF` from the same call, matching how a lot of real readers behave once the connection closes.

Fixes #827

---

## Checklist

- [x] Tests pass: `make test`


om7057:
Hi @amazingakai @NameHaibinZhang @txabman42. This one's in the streaming reader rather than the provider lookup: the last SSE chunk gets dropped if the stream closes without a trailing newline. Repro and fix are in the description; happy to walk through the approach if anything's unclear. Thanks!

Satya900:
@om7057 **Nice fix** — I checked the **`flushRemaining()`-on-`Read()`-error path** and it **correctly recovers the final chunk**, both when **the last data and the error arrive in the same `Read()` call** and when **the error comes in a separate call afterward**.

**One related gap I found while testing this:** `Close()` **still has the original bug**. It calls **`finalize()` directly without calling `flushRemaining()` first**:

```go
func (r *streamingReader) Close() error {
	if r.done.CompareAndSwap(false, true) {
		r.finalize()
	}
	...
}
```

So if a caller **stops reading and calls `Close()` before `Read()` ever returns a non-nil error** (e.g. it already got what it needed and doesn't bother draining to EOF), **the final chunk still gets silently dropped**, just through this path instead of `Read()`'s.

**I reproduced it:** a reader that **returns the final chunk with `err == nil`**, then **`Close()` is called without a further `Read()`**, and **`gen_ai.response.id`**, **`finish_reasons`**, and the **usage token attributes** all come back **empty**.

**Was that intentional** (assuming callers always drain to EOF), or **would it be worth adding `flushRemaining()` to `Close()` as well for symmetry?**

om7057:
> @om7057 **Nice fix** — I checked the **`flushRemaining()`-on-`Read()`-error path** and it **correctly recovers the final chunk**, both when **the last data and the error arrive in the same `Read()` call** and when **the error comes in a separate call afterward**.
> 
> **One related gap I found while testing this:** `Close()` **still has the original bug**. It calls **`finalize()` directly without calling `flushRemaining()` first**:
> 
> ```go
> func (r *streamingReader) Close() error {
> 	if r.done.CompareAndSwap(false, true) {
> 		r.finalize()
> 	}
> 	...
> }
> ```
> 
> So if a caller **stops reading and calls `Close()` before `Read()` ever returns a non-nil error** (e.g. it already got what it needed and doesn't bother draining to EOF), **the final chunk still gets silently dropped**, just through this path instead of `Read()`'s.
> 
> **I reproduced it:** a reader that **returns the final chunk with `err == nil`**, then **`Close()` is called without a further `Read()`**, and **`gen_ai.response.id`**, **`finish_reasons`**, and the **usage token attributes** all come back **empty**.
> 
> **Was that intentional** (assuming callers always drain to EOF), or **would it be worth adding `flushRemaining()` to `Close()` as well for symmetry?**

@Satya900 Yes, Close() has the same pattern. I traced through it and reproduced the issue: if a caller gets what they need and calls Close() before draining to EOF, the buffer is never flushed, so finish_reason and usage stay empty. The fix is symmetric, just add flushRemaining() before finalize() in Close() as well. Good find.

om7057:
@Satya900, the Close() method has the same issue. If someone calls Close() before draining the stream to EOF, that final chunk still gets silently dropped. It's the same fix: flush the buffer before we finalize. It would make sense to include it here for consistency.

Satya900 (changes requested):
Confirmed v2 and v3 both got the flushRemaining() fix in Close(), but the root streaming.go (v1) still doesn't have it, Close() there still calls finalize() directly. Looks like it got missed when applying the same fix across all three copies.

om7057:
@Satya900 Good catch! You found that v1 was missing the fix in Close(). Just amended the commit to add flushRemaining() before finalize() in all three versions now. Thanks for verifying each one.

om7057:
@Satya900 this is fixed as of f7f37f0. You were looking at b5107299, where v1's `Close()` did still call `finalize()` directly. All three versions call `flushRemaining()` first now.

I also added a regression test for the exact path you described, in each version. It uses a reader that hands back the final chunk with a nil error, so nothing on the `Read` path finalizes, then calls `Close()` without draining. Stripping `flushRemaining()` back out of `Close()` fails all three with the symptom you reported:

```
--- FAIL: TestStreamingReader_CloseWithoutDrainingFlushesFinalChunk
    attribute "gen_ai.response.id" not found
    Error: Not equal:  expected: 7  actual: 0
```

Could you take another look when you get a chance?

om7057:
@Satya900 I have addressed the requested change, please go through it once.

om7057:
@amazingakai @NameHaibinZhang, please review this PR once all the checks apart from project code coverage are passing!

Satya900 (approved):
@om7057  Verified the fix is complete now, all three files call flushRemaining() in Close(), ran TestStreamingReader_CloseWithoutDrainingFlushesFinalChunk and the full suite in all three modules, all pass. Nice work tracking this through both rounds.

One heads-up: main has moved 4 commits since this branch last synced, and the bundle blob hash no longer matches main's. Likely to show as a verify-bundle conflict once GitHub re-checks. Might be worth a rebase + regenerate before merge.

instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -70,6 +70,7 @@ func (r *streamingReader) Read(p []byte) (n int, err error) {
 	}
 
 	if err != nil && r.done.CompareAndSwap(false, true) {
+		r.flushRemaining()
 		r.finalize()
 	}
```
darccio, line 78: `flushRemaining()` runs on *any* read error, not just `io.EOF`. On a genuine
transport error the bytes still sitting in `lineBuffer` may be a truncated,
mid-chunk fragment rather than a complete final line — parsing it here risks
attaching stale/misleading `finish_reason`/usage attributes to a span that
actually ended in failure, not completion. Restricting the flush to the EOF
case keeps `finalize()` (span-ending) behavior unchanged for all errors while
only attempting to parse the buffer when the stream genuinely ran out cleanly.

```suggestion
	if err != nil && r.done.CompareAndSwap(false, true) {
		if err == io.EOF {
			r.flushRemaining()
		}
		r.finalize()
	}
```

instrumentation/github.com/openai/openai-go/v2/streaming.go
```diff
@@ -70,6 +70,7 @@ func (r *streamingReader) Read(p []byte) (n int, err error) {
 	}
 
 	if err != nil && r.done.CompareAndSwap(false, true) {
+		r.flushRemaining()
 		r.finalize()
 	}
```
darccio, line 78: Same as the v1 copy: gate the flush on `io.EOF` specifically so a non-EOF read
error doesn't get its (possibly truncated) buffered line parsed and attached
to the span as if the stream completed normally.

```suggestion
	if err != nil && r.done.CompareAndSwap(false, true) {
		if err == io.EOF {
			r.flushRemaining()
		}
		r.finalize()
	}
```

instrumentation/github.com/openai/openai-go/v3/streaming.go
```diff
@@ -70,6 +70,7 @@ func (r *streamingReader) Read(p []byte) (n int, err error) {
 	}
 
 	if err != nil && r.done.CompareAndSwap(false, true) {
+		r.flushRemaining()
 		r.finalize()
 	}
```
darccio, line 78: Same as the v1/v2 copies: gate the flush on `io.EOF` specifically so a
non-EOF read error doesn't get its (possibly truncated) buffered line parsed
and attached to the span as if the stream completed normally.

```suggestion
	if err != nil && r.done.CompareAndSwap(false, true) {
		if err == io.EOF {
			r.flushRemaining()
		}
		r.finalize()
	}
```

instrumentation/github.com/openai/openai-go/streaming_test.go
```diff
@@ -178,6 +178,106 @@ func TestStreamingReader_FirstTokenLatency(t *testing.T) {
...
+
+	spans := sr.Ended()
+	require.Len(t, spans, 1)
+
+	attrs := spans[0].Attributes()
+	assertAttribute(t, attrs, "gen_ai.response.id", "closed-early")
+	assertInt64Attribute(t, attrs, "gen_ai.usage.input_tokens", 7)
+	assertInt64Attribute(t, attrs, "gen_ai.usage.output_tokens", 2)
+	assertSliceAttribute(t, attrs, "gen_ai.response.finish_reasons", []string{"stop"})
+}
```
darccio, line 312: Both new regression tests hard-code `opChat`. `flushRemaining` routes through
`processChunk`, which dispatches to `processCompletionChunk` for `opCompletion`
— that path isn't covered by either new test. Worth adding a parallel case
(e.g. `TestStreamingReader_FinalChunkWithoutTrailingNewline_Completion`) using
`opCompletion` and the completions payload shape to cover both dispatch
branches through this fix.

The same applies to the v2/v3 copies.

instrumentation/github.com/openai/openai-go/streaming.go
darccio: Both call sites (`Read`'s error path and `Close`) duplicate the
`r.flushRemaining(); r.finalize()` pair. Folding the flush into `finalize()`
itself removes the duplication and guarantees any future third call site
can't forget it.

```suggestion
func (r *streamingReader) finalize() {
	r.flushRemaining()
```

instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -70,6 +70,7 @@ func (r *streamingReader) Read(p []byte) (n int, err error) {
...
 
 	if err != nil && r.done.CompareAndSwap(false, true) {
+		r.flushRemaining()
```
darccio: Now redundant once `finalize()` calls `flushRemaining()` itself (see comment
on line 91).

```suggestion
```

instrumentation/github.com/openai/openai-go/streaming.go
```diff
@@ -78,6 +79,7 @@ func (r *streamingReader) Read(p []byte) (n int, err error) {
 
 func (r *streamingReader) Close() error {
 	if r.done.CompareAndSwap(false, true) {
+		r.flushRemaining()
```
darccio: Now redundant once `finalize()` calls `flushRemaining()` itself (see comment
on line 91).

```suggestion
```

darccio (changes requested):

instrumentation/github.com/openai/openai-go/streaming.go
om7057: Removed.

instrumentation/github.com/openai/openai-go/streaming.go
om7057: Folded it in, but as a parameter rather than unconditionally: finalize(flush bool). Making it unconditional would flush on every error path, including non-EOF ones, which is the exact case the other comment above flags as unsafe. Read passes err == io.EOF, Close passes true.

instrumentation/github.com/openai/openai-go/streaming.go
om7057: Removed.

instrumentation/github.com/openai/openai-go/streaming_test.go
om7057: Added TestStreamingReader_FinalChunkWithoutTrailingNewline_Completion using opCompletion and a completions-shaped payload, same pattern as the existing chat test. Same test added to v2 and v3.

instrumentation/github.com/openai/openai-go/streaming.go
om7057: Fixed by gating the flush on err == io.EOF in Read, and folded the flush into finalize itself so the two call sites don't repeat the same two-line pattern:


func (r *streamingReader) finalize(flush bool) {
	if flush {
		r.flushRemaining()
	}
	...
}
Read calls r.finalize(err == io.EOF), Close calls r.finalize(true) unconditionally, same as before. Applied the same way to v2 and v3.

instrumentation/github.com/openai/openai-go/v2/streaming.go
om7057: Addressed.

instrumentation/github.com/openai/openai-go/v3/streaming.go
om7057: Addressed.

darccio (approved):
LGTM

om7057:
> LGTM

@darccio Thanks for the merge enablement. Also, I went through [issue](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/958) of triplication of streaming.go file that you just found, I can undertake that part of the duplicated code as a follow-up.

PR #837, opened 2026-07-26
docs: document OTEL_GLS_MAX_SPANS default value and valid range

shivv23:
## Description

Documents the default value (1000), valid range (positive integers), and silent-ignore behavior for `OTEL_GLS_MAX_SPANS` in `docs/configuration.md`.

## Motivation

The existing docs mention `OTEL_GLS_MAX_SPANS` and say "increasing beyond the default accommodates deeper call stacks" but never state what the default is. Users tuning this variable currently have to read the source code (`instrumentation/go.opentelemetry.io/otel/sdk/trace/otel_trace_context.go:25`) to find the default value of 1000. Invalid values are also silently ignored, which makes debugging harder without documentation.

## Changes

- Added default value: `1000`
- Added valid range: any positive integer (`> 0`)
- Noted that invalid/missing values are silently ignored
- Added a usage example

## Testing

Documentation-only change -- no code changes, no tests needed.


Satya900:
@shivv23 Checked this against otel_trace_context.go. Default, validation, and the silent-fallback behavior all match the code exactly. Also confirmed the GLS README doesn't already cover this, so it's not a duplicate. LGTM!

shivv23:
@Satya900 Sure, thanks for the review!

docs/configuration.md
```diff
@@ -102,6 +102,14 @@ through all call boundaries relies on GLS to propagate trace context. Increasing
 [GLS operation notes](../instrumentation/go.opentelemetry.io/otel/README.md) for
 the operational constraints.
 
+- **Default**: `1000`
```
amazingakai: Maybe we can just document the default value inline in the existing paragraph instead of adding a separate list. For example, "Increasing `OTEL_GLS_MAX_SPANS` beyond the default of `1000`...".

I'm not sure the valid values, invalid values, and `export` example add much value here, since they're already implied.

shivv23: fixed!

shivv23:
@amazingakai and @Satya900 All changes done! Please have a look. Thanks!

amazingakai (approved):

amazingakai:
@shivv23 Needs a rebase.

shivv23:
@amazingakai Please have a look on this PR.

PR #840, opened 2026-07-26
fix(instrumentation/gin): gin instrumentation cannot be disabled at runtime

### LINKED ISSUE #839: bug(gin): OTEL_GO_DISABLED_INSTRUMENTATIONS has no effect on gin instrumentation
### Describe the bug

The gin instrumentation does not check whether it is enabled before running. Setting OTEL_GO_DISABLED_INSTRUMENTATIONS=gin has no effect: the hooks still run, the span is still renamed from GET to GET /users/:id, http.route is still set, and nothing is logged to say the setting was ignored.

Every other instrumentation in the repo declares an instrumentationKey and an enabler, and checks it before doing any work. For example instrumentation/net/http/server/server_hook.go:

```go
type netHttpServerEnabler struct{}

func (n netHttpServerEnabler) Enable() bool {
	return runtime.Instrumented(instrumentationKey)
}
```

Grepping for runtime.Instrumented( turns up 17 call sites covering net/http (client and server), grpc (client and server), kafka-go (producer and consumer), database/sql, redis, mongo, k8s.io/client-go, openai-go (v1/v2/v3), anthropic-sdk-go, log, slog and logrus. instrumentation/github.com/gin-gonic/gin/ is the only instrumentation package with no key, no enabler and no check. gin.go there is four lines and only builds a logger, and BeforeNext / AfterNext in context_hook.go do their work unconditionally.

### Environment

- OS: Linux
- Go Version: 1.25.0
- Version: main @ 72d167c

### To Reproduce
1. Drop this into instrumentation/github.com/gin-gonic/gin/context_hook_test.go and run go test ./... from that module. It fails on current main:

2. 
```go
func TestBeforeNext_DisabledInstrumentation(t *testing.T) {
	t.Setenv("OTEL_GO_DISABLED_INSTRUMENTATIONS", "gin")

	sr, tr := setupContextTracer(t)
	_, span := tr.Start(context.Background(), "GET")
	c := newGinContextWithRoute(t, "GET", "/users/:id", "/users/42", span)

	ictx := hooktest.NewMockHookContext(c)
	BeforeNext(ictx, c)
	span.End()

	require.Len(t, sr.Ended(), 1)
	assert.Equal(t, "GET", sr.Ended()[0].Name(),
		"span name must not be rewritten when gin instrumentation is disabled")
}
```
3. See Error
```
--- FAIL: TestBeforeNext_DisabledInstrumentation (0.00s)
    Error:      	Not equal:
    	            	expected: "GET"
    	            	actual  : "GET /users/:id"
```

Image

The same thing shows up end to end. Build test/apps/ginserver with otelc, run it with OTEL_GO_DISABLED_INSTRUMENTATIONS=gin exported, hit a route, and the exported span still comes through named after the route template with http.route set.

Image

### Expected behavior

OTEL_GO_DISABLED_INSTRUMENTATIONS=gin should stop the gin hooks from touching the span, the same way it works for every other instrumentation. An allow-list that omits gin, for example OTEL_GO_ENABLED_INSTRUMENTATIONS=nethttp, should behave the same way.

### Additional context

I think this has gone unnoticed because gin does not create its own span. The span comes from the net/http server instrumentation and gin only enriches it, so disabling nethttp masks the problem: with no span in the request context trace.SpanFromContext returns a non-recording span and BeforeNext bails at the IsRecording() check. You only hit the gap if you want to keep HTTP server spans and drop just the gin enrichment, which seems reasonable to want. http.route is the attribute most likely to cause a cardinality problem on a large router, and these hooks fire once per middleware per request rather than once per request, since gin calls c.Next() at every layer of the chain.

The other half of this is that docs/instrument-guide.md never mentions runtime.Instrumented, the enabler pattern or instrumentationKey anywhere. Nothing in the guide tells someone adding a new instrumentation that the gate is required, so gin (#501) shipped without one. Fixing gin on its own leaves the next instrumentation free to do the same, so I would like to add the requirement to the guide in the same change.

Gin gets its own key (GIN, written as gin in the env var). This matches how kafka-go producer/consumer and grpc client/server are separate packages sharing one key, and it lets users keep nethttp spans while dropping route enrichment.

Related: #705 lists per-instrumentation enable/disable (go: <name>: enabled) as one of the first consumers of the declarative config work, and runtime.Instrumented is the mechanism that would sit underneath it, so it seems worth having the current one applied consistently first. #807 and #808 are also circling the same surface from the documentation side.

om7057:
## Description

The gin hooks never consulted `runtime.Instrumented`, so `OTEL_GO_DISABLED_INSTRUMENTATIONS=gin` did nothing. `BeforeNext` and `AfterNext` ran regardless, the span was still renamed from `GET` to `GET /users/:id`, and `http.route` was still set. Gin was the only instrumentation package in the repo without the enabler gate the other thirteen use.

`gin.go` now declares `instrumentationKey = "GIN"` and a `ginEnabler`, following the same shape as `net/http/server`, `redis` and `anthropic-sdk-go`. Both hooks return early when the instrumentation is disabled, and the check sits before either of them touches `nextDepthKey` so the depth counter they share stays balanced.

Two details worth flagging for review:

Gin carries its own key rather than sharing `NETHTTP`. It only enriches the span that the `net/http` server instrumentation creates, so sharing a key would leave "disable gin" with no way to express it. A separate key lets an operator keep HTTP server spans while dropping the route enrichment, which is the case that motivated the issue.

`BeforeNext` does not log on the disabled path, unlike `net/http/server`. Gin calls `c.Next()` once per middleware in the chain rather than once per request, so a per-call debug line would be noisy for no benefit.

This also adds a "Runtime Enable/Disable Gate" section to `docs/instrument-guide.md`. The guide never mentioned `runtime.Instrumented`, the enabler pattern or `instrumentationKey` anywhere, so nothing told a contributor adding an instrumentation that the gate was required. That omission is the reason one could ship without it, and fixing gin alone would leave the next one free to repeat it.

## Motivation

The gap is invisible unless you want HTTP server spans without gin's route enrichment. Disabling `nethttp` hides it, because with no span in the request context `trace.SpanFromContext` returns a non-recording span and `BeforeNext` bails at the `IsRecording()` check. Keeping the enrichment separately switchable matters because `http.route` is the attribute most likely to become a cardinality problem on a large router, and because these hooks run at every middleware layer rather than once per request.

Fixes #839

## Verification

Four unit tests cover the gate: disabled through the deny-list, absent from the allow-list, present in the allow-list so the gate cannot over-suppress, and `AfterNext` not recording `c.Error()` values when disabled. Three of the four fail without the source change; the allow-list case passes either way by design.

Checked end to end as well, building `test/apps/ginserver` with `otelc` and running it with the console exporter and `OTEL_GO_DISABLED_INSTRUMENTATIONS=gin`. Before the change the exported span still came through as `GET /hello/:name` carrying `http.route`, with `HTTP server instrumentation initialized` in the log confirming `nethttp` was live. After it, the span stays `GET` with no `http.route`.

`tool/data/otelc-bundle.tgz` regenerated and verified to match the sources.

---

## Checklist

- [x] Code formatted: `gofmt` clean on all changed files
- [x] Linters pass: `golangci-lint run --config .tools/golangci.yml` on the root and gin modules
- [x] Tests pass: unit tests across `./tool/...`, both `pkg` modules and all instrumentation modules. Integration and e2e suites left to CI.
- [ ] OpenTelemetry Registry updated (not applicable, no new instrumentation)
- [ ] I used AI, and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md).


amazingakai:
I'm confused why this was removed:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/501#discussion_r3320786438

cc @waqar2403

om7057:
> I'm confused why this was removed: [#501 (comment)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/501#discussion_r3320786438)
> 
> cc @waqar2403

Thanks for looking into this. I hadn't come across that discussion when I opened #839, and it changes the context. I said there that gin "shipped without" an enable/disable mechanism, which implied it was overlooked. That's not accurate. It was implemented and then intentionally removed during review, and I should have found that history first. Apologies for the noise.

What still stands out to me is the reasoning, that a common approach was needed first. At b6c6c1e, four days before that comment, Instrumented(instrumentationKey) was already in databasesql, redis, and both the grpc and nethttp client/server pairs. runtime.Instrumented itself goes back to #156 in December. It's 17 call sites across 13 packages today, with gin the only one outside it.

My guess is it didn't look like a convention because it isn't written down anywhere. docs/instrument-guide.md still doesn't mention runtime.Instrumented, instrumentationKey, or the enabler pattern, so reviewing a new instrumentation against the guide you'd reasonably conclude there wasn't one yet. That's the part I'd most like to land, whatever happens to gin.

Happy to drop the gin change and keep just the docs if you'd rather fold the rest into #705.

@txabman42 it was your call originally, what's your read?

waqar2403:
> I'm confused why this was removed: [#501 (comment)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/501#discussion_r3320786438)
> 
> cc @waqar2403

The gate was in the original PR and I removed it at review request, that's the thread @amazingakai linked. No issue at all with the gap being fixed now gin should have the same lever as everything else.

One thing I can confirm from the original design, since it came up in the PR body a separate key rather than sharing NETHTTP is the right call. The gin hooks don't create a span they enrich the one the net/http server instrumentation already creates, which is why they hook (*Context).Next after routing rather than ServeHTTP. So "disable gin" and "disable HTTP server spans" really are two different switches and sharing a key would collapse them.

Agree the docs part is the more valuable half. The enabler pattern isn't in instrument-guide.md today which is why it wouldn't have read as an established convention during that review.

Happy to review this if it helps.

om7057:
> > I'm confused why this was removed: [#501 (comment)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/501#discussion_r3320786438)
> > cc @waqar2403
> 
> The gate was in the original PR and I removed it at review request, that's the thread @amazingakai linked. No issue at all with the gap being fixed now gin should have the same lever as everything else.
> 
> One thing I can confirm from the original design, since it came up in the PR body a separate key rather than sharing NETHTTP is the right call. The gin hooks don't create a span they enrich the one the net/http server instrumentation already creates, which is why they hook (*Context).Next after routing rather than ServeHTTP. So "disable gin" and "disable HTTP server spans" really are two different switches and sharing a key would collapse them.
> 
> Agree the docs part is the more valuable half. The enabler pattern isn't in instrument-guide.md today which is why it wouldn't have read as an established convention during that review.
> 
> Happy to review this if it helps.

Thanks @waqar2403. Framing it as hooking (*Context).Next after routing rather than ServeHTTP is clearer than what I put in the PR body; I'll fold that into the comment on instrumentationKey if this lands.

Holding off on changes until @txabman42 has had a chance to weigh in, since it was his call originally.

instrumentation/github.com/gin-gonic/gin/context_hook.go
```diff
@@ -70,6 +76,12 @@ func BeforeNext(ictx hook.HookContext, c *gin.Context) {
 // AfterNext runs after (*gin.Context).Next returns. It records any errors
 // accumulated via c.Error() during request handling.
 func AfterNext(ictx hook.HookContext) {
+	// Gate before touching nextDepthKey so the depth counter stays balanced
+	// with BeforeNext, which returns at the same point when disabled.
+	if !enabler.Enable() {
+		return
+	}
+
 	c, ok := ictx.GetParam(0).(*gin.Context)
```
amazingakai: Let's be defensive here and save the value of `enabler.Enable()` from `BeforeNext` in the hook context, then reuse it in `AfterNext`. That way, even if the environment variable changes while the request is in flight, this `BeforeNext`/`AfterNext` pair remains balanced.

amazingakai (approved):
LGTM :heart: I only have one suggestion

om7057:
@amazingakai good call, done. BeforeNext now stores its Enable() result on the hook context via SetKeyData, and AfterNext reads it back instead of calling Enable() again, so the pair can't disagree if the environment variable changes mid-call. Added a regression test (TestAfterNext_UsesBeforeNextGateDecision) that flips the env var between the two calls and asserts AfterNext still honors BeforeNext's decision. Also added a note on this to the instrument-guide docs since it's a pattern any other paired before/after hook sharing one hook context would hit too.

om7057:
Hey @amazingakai / @open-telemetry/go-compile-instrumentation-approvers, wanted to flag the codecov/project/tool failure on this PR rather than ignore it, since I think it points to a real (if unrelated) config issue.

Why it's failing: this PR's own diff is fully covered (codecov/patch is 100%), so the failure isn't about the gin fix. Pulling the coverage report for this commit from Codecov's API, the number behind codecov/project/tool (64.70%) is a merged report spanning 115 files across tool/ (64), instrumentation/ (46), and pkg/ (5), not tool/ alone. Locally, go test ./tool/... -coverprofile=... gives ~78.8% for tool/ by itself, well above the 70% target in codecov.yml.

I checked other currently open PRs, and this isn't specific to my branch; it's failing the same way. Notably, #841 is literally "add unit test coverage for generic receiver matching" in tool/ast and still fails, if the check were correctly scoped to tool/'s own coverage, a pure test-addition PR should push it over 70%, not fail. That's what makes me think the flags: [tool] scoping in codecov.yml isn't actually restricting which uploaded sessions count toward this status.

My guess at the cause: test-integration.yaml and test-e2e.yaml both upload coverage via codecov-action without a flags: param:

# test-integration.yaml, test-e2e.yaml
- uses: codecov/codecov-action@...
  # no `files:` or `flags:`, auto-detects and uploads unflagged
If those unflagged uploads (which cover code across tool/, pkg/, and instrumentation/ exercised through integration/e2e runs) get folded into every flag-scoped project status instead of being excluded, that would explain why project: tool and project: pkg both see repo-wide numbers instead of their own path's coverage.

Proposed fix (Can open a separate PR for this, since it's a shared CI file, not scoped to this fix):

Add flags (or drop the coverage upload) to the codecov-action steps in test-integration.yaml/test-e2e.yaml, and/or
Add explicit flag_management with paths per flag in codecov.yml so unflagged/mismatched reports can't count toward a path-scoped status.
This isn't a required check, so it's not blocking the merge here, but I wanted to surface it since it looks like it's silently red on most open PRs right now. Let me know if you'd like me to open a dedicated PR for the Codecov fix.

om7057:
Correction to my comment above, I had the cause wrong.

The `flags: [tool]` scoping in `codecov.yml` is fine, and the unflagged integration/e2e uploads are not involved. The problem is in `test-unit.yaml`. `make test-unit/coverage` leaves `coverage-tool.txt`, `coverage-pkg.txt` and `coverage-instrumentation.txt` in the workspace, and the `files:` input does not restrict what gets uploaded. From the action's own input description at the pinned SHA:

> `files`: Comma-separated list of explicit files to upload. These will be added to the coverage files found for upload. If you wish to only upload the specified files, please consider using disable-search.

So the `flags: tool` step uploads all three profiles under the `tool` flag: 63 `tool/` files + 46 `instrumentation/` + 5 `pkg/` gives the 115 I saw. It shows up at the flag level rather than the status level. Codecov's API reports the stored `tool` flag coverage as 64.674%, so the flag data is already repo-wide before any status filtering runs.

That means my first suggestion (adding flags to integration/e2e) would have changed nothing, and the second (`flag_management` with paths) would have masked a bad upload server-side. Sorry for pointing in the wrong direction.

On why it only started failing now, comparing Codecov's reports for #799 and this PR:

| | #799 (Jul 24) | #840 (Jul 28) |
|---|---|---|
| files | 109 | 115 |
| statements | 9,001 | 11,199 |
| misses | 1,527 | 3,507 |
| coverage | 78.14% | 64.70% |

#799 was already at 109 files. The six extra files are the linodego v2 instrumentation from c5ab841, dominated by `public_methods_gen.go` (4,512 lines of generated wrappers), adding roughly 1,980 misses. This PR's `base_totals` is 64.67%, so `main` itself is now below target.

Two things that follow:

- #799's green check is stale (Jul 24, before c5ab841) and will turn red on its next push.
- #828 shows no codecov statuses at all, because it only touches `instrumentation/`, so neither upload step's `if:` fires.

Fix is one line per step, `disable_search: true`, opened as #857 against `main`. With the flag scoped correctly `go test ./tool/...` gives 78.8% over 63 files. Once #857 lands I can rebase this PR to confirm the check goes green.

None of this affects the gin change itself.

om7057:
Thank you @NameHaibinZhang for the merge!

PR #843, opened 2026-07-27
fix(net/http/server): unwrap response writers

rbroggi:
## Summary
- expose the wrapped ResponseWriter through Unwrap
- preserve ResponseController access to the original writer
- add a regression test and regenerate the embedded instrumentation bundle

## Motivation
Prometheus promhttp uses Unwrap to preserve ResponseController behavior through response-writer decorators. Without it, an otelc-instrumented controller-runtime metrics endpoint can return 404 even when /metrics is registered.

This complements #795, which corrects the separate http.Pusher implementation.

## Testing
- go test ./...
- make package


rbroggi:
This branch is now rebased onto #795 head `ab01469`. The PR must still target `main` because GitHub does not permit a fork PR branch as its base, but its history is stacked: `#795` Pusher commits followed by this `Unwrap` commit. Please merge #795 before this PR.

PR #844, opened 2026-07-27
fix(ast): check return error of file.Close() in WriteFile

vedant21-oss:
## Description
In `tool/internal/ast/parser.go`, `WriteFile` used `defer file.Close()`, ignoring any error returned when closing the file after writing AST output. Closing a writable file descriptor flushes buffered data; ignoring close errors can silently mask write failures or disk I/O errors.

This PR updates `WriteFile` to explicitly check the error returned by `file.Close()`.

## Fix Details
- Updated `WriteFile` in `tool/internal/ast/parser.go` to explicitly check and return any error from `file.Close()`.
- - Added unit test `TestWriteFile` in `tool/internal/ast/ast_test.go` to verify file creation and AST output writing.
## Checklist
- [x] PR title follows conventional commits format: `fix(ast): ...`
- [x] Code formatted & linters pass
- [x] Tests pass


Satya900:
@vedant21-oss Verified the fix: no leaked file handle, no double-close, both paths return correctly. Ran TestWriteFile locally, passes.

Checked WriteFileAtomic as a sibling path too. It already handles Close() correctly, so this was an isolated gap, not a broader pattern.

Test only covers the happy path, not an actual Close() failure, which is the specific case this PR fixes. WriteFile opens the file internally rather than accepting an injected writer, so simulating that isn't straightforward without a signature change.

LGTM.

vedant21-oss:
Thank you @Satya900 for the thorough review and verification!

tool/internal/ast/parser.go
```diff
@@ -97,12 +97,14 @@ func WriteFile(filePath string, root *dst.File) error {
...
 	}
-	defer file.Close()
 	r := decorator.NewRestorer()
-	err = r.Fprint(file, root)
-	if err != nil {
+	if err = r.Fprint(file, root); err != nil {
+		_ = file.Close()
 		return ex.Wrapf(err, "failed to write to file %s", filePath)
 	}
+	if err := file.Close(); err != nil {
```
amazingakai:
```suggestion
	if err = file.Close(); err != nil {
```

amazingakai (approved):
Looks reasonable to me. Good catch, Thanks ❤️

vedant21-oss:
Hi @amazingakai! 👋

I've fixed the CI linter issues and added test coverage:
- Fixed `govet` variable shadowing on `err` in `parser.go` by reusing the declared error variable.
- Fixed `gofumpt` formatting in `ast_test.go`.
- Added unit test `TestWriteFile_CreateError` to cover the error path in `WriteFile`.

All linters and unit tests are now passing. Ready for another review!

PR #845, opened 2026-07-27
docs(log): document sdk/trace build-graph requirement for log/slog and logrus correlation

### LINKED ISSUE #773: question: log/slog and logrus don't get trace_id/span_id without a direct sdk/trace import — is this expected?
## What I'm seeing

I've been experimenting with otelc's zero-code instrumentation and ran into something I'd like to double-check, in case I'm missing a setup step somewhere.

When an application relies purely on otelc's automatic instrumentation (`net/http`, `slog`, `logrus`, etc.) and doesn't import `go.opentelemetry.io/otel/sdk/trace` directly anywhere in its own source, I wasn't able to get `trace_id`/`span_id` to show up in `log/slog` or `logrus` output, even though a real span did appear to be active at the time. Adding a direct `sdk/trace` import (e.g. to construct a `TracerProvider` manually) made it work as expected.

I wanted to ask here first, in case this is a known constraint of the current design or I'm overlooking something on my end, before digging further.

## Reproduction 1: minimal zero-code app

A `net/http` handler with no OpenTelemetry imports at all:

```go
func greetHandler(w http.ResponseWriter, r *http.Request) {
	slog.InfoContext(r.Context(), "handling greet request")
	...
}
```

Building and running this with otelc:

- The generated `.otelc-build/matched.json` / `otel.instrumentation.go` never include `go.opentelemetry.io/otel/sdk/trace`, even though the binary does end up linking that package (via the automatically-injected SDK setup).
- At runtime, the automatic `net/http` server instrumentation produces a real span (confirmed via a console exporter — a genuine, non-zero trace/span ID is generated for the request), but the `slog.InfoContext` log line for that same request never carries `trace_id`/`span_id`.

## Reproduction 2: same effect on the existing integration tests

Out of curiosity, I tried editing both `test/apps/logslog/main.go` and `test/apps/logslogrus/main.go` to remove their direct `sdktrace`/`stdouttrace` imports and manual `otel.SetTracerProvider(tp)` calls, relying instead on `otel.GetTracerProvider().Tracer("test")` — i.e. the same zero-code path shown above. Both changes make their respective existing integration tests fail:

```
--- FAIL: TestLogsSlog
    logs_test.go:46: Should NOT be empty, but was []
    Expected trace_id to be injected into log messages

--- FAIL: TestLogsLogrus
    logs_test.go:74: Should NOT be empty, but was []
    Expected trace_id to be injected into logrus messages
```

In both cases, all other assertions in the test (log message content) still pass; only the `trace_id` assertion fails. Both `test/apps/logslog` and `test/apps/logslogrus` currently import `sdk/trace` directly, so I'm not sure whether that's intentional (e.g. to demonstrate a specific usage pattern) or incidental.

## Question

Is direct import of `go.opentelemetry.io/otel/sdk/trace` expected to be required for log/trace correlation to work (for both `log/slog` and `logrus`), even in an otherwise zero-code setup? If so, would it be worth documenting that as a requirement/limitation? If not, I'm happy to dig further into why the correlation doesn't kick in automatically.

hanshal101:
Fixes #773
Closes #773 

## Description

Documents a known limitation for `log`, `log/slog`, and `logrus` trace/span-id correlation: `go.opentelemetry.io/otel/sdk/trace` must already be part of the application's build dependency graph for `trace_id`/`span_id` to be injected into log output.

Adds a README to each of `instrumentation/log`, `instrumentation/log/slog`, and `instrumentation/github.com/sirupsen/logrus` explaining:
- how trace/span ids are attached (via `runtime.GetTraceAndSpanID()`, populated only once `sdk/trace`'s own instrumentation registers into `pkg/runtime`)
- that `otel.instrumentation.go` imports don't count, since that file is intentionally excluded from the build graph via `//go:build tools`
- the practical fix: a plain blank import of `go.opentelemetry.io/otel/sdk/trace` anywhere in the application's own source (outside a `tools`-tagged file) is enough — no need to actually construct or use a `TracerProvider`

## Checklist

- [x] Linters pass: `make lint/go`
- [x] Tests pass: `go test ./tool/...` (no code changes, docs only)
- [ ] Tests added for new functionality — n/a, docs only
- [x] Documentation updated
- [ ] OpenTelemetry Registry updated — n/a


hanshal101:
Hello @amazingakai , here do I need to write the testcases for the `resolveBootstrapDeps` to pass the check?

amazingakai (changes requested):
Okay, I now realize there are two separate problems here.

1. `otelc pin` / `AutoPin` do not automatically follow nested `otel.instrumentation.go` files and add `replace` directives for them. The proper fix for this is #585, but we could implement a simpler workaround for now.

2. The main problem is that imports added by `otel.instrumentation.go` are not included in the dependency graph (`go list -deps` / `go build -a -x -n`) because the file is marked with the `//go:build tools` build tag. This is not a bug, but rather an intentional design choice: instrumentation imports should not automatically become build dependencies, even though they are tracked in `go.mod`.

While the approach in this PR may work for this use case, I think it is fairly hacky.

Instead, I think we should document this limitation: `go.opentelemetry.io/otel/sdk/trace` must already be part of the user's build dependency graph for log instrumentation to work. Unless we come up with a cleaner solution, I think documenting the behavior is preferable.

We already document this for `otel/trace` (`SpanFromContext`) instrumentation:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/tree/main/instrumentation/go.opentelemetry.io/otel#3-trace-api-instrumentation

> If no span exists in GLS, or if the `go.opentelemetry.io/otel/sdk/trace` instrumentation was not applied (for example because the package was not present in the application's build graph), the original return value is left unchanged.

We'll need something similar for all three log instrumentations.

instrumentation/github.com/sirupsen/logrus/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Limitation: `go.opentelemetry.io/otel/sdk/trace` must be in the build graph
+
+**`go.opentelemetry.io/otel/sdk/trace` must already be part of the application's build dependency graph (as seen by `go list -deps` / `go build -a -x -n`) for this instrumentation to inject `trace_id`/`span_id`.**
+
+Instrumentation imports declared in an `otel.instrumentation.go` file are tracked in `go.mod` but are intentionally **not** added to the build's dependency graph — that file is marked with the `//go:build tools` build tag specifically so instrumentation imports do not become build dependencies themselves. So relying on `logrus`'s own instrumentation module to pull in `sdk/trace` does not work.
+
+In practice, this means your application must construct a real `sdk/trace.TracerProvider` somewhere in its own source (for example via `sdktrace.NewTracerProvider(...)` and `otel.SetTracerProvider(...)`), even if you don't otherwise call the SDK directly. Without that, spans are still created and exported correctly, but `logrus` output will not carry `trace_id`/`span_id`.
```
amazingakai: Let's document that simply a blank import is sufficient. All we need is for the package to be part of the application's build dependency graph. A blank import is enough to achieve that, as long as it is not behind a build tag like `//go:build tools` (as in `otel.instrumentation.go`).

amazingakai (commented):
Please update the PR title/desc.

amazingakai (approved):

PR #836, opened 2026-07-26
test(hook): add unit tests for MockHookContext in pkg/hook/hooktest

vizansh:
Closes #831

PR #848, opened 2026-07-27
fix(ast): check return error of file.Close() in WriteFile

### LINKED ISSUE #842: fix(ast): check return error of file.Close() in WriteFile to prevent lost AST write errors
### Description
In `tool/internal/ast/parser.go`, `WriteFile` relies solely on `defer file.Close()` when flushing AST data to disk via `decorator.Restorer.Fprint`. If an I/O error occurs during the flush/close step (e.g. disk space exhausted), `file.Close()` returns an error that is currently ignored, causing `WriteFile` to return success (`nil`).

### Proposed Solution
Explicitly invoke and check `file.Close()` after calling `r.Fprint(file, root)` and return any close error wrapped with `ex.Wrapf`.

vizansh:
Closes #842


tool/internal/ast/parser.go
```diff
@@ -103,6 +103,11 @@ func WriteFile(filePath string, root *dst.File) error {
 	if err != nil {
 		return ex.Wrapf(err, "failed to write to file %s", filePath)
 	}
+	if err := file.Close(); err != nil {
```
anand-242003, line 106: Thanks for the fix! I noticed that `defer file.Close()` is still present while `file.Close()` is also called explicitly after `r.Fprint()`. Since the deferred call will still run when the function returns, wouldn't `Close()` be invoked twice on the success path?

Would it make sense to remove the deferred call and handle the close explicitly instead?

anand-242003 (commented):
Thanks for the changes! I left one inline comment regarding the `file.Close()` handling.

amazingakai:
Thanks for this PR! There is another PR fixing this problem as well:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/844

vizansh:
Thanks for pointing that out @amazingakai and @anand-242003! Since [#844](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/844) already addresses [#842](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/842), I'll go ahead and close this PR to avoid duplication.

Please feel free to review my other open PR when you get a chance: [#836](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/836). Thanks!

PR #850, opened 2026-07-27
fix(instrumentation): add missing span.RecordError() in gRPC stats handlers

shivv23:
## Description

Adds the missing span.RecordError() call in both gRPC client and server HandleRPC methods. When an RPC fails, the error is now properly recorded as an exception event on the span.

## Problem

Both clientStatsHandler.HandleRPC and serverStatsHandler.HandleRPC set span status to Error on failed RPCs but never call span.RecordError(err). This means:

- Errors don't appear in the span's event list
- Stack traces are not captured
- Error details are lost (only status code remains)
- Debugging production gRPC issues is significantly harder

## Solution

Add span.RecordError(rs.Error) in the *stats.End case of both handlers, following the same pattern already used by 
et/http/client, database/sql, 
edis, and other instrumentations in this repository.

### Changes

**instrumentation/google.golang.org/grpc/client/client_hook.go** — Added span.RecordError(rs.Error) before span.End() when 
s.Error != nil

**instrumentation/google.golang.org/grpc/server/server_hook.go** — Same change on the server side

**instrumentation/google.golang.org/grpc/client/client_hook_test.go** — Updated TestClientStatsHandler_HandleRPC_WithError to verify the exception event is created

**instrumentation/google.golang.org/grpc/server/server_hook_test.go** — Added TestServerStatsHandler_HandleRPC_WithError (no error test existed for the server)

## Testing

All existing tests pass, plus new assertions verify that xception events are created when RPCs fail.

`ash
cd instrumentation/google.golang.org/grpc/client && go test ./...  # 13/13 PASS
cd instrumentation/google.golang.org/grpc/server && go test ./...   # 9/9 PASS
`

## Checklist

- [x] Code formatted
- [x] Linters pass
- [x] Tests pass
- [x] Tests added for new behavior
- [x] No breaking changes
- [x] No new dependencies


Satya900 (changes requested):
@shivv23 Fix logic looks solid, rs.Error != nil matches exactly when SetStatus sets Error, ran both tests, they pass.

One thing worth checking: the bundle has CRLF in 121 files. main's own bundle is clean, so this came from this PR's regeneration, likely from running make package on a Windows checkout. verify-bundle CI may fail on it. Easy fix though, rebuild from git cat-file -p per file instead of cp -r, that reads the LF blobs directly.

shivv23:
@Satya900 Hi! Made all required fixes. Have a look, please.

amazingakai (approved):
Good catch, thanks ❤️

Satya900 (approved):
@shivv23  Re-verified after the latest push: bundle is genuinely CRLF-clean now, grpc client/server tests pass, otelc builds fine. This is resolved. Updating to Approve.

shivv23:
@Satya900 Sure! If everything is fine, then let's get this merged. Thanks!

PR #775, opened 2026-07-23
feat(instrumentation): record gen_ai.client.operation.duration for an…

malladinagarjuna2:
## Description

The Anthropic middleware measured request latency (`start := time.Now()`) but
threw it away — it was passed to `handleNonStreamingResponse` as a discarded
`_ time.Time` parameter and never recorded. This adds the standard GenAI
`gen_ai.client.operation.duration` histogram (unit `s`) and records it on
**every** outcome — success, transport error, HTTP error, and the SSE
fallback — so error latency is observable too.

Implementation notes:
- Creates the meter + histogram in `initInstrumentation`, mirroring the
  existing tracer setup.
- Records via a single `defer` registered once a span exists, so pass-through
  (non-instrumented) requests don't emit a measurement.
- Removes the now-unused `ctx`/`start` parameters from
  `handleNonStreamingResponse`.
- `error.type` is intentionally left off the metric's dimensions for now;
  that enrichment is a follow-up (see #679).

## Motivation

`gen_ai.client.operation.duration` is a standard GenAI semantic-convention
metric that observability backends expect for latency dashboards and SLOs.
The instrumentation was structured to compute it but silently emitted
nothing, so the signal was missing entirely.

Refs #679

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format`
- [ ] Linters pass: `make lint`
- [ ] Tests pass: `make test`
- [x] Tests added for new functionality
- [x] Tests follow [testing guidelines](docs/testing.md)
- [ ] Documentation updated (if applicable)
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](docs/instrument-guide.md#4-register-the-instrumentation))

PR #851, opened 2026-07-27
feat(instrumentation/openai): record gen_ai.client.token.usage

pradhyum6144:
OpenAI calls already record latency but not token counts. This adds the `gen_ai.client.token.usage` histogram (unit `{token}`) to the openai-go instrumentation, so token cost is visible alongside the `gen_ai.client.operation.duration` metric that #775 added for Anthropic. It is one of the two standard GenAI client metrics in semconv.

Input and output tokens are recorded as separate measurements, each tagged with `gen_ai.token.type` (`input` or `output`) as semconv expects. Covers all three non-streaming operations, chat, text_completion and embeddings, across the v1, v2 and v3 packages.

The meter and histogram are created in `initInstrumentation` next to the tracer, following the same shape as #775. The v2 and v3 copies are generated from v1 so the three stay identical apart from package name and import path, which is the invariant `openai.yaml` relies on.

Zero versus absent: the parsed usage object is a pointer and `recordTokenUsage` takes a `*int64` per token type. A response that reports zero tokens records a zero data point; a response with no `usage` object at all records nothing. That keeps the data point count usable as "how many operations reported usage" rather than collapsing the two cases together. Embeddings report the output side as absent rather than zero, since the response has no `completion_tokens` field.

**Streaming is out of scope for this PR.** Since #990 the streaming reader is a shared module under `internal/streaming`, and its `StreamingReader` deliberately carries no context, model, provider or operation name so that one copy can serve v1, v2 and v3. The metric instrument lives in the versioned package, so the shared reader has no way to reach it. Wiring this up needs a recorder passed into `NewStreamingReader` from each version's bridge, which changes the shared module's API and is better reviewed on its own. Happy to open that as a follow-up, or fold it in here if you would rather it land together.

## Motivation

`gen_ai.client.token.usage` is one of the two standard GenAI client metrics in semconv. With duration landing in #775, token usage is the obvious next one so operators can see token cost per call, not just latency.

## Testing

Unit tests use a manual metric reader to assert the recorded per-type counts for chat, text_completion and embeddings, plus a test pinning the zero versus absent distinction. `go test ./...` passes for all three modules, and the bundle is regenerated.

---

## Checklist

- [x] Tests pass: `make test`


Satya900 (approved):
@pradhyum6144 Careful work here. Metric setup, attribute cardinality, and the slice-copying in recordTokenUsage (avoids mutating the shared baseAttrs) all check out. Ran the full test suite, passes.

Two things worth noting, not blockers: this uses r.inputTokens/r.outputTokens from streaming's finalize(), which #828 found has a bug (final chunk gets dropped if the stream ends without a trailing newline). This PR predates that fix, so the new metric will share the same edge case until it's rebased on top of it once #828 merges. Also, this only touches v1, not v2/v3, unlike the rest of this file. Intentional, or a planned follow-up?

Satya900:
@pradhyum6144  Please resolve the conflicts

darccio:
@pradhyum6144 Regenerate the bundle and please follow the instructions in the EasyCLA check.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -222,13 +223,36 @@ func handleNonStreamingResponse(
...
+		attrs = append(attrs, baseAttrs...)
+		attrs = append(attrs, semconv.GenAITokenType(tokenType))
+		tokenUsage.Record(ctx, tokens, metric.WithAttributes(attrs...))
+	}
+	if inputTokens > 0 {
+		record(inputTokens, "input")
+	}
+	if outputTokens > 0 {
+		record(outputTokens, "output")
 	}
```
darccio: `recordTokenUsage` only emits a data point when the count is `> 0`, so a response
with genuinely zero tokens of a type (or a response body that omits `usage`
entirely) produces the same observable result: no data point at all. That's fine
for histogram *sums*, but it means the data point *count* for
`gen_ai.client.token.usage` can't be used to infer "how many operations had usage
data" — zero-usage and missing-usage are indistinguishable downstream.

Is skipping on zero intentional (e.g. to avoid noise from embeddings' always-zero
output side), or should this record `0` whenever the field was actually present
in the response? If it's the former, worth a one-line comment here saying so
explicitly, since the current comment above `recordTokenUsage` doesn't call out
the zero-skip as deliberate.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -193,19 +194,19 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 			span.SetAttributes(semconv.GenAIRequestIsStream(true))
 			resp.Body = newStreamingReader(resp.Body, span, start, op)
 		} else {
-			handleNonStreamingResponse(ctx, resp, span, start, op)
+			handleNonStreamingResponse(ctx, resp, span, op, baseAttrs)
 		}
```
darccio, line 198: The PR description says token usage is recorded "in the streaming reader once
the stream finishes," but this branch never calls `recordTokenUsage` — the
streaming reader only sets span attributes (via the `streaming` package),
and `newStreamingReader` isn't given the histogram or `baseAttrs` at all.
`TestOtelMiddleware_StreamingResponse` confirms this: it asserts the
`gen_ai.usage.*` span attributes from a stream with a `usage` block, but never
checks the `gen_ai.client.token.usage` metric.

Since streaming is a common path for chat completions, this means token usage
goes unreported for a large share of real traffic. Either wire token recording
into the streaming finish path (mirroring how `gen_ai.usage.*` span attributes
are already set there), or update the PR description to say streaming is out
of scope for this change.

instrumentation/github.com/openai/openai-go/token_usage_test.go
```diff
@@ -0,0 +1,122 @@
...
+		}, nil
+	}
+
+	_, err := OtelMiddleware()(req, next)
+	require.NoError(t, err)
+
+	usage := tokenUsageByType(t, reader)
+	assert.Equal(t, int64(2), usage["input"], "input tokens")
+	assert.Len(t, usage, 1, "embeddings record only input tokens")
+}
```
darccio, line 122: Chat and Embedding both get a token-usage test, but `opCompletion` (text
completion) runs through the identical `recordTokenUsage` call in
`handleNonStreamingResponse` and has no coverage here. Worth adding a
`TestOtelMiddleware_RecordsTokenUsage_Completion` alongside these two so all
three non-streaming operations are verified consistently.

```suggestion
}

// TestOtelMiddleware_RecordsTokenUsage_Completion verifies a text completion
// call records one input and one output token.usage measurement.
func TestOtelMiddleware_RecordsTokenUsage_Completion(t *testing.T) {
	setupTestTracer(t)
	reader := setupTestMeter(t)

	req, _ := http.NewRequest(
		http.MethodPost,
		"http://api.openai.com/v1/completions",
		io.NopCloser(bytes.NewReader([]byte(`{"model":"gpt-3.5-turbo-instruct"}`))),
	)
	next := func(_ *http.Request) (*http.Response, error) {
		return &http.Response{
			StatusCode: http.StatusOK,
			Header:     http.Header{"Content-Type": []string{"application/json"}},
			Body: io.NopCloser(bytes.NewReader([]byte(
				`{"id":"cmpl-1","model":"gpt-3.5-turbo-instruct","choices":[{"finish_reason":"stop"}],` +
					`"usage":{"prompt_tokens":6,"completion_tokens":9,"total_tokens":15}}`))),
		}, nil
	}

	_, err := OtelMiddleware()(req, next)
	require.NoError(t, err)

	usage := tokenUsageByType(t, reader)
	assert.Equal(t, int64(6), usage["input"], "input tokens")
	assert.Equal(t, int64(9), usage["output"], "output tokens")
	assert.Len(t, usage, 2, "completion records both input and output")
}
```

darccio (commented):
`gen_ai.client.token.usage` is added only to the v1 openai-go instrumentation.
The v2 and v3 packages (`instrumentation/github.com/openai/openai-go/v2` and
`v3`) have their own independent `hook.go`/`middleware.go` and are untouched by
this PR, so token usage won't be reported for clients on those SDK versions.

Is this intentional scoping (e.g. a planned follow-up per version), or should
v2/v3 get the same histogram in this PR? If it's deliberate, worth saying so
explicitly in the PR description, since the title ("record
gen_ai.client.token.usage") reads as covering the openai-go instrumentation
generally rather than v1 specifically.

darccio (changes requested):

Satya900:
Hey @pradhyum6144 , the merge conflicts from before are still sitting there unresolved. Whenever you get a chance, could you rebase and clear them so this can move forward? Ping me if you hit anything weird or want a hand.

PR #853, opened 2026-07-28
fix(imports): check return error of file.Close() in WriteFile

vedant21-oss:
### Problem Description
In `tool/internal/imports/importcfg.go`, `WriteFile` uses `defer file.Close()`, which ignores the error returned by `file.Close()` when writing files:

```go
func (r *ImportConfig) WriteFile(filename string) error {
	file, err := os.Create(filename)
	if err != nil {
		return err
	}
	defer file.Close()

	return r.write(file)
}
```

If `r.write(file)` succeeds but flushing buffered writes during `file.Close()` fails, `WriteFile` returns `nil` (success), hiding the corrupted/incomplete file write.

### Solution
Follow the same error-handling pattern as #844 (`tool/internal/ast/parser.go`): explicitly check `file.Close()` upon successful write, close the file handle on write error, and wrap errors using `ex.Wrapf`.


Satya900:
@vedant21-oss Nice work on both fixes, verified this one too, same solid pattern as #844. Small tip for next time: since it was the identical bug in a different file, bundling both into one PR would've saved a review cycle. Not a blocker here, just something to keep in mind going forward. LGTM.

amazingakai (approved):
LGTM, Please fix the CI/Linter issues.

vedant21-oss:
Hi @amazingakai! 👋

I've resolved the CI linter issues and updated test coverage:
- Fixed `govet` variable shadowing on `err` in `importcfg.go`.
- Added unit test `TestWriteFile_CreateError` to cover the file creation error path in `WriteFile`.

All linters and unit tests are now passing cleanly. Thanks for reviewing!

Issue #543, opened 2026-05-27
feat(tool): rule schema linter (detect invalid selectors / modifiers / shapes)


kakkoyun:
Follow-up from #377 (review thread from @txabman42 on `docs/rules.md:260`).

## Context

Today, an unknown selector or modifier in a rule YAML file (e.g. `where: { fnc: Foo }` or `do: { inject_hook: ... }` — note the typos) is silently dropped. The rule then matches nothing and produces no instrumentation, with no error surfaced to the rule author.

## Problem

Typos and stale rule files are hard to detect. Rule authors get no feedback at load time. As the schema grows over OP4.1–OP4.6, the number of valid keys grows and the surface area for silent typos grows with it.

## Proposed work

- Depends on the central registry follow-up (see related issue).
- New entry point — either `otelc rules lint` subcommand or `make lint/rules` target — that walks all rule YAML files in the project and reports: unknown selectors, unknown modifiers, modifier-payload type mismatches, missing required fields, mutually exclusive combinations.
- Reports a structured diagnostic per finding (file, line, message, suggestion if a close-by valid name exists, à la "did you mean `inject_hooks`?").
- Wire into CI (`make lint`).

## References

- #377 inline thread on `docs/rules.md:260`

kakkoyun:
cc @amazingakai

kakkoyun:
We can also have a layer to generate json schema similar to https://github.com/DataDog/orchestrion/blob/main/internal/injector/config/schema.json for our yaml files.  

ref: 
- https://github.com/invopop/jsonschema

PR #854, opened 2026-07-28
refactor(tool/rule): central registry of valid selectors and modifiers

### LINKED ISSUE #542: feat(tool/rule): central registry of valid selectors and modifiers

Harsh63870:
## Summary
- Add `tool/internal/rule/schema.go` with `Selector`/`Modifier` types, constants, `IsValidSelector`/`IsValidModifier`, and `AllSelectors`/`AllModifiers`.
- Move existing `Sel*` / combinator string aliases onto that registry so call sites share one source of truth.
- Reject unknown `do` modifier keys at normalize time instead of silently dropping the modifier name.

## Motivation
Fixes #542. Prerequisite for the rule-schema linter (#543). Unknown modifiers previously fell through because `normalizeDo` discarded the modifier key and only kept the payload.

## Test plan
- [ ] CI: `tool/internal/rule` unit tests (schema + normalize unknown-modifier cases)
- [ ] Existing rule golden / setup parse paths still load known modifiers

## Notes for reviewers
- Modifier names match `docs/rules.md` (not the older names in the issue description).
- No constructor refactor (#541/#546 left untouched).
- `createRuleFromFields` still dispatches on flat field presence; this PR only adds the registry + validation.

### DIFF
--- a/tool/internal/rule/schema.go
+++ b/tool/internal/rule/schema.go
@@ -12,7 +12,7 @@ type Selector string
 // per docs/rules.md, declares the rule type.
 type Modifier string
 
-// Point selectors that may appear under where (or as legacy flat fields).
+// Selector keys in the structured rule schema (where selectors, file/combinator keys, plus top-level target/version).
 const (
 	SelectorTarget            Selector = "target"
 	SelectorVersion           Selector = "version"
@@ -74,7 +74,7 @@ const (
 	CombNot   = string(SelectorNot)
 
 	// RawField is the modifier-output key produced by normalize for raw rules
-	// (inject_code payload). It is not a where selector; match.go shares it.
+	// (inject_code payload). It is not a where selector; tool/internal/setup/match.go shares it.
 	RawField = "raw"
 )
 
@@ -88,40 +88,27 @@ const (
 // file group, and combinators). target/version are intentionally included so
 // IsValidSelector can distinguish "known but misplaced" from "unknown";
 // normalizeWhere still rejects them inside where.
-var selectors = map[Selector]struct{}{
-	SelectorTarget:            {},
-	SelectorVersion:           {},
-	SelectorFunc:              {},
-	SelectorRecv:              {},
-	SelectorStruct:            {},
-	SelectorFunctionCall:      {},
-	SelectorDirective:         {},
-	SelectorKind:              {},
-	SelectorIdentifier:        {},
-	SelectorSignature:         {},
-	SelectorSignatureContains: {},
-	SelectorResult:            {},
-	SelectorLastResult:        {},
-	SelectorParam:             {},
-	SelectorPattern:           {},
-	SelectorPlacement:         {},
-	SelectorFile:              {},
-	SelectorAllOf:             {},
-	SelectorOneOf:             {},
-	SelectorNot:               {},
+
+//nolint:gochecknoglobals // lookup set derived from AllSelectors(); intentionally package-level so IsValidSelector can query it without reconstruction
+var selectors = toSet(AllSelectors())
+
+// toSet builds a lookup set from an ordered slice, so the set can never
+// drift from the slice it is derived from.
+func toSet[T comparable](items []T) map[T]struct{} {
+	set := make(map[T]struct{}, len(items))
+	for _, item := range items {
+		set[item] = struct{}{}
+	}
+	return set
 }
 
-var modifiers = map[Modifier]struct{}{
-	ModifierInjectHooks:     {},
-	ModifierInjectCode:      {},
-	ModifierAddStructFields: {},
-	ModifierAddFile:         {},
-	ModifierWrapCall:        {},
-	ModifierExpandDirective: {},
-	ModifierAssignValue:     {},
-}
+//nolint:gochecknoglobals // lookup set derived from AllModifiers(); intentionally package-level so IsValidModifier can query it without reconstruction
+var modifiers = toSet(AllModifiers())
 
-// IsValidSelector reports whether s is a known where-clause key.
+// IsValidSelector reports whether s is part of the rule schema's selector
+// vocabulary. This includes target/version, which are valid selectors but
+// are rejected specifically when used inside a where block — see
+// normalizeWhere.
 func IsValidSelector(s string) bool {
 	_, ok := selectors[Selector(s)]
 	return ok
--- a/tool/internal/rule/schema.go
+++ b/tool/internal/rule/schema.go
@@ -1,0 +1,173 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package rule
+
+// Selector is a where-clause key recognized by the structured rule schema
+// (ADR-0003). Point selectors are hoisted to flat fields by Normalize; file
+// and combinator keys stay nested under where.
+type Selector string
+
+// Modifier is a do-clause key that names the instrumentation action and,
+// per docs/rules.md, declares the rule type.
+type Modifier string
+
+// Point selectors that may appear under where (or as legacy flat fields).
+const (
+	SelectorTarget            Selector = "target"
+	SelectorVersion           Selector = "version"
+	SelectorFunc              Selector = "func"
+	SelectorRecv              Selector = "recv"
+	SelectorStruct            Selector = "struct"
+	SelectorFunctionCall      Selector = "function_call"
+	SelectorDirective         Selector = "directive"
+	SelectorKind              Selector = "kind"
+	SelectorIdentifier        Selector = "identifier"
+	SelectorSignature         Selector = "signature"
+	SelectorSignatureContains Selector = "signature_contains"
+	SelectorResult            Selector = "result"
+	SelectorLastResult        Selector = "last_result"
+	SelectorParam             Selector = "param"
+	SelectorPattern           Selector = "pattern"
+	SelectorPlacement         Selector = "placement"
+	SelectorFile              Selector = "file"
+	SelectorAllOf             Selector = "all-of"
+	SelectorOneOf             Selector = "one-of"
+	SelectorNot               Selector = "not"
+)
+
+// do modifiers (docs/rules.md § modifier names → rule types).
+const (
+	ModifierInjectHooks     Modifier = "inject_hooks"
+	ModifierInjectCode      Modifier = "inject_code"
+	ModifierAddStructFields Modifier = "add_struct_fields"
+	ModifierAddFile         Modifier = "add_file"
+	ModifierWrapCall        Modifier = "wrap_call"
+	ModifierExpandDirective Modifier = "expand_directive"
+	ModifierAssignValue     Modifier = "assign_value"
+)
+
+// String forms used as YAML / map keys. Existing call sites (normalize,
+// match) keep using these aliases so the registry is a single source of truth
+// without forcing every consumer onto the typed Selector/Modifier APIs.
+const (
+	SelTarget            = string(SelectorTarget)
+	SelVersion           = string(SelectorVersion)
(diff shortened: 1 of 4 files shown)


Harsh63870:
Hi @kakkoyun, just a friendly ping whenever you have time to review this PR.
I noticed #770 is also addressing #542, so I'll leave it to the maintainers to decide which approach fits better.
For context, this PR:
- Introduces `tool/internal/rule/schema.go` as a typed registry for selectors and modifiers.
- Centralizes the shared selector/modifier definitions so normalization and matching use the same source of truth.
- Validates unknown `do` modifier keys during normalization (for both map and sequence forms), with tests covering the behavior.
- Uses the modifier names currently documented in `docs/rules.md`.
- Keeps the scope limited to #542 without touching #541 or #546.
The required checks are still waiting for maintainer workflow approval. If there are any changes you'd like to see, I'm happy to update the PR.
Thanks!
cc @amazingakai

om7057:
Worth a look at #846 before this goes much further, since the two overlap on the same vocabulary.

#846 adds `tool/cmd/otelc-lint`, which generates `tool/data/schema.json` from the rule structs' `json` and `jsonschema` tags, and wires `make lint/rules` to validate the instrumentation YAML against it. That generated schema already carries every name this registry declares. The point selectors come from the struct tags (`func`, `recv`, `signature`, `signature_contains`, `result`, `last_result`, `param`, plus `target` and `version` on `InstBaseRule`), and `schema.json` also contains `where`, `do`, the combinators `all-of`, `one-of`, `not`, and all seven `do` modifiers.

The two aren't doing the same job. A generated JSON Schema validates rule files from the outside, and typed Go constants are for in-process call sites like `normalize.go`. But the list of names would then live in two places with nothing keeping them in agreement, which is the drift #705 is trying to design away.

Might be worth deriving one from the other instead of maintaining both. No opinion on which direction that should go.

Harsh63870:
thanks @om7057 — good catch on the overlap with #846.
Agreed they solve different layers:
- this PR: typed Go registry for in-process call sites (`normalize`, future tooling helpers)
- #846: generated JSON Schema for external rule-file validation via `otelc-lint`
Also agreed that having the same vocabulary in two places without a sync story recreates the drift problem #542/#705 are aiming at.
Happy to follow maintainer preference here. A couple of options that keep this PR small:
1. Land this as the in-process registry for #542, and open a follow-up to derive one side from the other once #846’s shape is settled.
2. If maintainers want a single source of truth first, I’m happy to adjust this PR, rather than ship two unsynced lists.
I intentionally left #541/#546 out of scope here so this stays a focused registry + unknown-modifier rejection change. Open to direction from @kakkoyun / @amazingakai.

om7057:
@Harsh63870 Both read fine to me. Option 1 seems like the smaller ask, land the registry now since #542 doesn't depend on #846's shape, and let the sync question get settled once otelc-lint stabilizes rather than blocking on it now.

Harsh63870:
thanks @om7057, agreed, i'll keep this PR scoped to the #542 registry + unknown-modifier validation, and treat syncing with #846 / otelc-lint as a follow-up once that side settles, happy to open that follow-up issue after this merges if maintainers want it tracked separately.

tool/internal/rule/schema.go
```diff
@@ -0,0 +1,173 @@
...
+	SelectorResult:            {},
+	SelectorLastResult:        {},
+	SelectorParam:             {},
+	SelectorPattern:           {},
+	SelectorPlacement:         {},
+	SelectorFile:              {},
+	SelectorAllOf:             {},
+	SelectorOneOf:             {},
+	SelectorNot:               {},
+}
```
darccio: Three lists have to stay hand-synced here: the typed `Selector*` consts, this
map, and `AllSelectors()`. `TestAllSelectorsComplete` proves the map and the
slice agree, but nothing checks a newly-added const actually made it into
either — it would just silently never validate. Deriving the map from the
slice removes one of the three copies; the codebase already has precedent for
a small generic helper like this (see `addRule[T InstRule]` in base.go:165).

```suggestion
var selectors = toSet(AllSelectors())

// toSet builds a lookup set from an ordered slice, so the set can never
// drift from the slice it is derived from.
func toSet[T comparable](items []T) map[T]struct{} {
	set := make(map[T]struct{}, len(items))
	for _, item := range items {
		set[item] = struct{}{}
	}
	return set
}
```

tool/internal/rule/schema.go
```diff
@@ -0,0 +1,173 @@
...
+
+var modifiers = map[Modifier]struct{}{
+	ModifierInjectHooks:     {},
+	ModifierInjectCode:      {},
+	ModifierAddStructFields: {},
+	ModifierAddFile:         {},
+	ModifierWrapCall:        {},
+	ModifierExpandDirective: {},
+	ModifierAssignValue:     {},
+}
```
darccio: Same duplication as the `selectors` map above — derive from `AllModifiers()`
once `toSet` exists.

```suggestion
var modifiers = toSet(AllModifiers())
```

tool/internal/rule/schema.go
```diff
@@ -0,0 +1,173 @@
...
+
+// IsValidSelector reports whether s is a known where-clause key.
```
darccio: This doc comment says `IsValidSelector` reports a "known where-clause key,"
but `target`/`version` return `true` here while being explicitly illegal
inside `where` (rejected earlier in `normalizeWhere`, before this check ever
runs — see ADR-0003, "target and version stay at the top level"). The
selectors map comment above explains this is deliberate (so a future linter
can distinguish "known but misplaced" from "unknown"), but that nuance isn't
visible from the exported function's own doc comment, which is what a caller
outside this package will actually read.

```suggestion
// IsValidSelector reports whether s is part of the rule schema's selector
// vocabulary. This includes target/version, which are valid selectors but
// are rejected specifically when used inside a where block — see
// normalizeWhere.
```

tool/internal/rule/schema.go
```diff
@@ -0,0 +1,173 @@
...
+	SelPattern           = string(SelectorPattern)
+	SelPlacement         = string(SelectorPlacement)
+
+	WhereFile = string(SelectorFile)
+	CombAllOf = string(SelectorAllOf)
+	CombOneOf = string(SelectorOneOf)
+	CombNot   = string(SelectorNot)
+
+	// RawField is the modifier-output key produced by normalize for raw rules
+	// (inject_code payload). It is not a where selector; match.go shares it.
```
darccio: "match.go" here means tool/internal/setup/match.go, a different package —
worth naming it explicitly so this doesn't read as an intra-package
reference.

```suggestion
	// RawField is the modifier-output key produced by normalize for raw rules
	// (inject_code payload). It is not a where selector; tool/internal/setup/match.go shares it.
```

tool/internal/rule/schema_test.go
```diff
@@ -0,0 +1,69 @@
...
+	assert.Equal(t, string(SelectorFile), WhereFile)
+	assert.Equal(t, string(SelectorAllOf), CombAllOf)
+	assert.Equal(t, string(ModifierInjectHooks), "inject_hooks")
+	assert.Equal(t, string(ModifierInjectCode), "inject_code")
+	assert.Equal(t, string(ModifierAddStructFields), "add_struct_fields")
+	assert.Equal(t, string(ModifierAddFile), "add_file")
+	assert.Equal(t, string(ModifierWrapCall), "wrap_call")
+	assert.Equal(t, string(ModifierExpandDirective), "expand_directive")
+	assert.Equal(t, string(ModifierAssignValue), "assign_value")
+}
```
darccio, line 102: This test exhaustively checks all 7 modifier aliases but only spot-checks 3
of the 20 selector aliases (SelFunc, WhereFile, CombAllOf). The other 16
(SelTarget, SelVersion, SelRecv, SelStruct, SelFunctionCall, SelDirective,
SelKind, SelIdentifier, SelSignature, SelSignatureContains, SelResult,
SelLastResult, SelParam, SelPattern, SelPlacement, CombOneOf, CombNot) aren't
asserted here, and TestNormalize's where-hoisting cases don't exercise all of
them either (struct, function_call, directive, kind, identifier, pattern,
placement, one-of, not have no positive-path coverage in this package). Since
this PR's own description frames this as "moving" existing aliases onto the
registry, a transposition during that move (two adjacent aliases swapped)
would compile cleanly and only surface as a silent wrong-selector match at
runtime — this test is the one place designed to catch exactly that, and it
doesn't check most of them.

```suggestion
func TestStringAliasesMatchTypedConstants(t *testing.T) {
	selectorCases := []struct {
		alias string
		typed Selector
	}{
		{SelTarget, SelectorTarget},
		{SelVersion, SelectorVersion},
		{SelFunc, SelectorFunc},
		{SelRecv, SelectorRecv},
		{SelStruct, SelectorStruct},
		{SelFunctionCall, SelectorFunctionCall},
		{SelDirective, SelectorDirective},
		{SelKind, SelectorKind},
		{SelIdentifier, SelectorIdentifier},
		{SelSignature, SelectorSignature},
		{SelSignatureContains, SelectorSignatureContains},
		{SelResult, SelectorResult},
		{SelLastResult, SelectorLastResult},
		{SelParam, SelectorParam},
		{SelPattern, SelectorPattern},
		{SelPlacement, SelectorPlacement},
		{WhereFile, SelectorFile},
		{CombAllOf, SelectorAllOf},
		{CombOneOf, SelectorOneOf},
		{CombNot, SelectorNot},
	}
	for _, c := range selectorCases {
		assert.Equalf(t, string(c.typed), c.alias, "alias must equal string(%v)", c.typed)
	}

	modifierCases := []struct {
		alias string
		typed Modifier
	}{
		{"inject_hooks", ModifierInjectHooks},
		{"inject_code", ModifierInjectCode},
		{"add_struct_fields", ModifierAddStructFields},
		{"add_file", ModifierAddFile},
		{"wrap_call", ModifierWrapCall},
		{"expand_directive", ModifierExpandDirective},
		{"assign_value", ModifierAssignValue},
	}
	for _, c := range modifierCases {
		assert.Equalf(t, string(c.typed), c.alias, "alias must equal string(%v)", c.typed)
	}
}
```

darccio (changes requested):

Harsh63870:
@darccio @txabman42 @amazingakai, could you please review..

darccio (approved):

PR #841, opened 2026-07-27
test(tool/ast): add unit test coverage for generic receiver matching and stripGenericTypes

bhuvan-somisetty:
## Description

This PR adds unit test coverage for `stripGenericTypes` and generic receiver matching in `FindFuncDecl` within `tool/internal/ast/shared_test.go`.

Specifically, it adds:
- `TestStripGenericTypes` covering unqualified value/pointer receivers, single-parameter generic value/pointer receivers (`GenStruct[T]`, `*GenStruct[T]`), multi-parameter generic value/pointer receivers (`GenStruct[T, U]`, `*GenStruct[T, U]`), and unsupported AST expression fallbacks.
- Extended `TestFindFuncDecl` test cases covering generic receivers (`GenSingle`, `*GenSingle`, `GenMulti`, `*GenMulti`) as well as `InstRawRule` and `FilterDef` target types.

## Motivation

Fixes #833

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format`
- [x] Linters pass: `make lint`
- [x] Tests pass: `make test`
- [x] Tests added for new functionality
- [x] Tests follow [testing guidelines](docs/testing.md)
- [x] Documentation updated (if applicable)

PR #857, opened 2026-07-28
fix(ci): scope unit coverage uploads to their own flag

om7057:
## Description

`codecov/project/tool` is currently red on every open PR that touches `tool/`, and the number behind it is not `tool/`'s coverage.

`make test-unit/coverage` runs three sub-targets and leaves `coverage-tool.txt`, `coverage-pkg.txt` and `coverage-instrumentation.txt` in the workspace. Both upload steps name one file via `files:`, but that input does not restrict the upload. From the action's own input description at the pinned SHA:

> `files`: Comma-separated list of explicit files to upload. These will be added to the coverage files found for upload. If you wish to only upload the specified files, please consider using disable-search.

So the `flags: tool` step uploads all three profiles tagged as `tool`, and the `flags: pkg` step does the same. The `tool` flag ends up describing the whole repo: 115 files rather than the 63 under `tool/`. This is visible at the flag level, not just the status level. Codecov's API reports the stored `tool` flag coverage as 64.674%.

This sets `disable_search: true` on both steps so each uploads only the profile it names.

`codecov.yml` needs no change. The `flags:` scoping there was always correct, it was reading a mispackaged flag.

## Motivation

The mispackaging is not new. It became visible when the merged number crossed the 70% target. Comparing Codecov's reports for two PRs either side of that:

| | #799 (Jul 24) | #840 (Jul 28) |
|---|---|---|
| files | 109 | 115 |
| statements | 9,001 | 11,199 |
| misses | 1,527 | 3,507 |
| coverage | 78.14% (pass) | 64.70% (fail) |

#799's report was already at 109 files, so it was already measuring more than `tool/`. It happened to sit above target. The six extra files are the linodego v2 instrumentation added in c5ab841, dominated by `public_methods_gen.go` (4,512 lines of generated wrappers), contributing roughly 1,980 additional misses. #840's `base_totals` is 64.67%, so `main` itself is now below the target.

Two related symptoms this explains:

- #841 adds unit tests in `tool/ast` and still fails the check. It cannot pass, because most of the number it is judged on is not `tool/`.
- #799's green check is stale (Jul 24, before c5ab841) and will turn red on its next push.

With the flag scoped correctly, `go test ./tool/...` gives 78.8% over 63 files, above the 70% target.

## Verification

This PR touches only `.github/workflows/test-unit.yaml`, so the `tool` paths filter (`tool/**`) is false and the upload step's `if:` does not fire. `codecov/project/tool` will report "No coverage information found on head" here rather than 78.8%. The check is only exercised by a PR that touches `tool/`, so I can rebase #840 once this lands to confirm it goes green.

Locally: `yamlfmt -lint` and `actionlint` are clean on the changed file, and `go test ./tool/... -coverprofile` gives 78.8% over 63 files with only `tool/` paths in the profile.

---

## Checklist

- [x] Code formatted: `make format` (no Go changes; `yamlfmt` clean on the changed file)
- [x] Linters pass: `make lint` (`yamlfmt -lint` and `actionlint` clean on the changed file)
- [x] Tests pass: `make test` (no code change; `./tool/...` run locally for the coverage numbers above)
- [ ] Tests added for new functionality (n/a, CI configuration only)
- [ ] Tests follow [testing guidelines](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/testing.md) (n/a)
- [ ] Documentation updated (if applicable) (n/a)
- [ ] OpenTelemetry Registry updated (n/a, no new instrumentation)

[^I-know-my-stuff]:


om7057:
cc: @amazingakai @txabman42 since it touches the codecov upload steps from #227.

`files:` doesn't restrict the upload; it adds to what the action finds on its own, so both steps have been sending all three profiles that `make test-unit/coverage` leaves in the workspace. The `tool` flag is 115 files instead of the 63 under `tool/`, and Codecov's API reports the flag itself at 64.674%. That's why `codecov/project/tool` is red on every open PR touching `tool/`, including #841, which only adds tests in `tool/ast` and still can't pass.

Open question: `disable_search: true` on the upload steps, or per-flag `paths` under `flag_management` in `codecov.yml`? I went with the former since it fixes what gets uploaded rather than filtering it afterwards, but happy to switch if there's a preference.

anand-242003:
I verified that make test-unit/coverage produces the three coverage profiles mentioned in the PR and confirmed from the Codecov documentation that files supplements, rather than replaces, the automatic file search. Given that behavior, disable_search: true appears to be the correct way to scope each upload to its intended coverage profile. I wasn't able to verify the runtime upload behavior because the workflow is still awaiting maintainer approval.

om7057:
Thanks for checking @anand-242003, that matches what I found.

On the runs: every `pull_request` workflow on this PR is sitting at `action_required`, so nothing has executed yet. That's the gate for fork PRs that touch `.github/workflows`, which is why the upload behaviour isn't observable from here.

Worth noting it won't be observable even once they run. This PR only changes the workflow file, so the `tool` paths filter (`tool/**`) is false and the upload step's `if:` never fires. The check will report "No coverage information found on head" rather than 78.8%. Exercising it needs a PR that touches `tool/`.

om7057:
@NameHaibinZhang could you also please check this one as well, if it's needed in order to fix-codecov flag's scope?

darccio (approved):
LGTM, I'll validate on the linked PRs once this lands on `main`

PR #554, opened 2026-05-30
feat: add compile-time auto-instrumentation for MongoDB

SatyamPandey-07:
# Description

This PR adds compile-time auto-instrumentation support for the official MongoDB Go driver (`go.mongodb.org/mongo-driver`).

Following the existing instrumentation patterns used for Redis and `database/sql`, this implementation hooks into MongoDB client creation APIs and automatically injects OpenTelemetry's MongoDB `CommandMonitor` (`otelmongo`) during compilation. Applications using the MongoDB Go driver can now generate trace telemetry without requiring any source code changes.

### MongoDB Instrumentation

* Added `pkg/instrumentation/mongodb/mongodb.otelc.yaml`

  * Defines compile-time hook mappings for:

    * `mongo.Connect`
    * `mongo.NewClient`

* Added `pkg/instrumentation/mongodb/client_hook.go`

  * Registers MongoDB OpenTelemetry instrumentation during client initialization.
  * Initializes the OpenTelemetry SDK through `shared.SetupOTelSDK`.
  * Creates and injects an `otelmongo.CommandMonitor`.
  * Updates function parameters using `ictx.SetParam` to propagate modified options to the target application.

### Integration Testing

* Added a new sample application:

  * `test/apps/mongoclient`

* Added a dedicated integration test:

  * `test/integration/mongo_client_test.go`

The test suite uses a lightweight pure-Go mock MongoDB server that:

* Handles MongoDB handshake requests (`hello` / `isMaster`)
* Processes command requests
* Returns BSON responses for insert operations

This approach eliminates dependencies on:

* MongoDB binaries
* Docker containers
* External services

and provides deterministic, CI-friendly integration testing.

## Verification

The new integration test validates that telemetry spans are emitted correctly for MongoDB operations, including:

* `users.insert`
* `endSessions`

The following semantic convention attributes are verified:

* `db.system = mongodb`
* `db.operation`
* `db.name`
* `db.mongodb.collection`
* `net.peer.name`
* `net.transport`

### Test Execution

```bash
go test -v -tags integration ./integration/ -run TestMongoClient
```

### Output

```text
=== RUN   TestMongoClient
=== RUN   TestMongoClient/basic
--- PASS: TestMongoClient (76.77s)
    --- PASS: TestMongoClient/basic (76.77s)
PASS
ok  	github.com/open-telemetry/opentelemetry-go-compile-instrumentation/test/integration	77.270s
```

#### users.insert

* Kind: `Client`
* `db.system`: `mongodb`
[...]

PR #858, opened 2026-07-29
feat(instrumentation/mongo): add mongo driver v2 support

markcottrell:
## Description

This PR is inspired by https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/554 (and https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/769) and adds similar instrumentation for mongo-driver v2, as v1 has been deprecated for ~6 months. Equivalent unit and integration tests have also been added for the v2 instrumentation.

All of the description from https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/554 applies here, but only `mongo.Client.Connect` is hooked, as `mongo.NewClient` does not exist in v2.

The integration test client is updated to use either the v1 or v2 driver based on a command line parameter, and a new test has been added for v2 as the properties recorded by otelmongo v2 have different names. I also added a new test utility, `RequireAttributeContains`, to allow partial matching of span attributes in tests.

## Motivation

mongo-driver v1 has been deprecated for ~6 months, this will allow automatic compile time instrumentation where mongo-driver v2 is in use.

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelc.yaml
```diff
@@ -0,0 +1,8 @@
+mongodb_v2_hook_connect:
```
NameHaibinZhang, line 1: The new v2 instrumentation module is present in the source tree, but the shipped CLI bundle (`tool/data/otelc-bundle.tgz`, updated in this PR) still only contains the v1 module under `instrumentation_temp/go.mongodb.org/mongo-driver/mongo/...` — there is no `instrumentation_temp/.../v2/mongo` entry (verified via `tar -tzf`). Since the CLI extracts instrumentation from `instrumentation_temp`, this rule won't be available to users who rely on the shipped bundle, so the v2 feature is effectively source-build-only. Could you regenerate the bundle so it includes the v2 module tree as well?

test/integration/mongo_client_test.go
```diff
@@ -57,6 +57,46 @@ func TestMongoClient(t *testing.T) {
...
+			require.GreaterOrEqual(t, len(spans), 1, "expected at least 1 span (insert)")
+
+			// Verify insert span matching the actual attributes from otelmongo
+			insertSpan := testutil.RequireSpan(t, f.Traces(),
+				testutil.IsClient,
+				testutil.HasAttribute("db.operation.name", "insert"),
+			)
+
+			// Assert MongoDB specific semantic conventions attributes
+			testutil.RequireAttribute(t, insertSpan, "db.system.name", "mongodb")
```
NameHaibinZhang, line 90: This test asserts the current DB semconv keys (`db.system.name`, `db.operation.name`, `db.namespace`, `db.collection.name`, `network.peer.address`, `network.transport`), but `schemas/otelc/groups/mongo.yaml` still only declares the legacy v1 attributes (`db.system`, `db.operation`, `db.name`, `db.mongodb.collection`, `net.peer.name`, `net.transport`). Since the registry is meant to be the source of truth for what otelc emits, it would be good to extend `mongo.yaml` with a v2 group covering the attribute keys asserted here (and update its comments to mention both v1 and v2). `make lint-schema` should still pass after the update.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"go.mongodb.org/mongo-driver/v2/mongo/options"
+	"go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo"
+	"go.opentelemetry.io/otelc/pkg/hook"
+	"go.opentelemetry.io/otelc/pkg/runtime"
+)
+
+const (
+	instrumentationKey = "MONGODB"
```
NameHaibinZhang: Both the v1 and v2 mongo instrumentation share `instrumentationKey = "MONGODB"`, so `OTEL_GO_ENABLED_INSTRUMENTATIONS` / `OTEL_GO_DISABLED_INSTRUMENTATIONS` will always toggle both drivers together — there's no way to instrument only v2 while keeping v1 uninstrumented (or vice versa) during a staged migration. Other versioned modules use a version-scoped key (e.g. `AWS_SDK_V2`). If sharing the key is intentional, a short doc note would help; otherwise consider a dedicated key such as `MONGODB_V2`.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/go.mod
```diff
@@ -0,0 +1,78 @@
...
+go 1.25.0
+
+replace go.opentelemetry.io/otelc/pkg => ../../../../../pkg
+
+replace go.opentelemetry.io/otelc/pkg/runtime => ../../../../../pkg/runtime
+
+require (
+	github.com/stretchr/testify v1.11.1
+	go.mongodb.org/mongo-driver/v2 v2.8.0
+	go.opentelemetry.io/contrib/instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelmongo v0.0.0-20260728191659-5f30960897bf
```
NameHaibinZhang, line 13: This depends on the v2 otelmongo via a pseudo-version (`v0.0.0-20260728191659-5f30960897bf`), whereas the v1 module pins a stable tag (`v0.52.0`). That works, but automated dependency tooling generally won't propose upgrades for pseudo-versions, so switching to the tagged release once contrib publishes one will require a manual step. Worth a brief comment here (or a release-checklist item) noting this is temporary until a stable v2 otelmongo tag is available.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go
```diff
@@ -0,0 +1,49 @@
...
+	// If no options were provided, create a default options struct
+	if len(opts) == 0 {
+		opts = []*options.ClientOptions{
+			options.Client(),
+		}
+	}
+
+	// Inject monitor to all existing options
+	for _, opt := range opts {
+		if opt.Monitor == nil {
```
NameHaibinZhang, line 45: If a caller does `mongo.Connect(nil)`, `opts` becomes `[]*options.ClientOptions{nil}` and `opt.Monitor` dereferences a nil pointer here (the trampoline's recover will swallow the panic, but the monitor is then never injected). A small guard keeps the hook robust:

```go
for _, opt := range opts {
    if opt == nil {
        continue
    }
    if opt.Monitor == nil {
        opt.SetMonitor(monitor)
    }
}
```

Same pattern exists in the v1 hook, so feel free to treat this as optional here.

NameHaibinZhang (commented):
Thanks for adding mongo-driver v2 support! The hook implementation itself looks solid — the variadic signature matches `Connect(opts ...*options.ClientOptions)` in v2.8.0, the `SetParam` write-back follows the same pattern as the v1 hook, and the enable/disable wiring works correctly. Unit and integration coverage of the hook behavior is good.

A few gaps around the contract/registry and distribution side, plus two smaller notes — see the inline comments. Two additional items that can't be attached inline because the files aren't part of this diff:

- `schemas/otelc/README.md`: the "Instrumentation coverage" table still only lists `go.mongodb.org/mongo-driver/mongo`. Please add a row for the v2 module (similar to how the openai v1/v2/v3 modules are listed per version), since a missing row signals an undeclared instrumentation per the registry conventions.
- `docs/api-design-and-project-structure.md`: the instrumentation module list only mentions the v1 mongo module; please add `go.mongodb.org/mongo-driver/v2/mongo` so the project structure doc reflects both supported driver versions.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/otelc.yaml
markcottrell: thanks, I thought I'd updated the bundle but must have missed it. I just pushed a commit with an updated bundle and confirmed the v2 files are now present:
```
$ tar tf tool/data/otelc-bundle.tgz | grep v2/mongo
instrumentation_temp/go.mongodb.org/mongo-driver/v2/mongo/
instrumentation_temp/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go
instrumentation_temp/go.mongodb.org/mongo-driver/v2/mongo/client_hook_test.go
instrumentation_temp/go.mongodb.org/mongo-driver/v2/mongo/go.mod
instrumentation_temp/go.mongodb.org/mongo-driver/v2/mongo/go.sum
instrumentation_temp/go.mongodb.org/mongo-driver/v2/mongo/otelc.yaml
```

test/integration/mongo_client_test.go
markcottrell: I added a section with the v2 span attributes

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go
markcottrell: updated to use `MONGODB_V2` instead, as suggested

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/go.mod
markcottrell: I added a comment in go.mod r.e. the pseudo version, let me know if that's sufficient

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go
markcottrell: I added the suggested guard and a test, thanks for catching it

NameHaibinZhang (approved):

NameHaibinZhang:
Hi @markcottrell — thanks for addressing all the review points; the instrumentation key, nil-option guard, pseudo-version comment, schema v2 group, docs, and bundle are all in good shape now.

The branch had fallen behind `main` (which caused the merge conflict), so I've rebased it onto the latest `main` and regenerated `tool/data/otelc-bundle.tgz` (needed so `verify-bundle` stays green after the rebase). This was a force-push to your PR branch, so please `git fetch` + `git reset --hard` your local branch before pushing again to avoid clobbering it — apologies if you had any unpushed local work.

The PR now shows as mergeable; just needs a maintainer approval.

NameHaibinZhang (approved):

PR #860, opened 2026-07-29
fix(tool/setup): close file handle when bundle extraction copy fails

### LINKED ISSUE #859: bug(setup): file handle leaked when bundle extraction fails partway through a copy
### Describe the bug

`extract()` in `tool/internal/setup/extract.go` opens the destination file for a tar entry before copying its contents, but only closes the handle on the success path:

```go
case tar.TypeReg:
    file, err := os.OpenFile(targetPath, os.O_CREATE|os.O_TRUNC|os.O_RDWR,
        fileInfo.Mode())
    if err != nil {
        return ex.Wrap(err)
    }

    _, err = io.CopyN(file, tarReader, header.Size)
    if err != nil {
        return ex.Wrap(err)
    }
    err = file.Close()
    if err != nil {
        return ex.Wrap(err)
    }
```

If `io.CopyN` fails — truncated/corrupt gzip stream, disk full mid-write, an unexpected EOF from the tar reader — the function returns on the error from `CopyN` without ever calling `file.Close()`. The open `*os.File` is leaked for the rest of the process.

### Environment

- OS: All (Linux, macOS, Windows)
- Version: main

### To Reproduce

1. Call `extract()` (via `extractGZip`) with a tar stream where a regular-file entry's declared `header.Size` is larger than the bytes actually available in the (possibly gzip-truncated) stream, so `io.CopyN` errors before finishing.
2. `extract()` returns the wrapped copy error, as expected.
3. Check open file descriptors for the process (`lsof -p <pid>` / `/proc/<pid>/fd`) — the destination file opened at the top of the `tar.TypeReg` branch is still open.

### Expected behavior

The file handle should be closed whether or not `io.CopyN` succeeds, so a failed extraction doesn't leak a descriptor or leave a dangling open handle on a partially-written file.

### Additional context

- Affected code: `tool/internal/setup/extract.go`, `extract()`, the `tar.TypeReg` branch (currently around lines 57-71).
- This runs on every `otelc setup`/build while unpacking the built-in instrumentation bundle (`extractOtelcBundle` → `extractGZip` → `extract`), so a corrupted or truncated bundle triggers the leak.
- #530 fixed truncation of previously-extracted files in this same function but didn't add close-on-error handling.
- `extractGZip` already uses `defer gzReader.Close()` a few lines below for the gzip reader — the same pattern (open, `defer file.Close()`, then just check the deferred error or copy error) would fix this too.

bhuvan-somisetty:
## Description

`extract()` in `tool/internal/setup/extract.go` opens the destination file for a regular tar entry, then copies the entry's bytes into it with `io.CopyN`. The file handle was only closed on the success path — if `io.CopyN` returned an error (truncated/corrupt bundle stream, disk full mid-write, unexpected EOF), the function returned immediately and the open `*os.File` leaked for the rest of the process.

This closes the file on the copy-error path too, mirroring the existing `_ = dstFile.Close()` pattern already used in `util.CopyFile` for the same situation.

Also adds `TestExtract_ClosesFileOnCopyError`, which builds a tar entry whose declared size is larger than the bytes actually written, so `io.CopyN` fails partway through. Without the fix, this test fails on Windows because `os.Remove` on the target file returns "used by another process" while the leaked handle is still open in-process; with the fix it passes.

## Motivation

Every `otelc setup`/build run extracts the built-in instrumentation bundle through this code path (`extractOtelcBundle` → `extractGZip` → `extract`), so a corrupted or truncated bundle would leak a file descriptor with no cleanup. #530 fixed truncation of previously-extracted files in this same function but didn't add close-on-error handling for the copy step.

Fixes #859

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/extract.go
+++ b/tool/internal/setup/extract.go
@@ -63,6 +63,7 @@ func extract(tarReader *tar.Reader, header *tar.Header, targetPath string) error
 
 		_, err = io.CopyN(file, tarReader, header.Size)
 		if err != nil {
+			_ = file.Close()
 			return ex.Wrap(err)
 		}
 		err = file.Close()
--- a/tool/internal/setup/extract_test.go
+++ b/tool/internal/setup/extract_test.go
@@ -107,6 +107,41 @@ func TestExtract_TruncatesExistingFile(t *testing.T) {
 	require.Equal(t, string(newContent), string(bs))
 }
 
+func TestExtract_ClosesFileOnCopyError(t *testing.T) {
+	tmpDir := t.TempDir()
+
+	targetPath := filepath.Join(tmpDir, "rules.yaml")
+
+	// Build a tar header that promises more bytes than the tar stream
+	// actually contains, so io.CopyN fails partway through the copy.
+	var tarBuf bytes.Buffer
+	tw := tar.NewWriter(&tarBuf)
+
+	err := tw.WriteHeader(&tar.Header{
+		Name:     "rules.yaml",
+		Mode:     0o644,
+		Size:     100,
+		Typeflag: tar.TypeReg,
+	})
+	require.NoError(t, err)
+
+	_, err = tw.Write([]byte("short"))
+	require.NoError(t, err)
+
+	tr := tar.NewReader(bytes.NewReader(tarBuf.Bytes()))
+
+	header, err := tr.Next()
+	require.NoError(t, err)
+
+	err = extract(tr, header, targetPath)
+	require.Error(t, err)
+
+	// If extract() left the destination file open, removing it here
+	// fails on Windows (the OS refuses to delete a file that still has
+	// an open handle in this process).
+	require.NoError(t, os.Remove(targetPath))
+}
+
 func TestExtract_Directory(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Satya900 (approved):
@bhuvan-somisetty  Verified the fix and test. Real leaked handle on the io.CopyN error path, one-line fix, and the os.Remove-fails-on-open-handle trick in the test is a clean, portable way to prove the fix works. LGTM.

bhuvan-somisetty:
Thanks for the review and for double checking the test logic, appreciate it!

darccio (commented):
LGTM

bhuvan-somisetty:
Looked into the codecov/project/tool failure — it's not from this change. The diff itself is fully covered (patch coverage 100%) and even nudges overall tool coverage up slightly (64.67% -> 64.69%), it's just that the module as a whole has been sitting below the 70% target for a while. Branch is also a bit behind main, updating it won't close a 5-point gap on its own. Didn't want to bolt on a pile of unrelated tests just to force this green on a one-line fix, so flagging it here in case a maintainer wants to override/adjust the check for this one.

darccio (approved):

PR #866, opened 2026-07-29
fix(instrumentation/openai): skip span creation when model is empty

### LINKED ISSUE #865: fix(instrumentation/openai): span still created when model is empty
### Describe the bug

`OtelMiddleware()` in the OpenAI GenAI instrumentation creates a span even when the request model can't be determined - either because the request body fails to parse (invalid JSON, or truncated by the `maxRequestBodySize` cap on a large multimodal payload) or because the body is valid JSON that simply omits the `model` field. In both cases the span still gets created, named `"chat "` (note the trailing space) with `gen_ai.request.model = ""`.

Affected files: `instrumentation/github.com/openai/openai-go{,/v2,/v3}/middleware.go`

### Environment

### To Reproduce

Send a chat completion request with a body that's valid JSON but missing `model`:

```go
req, _ := http.NewRequest("POST", "http://api.openai.com/v1/chat/completions",
    io.NopCloser(bytes.NewReader([]byte(`{"max_tokens":10}`))))
```

The middleware creates a span named `"chat "` with `gen_ai.request.model` set to an empty string, instead of skipping instrumentation.

### Expected behavior

When the model can't be determined, the request should pass through uninstrumented, the same way the middleware already handles other cases where there's nothing meaningful to attach (unknown operation, nil body, request read error).

### Additional context

The Anthropic instrumentation had the identical bug and was fixed with the same approach in #805 (`if model == "" { return next(req) }`). This issue is the OpenAI-side equivalent, found while reviewing that PR.

I'd like to work on this - the fix is a one-line guard applied to all three copies of `middleware.go` (v1/v2/v3), plus two test cases (invalid JSON, valid JSON missing the field) mirroring the ones added for the Anthropic fix.

Satya900:
## Description

`OtelMiddleware()` in the OpenAI GenAI instrumentation created a span even when the request model couldn't be determined - either because the request body failed to parse (invalid JSON, or truncated by the `maxRequestBodySize` cap on a large multimodal payload) or because the body was valid JSON that simply omitted the `model` field. In both cases the span still got created, named `"chat "` (trailing space) with `gen_ai.request.model = ""`.

This adds a guard: when `model == ""`, the request passes through uninstrumented, consistent with the other pass-through cases already in the middleware (unknown operation, nil body, request read error). Applied to all three copies of `middleware.go` (v1, v2, v3).

## Motivation

A span with an empty model attribute is malformed telemetry, it can't be attributed to any model in a backend, and the `"chat "` name is visibly broken. The Anthropic instrumentation had the identical bug, fixed the same way in #805. This is the OpenAI-side equivalent, found while reviewing that PR.

Fixes #865

## Testing

Added two cases per module (invalid JSON, valid JSON missing `model`), both asserting no span is created and the request still reaches the transport. `go test ./...` and scoped `golangci-lint` pass clean in all three affected modules.

---

## Checklist

- [x] Tests pass: `make test`


darccio:
@Satya900 Please address the failing CI check and the conflict.

Satya900:
@darccio  conflicts are resolved

darccio (approved):
LGTM

PR #867, opened 2026-07-29
fix(tool): handle os.Getwd error to prevent silent empty path fallback

### LINKED ISSUE #864: bug(util): GetOtelcWorkDir silently drops os.Getwd error and falls back to empty string
### Describe the bug

In `tool/util/shared.go`, `GetOtelcWorkDir()` silently ignores the error returned by `os.Getwd()`. If resolving the working directory fails (which is common in CI/CD containers or if the directory is removed mid-execution), it falls back to an empty string. This causes downstream path operations to scatter build artifacts into arbitrary relative locations instead of failing cleanly.

**Proposed Solution:**
Add an explicit error check for `os.Getwd()` to fail fast with a clear error message instead of silently returning an empty string.

sudo-muneeb:
## Description

Modifies `GetOtelcWorkDir()` in `tool/util/shared.go` to explicitly check the error returned by `os.Getwd()`. If it fails, the tool now logs a critical error to `stderr` and exits with status `1` instead of silently ignoring the failure.

## Motivation

Previously, the error from `os.Getwd()` was silently dropped using the blank identifier `_`. In CI/CD containers or environments with dropped volume mounts, failing to resolve the working directory resulted in an empty string (`""`). This silently corrupted downstream `filepath.Join` operations by generating relative paths, which scattered build artifacts into unpredictable locations. Failing fast with a clear diagnostic message prevents this silent build corruption.

Fixes #864

## Checklist

- [x] Tests pass: `make test`


Satya900 (approved):
@sudo-muneeb  Verified ex.Fatal does a hard os.Exit(1), and tool/util is never imported into pkg/ or instrumentation/, so this only affects otelc's own build-time execution, not a user's running app. Correct fix. Build and tests pass.

**LGTM.**

darccio (approved):
LGTM

PR #871, opened 2026-07-29
fix(tool/setup): check file.Close() error in SetupPhase.store

### LINKED ISSUE #870: fix(tool/setup): matched.json can be silently corrupted if Close() fails after write
### Describe the bug

`SetupPhase.store()` opens `matched.json` with `os.Create`, writes the marshaled JSON via `file.Write`, and relies on a bare `defer file.Close()` with no check on its return value. Closing a writable file is what actually flushes buffered data to disk - if that flush fails (disk full, quota exceeded, a network filesystem hiccup), the error is silently discarded and `store()` still returns `nil`, so the caller believes the write succeeded while the on-disk file may be corrupted or incomplete.

Affected file: `tool/internal/setup/store.go`

### Environment

### To Reproduce

Trigger a `Close()` failure on the matched-rules file during `otelc setup` (e.g. disk full, or a filesystem that rejects the final flush) after a successful `Write()`. `store()` returns `nil` regardless.

### Expected behavior

If closing the file fails after writing, `store()` should return an error rather than silently reporting success.

### Additional context

`matched.json` isn't just an internal artifact - `docs/configuration.md`'s "Verifying Your Configuration" section explicitly tells users to inspect this file directly to confirm their instrumentation rules matched as expected. A silently corrupted file here would be misleading in exactly the workflow the docs recommend.

I'd like to work on this - the fix explicitly checks the `Close()` error on the success path, and closes best-effort before returning on the earlier marshal/write error paths.

Satya900:
## Description

`SetupPhase.store()` opened `matched.json` with `os.Create` and relied on a bare `defer file.Close()`, never checking its return value, then wrote the JSON via `file.Write`. This adds an explicit `Close()` error check on the success path, and closes best-effort before returning on the earlier marshal/write error paths.

## Motivation

Closing a writable file is what actually flushes buffered data to disk. If that flush fails (disk full, quota, a network filesystem hiccup) after a successful `Write()`, the error was silently discarded and `store()` still returned `nil`, so the caller believed the write succeeded while the on-disk file could be corrupted or incomplete.

This isn't just an internal artifact: `docs/configuration.md`'s "Verifying Your Configuration" section explicitly tells users to inspect `matched.json` directly to confirm their instrumentation rules matched as expected. A silently corrupted file here is misleading in exactly the workflow the docs recommend.

Fixes #870

## Testing

Added `TestStore`, verifying a full round-trip (write, close, read back, unmarshal) succeeds and produces valid JSON. `go test ./...` and scoped `golangci-lint` pass clean for the affected package.

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/setup/store.go
+++ b/tool/internal/setup/store.go
@@ -91,17 +91,22 @@ func (sp *SetupPhase) store(ctx context.Context, matched []*rule.InstRuleSet, mo
 	if err != nil {
 		return ex.Wrapf(err, "failed to create file %s", f)
 	}
-	defer file.Close()
 
 	bs, err := json.Marshal(matched)
 	if err != nil {
+		_ = file.Close()
 		return ex.Wrapf(err, "failed to marshal rules to JSON")
 	}
 
-	_, err = file.Write(bs)
-	if err != nil {
+	if _, err = file.Write(bs); err != nil {
+		_ = file.Close()
 		return ex.Wrapf(err, "failed to write JSON to file %s", f)
 	}
+
+	if err = file.Close(); err != nil {
+		return ex.Wrapf(err, "failed to close file %s", f)
+	}
+
 	sp.Info("Stored matched sets", "path", f)
 	return nil
 }


darccio:
@Satya900 Please address the conflicts.

Satya900:
@darccio yup checking that immediately

Satya900:
@darccio all conflicts have been resolved

Satya900:
@darccio  again there are conflicts resolving it now

Satya900:
Closing in favor of [#941](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/941), merged today, which fixes the same bug ([#870](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/870)) with a more thorough approach: it switches store() to util.WriteFileAtomic instead of manual os.Create/Write/Close(). That checks the Close() error internally, and also protects against partial-write corruption if the process dies mid-write, which this PR didn't handle.

PR #873, opened 2026-07-29
fix(instrumentation/linodego): derive server.address from client's configured host

### LINKED ISSUE #872: bug(instrumentation/linodego): server.address always hardcoded to api.linode.com, ignoring SetBaseURL/UseURL
### Describe the bug

The linodego v2 instrumentation always records `server.address = api.linode.com` on every span and metric it produces, even when the instrumented application has pointed the `linodego.Client` at a different host via `SetBaseURL` or `UseURL` (e.g. a specific Linode API gateway, a regional/mirrored endpoint, or a mock server used in tests).

This happens in two places:

1. `instrumentation/github.com/linode/linodego/v2/semconv/client.go` - `LinodegoOperationTraceAttrs` and `MetricAttributes` both call `semconv.ServerAddress(DefaultServerAddress)` unconditionally, where `DefaultServerAddress = "api.linode.com"` is a hardcoded constant. Neither function takes the client's actual host as input.
2. `instrumentation/github.com/linode/linodego/v2/hook.go` - `BeforeDoRequest` builds a `semconv.LinodegoRequest{Method: method, Endpoint: endpoint}` and never sets `ServerAddress`, so `LinodegoRequestTraceAttrs` also falls back to the same hardcoded default for the per-request "doRequest" spans.

So both the public-method operation spans (`linodego.<Method>`), the per-request spans (`{METHOD} {endpoint}`), and the `linodego.client.operation.duration` metric all report `server.address = api.linode.com` regardless of what host the client is actually talking to.

### Environment

- linodego version: `github.com/linode/linodego/v2 v2.4.1`

### To Reproduce

1. Configure a client against a non-default host:
   ```go
   client := linodego.NewClient(httpClient)
   client.SetBaseURL("https://my-mock-linode-api.example.com")
   client.GetInstance(ctx, 12345)
   ```
2. Inspect the emitted spans/metrics for `GetInstance` (and the underlying `doRequest` call).
3. `server.address` is reported as `api.linode.com` on every span and on the `linodego.client.operation.duration` metric, even though the client is actually talking to `my-mock-linode-api.example.com`.

Note: `linodego.Client` has no exported getter for the configured host - `SetBaseURL`/`UseURL` only update unexported fields (`baseURL`, `hostURL`) - so the fix isn't a one-line "read the field back"; it needs a plan for how the correct host is threaded from the client into the hooks (candidates: capture it at `SetBaseURL`/`UseURL`-adjacent call sites if those are ever instrumented, derive it from the resolved request in `doRequest`, or accept a documented limitation for the operation-level span/metric while fixing the request-level span).

### Expected behavior

`server.address` should reflect the host the `linodego.Client` is actually configured to call, falling back to `api.linode.com` only when no other information is available (e.g. before the request has been resolved).

### Additional context

This instrumentation package was added recently (#710) and is still fairly new, so this hasn't been reported yet. I'd like to pick this up - happy to start with the `doRequest` per-request span (where the real endpoint is already flowing through `BeforeDoRequest`) and then decide, with maintainer input, whether the operation-level span/metric can reuse that or needs a separate approach.

bhuvan-somisetty:
## Description

The linodego v2 instrumentation always tagged `server.address` as `api.linode.com` on both the public-method operation spans (`linodego.<Method>`) and per-request spans (`{METHOD} {endpoint}`), and on the `linodego.client.operation.duration` metric - even when the instrumented app pointed the `linodego.Client` at a different host via `SetBaseURL`/`UseURL`.

This PR:

- Adds `hostFromClient` in `hook.go`, which reads the API host the client is actually configured to call. `linodego.Client` has no exported getter for this (`SetBaseURL`/`UseURL` only update unexported `baseURL`/`hostURL` fields), so it's read via reflection off the already-available `*linodego.Client` receiver, falling back to `""` (and therefore the previous default) if the field is ever renamed or missing.
- Threads the resolved host into `BeforeDoRequest` (per-request spans) and `beforeAPICall`/`afterAPICall` (operation spans + the duration metric).
- Updates `semconv.LinodegoOperationTraceAttrs` and `semconv.MetricAttributes`/`RecordOperationDuration` to accept a `serverAddress` parameter instead of hardcoding `DefaultServerAddress`.
- Adds test coverage for a client configured with a non-default host, on both the request-level span and the operation-level span/metric, plus direct coverage of `hostFromClient`.

## Motivation

Fixes #872

Anyone using a custom Linode API endpoint (private API gateway, a specific region, or a mock server in tests) currently gets telemetry that reports the wrong `server.address`, which is misleading for anyone using traces/metrics to understand where requests actually went.

---

## Checklist

- [x] Tests pass: `make test`


darccio:
@bhuvan-somisetty CI is failing significantly in this PR. Please check it.

bhuvan-somisetty:
@darccio fixed. Turns out the archive drift was from my local Windows checkout (CRLF line endings + a mismatched Go toolchain vs. the CI-pinned one) producing a different reproducible bundle than Linux CI rebuilt it properly. Also found the actual server.address fix had exposed a stale assertion in the linodego integration test (it was hardcoded to "api.linode.com" instead of the mock server's real host), so I updated that too. All green now.

instrumentation/github.com/linode/linodego/v2/hook.go
```diff
@@ -66,6 +69,38 @@ func (linodegoEnabler) Enable() bool {
...
+	if !f.IsValid() || f.Kind() != reflect.String {
+		return ""
+	}
+
+	parsed, err := url.Parse(f.String())
+	if err != nil || parsed.Host == "" {
+		return ""
+	}
+	return parsed.Host
+}
```
amazingakai: This runs reflect on every single request. Since otelc supports file rules, we should probably inject a file into `linodego` to read `c.hostURL` directly to avoid the hot path overhead.   
                                                                                                                                                                                             
 To keep unit tests working (since otelc rules don't run during `go test`), we could put this reflect logic in a package-level func variable as a fallback. Then drop a `//go:build ignore`  
 file in this package that gets injected via `add_file`, and have its `init()` use `go:linkname` to swap the func pointer to just return `c.hostURL` directly.                               
                                                                                                                                                                                             
Can you try exploring that direction? It would be great to drop the reflect overhead here if possible.

amazingakai (changes requested):

instrumentation/github.com/linode/linodego/v2/hook.go
```diff
@@ -44,19 +47,31 @@ const (
...
+	tracer     trace.Tracer
+	metrics    semconv.Metrics
+	initOnce   sync.Once
+	getHostURL = defaultHostURLGetter
 )
 
+// setHostURLGetter is called via go:linkname by host_extractor.go (injected into package linodego)
+// to replace the reflection fallback with direct access to Client.hostURL.
+//
+//go:linkname setHostURLGetter go.opentelemetry.io/otelc/instrumentation/github.com/linode/linodego/v2.setHostURLGetter
```
amazingakai: This isn't needed?

amazingakai (changes requested):

bhuvan-somisetty:
Hi @amazingakai,

Good catch  you're completely right. The `//go:linkname` directive and `unsafe` import in `hook.go` were redundant since `setHostURLGetter` is defined directly in this package and linked from `host_extractor.go`.

I've removed them, rebased onto `main`, and regenerated the bundle. All tests and linters are passing green.

Thanks for pointing that out!

PR #874, opened 2026-07-29
fix(setup): warn when version-gated rules skip due to unresolved dep …

### LINKED ISSUE #777: bug(setup): version-gated rules silently skipped when dependency is built from local/replace path
### Describe the bug

Version-gated instrumentation rules are **silently skipped** when a dependency is compiled from a **local checkout** or via a **`replace` directive** in `go.mod`.

During setup, `otelc` derives each dependency's version by regex-parsing the compile source path for a module-cache segment (`...@vX.Y.Z/...`). Local and replaced paths do not contain that pattern, so `Dependency.Version` is left empty.

When a rule has a non-empty `version:` field, `matchVersion` calls `VersionInRange("", "<range>")`, which returns **false**. The rule is filtered out in both the **setup** phase (`match.go`) and the **pin** phase (`pin.go`) with **no error or warning**.

This is especially painful when developing instrumentation locally with `replace`, or when troubleshooting version-gated rules - the docs say to check `go list -m`, but the tool never resolves the version that way.

---

### Environment

- OS: Linux
- Go Version: 1.25.0
- Version: `otelc` main branch / v1.0.1

---

### To Reproduce

**Minimal path-based repro (no full build required):**

`findModVersion` only works for module-cache paths:

| Source path | `findModVersion()` result |
|---|---|
| `/go/pkg/mod/github.com/foo/bar@v1.2.3/pkg/foo.go` | `v1.2.3` |
| `/home/user/projects/bar/pkg/foo.go` | `""` (empty) |
| `/tmp/myapp/vendor/github.com/foo/bar/pkg/foo.go` | `""` (empty) |

Relevant code:

```go
// tool/internal/setup/find.go
var versionRegexp = regexp.MustCompile(`@v\d+\.\d+\.\d+(-.*?)?/`)

func findModVersion(path string) string {
    version := versionRegexp.FindString(filepath.ToSlash(path))
    if version == "" {
        return ""
    }
    return version[1 : len(version)-1]
}
```

Empty version causes version-gated rules to never match:

```go
// tool/util/shared.go
VersionInRange("", "v1.0.0")        // → false
VersionInRange("", "v1.0.0,v2.0.0") // → false
```

**End-to-end repro sketch:**

1. Create a module that depends on a version-gated library and uses `replace`:

```go
// go.mod
module example.com/app

go 1.25.0

require github.com/redis/go-redis/v9 v9.0.0

replace github.com/redis/go-redis/v9 => ./local-redis
```

2. Ensure `./local-redis` is a local checkout of the library (no `@vX.Y.Z/` in compile paths).

3. Build with `otelc` against a rule that targets `github.com/redis/go-redis/v9` with a non-empty `version:` range (as in `instrumentation/github.com/redis/go-redis/v9/otelc.yaml`).

4. Observe: instrumentation for that dependency is **not applied**, and there is **no log/error** explaining that the dependency version could not be resolved.

5. Compare: remove the `replace` directive (or build against the module cache) - the same rule matches when the compile path contains `@v9.x.x/`.

---

### Expected behavior

- `Dependency.Version` should reflect the module version from the build's module graph (e.g. as reported by `go list -m <module-path>`), not only from compile command file paths.
- When a version cannot be resolved, `otelc` should at minimum **warn** (or fail loudly in debug mode) rather than silently dropping version-gated rules.
- Behavior under `replace` directives should be explicitly defined (e.g. use the `require` version from `go.mod`, or document that purely local replacements have no semver).

---

### Additional context

**Affected code paths:**
- `tool/internal/setup/find.go` : `findModVersion`, `findGoSources` (sets `dep.Version`)
- `tool/internal/setup/match.go` :  `matchVersion` filters rules during setup
- `tool/internal/setup/pin.go` :  same `VersionInRange` check when deciding which instrumentation packages to pin

Would be working on this issue once the approach is finalized.

Harsh63870:
## Description

When a dependency is compiled from a local checkout, `replace` path, or vendor tree, `findModVersion` cannot extract a semver from the compile source path (there is no `@vX.Y.Z/` module-cache segment). `Dependency.Version` stays empty.

If a matching instrumentation rule also has a non-empty `version:` / version range, both:

- setup matching (`tool/internal/setup/match.go`)
- pin import selection (`tool/internal/setup/pin.go` → `matchInstrumentationImports`)

call `VersionInRange("", "<range>")`, which returns `false`, and the rule is skipped with a bare `continue`.

This PR does **not** change that matching behavior. It only makes the skip visible:

- setup: `SetupPhase.Warn(...)` when a version-gated rule is skipped because the dependency version is unresolved
- pin: `logger.WarnContext(...)` via a warn callback when the same condition drops an instrumentation import
- shared helper `unresolvedVersionSkip(depVersion, ruleVersion)` so both sites use the same condition

Tests added/updated:

- `TestFindModVersion` (module-cache, pre-release, Windows path, local, vendor, empty)
- `TestUnresolvedVersionSkip`
- `TestMatchInstrumentationImports` cases for empty dep version with/without a rule version range
- `TestMatchInstrumentationImports_WarnsOnUnresolvedVersion`
- `TestVersionInRange` empty-version cases in `tool/util`

## Motivation

Today a developer using `replace` (or a local path) against a version-gated instrumentation gets **no instrumentation and no explanation**. That matches the bug report in #777 and makes local instrumentation development hard to debug.

Full version resolution from the module graph (`go list -m`, replace/workspace semantics) still needs maintainer design agreement. This PR is the small, uncontroversial first step discussed on the issue: **make the silent skip loud**, without preempting how versions should be resolved later.

Fixes #777 (diagnostics / visibility portion; version-resolution follow-up remains open)

---

## Checklist

- [ ] Tests pass: `make test`


Harsh63870:
Hi @kakkoyun, friendly ping when you have a moment.
This is the small #777 follow-up that only warns when a version-gated rule is skipped because Dependency.Version is empty (replace/local/vendor). Matching behavior is unchanged; full go list -m resolution is left for a later design decision.
Required checks are still waiting on maintainer workflow approval. Happy to adjust based on review. Thanks!

tool/internal/setup/match.go
darccio, line 202: This warns once per skipped rule per dependency. A `replace`d/local dep with several version-gated rules targeting it will emit one warning per rule, all saying the same actionable thing ("dependency version is unresolved"). Was firing per-rule (identifies exactly which instrumentation was dropped) a deliberate choice over firing once per dependency (less noisy for the common multi-rule case)?

darccio, line 202: `runMatch` takes `ctx context.Context` but the warning here uses `sp.Warn` (no context), while the sibling warning added by this same PR in `matchInstrumentationImports`'s caller uses `logger.WarnContext(ctx, ...)`. Consistent with `SetupPhase`'s existing helpers (which are all non-context), but inconsistent with the other half of this PR — worth aligning, or is the difference intentional?

tool/internal/setup/match.go
```diff
@@ -190,6 +197,13 @@ func (sp *SetupPhase) runMatch(
 	filteredRules := make([]rule.InstRule, 0, len(relevantRules))
 	for _, r := range relevantRules {
 		if !matchVersion(dep, r) {
+			if unresolvedVersionSkip(dep.Version, r.GetVersion()) {
+				sp.Warn("skipping version-gated rule because dependency version is unresolved",
+					"rule", r.GetName(),
+					"dep", dep.ImportPath,
+					"version_range", r.GetVersion(),
+				)
+			}
```
darccio, line 198: This warning has no dedicated test. The pin-path equivalent gets `TestMatchInstrumentationImports_WarnsOnUnresolvedVersion`, and this package already has the machinery for it — `logger_delegate_test.go` exercises `sp.Warn`, and `TestMatchDeps_NoMatchesWarning` tests a sibling warning the same way. Worth adding an equivalent test here so both skip sites added by this PR are covered.

tool/internal/setup/pin_test.go
```diff
@@ -554,12 +588,37 @@ func TestMatchInstrumentationImports(t *testing.T) {
...
+	got := matchInstrumentationImports(deps, rules, func(msg string, args ...any) {
+		warned = true
+		warnedMsg = msg
+		_ = args
+	})
+
+	require.Empty(t, got)
+	require.True(t, warned)
+	require.Contains(t, warnedMsg, "unresolved")
+}
```
darccio, line 689: This test only covers a single rule under one modPath, so it can't catch a real false positive: when a modPath has multiple rules (e.g. targeting different subpackages or version ranges), one rule's unresolved-version skip fires this warning immediately, even if another rule for the same modPath later sets `imports[modPath] = true`. Concretely — dep `example.com/foo/v1` with empty version matched by a version-gated rule triggers the warning, while dep `example.com/foo/v1/sub` with a resolved version matched by a second, unversioned rule under the same modPath still adds it to `imports`. Net result: `imports["example.com/instrumentation/foo"] == true` and `warned == true` — the log claims the instrumentation was skipped when it was actually imported.

Worth extending this test (or adding a sibling case) with two deps/rules sharing one modPath, one resolving and one not, asserting that no warning fires once the modPath ends up in the returned `imports` map.

tool/internal/setup/find_test.go
```diff
@@ -412,3 +412,56 @@ echo nothing useful
...
+		{
+			name: "module cache path with pre-release",
+			path: "/go/pkg/mod/github.com/foo/bar@v1.2.3-rc.1/pkg/foo.go",
+			want: "v1.2.3-rc.1",
+		},
+		{
+			name: "windows-style module cache path",
+			path: `C:\go\pkg\mod\github.com\foo\bar@v9.0.0\client.go`,
+			want: "v9.0.0",
+		},
```
darccio, line 460: This case could fail on non-Windows: `filepath.ToSlash` is a no-op when the host separator is already `/`, so on Linux/macOS the backslashes in this literal path survive untouched, `versionRegexp` (which requires a trailing `/` after the version) never matches, and `findModVersion` returns `""` instead of `"v9.0.0"`. I confirmed this by running the test and by checking `filepath.ToSlash` directly on darwin — it returns the input unchanged.

```suggestion
		{
			name: "windows-style module cache path",
			path: strings.ReplaceAll(`C:\go\pkg\mod\github.com\foo\bar@v9.0.0\client.go`, `\`, "/"),
			want: "v9.0.0",
		},
```

This exercises the same regexp-matching behavior `findModVersion` needs (a path already in `/`-separated form, as it would be after a real Windows host's `filepath.ToSlash` runs), without depending on the test-runner's GOOS. Needs `strings` added to imports in this file if not already present.

tool/internal/setup/pin.go
```diff
@@ -306,7 +306,11 @@ func ensureOtelcRequire(moduleDir, version string) (bool, error) {
 	return true, nil
 }
 
-func matchInstrumentationImports(deps []*Dependency, ruleset map[string][]yamlRule) map[string]bool {
+func matchInstrumentationImports(
+	deps []*Dependency,
+	ruleset map[string][]yamlRule,
+	warn func(msg string, args ...any),
+) map[string]bool {
```
darccio, line 313: The `warn func(msg string, args ...any)` callback here vs. `sp.Warn` in `runMatch` (match.go) isn't just a style difference — `matchInstrumentationImports` is a free function with no logger, while `runMatch` is a `*SetupPhase` method that already has one, so each shape is the natural fit locally. But given the finding above (the two call sites can react to the same underlying event and disagree), it might be worth treating this as one warning strategy owned by one site rather than two call shapes independently reporting the same condition.

tool/internal/setup/match.go
darccio, line 202: Only the boolean condition (`unresolvedVersionSkip`) is shared between this call site and its mirror in `pin.go`'s `matchInstrumentationImports` — the log message text and key names are hand-duplicated at each. A shared helper (e.g. `warnUnresolvedVersionSkip(warn func(string, ...any), dep, rule, modPath)`) would keep the two messages in sync and give a single place to land the pin/setup interaction fix above.

tool/internal/setup/match.go
```diff
@@ -112,6 +112,13 @@ func matchVersion(dependency *Dependency, rule rule.InstRule) bool {
 	return util.VersionInRange(dependency.Version, rule.GetVersion())
 }
 
+// unresolvedVersionSkip reports whether a version-gated rule is being skipped
+// because the dependency version could not be resolved from its compile path
+// (common for replace/local/vendor sources that lack an @vX.Y.Z/ segment).
+func unresolvedVersionSkip(depVersion, ruleVersion string) bool {
+	return depVersion == "" && ruleVersion != ""
+}
```
darccio: Two smaller points on this helper:

Placement — it's defined in `match.go` but consumed by both this file and `pin.go`'s `matchInstrumentationImports`. Not a build issue (same package), just a discoverability one: a reviewer skimming `pin.go` wouldn't expect this predicate's definition to live in a file named after the other phase.

Implicit precondition — this is only correct to call from a branch where the version check has already failed. It doesn't itself verify that the rule would otherwise have matched; nothing stops a future caller from invoking it standalone and misreading the result. Worth documenting the precondition explicitly:

```suggestion
// unresolvedVersionSkip reports whether a version-gated rule is being skipped
// because the dependency version could not be resolved from its compile path
// (common for replace/local/vendor sources that lack an @vX.Y.Z/ segment).
//
// Precondition: call only after the rule has already failed its version check
// (matchVersion / util.VersionInRange returned false) — this alone does not
// verify that the rule would otherwise have matched.
func unresolvedVersionSkip(depVersion, ruleVersion string) bool {
	return depVersion == "" && ruleVersion != ""
}
```

darccio (changes requested):

tool/internal/setup/match.go
Harsh63870: Kept per-rule on purpose for the setup path. In runMatch each version-gated rule is dropped independently, so naming the rule is the actionable signal. A single per-dep warning would hide which instrumentations were skipped. Pin is different, see the deferred-warning change below.

Harsh63870: Intentional, SetupPhase already owns non-context helpers (Info/Warn/Error/Debug), so setup stays on sp.Warn. Pin keeps WarnContext via the callback because matchInstrumentationImports has no logger of its own.

tool/internal/setup/match.go
Harsh63870: Added TestRunMatch_WarnsOnUnresolvedVersion, captures sp.Warn via a slog TextHandler and asserts the unresolved-version message for version-gated rules with an empty Dependency.Version.

tool/internal/setup/pin_test.go
Harsh63870: good catch, fixed. Pin now records unresolved skips during matching and only warns afterward if that modPath is still absent from the returned imports map, added a regression case: one unresolved dep + one matched unversioned rule under the same instrumentation module asserts import happens and no warning fires.

tool/internal/setup/find_test.go
Harsh63870: updated to a /-separated path (C:/go/pkg/mod/...@v9.0.0/...) so it exercises the same shape filepath.ToSlash produces on windows without depending on GOOS.

tool/internal/setup/pin.go
Harsh63870: aligned with the pin false-positive fix: pin only reports when the instrumentation was actually skipped,  setup still reports per dropped rule. Shared message/keys live in warnUnresolvedVersionSkip so both sites stay in sync.

tool/internal/setup/match.go
Harsh63870: done, extracted warnUnresolvedVersionSkip (plus unresolvedVersionSkip) into version.go, both call sites use it; setup passes "rule", pin passes "instrumentation".

tool/internal/setup/match.go
Harsh63870: moved unresolvedVersionSkip to version.go (shared by match and pin) and documented the precondition: only call after the version check has already failed.

Harsh63870:
thanks @darccio, all review comments are addressed in the latest pushes, matching behavior is unchanged; this stays the diagnostics-only slice of #777, happy to tweak anything that still looks off.

Harsh63870:
@darccio, @amazingakai , @txabman42 , TestGRPCServer/telemetry_flush_on_SIGTERM timing out on span flush, unrelated to this PR's diff, only touches setup/pin version-gating. SIGINT sibling case passes fine, same flush path. Reran and got, looks like a flaky test on this runner, not something my change could cause.

Harsh63870:
@darccio, all CI checks are now passing, including the required checks, the review comments have also been addressed in the latest changes, thanks!

darccio (approved):
LGTM

PR #878, opened 2026-07-29
fix: certain import patterns were not properly identified

hannahkm:
## Description

Enables `*dst.SelectorExpr` to match not only on single-segment import paths (common examples being `io`, `os`,  `context`, etc) but also to match multi-segment paths with their common names/alias (e.g. `net/http` with `http`, `althttp net/http` with `althttp`, etc). 

Changes are mostly in `typename.go` -- other files are tests or changes to function calls due to a new parameter.

## Motivation

Allows matching on more packages.

Fixes #

---

## Checklist

- [ ] Tests pass: `make test`

[^I-know-my-stuff]:


tool/internal/ast/typename.go
```diff
@@ -85,9 +102,51 @@ func fieldListContainsType(fields *dst.FieldList, typeStr string) (bool, error)
...
+
+// importPathTail returns the last path segment of an import path, which is
+// conventionally (but not necessarily) that package's default identifier,
+// e.g. "net/http" -> "http".
+func importPathTail(path string) string {
+	if i := strings.LastIndexByte(path, '/'); i >= 0 {
+		return path[i+1:]
+	}
+	return path
+}
```
darccio: `importPathTail` treats the last path element as the package's local name, but for a
module with a major-version suffix that element is the version, not the name. Two
consequences, both verified against this branch:

1. **Versioned modules never match.** For `import "github.com/redis/go-redis/v9"`,
   `importAliasMap` returns `{"v9": "github.com/redis/go-redis/v9"}`. A filter of
   `*github.com/redis/go-redis/v9.Client` against `func handle(c *redis.Client)`
   looks up `imports["redis"]`, misses, then falls back to comparing `"v9" == "redis"` —
   so `FindFuncDecl` returns `false`. Same silent-non-match failure mode this PR fixes
   for `net/http`, still open for any `/vN` module.

2. **Distinct imports collide.** This is valid Go (real names `foo` and `bar`):

   ```go
   import (
       "github.com/a/foo/v2"
       "github.com/b/bar/v2"
   )
   ```

   Both key as `v2`, so one silently overwrites the other and the map ends up with a
   single entry — neither real name is present. The tail heuristic manufactures a
   collision the language guarantees cannot exist.

Stripping the version element fixes both with no added I/O (the collision dissolves
because the keys become `foo` and `bar`). It cannot fix the case where a package simply
declares a name unrelated to its path (`go-redis` → `redis`), so the doc comment now
states that limitation explicitly rather than leaving it implied.

```suggestion
// importPathTail returns the local identifier conventionally used to reference
// an import path: its last segment, ignoring a Go module major-version suffix
// ("/v2".."/vN", or gopkg.in's ".vN"), which is part of the module path but not
// of the package name, e.g. "net/http" -> "http", "github.com/x/jwt/v5" -> "jwt".
//
// This is a convention, not a guarantee: a package may declare a name unrelated
// to its path (e.g. "github.com/redis/go-redis/v9" declares "redis"). Such
// packages are matched only when the importing file aliases them explicitly.
func importPathTail(path string) string {
	if i := strings.LastIndexByte(path, '/'); i >= 0 && isMajorVersion(path[i+1:]) {
		path = path[:i]
	}
	if i := strings.LastIndexByte(path, '/'); i >= 0 {
		path = path[i+1:]
	}
	// gopkg.in style: "yaml.v3" -> "yaml".
	if i := strings.LastIndexByte(path, '.'); i >= 0 && isMajorVersion(path[i+1:]) {
		path = path[:i]
	}
	return path
}

// isMajorVersion reports whether s is a module major-version element ("v2", "v11").
func isMajorVersion(s string) bool {
	if len(s) < 2 || s[0] != 'v' {
		return false
	}
	for _, c := range []byte(s[1:]) {
		if c < '0' || c > '9' {
			return false
		}
	}
	return true
}
```

tool/internal/ast/typename.go
```diff
@@ -37,30 +39,44 @@ func parseTypeName(s string) (parsedTypeName, error) {
...
+		if ident.Path != "" {
+			// Populated by a resolving decorator; already the real import path.
+			return t.importPath == ident.Path && t.name == n.Sel.Name
+		}
+		if resolved, importOk := imports[ident.Name]; importOk {
+			return t.importPath == resolved && t.name == n.Sel.Name
+		}
+		// No import context resolved this qualifier (nil/incomplete imports
+		// map): fall back to comparing against importPath's last segment
+		return importPathTail(t.importPath) == ident.Name && t.name == n.Sel.Name
```
darccio: Worth tightening this comment, because it currently reads as a resilience net and
invites a wrong assumption. With a non-nil `imports` map the fallback can never be
what produces a match:

- If the file imports this path unaliased, `importPathTail(path)` is already a key, so
  the lookup above **hits** and we never reach here.
- If the lookup misses, the key was derived wrongly — and this line compares against
  that same wrong tail, so it fails identically.

Since a successful tail comparison (`ident.Name == importPathTail(t.importPath)`) also
implies `ident.Name` is a map key, the map would always have hit first. So the only
case this branch actually serves is `imports == nil` (hand-built test nodes) — it is
*not* a safety net for imports the map keyed incorrectly, which is what "incomplete"
suggests.

```suggestion
		// No import context at all (imports == nil, e.g. hand-built AST nodes in
		// tests with no backing *dst.File): compare against importPath's last
		// segment. Note this cannot rescue a mis-keyed map — a tail match here
		// would imply ident.Name is a key, so the lookup above would have hit.
		return importPathTail(t.importPath) == ident.Name && t.name == n.Sel.Name
```

tool/internal/ast/typename.go
```diff
@@ -85,9 +102,51 @@ func fieldListContainsType(fields *dst.FieldList, typeStr string) (bool, error)
...
+func importAliasMap(file *dst.File) map[string]string {
+	if file == nil {
+		return nil
+	}
+	aliases := make(map[string]string, len(file.Imports))
+	for _, imp := range file.Imports {
+		path, err := strconv.Unquote(imp.Path.Value)
+		if err != nil {
+			continue
+		}
```
darccio, line 137: `imp.Path` is dereferenced without a nil check. The equivalent loop in
`tool/internal/imports/parseFile` guards it (`if !isImport || importSpec.Path == nil`),
so this is inconsistent with the existing precedent even if `file.Imports` entries are
always non-nil in practice.

```suggestion
	for _, imp := range file.Imports {
		if imp.Path == nil {
			continue
		}
		path, err := strconv.Unquote(imp.Path.Value)
		if err != nil {
			continue
		}
```

tool/internal/ast/typename.go
```diff
@@ -85,9 +102,51 @@ func fieldListContainsType(fields *dst.FieldList, typeStr string) (bool, error)
...
 }
+
+// importAliasMap builds a map from the local identifier used to reference an
+// imported package within file (its explicit alias, or its default package
+// name when unaliased) to that package's real import path. It correctly disambiguates:
+//   - aliased imports (e.g. `import althttp "net/http"`)
+//   - distinct import paths that happen to share a last path segment (e.g.
+//     "text/template" vs "html/template", both conventionally "template")
+//
+// Returns nil when file is nil.
```
darccio, line 124: Worth recording *why* this duplicates `tool/internal/imports.parseFile`, which already
builds an alias→path map. I checked and `ast` → `internal/imports` would be acyclic, so
nothing structural prevents reuse — which means the next reader may well try to unify
them. The reason not to: `parseFile` resolves unaliased imports via
`pkgload.ResolvePackageName`, a `go/packages` load per import per file that `ex.Fatalf`s
on failure. That's fine in the instrument phase but not on the setup/match path, which
runs across every compiled package in the build.

```suggestion
// importAliasMap builds a map from the local identifier used to reference an
// imported package within file (its explicit alias, or its default package
// name when unaliased) to that package's real import path. It correctly disambiguates:
//   - aliased imports (e.g. `import althttp "net/http"`)
//   - distinct import paths that happen to share a last path segment (e.g.
//     "text/template" vs "html/template", both conventionally "template")
//
// This deliberately duplicates tool/internal/imports.parseFile rather than reusing
// it: that resolves unaliased imports with pkgload.ResolvePackageName (a
// go/packages load that ex.Fatalf's on failure), which is too costly and too fatal
// for the setup/match path, where this runs for every compiled package in the build.
// The cost is that the default name here is a syntactic guess; see importPathTail.
//
// Returns nil when file is nil.
```

tool/internal/ast/shared.go
```diff
@@ -156,7 +156,7 @@ func FindFuncDecl[R rule.InstFuncRule | rule.InstRawRule | rule.FilterDef](
 	if !ok {
 		return nil, false, ex.Newf("unexpected %T value", r)
 	}
-	ok, err := funcDeclMatchesFilters(funcDecl, rr)
+	ok, err := funcDeclMatchesFilters(funcDecl, rr, importAliasMap(root))
```
darccio: `importAliasMap(root)` is evaluated as an argument, so it runs before
`funcDeclMatchesFilters` checks whether any type filter is set — and that function's own
doc says "Returns true when no sub-filters are set", which is exactly when the map is
pure waste. Rules with only `func:`/`recv:` pay a map allocation plus a walk of every
import for nothing.

It's already guarded by the cheap name+recv match so the cost is small; flagging it
because it's cheap to avoid, not because it's hot. Passing `root` and building the map
lazily inside `funcDeclMatchesFilters` (only once a filter needs it) would also drop a
parameter from the five helpers below.

tool/internal/ast/shared.go
```diff
@@ -328,11 +328,11 @@ func AddStructField(decl dst.Decl, name, t string) {
 // Matching uses structural comparison of dst.Expr nodes (no type checker).
 // For the scalar-type filters this means an exact type-name match rather than
 // full interface-satisfaction checking.
```
darccio, line 346: This paragraph describes the matching semantics but not the change this PR makes to
them: qualifier resolution now depends on the enclosing file's import declarations, so
the same `funcDecl` can match or not depending on which file it came from. The new
parameter is documented well on `matches`, but this is the doc a rule author reading the
filter contract lands on first.

```suggestion
// Matching uses structural comparison of dst.Expr nodes (no type checker).
// For the scalar-type filters this means an exact type-name match rather than
// full interface-satisfaction checking.
//
// Qualified type names are resolved against imports, which maps the local
// identifier used at a use site to its real import path (see importAliasMap).
// Matching is therefore relative to the enclosing file's import declarations.
```

tool/internal/ast/typename_test.go
```diff
@@ -114,20 +114,157 @@ func TestTypeNameMatches(t *testing.T) {
...
+
+func f(r *http.Request, r2 *althttp.Request) {}
+`)
+		require.NoError(t, err)
+
+		imports := importAliasMap(file)
+		assert.Equal(t, "net/http", imports["http"])
+		assert.Equal(t, "net/http", imports["althttp"])
+		assert.NotContains(t, imports, "_")
+		assert.NotContains(t, imports, ".")
```
darccio, line 286: Since the point of this case is that blank and dot imports are *excluded*, asserting the
exact size would catch an unexpected extra entry (e.g. a stray tail-derived key) that
`NotContains` on two specific keys lets through.

```suggestion
		imports := importAliasMap(file)
		assert.Len(t, imports, 2)
		assert.Equal(t, "net/http", imports["http"])
		assert.Equal(t, "net/http", imports["althttp"])
		assert.NotContains(t, imports, "_")
		assert.NotContains(t, imports, ".")
```

tool/internal/ast/typename_test.go
```diff
@@ -114,20 +114,157 @@ func TestTypeNameMatches(t *testing.T) {
...
+		assert.True(t, textTemplate.matches(node, importsText))
+		assert.False(t, htmlTemplate.matches(node, importsText))
+
+		importsHTML := map[string]string{"template": "html/template"}
+		assert.False(t, textTemplate.matches(node, importsHTML))
+		assert.True(t, htmlTemplate.matches(node, importsHTML))
+	})
+}
+
+func TestImportAliasMap(t *testing.T) {
```
darccio, line 261: Please add coverage for the version-suffix cases alongside the fix to
`importPathTail` — both the resolution and the collision, since they're the two
distinct symptoms:

```go
	t.Run("module version suffix is not the package name", func(t *testing.T) {
		p := NewAstParser()
		file, err := p.ParseSource(`package main

import (
	"github.com/golang-jwt/jwt/v5"
	"gopkg.in/yaml.v3"
)

func f(t *jwt.Token, n *yaml.Node) {}
`)
		require.NoError(t, err)

		imports := importAliasMap(file)
		assert.Equal(t, "github.com/golang-jwt/jwt/v5", imports["jwt"])
		assert.Equal(t, "gopkg.in/yaml.v3", imports["yaml"])
	})

	t.Run("distinct modules sharing a version suffix do not collide", func(t *testing.T) {
		p := NewAstParser()
		file, err := p.ParseSource(`package main

import (
	"github.com/a/foo/v2"
	"github.com/b/bar/v2"
)

func f(x *foo.T, y *bar.T) {}
`)
		require.NoError(t, err)

		imports := importAliasMap(file)
		assert.Len(t, imports, 2)
		assert.Equal(t, "github.com/a/foo/v2", imports["foo"])
		assert.Equal(t, "github.com/b/bar/v2", imports["bar"])
	})
```

A case pinning the *unsupported* mode would also be valuable, so the limitation is
visible rather than folklore — `github.com/redis/go-redis/v9` resolves to `go-redis`,
not `redis`, and matches only when aliased explicitly.

darccio (changes requested):
Nice fix — the diagnosis is exactly right, and the 3-tier resolution in the
`*dst.SelectorExpr` branch is well-ordered: a decorator-populated `ident.Path` wins over
the import map, which wins over the syntactic fallback. Restoring the `ident.Path` case
is a real bonus; the old code returned `false` outright whenever it was set.

I verified the threading is complete and closed: all six threaded helpers pass `imports`
through, `FindFuncDecl` is the only production entry point, and the fix reaches both the
setup/match and instrument phases. I also confirmed `rule.FilterDef` has no type-shaped
predicates, so there's no second matcher left carrying the old bug. `go vet` and
`go test ./tool/internal/ast/` are clean on this branch.

Two things I'd single out as worth preserving. `TestTypeNameMatches_PreResolvedIdentPath`
seeds a deliberately *conflicting* imports map to prove precedence instead of just
asserting a happy path, and the `text/template` vs `html/template` case asserts both
directions — that's the right instinct for a disambiguation fix. The `mustMatch` helper
comment explaining why it passes `nil` is also good; a bare `nil` there would read as an
oversight later.

The one substantive gap is the module version suffix (comment on `importPathTail`).
Everything there is a false negative, never a wrong match, so there's no
mis-instrumentation risk — but it's the same silent-non-match failure mode this PR is
fixing, just still open for `/vN` modules, and rule authors get no error when it bites.

Minor and non-blocking: the `imports` parameter would read better as `aliasToPath` or
`importAliases` across the six signatures — it names the map's semantics, which matters
because the direction (local name → path, not the reverse) isn't obvious at the call
sites.

darccio (approved):
LGTM. Conflict on test cases, so it's fine. @hannahkm, please fix whenever you are able. Thanks!

kakkoyun (approved):

PR #880, opened 2026-07-30
fix(instrumentation/linodego): record error.type on operation.duration when no status code is available

### LINKED ISSUE #879: bug(instrumentation/linodego): operation.duration metric can't distinguish non-HTTP failures from successes
### Describe the bug

The `linodego.client.operation.duration` histogram cannot distinguish a successful public API call from one that failed with a non-HTTP error (timeout, connection refused, DNS failure, context cancellation, etc.).

`afterAPICall` in `instrumentation/github.com/linode/linodego/v2/public_method.go` computes the status code via `finishSpanWithError`:

```go
statusCode := finishSpanWithError(span, err)
if !start.IsZero() {
    metrics.RecordOperationDuration(ctx, time.Since(start).Seconds(), operation, statusCode, serverAddress)
}
```

`finishSpanWithError` (`instrumentation/github.com/linode/linodego/v2/hook.go`) returns `0` in two different situations:

1. `err == nil` (the call actually succeeded), and
2. `err != nil` but `semconv.StatusCodeFromError(err)` can't extract an HTTP status code from it (e.g. the client never got a response at all).

`semconv.MetricAttributes` (`instrumentation/github.com/linode/linodego/v2/semconv/client.go`) only attaches `http.response.status_code` when `statusCode > 0`, and never records any error indicator otherwise:

```go
if statusCode > 0 {
    attrs = append(attrs, semconv.HTTPResponseStatusCode(statusCode))
}
```

So a request that times out or fails before a response is received is recorded in the histogram with exactly the same attribute set as a successful call. This existing test documents the ambiguity rather than a real "success" case — `ListRegions` here stands in for any call with `statusCode == 0`, which includes failures:

```go
// instrumentation/github.com/linode/linodego/v2/semconv/client_test.go
// Success path without a status code omits the status attribute.
m = attrsToMap(MetricAttributes("ListRegions", 0, ""))
```

The span itself is fine — `span.RecordError` / `span.SetStatus(codes.Error, ...)` still run in `finishSpanWithError` for the no-status-code branch — only the metric loses the failure signal.

### Environment

- Go Version: 1.24+
- Package: `go.opentelemetry.io/otelc/instrumentation/github.com/linode/linodego/v2`

### To Reproduce

1. Point a `linodego.Client` at an unreachable/non-responsive host (or otherwise force a network-level error, e.g. via a proxy that resets the connection) and call any instrumented method, such as `client.ListRegions(ctx)`.
2. Also call the same method successfully against a real endpoint.
3. Inspect the collected `linodego.client.operation.duration` metric for both calls.
4. Both data points carry the same attribute set (`server.address`, `code.function.name`) — there is no way to filter/alert on the failed one using this metric alone.

### Expected behavior

Per the HTTP metrics semantic conventions, a duration metric should carry an `error.type` attribute whenever the operation did not succeed, even when no HTTP status code is available (e.g. `error.type` set to a low-cardinality value derived from the error, such as `"timeout"`/`"other"`, or at minimum the Go error type name). This lets the failed call be distinguished from a successful one in the histogram without relying on trace data.

### Additional context

This only affects the `linodego.client.operation.duration` metric; the per-call span already records the error correctly via `RecordError`/`SetStatus`. The fix likely belongs in `finishSpanWithError` (to also signal "failed, no status code" back to the caller) and `semconv.MetricAttributes`/`RecordOperationDuration` (to accept and attach that signal as `error.type`).

bhuvan-somisetty:
## Description

`finishSpanWithError` in `instrumentation/github.com/linode/linodego/v2/hook.go` now returns a second value, `errorType`, alongside the existing `statusCode`:

- when the call succeeded, `errorType` is `""`
- when an HTTP status code could be extracted from the error, `errorType` is that code as a string (unchanged span behavior, now also surfaced to the metric)
- when no status code is available (timeout, connection refused, DNS failure, context cancellation, etc.), `errorType` is the Go type name of the error, using the same convention as `instrumentation/net/http/semconv`'s `HTTPClient.ErrorType`

`semconv.MetricAttributes` and `semconv.Metrics.RecordOperationDuration` take this value and attach it as `error.type` on the `linodego.client.operation.duration` histogram whenever it's non-empty. `public_method.go`'s `afterAPICall` was updated to pass the new return value through.

I also added a new `semconv.ErrorType(err error)` helper (mirrors the net/http one) instead of duplicating the reflection logic, and covered it plus the new code paths with unit tests in `hook_test.go` and `semconv/client_test.go`.

## Motivation

Before this change, `finishSpanWithError` returned `0` both when a call succeeded and when it failed without a status code, and `MetricAttributes` only attached `http.response.status_code` when that value was `> 0`. That meant a request that timed out or never got a response was recorded in `linodego.client.operation.duration` with the exact same attributes as a successful call — there was no way to filter or alert on the failure using this metric alone. The span itself already recorded the error correctly (`RecordError` / `SetStatus`), so this only affected the metric.

Fixes #879

---

## Checklist

- [x] Tests pass: `make test`


Labreo:
Tested locally and tests pass cleanly (`go test ./...` in the linodego package).

One quick edge case on `ErrorType(err)`: since it uses `reflect.TypeOf(err)` directly, wrapped errors (e.g. `fmt.Errorf("...: %w", err)`) will report as `*fmt.wrapError` rather than the root cause (like `*net.OpError` or `context.deadlineError`). 

Should we unwrap the error chain first before checking the type, or is `*fmt.wrapError` expected here?

bhuvan-somisetty:
Good catch, @Labreo you were right, `ErrorType(err)` was typing the outermost wrapper. Fixed in df68b78: it now unwraps the chain with `errors.Unwrap` in a loop before calling `reflect.TypeOf`, so a `fmt.Errorf("...: %w", err)` around a `*net.OpError`/`context.deadlineExceeded` now reports the root cause instead of `*fmt.wrapError`.

Added a test case covering the wrapped scenario. Ran `go vet ./...` and `go test -v -cover ./...` locally everything passes, semconv package coverage is still 96.6%.

darccio:
@bhuvan-somisetty Please regenerate the bundle.

bhuvan-somisetty:
@darccio done, rebased on main and regenerated the bundle. CI is green.

instrumentation/github.com/linode/linodego/v2/hook.go
amazingakai, line 141: Maybe we should use `finishSpanWithError` here as well?

amazingakai (approved):
LGTM ❤️

PR #883, opened 2026-07-30
fix(tool): resolve canFlattenTJump false positives on unrelated identifier matches

### LINKED ISSUE #882: bug(tool/optimize): canFlattenTJump disables TJump flattening on false positive identifier matches
### Component(s)
`tool/internal/instrument`

### What happened?

**Describe the bug**
In the compiler's trampoline optimization pass, `canFlattenTJump` attempts to verify if a trampoline-jump-if (TJump) can be flattened into sequential execution by asserting that the `HookContext` does not call `SetSkipCall`. 

However, the current AST inspection uses a naive `strings.Contains` across all `*dst.Ident` nodes in the hook function:
```go
// tool/internal/instrument/optimize.go
if ident, ok := node.(*dst.Ident); ok {
    if strings.Contains(ident.Name, trampolineSetSkipCallName) {
        found = true
        return false
    }
}
```

Because `strings.Contains` matches any identifier containing the substring `"SetSkipCall"`, the optimization pass triggers a false positive and **completely disables TJump flattening** if:
1. The user defines a local variable like `mySetSkipCall`.
2. The user calls a completely unrelated function containing the substring, such as `req.Header.SetSkipCall()`.
3. The user has a struct field `SetSkipCallback`.

Since `canFlattenTJump` subsequently runs an escape analysis to ensure the `HookContext` parameter is *only* used as a direct method receiver (e.g., `ictx.Method()`), we have a mathematical guarantee that any valid call to `SetSkipCall` on the context must be an exact `*dst.SelectorExpr` where `X` is the hook context parameter and `Sel` is `SetSkipCall`. 

The global string search is both computationally wasteful (visiting every identifier unnecessarily) and architecturally unsafe, leading to silent de-optimization of hot paths.

**Steps to reproduce**
Define an instrumentation hook that does not skip the call, but happens to use the string `"SetSkipCall"` in an unrelated identifier:
```go
func BeforeAPICall(ictx hook.HookContext, req *http.Request) {
    var previousSetSkipCall bool // Falsely triggers strings.Contains
    _ = previousSetSkipCall
}
```
The compiler will refuse to flatten the TJump.

**What did you expect to see?**
The AST optimization pass should properly verify `*dst.CallExpr` -> `*dst.SelectorExpr` to ensure the receiver is strictly the `HookContext` parameter before disabling flattening.

### Additional context
Since `canFlattenTJump` already extracts `hookContextParam` from the function signature (`hookFunc.Type.Params.List[0].Names[0].Name`), we can surgically target `ictx.SetSkipCall()` rather than globally grepping all identifiers. 

I am working on a PR to resolve this AST traversal logic.

Aryanbhargava18:
## Description

This PR refactors the AST identification logic inside `canFlattenTJump` (`tool/internal/instrument/optimize.go`) to use strict type-checked AST traversal (`*dst.CallExpr` -> `*dst.SelectorExpr`) rather than a global `strings.Contains` substring match.

Previously, the compiler pass attempted to detect whether a HookContext parameter called `SetSkipCall` by checking if any identifier in the entire hook function contained the substring `"SetSkipCall"`. This caused the optimization pass to falsely disable trampoline-jump-if flattening whenever a user defined unrelated variables (e.g., `mySetSkipCall`) or invoked similarly named methods on unrelated objects.

By specifically inspecting the AST for a method invocation where the receiver explicitly matches the extracted `hookContextParam` identifier, we eliminate false-positive de-optimizations. 

## Motivation

Fixes #882

The previous string-matching approach silently de-optimized performance-critical trampoline hooks when perfectly valid Go code happened to share a substring with the reserved `SetSkipCall` method. Since the subsequent escape analysis phase guarantees that `hookContextParam` is only used as a method receiver (otherwise flattening is legitimately aborted), we have a mathematical guarantee that any valid call to `SetSkipCall` will be parsed as a `SelectorExpr` on the context object. 

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/instrument/optimize.go
+++ b/tool/internal/instrument/optimize.go
@@ -210,17 +210,20 @@ func canFlattenTJump(hookFunc *dst.FuncDecl) bool {
 	// If found, the trampoline-jump-if cannot be flattened.
 	found := false
 	dst.Inspect(hookFunc, func(node dst.Node) bool {
-		if call, ok := node.(*dst.CallExpr); ok {
-			if sel, ok := call.Fun.(*dst.SelectorExpr); ok {
-				if id, ok := sel.X.(*dst.Ident); ok && id.Name == hookContextParam {
-					if sel.Sel.Name == trampolineSetSkipCallName {
-						found = true
-						return false
-					}
-				}
-			}
-		}
 		if found {
+			return false
+		}
+		call, isCall := node.(*dst.CallExpr)
+		if !isCall {
+			return true
+		}
+		sel, isSel := call.Fun.(*dst.SelectorExpr)
+		if !isSel {
+			return true
+		}
+		id, isID := sel.X.(*dst.Ident)
+		if isID && id.Name == hookContextParam && sel.Sel.Name == trampolineSetSkipCallName {
+			found = true
 			return false
 		}
 		return true
--- a/tool/internal/instrument/optimize.go
+++ b/tool/internal/instrument/optimize.go
@@ -4,8 +4,6 @@
 package instrument
 
 import (
-	"strings"
-
 	"github.com/dave/dst"
 
 	"go.opentelemetry.io/otelc/tool/ex"
@@ -205,14 +203,21 @@ func removeAfterTrampolineDecl(targetFile *dst.File, tjump *TJump) error {
 // 1. SetSkipCall is never called (so skip is always false)
 // 2. The HookContext parameter is only used as a receiver for method calls
 func canFlattenTJump(hookFunc *dst.FuncDecl) bool {
-	// Check if the hook function contains any "SetSkipCall" string
-	// If found, the trampoline-jump-if cannot be flattened
+	// The hook context parameter is always the first parameter of the hook function.
+	hookContextParam := hookFunc.Type.Params.List[0].Names[0].Name
+
+	// Check if the hook function explicitly calls SetSkipCall on the hook context.
+	// If found, the trampoline-jump-if cannot be flattened.
 	found := false
 	dst.Inspect(hookFunc, func(node dst.Node) bool {
-		if ident, ok := node.(*dst.Ident); ok {
-			if strings.Contains(ident.Name, trampolineSetSkipCallName) {
-				found = true
-				return false
+		if call, ok := node.(*dst.CallExpr); ok {
+			if sel, ok := call.Fun.(*dst.SelectorExpr); ok {
+				if id, ok := sel.X.(*dst.Ident); ok && id.Name == hookContextParam {
+					if sel.Sel.Name == trampolineSetSkipCallName {
+						found = true
+						return false
+					}
+				}
 			}
 		}
 		if found {
@@ -226,7 +231,6 @@ func canFlattenTJump(hookFunc *dst.FuncDecl) bool {
 
 	// Check if the hook context parameter escapes (used for non-method calls)
 	escape := false
-	hookContextParam := hookFunc.Type.Params.List[0].Names[0].Name
 	if hookContextParam == ast.IdentIgnore {
 		// If the parameter is ignored, it doesn't escape because it is not used
 		return true
--- a/tool/internal/instrument/optimize.go
+++ b/tool/internal/instrument/optimize.go
@@ -204,20 +204,26 @@ func removeAfterTrampolineDecl(targetFile *dst.File, tjump *TJump) error {
 // 2. The HookContext parameter is only used as a receiver for method calls
 func canFlattenTJump(hookFunc *dst.FuncDecl) bool {
 	// The hook context parameter is always the first parameter of the hook function.
-	hookContextParam := hookFunc.Type.Params.List[0].Names[0].Name
+	hookContextParam := ""
+	if len(hookFunc.Type.Params.List[0].Names) > 0 {
+		hookContextParam = hookFunc.Type.Params.List[0].Names[0].Name
+	}
 
-	// Check if the hook function explicitly calls SetSkipCall on the hook context.
-	// If found, the trampoline-jump-if cannot be flattened.
+	if hookContextParam == "" {
+		return true // Unnamed parameter cannot be referenced
+	}
+
+	// Check if the hook function references SetSkipCall on the hook context.
+	// If found, the trampoline-jump-if cannot be flattened. Match the bare
+	// selector rather than requiring a CallExpr, so the method-value form
+	// (f := ctx.SetSkipCall) is caught too: the escape analysis below treats
+	// it as a valid receiver use and would otherwise allow flattening.
 	found := false
 	dst.Inspect(hookFunc, func(node dst.Node) bool {
 		if found {
 			return false
 		}
-		call, isCall := node.(*dst.CallExpr)
-		if !isCall {
-			return true
-		}
-		sel, isSel := call.Fun.(*dst.SelectorExpr)
+		sel, isSel := node.(*dst.SelectorExpr)
 		if !isSel {
 			return true
 		}
--- a/tool/internal/instrument/optimize.go
+++ b/tool/internal/instrument/optimize.go
@@ -204,14 +204,7 @@ func removeAfterTrampolineDecl(targetFile *dst.File, tjump *TJump) error {
 // 2. The HookContext parameter is only used as a receiver for method calls
 func canFlattenTJump(hookFunc *dst.FuncDecl) bool {
 	// The hook context parameter is always the first parameter of the hook function.
-	hookContextParam := ""
-	if len(hookFunc.Type.Params.List[0].Names) > 0 {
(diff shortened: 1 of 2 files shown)


waqar2403 (commented):
@Aryanbhargava18 Looks Good. I confirmed both cases from the description flip from false to true.

One thing worth looking. The check only matches a `CallExpr` whose `Fun` is a `SelectorExpr`, so a reference to `ctx.SetSkipCall` that isn't in call position gets missed. The escape check below doesn't catch it either, since it treats any `ctx.<something>` selector as valid receiver usage and stops descending. Comparing against main:

```
f := ctx.SetSkipCall; f(true)     main=false  PR=true
register(ctx.SetSkipCall)         main=false  PR=true
s.fn = ctx.SetSkipCall            main=false  PR=true
```

Nothing in the repo writes hooks this way today. The only real `SetSkipCall` caller is `instrumentation/go.opentelemetry.io/otel/hook.go` and it uses the direct form. But `SetSkipCall` is exported in `pkg/hook`, so user hooks could, and `flattenTJump` sets `ifStmt.Cond = ast.BoolFalse()` and empties the then block, so a wrong `true` would drop the skip. The old `strings.Contains` was imprecise but erred toward not flattening.

Matching the `SelectorExpr` directly closes it:

```go
sel, isSel := node.(*dst.SelectorExpr)
if !isSel {
    return true
}
id, isID := sel.X.(*dst.Ident)
if isID && id.Name == hookContextParam && sel.Sel.Name == trampolineSetSkipCallName {
    found = true
    return false
}
return true
```

The cases above go back to false the two you're fixing still return true and the package tests pass.

Separately: hoisting `hookFunc.Type.Params.List[0].Names[0].Name` to the top puts it above the `if found { return false }` exit so it runs on inputs that used to return early. `func hookFunc(HookContext, string) { other.SetSkipCall(true) }` returns false on main but panics here, same root cause as #788. #789 guards that exact line, and merging the two branches conflicts in both `optimize.go` and `optimize_test.go`, so the ordering is worth sorting out.

tool/internal/instrument/optimize.go
```diff
@@ -205,19 +203,29 @@ func removeAfterTrampolineDecl(targetFile *dst.File, tjump *TJump) error {
...
 			return false
 		}
+		call, isCall := node.(*dst.CallExpr)
+		if !isCall {
+			return true
+		}
+		sel, isSel := call.Fun.(*dst.SelectorExpr)
+		if !isSel {
+			return true
+		}
```
darccio: The switch from substring matching to a strict `CallExpr` → `SelectorExpr` walk fixes the false positives in #882, but it also narrows detection in one case the old code did cover: a **method value** of `SetSkipCall`.

```go
func BeforeAPICall(ctx hook.HookContext, arg string) {
    f := ctx.SetSkipCall
    f(true)          // the hook genuinely wants to skip the call
}
```

`f := ctx.SetSkipCall` is a bare `SelectorExpr`, not the `.Fun` of a `CallExpr`, so this pass misses it. The escape analysis below doesn't catch it either: it treats *any* selector whose receiver is the context param as a valid method-receiver use and stops descending (optimize.go:246-253), so `ctx` is never flagged as escaping. Net result is `canFlattenTJump` returning `true` and the TJump getting flattened even though the hook does set skip — the condition is rewritten to `false`, the skip request is silently discarded, and the original function still runs.

I checked this against both revisions: on the merge base `canFlattenTJump` returns `false` here (safe), on this branch it returns `true`.

This seems worth weighing because the two failure modes aren't symmetric. The bug being fixed was a *performance* problem — a false positive that lost an optimization but stayed correct. This one is a *correctness* problem: a false negative that miscompiles the hook's control flow, surfacing at runtime with no diagnostic. The pattern is admittedly rare in real hook code, but given the header comment's warning that this pass "is fragile as it really heavily depends on the structure of trampoline-jump-if" and that "any change in tjump should be carefully examined", staying conservative looks like the better trade.

Matching the bare selector instead of requiring the `CallExpr` wrapper closes the gap while keeping the entire #882 fix — the receiver-identity check still lets `mySetSkipCall`, `req.Header.SetSkipCall()` and `SetSkipCallback` flatten. It's also slightly less code, since the second type assertion goes away:

```suggestion
	// Check if the hook function references SetSkipCall on the hook context.
	// If found, the trampoline-jump-if cannot be flattened. Match the bare
	// selector rather than requiring a CallExpr, so the method-value form
	// (f := ctx.SetSkipCall) is caught too: the escape analysis below treats
	// it as a valid receiver use and would otherwise allow flattening.
	found := false
	dst.Inspect(hookFunc, func(node dst.Node) bool {
		if found {
			return false
		}
		sel, isSel := node.(*dst.SelectorExpr)
		if !isSel {
			return true
		}
```

With this applied I get: `gofmt` clean, `go build ./tool/...` and `go vet ./tool/internal/instrument/...` pass, and the full `tool/internal/instrument` suite passes including the new `flatten despite unrelated SetSkipCall` case. A table entry for the method-value form (expecting `canFlatten: false`) would lock the behaviour in.

tool/internal/instrument/optimize_test.go
```diff
@@ -503,6 +503,18 @@ func TestFlattenTJump(t *testing.T) {
...
+			hookSrc: `package main
+			func hookFunc(ctx HookContext, arg1 string) {
+				var mySetSkipCall string
+				_ = mySetSkipCall
+				req.Header.SetSkipCall()
+			}`,
+			canFlatten:    true,
+			removedOnExit: false,
+			validate:      nil,
+		},
```
darccio, line 536: Two things worth tightening in this test table.

First, this case bundles two distinct false-positive shapes from #882 into one entry (a same-named local var, and a same-named method on an unrelated receiver). Splitting them makes a future regression easier to diagnose — right now if either detection path breaks, only "the combined case" fails, not which half.

Second, every case in this table names the context parameter `ctx`. Nothing proves the receiver-identity check is bound to the *actual* parameter name rather than a hardcoded string — a case using a differently-named param (e.g. `ictx`) would close that gap cheaply.

```suggestion
		{
			name: "flatten despite unrelated identifier substring",
			hookSrc: `package main
			func hookFunc(ctx HookContext, arg1 string) {
				var mySetSkipCall string
				_ = mySetSkipCall
			}`,
			canFlatten:    true,
			removedOnExit: false,
			validate:      nil,
		},
		{
			name: "flatten despite unrelated receiver call",
			hookSrc: `package main
			func hookFunc(ctx HookContext, arg1 string) {
				req.Header.SetSkipCall()
			}`,
			canFlatten:    true,
			removedOnExit: false,
			validate:      nil,
		},
		{
			name: "flatten with differently named context param",
			hookSrc: `package main
			func hookFunc(ictx HookContext, arg1 string) {
				req.Header.SetSkipCall()
			}`,
			canFlatten:    true,
			removedOnExit: false,
			validate:      nil,
		},
	}
```

darccio (commented):
The PR checklist leaves "Tests added for new functionality" unchecked, but `optimize_test.go` does add a new table case for this fix. Worth ticking that box so the checklist reflects what actually shipped.

darccio (changes requested):

tool/internal/instrument/optimize.go
Aryanbhargava18: Dropped the `CallExpr` wrapper to match the bare `SelectorExpr` directly. This catches the `f := ctx.SetSkipCall` assignment without breaking the escape analysis. Also added a bounds check on `Names` to prevent panics on unnamed parameters as @waqar2403 suggested.

tool/internal/instrument/optimize_test.go
Aryanbhargava18: Split up the test cases exactly as suggested and added a specific check for the method-value assignment.

Aryanbhargava18:
@darccio I have pushed the updates addressing both the CallExpr removal and the test table splitting. Could you please take another look and dismiss the blocking review if everything looks good?

Aryanbhargava18:
@waqar2403 added the bounds check on Names to handle the early return. thanks.

tool/internal/instrument/optimize.go
```diff
@@ -205,19 +203,35 @@ func removeAfterTrampolineDecl(targetFile *dst.File, tjump *TJump) error {
...
-	// If found, the trampoline-jump-if cannot be flattened
+	// The hook context parameter is always the first parameter of the hook function.
+	hookContextParam := ""
+	if len(hookFunc.Type.Params.List[0].Names) > 0 {
+		hookContextParam = hookFunc.Type.Params.List[0].Names[0].Name
+	}
+
+	if hookContextParam == "" {
+		return true // Unnamed parameter cannot be referenced
+	}
```
amazingakai: #789 includes the same fix, let's remove it from here.

amazingakai (approved):
Looks good. Thanks for the fix ❤️

darccio (approved):

PR #884, opened 2026-07-30
refactor(util): remove util.WriteFile in favor of os.WriteFile

### LINKED ISSUE #881: fix(util): remove process-terminating ex.Fatal call from util.WriteFile
### Problem Description
`util.WriteFile` returns an `error` signature, but inside its deferred cleanup function, if `file.Close()` returns a non-nil error, it executes `ex.Fatal(err)`. This terminates the process immediately (`os.Exit(1)`).

Because `util.WriteFile` is a general helper function with an `error` return signature, callers expect file I/O errors to be returned and handled gracefully higher up in the call stack rather than causing an unrecoverable process crash.

### Target Location
- `tool/util/sys.go` (lines 147-164)
### Proposed Solution
- Remove `ex.Fatal(err)` from `util.WriteFile`.
- - Check write errors and `file.Close()` errors explicitly, returning wrapped errors via `ex.Wrap`.

vedant21-oss:
## Description

Removes the `util.WriteFile` helper function from `tool/util/sys.go` and updates its single caller in `tool/internal/instrument/instrument_test.go` to use Go's standard library `os.WriteFile` directly.

## Motivation

Fixes #881
`util.WriteFile` was a custom wrapper only used in `instrument_test.go`. Replacing it with standard `os.WriteFile` simplifies the code and eliminates unnecessary utility wrappers.

## Checklist

- [x] Tests pass: `make test`


amazingakai:
The only place where this is used is `instrument_test.go`, can we remove this maybe and use `os.WriteFile` there?

vedant21-oss:
Fixed in 0657c8c — turned out util.WriteFile only had one caller (instrument_test.go), so I just dropped it and switched to os.WriteFile directly.

amazingakai:
Please also update the PR title/description.

amazingakai (approved):
LGTM, Thanks ❤️

PR #887, opened 2026-07-30
fix(tool): struct rules match non-struct types and add fields to the wrong struct

### LINKED ISSUE #886: bug(tool): struct rules match non-struct types and add fields to the wrong struct
`FindStructDecl` is an alias for `FindTypeDecl`:

```go
func FindStructDecl(root *dst.File, structName string) *dst.GenDecl {
	return FindTypeDecl(root, structName)
}
```

It never checks that the named type is a struct, and it returns the enclosing `GenDecl` rather
than the spec that matched. Two problems follow from that.

**Non-struct types match.** `StructFilter.Match` and the `InstStructRule` case in `matchOneRule`
both decide a match with `FindStructDecl(...) != nil`, so a struct rule targeting

```go
type Reader interface {
	Read() error
}
```

is collected as a match. It then fails in `AddStructField`, where
`util.AssertType[*dst.StructType]` calls `ex.Fatalf`:

```
Type assertion failed: got *dst.InterfaceType, expected *dst.StructType
```

The message identifies neither the rule nor the file.

**Grouped type blocks resolve to the wrong spec.** `AddStructField` reaches the struct through
`gen.Specs[0]`. For

```go
type (
	Config struct {
		Name string
	}
	Client struct {
		Addr string
	}
)
```

a struct rule targeting `Client` appends the new field to `Config`, and reports no error.

Expected: a rule whose target is not a struct does not match, and a rule targeting a name inside a
grouped `type ( ... )` block resolves to the spec with that name.

I have a fix and tests for both, happy to open a PR.

dobbydobap:
## Description

`FindStructDecl` was an alias for `FindTypeDecl`. It never checked that the named type was a struct, and it returned the enclosing `GenDecl` rather than the spec that matched.

This adds `FindStructType`, which resolves the spec by name and returns the `*dst.StructType` only when the named type really is a struct. `AddStructField` now takes the `*dst.StructType` directly, so it can no longer reach the wrong spec, and `StructFilter.Match` and the struct case in `matchRule` both use `FindStructType` so a struct rule no longer matches a non-struct type. `FindStructDecl` has no callers left and is removed. `FindTypeDecl` is unchanged and still used elsewhere.

`applyStructRule` keeps a nil check and returns a clear error, so a rule that slips through still reports rather than failing obscurely.

Tests cover a grouped type block where the target is not the first spec, and a rule whose target names a non-struct type.

## Motivation

Neither failure produced a useful diagnostic. A rule targeting an interface was collected as a match and only failed later on a type assertion that named neither the rule nor the file, and in a grouped `type ( ... )` block a field could be added to a struct the rule never named without any error at all.

### Behaviour change worth flagging

`has_struct` now matches only real struct types, which widens instrumentation in one
case: a `not: has_struct: X` rule where `X` is an interface. Previously `has_struct`
matched the interface and `not` excluded the file; now it does not match, so `not`
succeeds and the file is instrumented. This is the documented behaviour -- `base.go`
describes `has_struct` as "match files that declare this struct type" -- so the old
result was the deviation, but it is a silent change in the wider direction. It is now
noted on the field doc and covered in `TestStructFilter_Match`.

Fixes #886

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


anand-242003 (approved):
I reviewed this change and verified the two issues described in #886. Replacing `FindStructDecl` with `FindStructType` ensures struct rules only match actual `*dst.StructType`s and correctly resolves the matching `TypeSpec` within grouped `type (...)` declarations instead of relying on `Specs[0]`. I also reviewed the added regression tests covering interfaces, grouped declarations, and `StructFilter`, and they appear to exercise the original failure cases well. The refactoring also improves error handling by returning a descriptive error instead of failing later through a type assertion.

tool/internal/ast/shared_test.go
```diff
@@ -232,3 +232,48 @@ func TestFindNamedDecl(t *testing.T) {
...
+		src, err := NewAstParser().ParseSource(`package main
+type (
+	First  = int
+	Second struct{ a int }
+)
+`)
+		require.NoError(t, err)
+		assert.Nil(t, FindStructType(src, "First"))
+		assert.NotNil(t, FindStructType(src, "Second"))
+	})
```
darccio, line 264: `TestFindStructType` covers plain, interface, alias, missing, and grouped-block
cases, but not generic structs — and generics are the one shape where the
`ts.Type` assertion could plausibly have gone wrong, since the type parameters
live on `ts.TypeParams` rather than in `ts.Type`.

I checked this against the branch and the behaviour is already correct:
`Gen[T any]` and `GenMulti[K comparable, V any]` both resolve to their
`*dst.StructType`, and `AddStructField` appends to them as expected. So this
isn't a bug — it's just worth locking in, so a future refactor of
`FindStructType` can't silently regress generic targets.

```suggestion
	})

	t.Run("resolves generic structs", func(t *testing.T) {
		src, err := NewAstParser().ParseSource(`package main
type Gen[T any] struct{ v T }
type GenMulti[K comparable, V any] struct{ m map[K]V }
`)
		require.NoError(t, err)
		assert.NotNil(t, FindStructType(src, "Gen"))
		assert.NotNil(t, FindStructType(src, "GenMulti"))
	})
```

tool/internal/setup/filter.go
```diff
@@ -93,7 +93,7 @@ type StructFilter struct {
 }
 
 func (f *StructFilter) Match(ctx *MatchContext) bool {
-	return ast.FindStructDecl(ctx.AST, f.Struct) != nil
+	return ast.FindStructType(ctx.AST, f.Struct) != nil
```
darccio, line 98: Worth calling out explicitly, because it's the one direction in which this fix
makes instrumentation *wider* rather than narrower.

`has_struct` is reachable under `not` (there are tests for exactly that shape:
`filter_test.go:603` uses `Not{HasStruct: "Mock"}`, `match_test.go:915` uses
`Not{HasStruct: "MockConn"}`). For a user who writes `not: has_struct: MockConn`
to skip mock files, and whose `MockConn` is an *interface* — which is the common
shape for Go mocks, and what those fixture names suggest — the behaviour flips:

- before: `has_struct` matched the interface, so `not` **excluded** the file
- after:  `has_struct` no longer matches, so `not` **includes** the file

The file the user meant to skip now gets instrumented. I think the new behaviour
is right — `base.go:46` already documents this as "match files that declare this
struct type", so the old behaviour was the deviation — but it's a silent change
in the riskier direction, and neither of those two tests would catch it (both
only assert that the combinator builds the right filter type, not how it
matches).

Could this get a line in the PR description, and ideally a note on the
`has_struct` field doc in `base.go`? A reader hitting this six months from now
will otherwise file it as a regression.

```suggestion
	// Only true structs match: an interface or alias of the same name does not.
	// Note this inverts under `not` — a `not: has_struct: X` rule where X is an
	// interface now includes files it previously excluded.
	return ast.FindStructType(ctx.AST, f.Struct) != nil
```

tool/internal/instrument/apply_struct_test.go
```diff
@@ -0,0 +1,64 @@
...
+	r := &rule.InstStructRule{
+		InstBaseRule: rule.InstBaseRule{Name: "add_field"},
+		Struct:       "Foo",
+		NewField:     []*rule.InstStructField{{Name: "Traced", Type: "bool"}},
+	}
+
+	err := newTestPhase().applyStructRule(context.Background(), r, file)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "Foo")
```
darccio: This asserts only that the error mentions `Foo`, but the point of the change is
that the *old* failure was uninformative — it died on a type assertion naming
neither the rule nor the file. A test proving the new diagnostic is good should
assert the discriminating half of it; as written this still passes if the
message regresses to bare `"Foo"`.

The sibling test does assert the full text — `apply_func_test.go:43` uses
`assert.Contains(t, err.Error(), "can not find function Target")`.

I ran this against the branch; the produced string is exactly
`can not find struct "Foo" (missing, or not a struct type)`.

```suggestion
	assert.Contains(t, err.Error(), `can not find struct "Foo"`)
	assert.Contains(t, err.Error(), "not a struct type")
```

tool/internal/setup/filter_test.go
```diff
@@ -114,12 +114,16 @@ func TestStructFilter_Match(t *testing.T) {
...
+type Reader interface{ Read() error }
 func NotAStruct() {}
 `)
 
 	if !(&setup.StructFilter{Struct: "Server"}).Match(ctx) {
 		t.Fatal("StructFilter.Match(Server) = false, want true")
 	}
+	if (&setup.StructFilter{Struct: "Reader"}).Match(ctx) {
+		t.Fatal("StructFilter.Match(Reader) = true, want false")
+	}
```
darccio, line 126: Good addition. Could it go one step further and cover the negated form?

`has_struct` is reachable under `not`, and that's where this change is
observable in the riskier direction: a `not: has_struct: Reader` filter over a
file declaring `Reader` as an interface used to *exclude* that file and now
*includes* it. Files a user meant to skip start getting instrumented.

The existing negation tests don't catch this — `filter_test.go:603` and
`match_test.go:915` both assert only that the combinator builds the right filter
*type*, never how it matches. So the one behavioural flip this PR introduces is
currently untested.

The fixture here already declares `Reader`, so it's two lines. Verified compiling
and passing on this branch:

```suggestion
	if (&setup.StructFilter{Struct: "Reader"}).Match(ctx) {
		t.Fatal("StructFilter.Match(Reader) = true, want false")
	}
	// Interfaces no longer match, so `not: has_struct` now *includes* files
	// declaring an interface of that name, where it previously excluded them.
	if !(&setup.Not{Inner: &setup.StructFilter{Struct: "Reader"}}).Match(ctx) {
		t.Fatal("Not(StructFilter(Reader)).Match = false, want true")
	}
```

darccio (changes requested):

tool/internal/ast/shared_test.go
dobbydobap: Added, thanks. Agreed this is the shape most likely to break under a refactor:
the type parameters sit on ts.TypeParams, so ts.Type stays a plain
*dst.StructType and the assertion resolves. Locked in as a subtest.

tool/internal/setup/filter.go
dobbydobap: Good catch, and I agree the new behaviour is the right one -- base.go already
describes has_struct as "match files that declare this struct type".

I checked the golden fixtures too: not-filter-no-match declares
`type MockConn struct{}`, so that suite is silent on the interface case as well.
Comment applied here, note added to the has_struct field doc in base.go, and a
paragraph added to the PR description.

tool/internal/setup/filter_test.go
dobbydobap: Applied as suggested. You're right that the fixture already had Reader, so this
was cheaper than the separate test I'd started writing -- I dropped that in
favour of this. The negated assertion is the one that actually fails on old
main, so it's the part that earns its place.

tool/internal/instrument/apply_struct_test.go
dobbydobap: Applied. The message is exactly `can not find struct "Foo" (missing, or not a
struct type)`, and you're right that the old assertion would have survived a
regression to bare "Foo". Now matches the apply_func_test.go style.

darccio (approved):

PR #888, opened 2026-07-30
fix(instrumentation/linodego): attach error.type on doRequest span for non-HTTP failures

### LINKED ISSUE #885: bug(instrumentation/linodego): doRequest span missing error.type attribute for non-HTTP failures
### Describe the bug

The per-request `doRequest` span in the linodego v2 instrumentation never gets an `error.type` attribute when the underlying error carries no HTTP status code (timeout, connection refused, DNS failure, context cancellation, etc.). The higher-level public-method span already handles this correctly, so the two span layers this instrumentation produces are inconsistent with each other.

`AfterDoRequest` in `instrumentation/github.com/linode/linodego/v2/hook.go`:

```go
if err != nil {
    span.RecordError(err)
    if code, ok := semconv.StatusCodeFromError(err); ok {
        span.SetAttributes(semconv.LinodegoErrorTraceAttrs(err)...)
        if sc, desc := semconv.HTTPClientStatus(code); sc != codes.Unset {
            span.SetStatus(sc, desc)
        }
    } else {
    }
    logger.Debug("AfterDoRequest error", "error", err)
}
```

When `StatusCodeFromError` can't extract a code, the `else` branch only calls `span.SetStatus` - it never attaches `semconv.ErrorType(err)` as a span attribute.

Compare this with `finishSpanWithError`, used for the public-method span layer, a few lines below in the same file:

```go
et := semconv.ErrorType(err)
span.SetAttributes(et)
```

`finishSpanWithError` always sets `error.type` on the no-status-code path. `AfterDoRequest` doesn't, even though `semconv.ErrorType` is already available in the same package and is exactly what's needed here.

### Environment

- linodego version: `github.com/linode/linodego/v2 v2.4.1`
- Package: `go.opentelemetry.io/otelc/instrumentation/github.com/linode/linodego/v2`

### To Reproduce

1. Point a `linodego.Client` at an unreachable host so `doRequest` fails before a response is received:
   ```go
   client.GetInstance(ctx, 12345) // times out / connection refused
   ```
2. Inspect the `{METHOD} {endpoint}` span produced by `BeforeDoRequest`/`AfterDoRequest` (the doRequest span, not the outer `linodego.GetInstance` operation span).
3. The span status is correctly `codes.Error` with the error message as description, but it has no `error.type` attribute - there's no way to filter/group these spans by failure kind (e.g. distinguish a timeout from a DNS failure) without parsing the status description string.
4. The outer `linodego.GetInstance` span for the same failed call does carry `error.type`, since it goes through `finishSpanWithError` instead.

### Expected behavior

`AfterDoRequest` should attach `semconv.ErrorType(err)` to the span in the no-status-code branch, the same way `finishSpanWithError` does, so both span layers this instrumentation produces carry the same error attributes for the same failure.

### Additional context

Existing test `TestBeforeAfterDoRequest_PlainError` in `hook_test.go` only asserts `Status().Code`/`Status().Description` for this path, so it wouldn't currently catch a missing `error.type` attribute - that assertion should be added alongside the fix.

I'd like to pick this up.

bhuvan-somisetty:
## Description

`AfterDoRequest` in `instrumentation/github.com/linode/linodego/v2/hook.go` sets the doRequest span's status to `codes.Error` when a request error carries no HTTP status code (timeout, connection refused, DNS failure, context cancellation, etc.), but never attaches an `error.type` attribute for that case. `finishSpanWithError`, used for the public-method span a few lines below in the same file, already does this via `semconv.ErrorType(err)`. This PR brings `AfterDoRequest` in line with it so both span layers this instrumentation produces carry the same error attributes for the same failure.

## Motivation

Without `error.type` on the doRequest span, there's no way to filter or group these failed requests by failure kind (timeout vs. DNS vs. connection refused, etc.) without parsing the free-text status description. The outer `linodego.<Method>` span for the same call already carries this attribute, so the two layers were inconsistent for the exact same underlying error.

Fixes #885

---

## Checklist

- [x] Tests pass: `make test`

Note: this branches off #880 since it reuses `semconv.ErrorType`, which that PR introduces - the diff here will shrink to just the `hook.go`/`hook_test.go` changes once #880 lands. Ran `go build ./...`, `go vet ./...`, and `go test -v -cover ./...` locally in the linodego package - all pass, coverage unchanged (96.6% semconv, 8.5% package).


darccio:
@bhuvan-somisetty CI is failing. Can you check it, please?

bhuvan-somisetty:
@darccio good catch, thanks. Branch was way behind main so the archive drifted — rebased and regenerated the bundle, all green now.

amazingakai:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/880#discussion_r3740884571 should address this. Since this is a small change, let's keep it all in one place. Closing this PR!

PR #889, opened 2026-07-30
fix(genai): record error for HTTP 4xx+ responses in OpenAI and Anthropic middleware

shivv23:
## Description

Adds missing `span.RecordError()` calls in the HTTP 4xx/5xx error branch of the OpenAI (v1, v2, v3) and Anthropic middleware. When the API returns an error status code, the error is now recorded as an exception event on the span.

## Problem

All four middleware versions (`openai-go`, `openai-go/v2`, `openai-go/v3`, `anthropic-sdk-go`) correctly call `span.RecordError(err)` in the transport error branch (`err != nil`), but the HTTP 4xx/5xx branch (`resp.StatusCode >= 400`) only calls `span.SetStatus(codes.Error, ...)` without `RecordError`. This means:

- API-level errors (rate limits, auth failures, server errors) don't appear in the span's event list
- Stack traces are not captured for API errors
- Debugging production issues with API errors is significantly harder

## Solution

Add `span.RecordError(fmt.Errorf(resp.Status))` in the 4xx+ branch of each middleware, following the same pattern used by the transport error branch and other instrumentations in this repository (database/sql, redis, gRPC).

### Changes

| File | Change |
|------|--------|
| `instrumentation/github.com/openai/openai-go/middleware.go` | Added `span.RecordError` to 4xx+ branch |
| `instrumentation/github.com/openai/openai-go/v2/middleware.go` | Same |
| `instrumentation/github.com/openai/openai-go/v3/middleware.go` | Same |
| `instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go` | Same |
| `instrumentation/github.com/openai/openai-go/middleware_integration_test.go` | Added `TestOtelMiddleware_HTTPErrorStatus` + event assertions on existing `TestOtelMiddleware_NextError` |
| `instrumentation/github.com/openai/openai-go/v2/middleware_integration_test.go` | Same |
| `instrumentation/github.com/openai/openai-go/v3/middleware_integration_test.go` | Same |
| `instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go` | Added event assertions on existing `TestOtelMiddleware_HTTPError` and `TestOtelMiddleware_TransportError` |

## Testing

All existing tests pass, plus new assertions verify that `exception` events are created for both HTTP error statuses and transport errors.

```
cd instrumentation/github.com/openai/openai-go && go test ./...  # PASS
cd instrumentation/github.com/openai/openai-go/v2 && go test ./...  # PASS
cd instrumentation/github.com/openai/openai-go/v3 && go test ./...  # PASS
cd instrumentation/github.com/anthropics/anthropic-sdk-go && go test ./...  # PASS
```

## Checklist

- [x] Code formatted
- [x] Linters pass
- [x] Tests pass
- [x] Tests added for new behavior
- [x] No breaking changes
- [x] No new dependencies


instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -171,6 +172,7 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 		}
 
 		if resp.StatusCode >= 400 {
+			span.RecordError(fmt.Errorf("%s", resp.Status))
```
NameHaibinZhang: Consider `errors.New(resp.Status)` instead of `fmt.Errorf("%s", resp.Status)` — it avoids the extra `fmt` import and is the idiomatic way to construct a plain-text error. Also note the resulting `exception.type` will be `*errors.errorString`, which carries little aggregation value; a small typed error (similar to the status error used in the linodego instrumentation) would make exception events more useful.

Separately, this makes GenAI middlewares diverge from the `net/http` client hook, which only records exception events for transport errors (status-code failures are represented via span status + `error.type` only). Under a 429 storm this produces one exception event per request, so it may be worth confirming this representation is intended project-wide. Same applies to v2/v3 and the Anthropic middleware.

instrumentation/github.com/openai/openai-go/middleware_integration_test.go
```diff
@@ -242,6 +243,46 @@ func TestOtelMiddleware_NextError(t *testing.T) {
 
 	span := spans[0]
 	assert.Equal(t, "chat gpt-4", span.Name())
+
+	events := span.Events()
+	require.Len(t, events, 1, "expected exception event for transport error")
+	assert.Equal(t, "exception", events[0].Name)
+}
+
+func TestOtelMiddleware_HTTPErrorStatus(t *testing.T) {
```
NameHaibinZhang, line 252: The new tests only exercise 429. Consider adding a 5xx case (e.g. 500) to pin the contract for server errors, and a streaming case (`Content-Type: text/event-stream` with an error status) since the `StatusCode >= 400` branch now short-circuits before the streaming path — that interaction is currently untested. Same applies to the v2/v3 test files.

NameHaibinZhang (commented):
Thanks for the fix. The change is consistent across all four middlewares, the bundle is regenerated correctly, and there are no API or dependency changes. Tests validate the new exception events. Left a couple of suggestions on error construction/cross-module consistency and test coverage.

instrumentation/github.com/openai/openai-go/middleware.go
shivv23: Fixed!

instrumentation/github.com/openai/openai-go/middleware_integration_test.go
shivv23: Fixed!

shivv23:
@NameHaibinZhang All changes made. Please have a look.

NameHaibinZhang (approved):

PR #892, opened 2026-07-30
docs(rfcs): propose ADR 0007 for genai.Observer architecture

Aryanbhargava18:
This ADR proposes the `genai.Observer` architecture to centralize streaming memory bounds, enforce zero-overhead content capture limits, and unify semantic convention mapping across AI SDKs.

It directly addresses prior feedback by including explicit migration paths away from duplicated `io.TeeReader` middleware, backwards compatibility guarantees, and a strict exclusion of non-HTTP protocols like MCP.

*(Note: Replacing the previously closed PR #890. This version includes an ultra-verified `ExtractedData` interface that correctly handles Anthropic's caching metrics and explicitly addresses the SSE protocol framing discrepancies between OpenAI's `data:` and Anthropic's `event:` structures.)*

### DIFF
--- a/docs/adr/0007-genai-observer-architecture.md
+++ b/docs/adr/0007-genai-observer-architecture.md
@@ -1,0 +1,73 @@
+# 7. GenAI Observer Architecture
+
+Date: 2026-07-30
+
+## Status
+
+PROPOSED
+
+## Context
+
+Instrumenting AI SDKs (OpenAI, Anthropic) requires parsing Server-Sent Events (SSE) to aggregate telemetry like input/output token usage and finish reasons. Additionally, capturing streaming message content (`gen_ai.content.completion`) poses a high OOM risk if unbounded string deltas are accumulated in memory.
+
+Currently, this requires complex state machines inside the HTTP middleware of every SDK (e.g., `openai-go/streaming.go` is ~260 lines of custom `io.TeeReader` buffer management and content concatenation limits). As we expand to support Gemini and Agent frameworks (LangChain), duplicating this buffer management and span lifecycle across every package will lead to memory leaks, inconsistent truncation limits, and semantic convention drift.
+
+## Decision
+
+Introduce `pkg/genai.Observer` to centralize telemetry mapping, span lifecycle, and memory bounding. 
+
+### 1. The `StreamAdapter` for HTTP Middleware
+For SDKs operating at the HTTP layer, the Observer provides a `StreamAdapter` that takes ownership of the `*http.Response.Body`. It safely buffers SSE chunks, enforces a hard global memory limit (`maxResponseBodySize`) for content concatenation (preventing OOMs), calculates Time-To-First-Token (TTFT), and finalizes the OpenTelemetry span.
+
+### 2. SDKs as JSON Extractors
+HTTP SDK instrumentations are reduced to lightweight JSON mappers. They provide a callback to the `StreamAdapter`:
+
+```go
+type ExtractedData struct {
+    ID                 string
+    Model              string
+    PromptTokens       int64
+    CompletionTokens   int64
+    TotalTokens        int64
+    FinishReasons      []string
+    ContentDelta       string
+    ProviderAttributes []attribute.KeyValue
+}
+
+type ChunkExtractor func(rawEvent []byte) ExtractedData
+```
+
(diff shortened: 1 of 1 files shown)


Aryanbhargava18:
Rebased onto latest `main` to clear the stale bot.

Updated Section 2 to clarify the streaming boundary based on the recent Anthropic and Gemini discussions.

The current `internal/streaming.StreamingReader` assumes single-line `data:` chunks. Anthropic uses multi-line event frames (`event: message_start\ndata: ...\n\n` for input tokens, `event: message_delta\ndata: ...\n\n` for output tokens and stop reasons).

Instead of each provider maintaining its own `io.TeeReader` buffer loop, `Observer.WrapStream()` chunks on standard `\n\n` SSE message boundaries. The raw frame is passed straight to the provider's `ChunkExtractor` callback, while raw buffer management, TTFT calculation, and the 16 KiB UTF-8 content capture bounds remain centralized.

For non-HTTP calls (like MCP tool execution over stdio or LangChain chains), the compiler injects `Before`/`After` hooks that call `Observer.ObserveHook()` directly with the extracted metadata.

docs/adr/0007-genai-observer-architecture.md
```diff
@@ -0,0 +1,115 @@
...
+
+func (o *Observer) WrapStream(ctx context.Context, body io.ReadCloser, span trace.Span, ext ChunkExtractor) io.ReadCloser
+
+func (o *Observer) ObserveHook(ctx context.Context, span trace.Span, data ExtractedData)
+```
+
+### 2. HTTP SDK Instrumentation (OpenAI, Anthropic, Gemini)
+
+For SDKs operating over HTTP SSE transports:
+* **Frame Splitting vs Event Decoding:** `Observer.WrapStream()` delimits stream chunks on standard SSE message boundaries (`\n\n`). This makes the transport reader protocol-agnostic: OpenAI single-line payloads (`data: {...}\n\n`) and Anthropic multi-line frames (`event: message_start\ndata: {...}\n\n`) are both delivered intact to the provider's `ChunkExtractor`.
```
ADITYA-CODE-SOURCE, line 86: Does `\n\n` mean any valid SSE blank-line boundary, including LF, CRLF, and CR, or specifically two LF bytes? It would also be helpful to clarify whether delimiters split across multiple `Read` calls are supported.

docs/adr/0007-genai-observer-architecture.md
```diff
@@ -0,0 +1,115 @@
...
+* **State Machine Dispatch:** The `ChunkExtractor` decodes provider-specific fields (e.g. Anthropic's `event: message_start` input tokens vs `event: message_delta` output tokens) and normalizes them into `ExtractedData`.
+* **Prompt Cache Normalization:** Cache read and creation metrics are mapped directly into `gen_ai.usage.cache_read.input_tokens` and `gen_ai.usage.cache_creation.input_tokens` without double-counting (OpenAI includes cached tokens in `prompt_tokens`, while Anthropic reports them separately).
```
ADITYA-CODE-SOURCE, line 88: Are the token counts emitted by `ChunkExtractor` cumulative values or per-event deltas? How does the observer distinguish a missing value from an actual zero? Also, does `PromptTokens` include cache-read and cache-creation tokens, or are those normalized elsewhere?

docs/adr/0007-genai-observer-architecture.md
```diff
@@ -0,0 +1,115 @@
...
+* **Time-To-First-Token (TTFT):** Measured automatically upon the first non-empty content delta.
+* **Bounded Content Capture:** Enforces opt-in privacy constraints (`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`) and strict 16 KiB / UTF-8 boundary truncation centrally, preventing memory accumulation leaks and malformed runes.
```
ADITYA-CODE-SOURCE, line 90: The 16 KiB limit seems to apply to content retained for telemetry, but an incomplete SSE frame may still need to be buffered while waiting for its boundary. Is there a separate limit for that parser buffer? If a frame exceeds it, observation should be dropped safely without changing the bytes or errors returned to the application.

ADITYA-CODE-SOURCE (changes requested):

PR #921, opened 2026-08-02
test: fix Windows test failures due to POSIX shell assumptions

### LINKED ISSUE #920: test(tool/internal/setup): TestListBuildPlan fails on Windows due to hardcoded sh executable
### Describe the bug

Running unit tests for package `tool/internal/setup` on Windows causes `TestListBuildPlan` to fail. The test mock in `find_test.go` attempts to simulate `go build -a -x -n` output by executing `sh -c "<script>"`, hardcoding `sh` and POSIX heredoc syntax (`cat <<'EOF' >&2 ... EOF`).

On Windows environments where `sh` is not available in `%PATH%`, `exec.Command("sh", ...)` fails immediately with `exec: "sh": executable file not found in %PATH%`.

### Environment

 - OS: Windows (Windows 10 / 11 / Windows Server)
 - Go Version: go1.25+
 - Version: `main`

### To Reproduce

1. Clone the repository on a Windows environment (Command Prompt / PowerShell / CI Windows runner).
2. Run unit tests for `tool/internal/setup`:
   ```cmd
   go test -v ./tool/internal/setup
   ```
3. Observe the test failure output:
   ```text
   === RUN   TestListBuildPlan/filters_compile_and_cgo_commands
       find_test.go:409: 
           Error Trace:    c:/Users/Vishal/Code_File/Projects/opentelemetry-go-compile-instrumentation/tool/internal/setup/find_test.go:409
           Error:          Received unexpected error:
                           failed to run build plan: 
                           : exec: "sh": executable file not found in %PATH%
           Test:           TestListBuildPlan/filters_compile_and_cgo_commands
   --- FAIL: TestListBuildPlan (1.99s)
       --- FAIL: TestListBuildPlan/filters_compile_and_cgo_commands (0.62s)
       --- FAIL: TestListBuildPlan/passes_additional_build_args (0.30s)
       --- FAIL: TestListBuildPlan/returns_build_failure (0.27s)
       --- FAIL: TestListBuildPlan/empty_build_plan (0.24s)
       --- FAIL: TestListBuildPlan/ignores_malformed_compile_lines (0.23s)
       --- FAIL: TestListBuildPlan/test_subcommand_lists_a_go_test_plan (0.33s)
   ```

### Expected behavior

Unit tests in `tool/internal/setup` should execute and pass cleanly across all supported operating systems (Linux, macOS, and Windows) without assuming a POSIX shell environment.

#### Root Cause Analysis

In [tool/internal/setup/find_test.go](file:///c:/Users/Vishal/Code_File/Projects/opentelemetry-go-compile-instrumentation/tool/internal/setup/find_test.go#L382-L397):

```go
execCommandContext = func(
    _ context.Context,
    name string,
    args ...string,
) *exec.Cmd {
    assert.Equal(t, "go", name)
    assert.Equal(t, tt.expectedGoCmd, args)

    script := "cat <<'EOF' >&2\n" + tt.buildPlan + "\nEOF\n"
    if tt.buildFails {
        script += "\nexit 1\n"
    }

    return exec.Command("sh", "-c", script)
}
```

Line 395 hardcodes `exec.Command("sh", "-c", script)`. On native Windows, `sh` is not present in `%PATH%`, causing the sub-process invocation to fail before writing the expected build plan output to `stderr`.

#### Suggested Fix

Replace `exec.Command("sh", ...)` with a standard Go test helper process pattern (`TestHelperProcess` re-executing `os.Executable()`) to print `tt.buildPlan` to `stderr` and exit with status code `0` or `1`. This provides 100% cross-platform compatibility across Windows, Linux, and macOS without relying on external system binaries.

vishalmore90:
Fixes #920

### What changed
Two test files assumed POSIX-only shell tooling that does not exist on Windows:

- `tool/internal/setup/find_test.go`: `TestListBuildPlan` mocked
  `execCommandContext` with `exec.Command("sh", "-c", script)`.
  On Windows there is no `sh`, so all 6 sub-tests failed.

- `tool/util/sys_test.go`: `TestRunCmd` used `echo` as the test
  command. `echo` is a shell built-in on Windows and is not
  exec-able as a standalone binary.

### How it was fixed

**find_test.go** — adopted the standard [TestHelperProcess pattern]
used in the Go standard library: the test binary re-invokes itself
with a dedicated helper function that reads the desired build plan and
exit behavior from environment variables (`GO_WANT_HELPER_PROCESS`,
`GO_HELPER_BUILD_PLAN`, `GO_HELPER_BUILD_FAILS`). No `sh` required.

**sys_test.go** — replaced `echo` with `go version` / `go env GOOS`,
which are always available wherever Go is installed.

### Testing
- `go test -v -run "TestListBuildPlan|TestRunCmd" ./tool/internal/setup/... ./tool/util/...`
  passes on Windows (all 9 targeted sub-tests pass).
- `go test ./tool/util/...` passes fully on Windows.

### DIFF
--- a/tool/internal/setup/find_test.go
+++ b/tool/internal/setup/find_test.go
@@ -5,6 +5,7 @@ package setup
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -15,6 +16,20 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/otelc/tool/util"
 )
+
+func TestHelperProcess(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+	plan := os.Getenv("GO_HELPER_BUILD_PLAN")
+	if plan != "" {
+		fmt.Fprintln(os.Stderr, plan)
+	}
+	if os.Getenv("GO_HELPER_BUILD_FAILS") == "1" {
+		os.Exit(1)
+	}
+	os.Exit(0)
+}
 
 func TestParseCdDir(t *testing.T) {
 	tests := []struct {
@@ -380,19 +395,24 @@ echo nothing useful
 			t.Setenv(util.EnvOtelcWorkDir, tempDir)
 
 			execCommandContext = func(
-				_ context.Context,
+				ctx context.Context,
 				name string,
 				args ...string,
 			) *exec.Cmd {
 				assert.Equal(t, "go", name)
 				assert.Equal(t, tt.expectedGoCmd, args)
 
-				script := "cat <<'EOF' >&2\n" + tt.buildPlan + "\nEOF\n"
+				exe, err := os.Executable()
+				require.NoError(t, err)
+				cmd := exec.CommandContext(ctx, exe, "-test.run=^TestHelperProcess$")
+				cmd.Env = append(os.Environ(),
+					"GO_WANT_HELPER_PROCESS=1",
+					"GO_HELPER_BUILD_PLAN="+tt.buildPlan,
+				)
 				if tt.buildFails {
-					script += "\nexit 1\n"
+					cmd.Env = append(cmd.Env, "GO_HELPER_BUILD_FAILS=1")
 				}
-
-				return exec.Command("sh", "-c", script)
+				return cmd
 			}
 
 			subcommand := tt.subcommand
--- a/tool/internal/setup/find_test.go
+++ b/tool/internal/setup/find_test.go
@@ -394,6 +394,9 @@ echo nothing useful
 
 			t.Setenv(util.EnvOtelcWorkDir, tempDir)
 
+			exe, err := os.Executable()
+			require.NoError(t, err)
+
 			execCommandContext = func(
 				ctx context.Context,
 				name string,
@@ -402,8 +405,6 @@ echo nothing useful
 				assert.Equal(t, "go", name)
 				assert.Equal(t, tt.expectedGoCmd, args)
 
-				exe, err := os.Executable()
-				require.NoError(t, err)
 				cmd := exec.CommandContext(ctx, exe, "-test.run=^TestHelperProcess$")
 				cmd.Env = append(os.Environ(),
 					"GO_WANT_HELPER_PROCESS=1",
--- a/tool/util/sys_test.go
+++ b/tool/util/sys_test.go
@@ -20,13 +20,13 @@ func TestRunCmd(t *testing.T) {
 		expectErr bool
 	}{
 		{
-			name:      "simple echo command",
-			args:      []string{"echo", "hello"},
+			name:      "simple go command",
+			args:      []string{"go", "version"},
 			expectErr: false,
 		},
 		{
 			name:      "command with multiple arguments",
-			args:      []string{"echo", "hello", "world"},
+			args:      []string{"go", "env", "GOOS"},
 			expectErr: false,
 		},
 		{


anand-242003 (commented):
Nice fix. One question: do the existing GitHub Actions already execute these tests on windows-latest? Since this regression was Windows-specific, having at least one Windows job covering these packages would help ensure it doesn't silently return. If that's already covered elsewhere, then this looks good.

darccio:
> One question: do the existing GitHub Actions already execute these tests on windows-latest? Since this regression was Windows-specific, having at least one Windows job covering these packages would help ensure it doesn't silently return. If that's already covered elsewhere, then this looks good.

@anand-242003 It should be covered by the multiple tests on Windows:
* [E2E Tests / Test (go windows-latest amd64)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/30755128237/job/91632264120?pr=921)
* [Unit Tests / Unit Tests (go windows-latest amd64)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/30755128204/job/91632295735?pr=921)
* [Integration Tests / Integration Tests (go windows-latest amd64)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/30755128198/job/91632155695?pr=921)
* [LatestLibBuild Tests / LatestBuildTest Tests (go windows-latest amd64)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/30755128213/job/91632299954?pr=921)

Update: this is only covered by [Unit Tests / Unit Tests (go windows-latest amd64)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/30755128204/job/91632295735?pr=921) as it's the only job that builds `./tool/...`.

anand-242003:
Thanks for clarifying! I missed that those Windows jobs already cover these packages.

darccio (approved):

darccio:
For future travelers:

**The tests are already run on Windows.** `Unit Tests (go windows-latest amd64)` runs
`make test-unit` → `test-unit/tool` → `go test ./tool/...`, which covers both
`tool/internal/setup` and `tool/util`. (Correction to my earlier comment: of the four jobs
I listed, only that one does. E2E / Integration / LatestLibBuild run `test/...` and never
execute `TestListBuildPlan` or `TestRunCmd`.)

**But that job passed before this fix.** In the last pre-fix `main` run, the Windows unit
job reports `--- PASS: TestListBuildPlan` (all 6 subtests) and
`--- PASS: TestRunCmd/simple_echo_command`, with `hello` in the captured output — so the
`sh -c` mock and the bare `echo` really did work on the runner.

**Why:** `windows-latest` is not a plain Windows box. `actions/runner-images` installs Git
for Windows with `/o:PathOption=CmdTools` (`images/windows/scripts/build/Install-Git.ps1`)
and additionally does `Add-MachinePathItem "C:\Program Files\Git\bin"`. `CmdTools` is the
installer option *"Use Git and optional Unix tools from the Command Prompt"*, which appends
`Git\cmd`, `Git\mingw64\bin` **and `Git\usr\bin`** to the machine PATH. `Git\usr\bin` is the
MSYS2 userland — `sh.exe`, `bash.exe`, `cat.exe`, `echo.exe`, `tee.exe`, `find.exe` — so
every process inherits it: pwsh → GNU Make → the `go test` binary → the mocked subprocess.
#920 was reported from a machine using the default (recommended) `Cmd` option, which adds
only `Git\cmd` (git wrappers), hence `exec: "sh": executable file not found in %PATH%`.

Same reason `make test-unit` works on the runner at all: the recipes rely on
`set -euo pipefail`, `tee`, `find`, `cp -r`, `trap`, and GNU Make picks up `sh.exe` from
that same PATH.

So the Windows coverage exists, but it is structurally unable to reproduce a
"no POSIX shell on PATH" failure — adding more Windows jobs of the current shape wouldn't
help. Catching this class of regression would require a job that scrubs the Git/MSYS
directories out of `$env:PATH` and invokes `go test ./tool/...` directly (not through make,
which itself needs `sh`). That's a separate CI-hardening discussion and shouldn't block this
PR: removing the dependency at the source, as done here, is the right fix either way.

PR #927, opened 2026-08-03
fix(setup): close file handle on stale lock verification in tryAcquire

### LINKED ISSUE #926: fix(setup): close file handle on stale lock verification in tryAcquire
### Problem
In `tool/internal/setup/lock.go`, `tryAcquire` attempts to acquire the build lock using `flock.New(path)`. After acquiring the lock, `lockFileIsCurrent` checks whether the held lock handle matches the file at `path` on disk (guarding against races where another process unlinked the file prior to release).

If `lockFileIsCurrent` returns `!current` or returns an error, `tryAcquire` calls `_ = lock.Unlock()`.

However, in `gofrs/flock`, `lock.Unlock()` only unlocks the flock advisory mechanism but does NOT close the underlying file descriptor (`os.File`). Unlike earlier error paths in `tryAcquire` (lines 133 & 140) which explicitly call `_ = lock.Close()`, lines 145 & 149 omit `lock.Close()`, leaking open file handles when stale lock verification retries occur.

### Proposed Solution
Ensure `_ = lock.Close()` is called whenever `!current` or a `lockFileIsCurrent` error occurs:

```go
if err != nil {
    _ = lock.Unlock()
    _ = lock.Close()
    return nil, false, leftover, err
}
if !current {
    _ = lock.Unlock()
    _ = lock.Close()
    return nil, false, leftover, nil
}
```

vedant21-oss:
## Description
In `tool/internal/setup/lock.go`, `tryAcquire` attempts to acquire the build lock using `flock.New(path)`. After acquiring the lock, `lockFileIsCurrent` checks whether the held lock handle matches the file at `path` on disk.

If `lockFileIsCurrent` returns `!current` or returns an error, `tryAcquire` calls `_ = lock.Unlock()`. However, `lock.Unlock()` in `gofrs/flock` releases the flock advisory lock mechanism but leaves the underlying file handle (`os.File`) open. Omitting `lock.Close()` causes open file descriptor leaks when stale lock retries occur.

Fixes #926

## Fix Details
- Added `_ = lock.Close()` when `lockFileIsCurrent` returns an error or `!current` in `tryAcquire`.
- Ran unit test suite in `tool/internal/setup` (100% PASS).

## Checklist
- [x] PR title follows conventional commits format: `fix(setup): ...` 
- [x] Code formatted & linters pass
- [x] Tests pass

### DIFF
--- a/tool/internal/setup/lock_test.go
+++ b/tool/internal/setup/lock_test.go
@@ -10,7 +10,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/gofrs/flock"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -148,36 +147,5 @@ func TestAcquireBuildLockMissingWorkDirIsNoop(t *testing.T) {
 	require.NoError(t, err)
 	release()
 	assert.False(t, util.PathExists(missing), "no-op acquisition must not create the work dir")
-}
-
-func TestTryAcquireStaleLockCleanup(t *testing.T) {
-	lockTestDir(t)
-	path := buildLockPath()
-
-	// Handle A locks original file
-	lockA := flock.New(path)
-	acquiredA, err := lockA.TryLock()
-	require.NoError(t, err)
-	require.True(t, acquiredA)
-
-	// Remove original file on disk to simulate deletion race
-	require.NoError(t, os.Remove(path))
-
-	// Handle B creates and locks a new file at the same path
-	lockB := flock.New(path)
-	acquiredB, err := lockB.TryLock()
-	require.NoError(t, err)
-	require.True(t, acquiredB)
-
-	// lockA is now stale because path points to lockB's inode
-	current, err := lockFileIsCurrent(path, lockA)
-	require.NoError(t, err)
-	assert.False(t, current, "stale lock handle must not be current")
-
-	_ = lockA.Unlock()
-	_ = lockA.Close()
-
-	_ = lockB.Unlock()
-	_ = lockB.Close()
 }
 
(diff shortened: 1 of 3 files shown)


tool/internal/setup/lock_test.go
```diff
@@ -148,3 +149,35 @@ func TestAcquireBuildLockMissingWorkDirIsNoop(t *testing.T) {
...
+	require.NoError(t, os.Remove(path))
+
+	// Handle B creates and locks a new file at the same path
+	lockB := flock.New(path)
+	acquiredB, err := lockB.TryLock()
+	require.NoError(t, err)
+	require.True(t, acquiredB)
+
+	// lockA is now stale because path points to lockB's inode
+	current, err := lockFileIsCurrent(path, lockA)
```
anand-242003: Thanks for adding this regression test. One thing I noticed is that it exercises lockFileIsCurrent directly, but it doesn't actually execute the tryAcquire cleanup path modified by this PR. Since the test manually calls lockA.Close(), removing the new lock.Close() calls in tryAcquire would not cause this test to fail. Would it be possible to add a regression test that invokes tryAcquire and verifies this cleanup path directly?

anand-242003 (changes requested):

vedant21-oss:
Hi @anand-242003, thanks for the review!

Good point — I have updated `TestTryAcquireStaleLockCleanup` in `tool/internal/setup/lock_nonwindows_test.go` to invoke `tryAcquire` directly under an unlinked file condition. This directly exercises the `!current` cleanup branch inside `tryAcquire` and verifies that it returns `(nil, false, leftover, nil)` with the file handle properly closed.

Commit `d8442aa` has been pushed to the PR!

anand-242003 (approved):
Thanks for addressing the feedback. I verified the updated test now exercises the tryAcquire cleanup path directly, and the targeted test passes locally.

darccio (approved):
LGTM

PR #928, opened 2026-08-03
test(ex): cover Fatal and Fatalf exit paths

vignesh2027:
## Description

`Fatal` and `Fatalf` in `tool/ex/error.go` were the only two functions in the package with 0% test coverage, because they call `os.Exit` and so can't be run directly in the test process.

This PR adds subprocess-based tests using the standard Go re-exec pattern: `TestMain` runs a specific fatal case when `EX_FATAL_CASE` is set, and each test spawns the test binary with that variable, then asserts on the child's exit code and stderr. This covers all three branches of `Fatal` (nil error, joined/multi-unwrap error, single error) plus the `Fatalf` wrapper.

## Result

`tool/ex` coverage goes from **76.4% → 98.2%**, with `Fatal` and `Fatalf` both at 100%.

- `go test ./tool/ex/...` passes
- `golangci-lint run ./tool/ex/...` reports 0 issues
- `go vet` and `gofmt` clean

No production code changed — tests only.

### DIFF
--- a/tool/ex/fatal_test.go
+++ b/tool/ex/fatal_test.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"io"
 	"os"
+	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


darccio (approved):
LGTM

PR #929, opened 2026-08-03
fix(tool): preserve spaces in CGO build-plan directories

### LINKED ISSUE #912: bug(tool/setup): preserve spaces when parsing CGO build-plan directories
## Problem

`otelc` parses the build plan produced by:

```bash
go build -a -x -n
```

When a CGO project is located in a directory whose path contains spaces, `parseCdDir` truncates the directory name.

For example, for a project located at:

```text
/tmp/test project with spaces
```

Go emits the following build-plan line:

```text
cd /tmp/test project with spaces
```

However, the current implementation in `tool/internal/setup/find.go` splits the text after `cd` at the first space and returns:

```text
/tmp/test
```

instead of the complete path:

```text
/tmp/test project with spaces
```

## Impact

`findDeps` stores this parsed directory as `currentDir` and associates it with the CGO object directory. Later, `resolveCgoFile` uses that mapping to locate the original Go source file.

When the directory is truncated, `otelc` searches in the wrong location. As a result, CGO source files may not be resolved, and applicable instrumentation rules can be skipped.

## Related edge cases

While reviewing the parser, I also found two related cases:

* A plain `cd` line can panic because `parseCdDir` evaluates `line[3:]`.
* An unrelated command beginning with `cd`, such as `cdrom`, is incorrectly treated as a directory change.

## Reproduction

```bash
mkdir -p "/tmp/test project with spaces"
cd "/tmp/test project with spaces"
go build -a -x -n 2>&1
```

The output includes an unquoted directory-change line similar to:

```text
cd /tmp/test project with spaces
```

Passing this line to `parseCdDir` currently returns only `/tmp/test`.

## Expected behavior

`parseCdDir` should:

* Preserve the complete directory path after a valid `cd ` prefix.
* Reject an empty `cd` command without panicking.
* Ignore unrelated commands that merely begin with `cd`, such as `cdrom`.

Regression coverage should be added to `TestParseCdDir` and `TestFindCommands`.

## Coordination

PR #657 also changes `tool/internal/setup/find.go` and `tool/internal/setup/find_test.go` as part of its `-C` support.

I discussed the overlap with Azhar on Slack. He recommended opening this issue first so maintainers can decide whether the fix should be included in #657 or submitted as a separate PR.

I’m happy to implement the fix once the preferred scope is confirmed.

ADITYA-CODE-SOURCE:
## Description

Updates `parseCdDir` to preserve the complete directory path emitted by the Go build plan, including paths containing spaces.

The parser now:

- Accepts only a valid `cd ` prefix.
- Preserves the full directory path.
- Rejects empty `cd` commands without panicking.
- Ignores unrelated commands such as `cdrom`.

Regression tests cover paths with spaces, malformed commands, trailing whitespace, Windows paths, and `findCommands` filtering.

## Motivation

For CGO packages, the parsed directory is associated with the generated object directory and later used to resolve the original Go source file.

Previously, a path such as `/tmp/test project` was truncated to `/tmp/test`. This could prevent CGO source resolution and cause applicable instrumentation rules to be skipped.

Fixes #912


tool/internal/setup/find_test.go
```diff
@@ -201,6 +228,29 @@ cd /home/user/project/pkg/cgopkg
...
+		{
+			name: "malformed cd commands ignored",
+			buildPlanContent: `
+cd
+` + "cd   \n" + `
+cdrom
+cdrom /project/src
+`,
+			expectedCommands: nil,
+		},
```
darccio, line 253: `parseCdDir` and `findCommands` are both covered for spaced directories, but the actual regression from #912 lives one level up: `findDeps` (find.go:228-248) is what stores the parsed `cd` dir as `currentDir`, maps it to `objDir` in `cgoObjDirs`, and later `resolveCgoFile` joins it to resolve the original `.go` file. None of the current tests — old or new — drive that path end-to-end, and `findDeps` has no dedicated test at all.

Consider adding a case (in this file, or a new `TestFindDeps`) that feeds a build plan with a spaced `cd` dir through a `cgo`/`compile` pair and asserts the resolved source file, to directly guard the behavior this PR fixes.

darccio (changes requested):

ADITYA-CODE-SOURCE:
@darccio I added an end-to-end findDeps regression test for a source directory with spaces. It verifies that the CGO object directory resolves back to the correct original Go file. I also rebased the branch on the latest main.

darccio (approved):

PR #931, opened 2026-08-03
chore(test): add multi-process e2e for HTTP→SQL propagation

### LINKED ISSUE #924: chore(test): add multi-process e2e for HTTP→SQL propagation
## Summary

Per [docs/testing.md](docs/testing.md), E2E tests spin up **multiple instrumented processes** and assert they form one coherent trace (shared trace ID).

`test/apps/httpserverdbclient` is documented as a combined HTTP server + SQL client app **for e2e testing**, but today it is only exercised by an **integration** test:

- File: `test/integration/httpserver_dbclient_test.go`
- The request is made with uninstrumented `http.DefaultClient`
- Asserts 1 trace with 2 spans: HTTP server → SQL client (in-process parent/child)

That does **not** verify cross-process propagation from an instrumented HTTP client into the HTTP→SQL chain.

Existing E2E coverage:

| E2E Test | File |
|---|---|
| HTTP client + server | `http_test.go` |
| gRPC client + server | `grpc_test.go` |
| Kafka producer + consumer | `kafka_test.go` |

## Proposed work

Add `test/e2e/httpserver_dbclient_test.go` (name flexible), mirroring `test/e2e/http_test.go`:

1. `BuildAndStart("httpserverdbclient", "-front-port=<port>")`
2. `BuildAndRun("httpclient", "-addr", "http://127.0.0.1:<port>", ...)`
3. Assert **one** distributed trace with **3** spans:
   - instrumented HTTP **client**
   - HTTP **server** (frontend)
   - SQL **client** (frontend → DB)
4. Assert parent/child links and shared trace ID across the chain

### Files
- Add: `test/e2e/httpserver_dbclient_test.go`
- Reuse (no new apps): `test/apps/httpclient`, `test/apps/httpserverdbclient`

### No new instrumentation needed
HTTP client, HTTP server, and `database/sql` instrumentation already exist.

## Checklist
- [ ] Multi-process E2E: instrumented `httpclient` → `httpserverdbclient`
- [ ] `RequireTraceCount(1)` and 3 spans in the chain (HTTP client, HTTP server, SQL client)
- [ ] Assert shared trace ID + parent/child relationships
- [ ] `make all` / e2e suite passes

## References
- [docs/testing.md](docs/testing.md) — E2E definition
- `test/e2e/http_test.go`
- `test/integration/httpserver_dbclient_test.go`
- `test/apps/httpserverdbclient/`
- `test/apps/httpclient/`

I'd like to implement this after feedback on scope.

sanjana2505006:
## Description

Adds a multi-process e2e test that starts an instrumented `httpserverdbclient`, calls it with an instrumented `httpclient`, and asserts one distributed trace with three linked spans: HTTP client → HTTP server → SQL client.

## Motivation

`httpserverdbclient` is documented for e2e testing, but was only covered by an integration test that uses an uninstrumented `http.DefaultClient`. That misses cross-process context propagation into the HTTP→SQL chain. This change closes that gap per `docs/testing.md`.

Fixes #924

---

## Checklist

- [x] Tests pass: `make test`


test/e2e/httpserver_dbclient_test.go
```diff
@@ -0,0 +1,60 @@
...
+	f.RequireSpansPerTrace(3)
+
+	httpClientSpan := testutil.RequireSpan(t, f.Traces(),
+		testutil.IsClient,
+		testutil.HasAttributeContaining(string(semconv.URLFullKey), "/hello"),
+	)
+	httpServerSpan := testutil.RequireSpan(t, f.Traces(),
+		testutil.IsServer,
+		testutil.HasAttribute(string(semconv.URLPathKey), "/hello"),
+	)
```
darccio, line 46: The server span is matched on method/path but not on outcome, so this test passes on a failed request. If `db.QueryContext` returns an error the handler responds 500, yet the SQL client span is still created — you'd still get one trace, three spans, and intact parent/child links, and the assertions below would all succeed while the scenario actually failed.

Pinning the status code closes that hole and matches what `RequireHTTPServerSemconv` checks in `http_test.go`:

```suggestion
	httpServerSpan := testutil.RequireSpan(t, f.Traces(),
		testutil.IsServer,
		testutil.HasAttribute(string(semconv.URLPathKey), "/hello"),
		testutil.HasAttribute(string(semconv.HTTPResponseStatusCodeKey), int64(200)),
	)
```

test/e2e/httpserver_dbclient_test.go
```diff
@@ -0,0 +1,60 @@
...
+	sqlClientSpan := testutil.RequireSpan(t, f.Traces(),
+		testutil.IsClient,
+		testutil.HasAttribute("db.operation.name", "SELECT"),
```
darccio: Small consistency nit: this is the only attribute key in the test written as a bare string — the HTTP matchers above use `string(semconv.…Key)`. The typed key gives you compile-time breakage instead of a silent no-match if the constant is ever renamed.

Entirely optional, since `db_client_test.go` uses both styles.

```suggestion
		testutil.HasAttribute(string(semconv.DBOperationNameKey), "SELECT"),
```

darccio (changes requested):
Thanks for this — it closes exactly the gap #924 described, and the structure is right. A few things I want to call out as done well: using `FreePort` + `WaitForTCP` instead of a hardcoded port is more robust than the `http_test.go` it mirrors, the two client spans are disambiguated unambiguously (`url.full` vs `db.operation.name`, neither can match the other span), and keeping the assertions focused on propagation rather than re-checking semconv is the right scoping — that coverage already lives in `http_test.go` and `db_client_test.go`.

One thing needed fixing before this could run: `testutil.WaitForSpanFlush` doesn't exist anywhere in the repo, so the whole `e2e` build tag failed to compile with `undefined: testutil.WaitForSpanFlush`. Since you'd enabled maintainer edits, I pushed the fix directly to your branch (`aa57c435`) rather than round-tripping it.

The fix uses `f.WaitForSpans(3)`, matching `test/integration/httpserver_dbclient_test.go`. Worth noting the wait is genuinely needed and not just a compile fix: `BuildAndRun` returns when the *client* process exits, but the server span ends after the response is written and may still be in flight, so asserting immediately can observe only two of the three spans. There's no count-free "wait for flush" primitive because the collector is a passive sink — it can't know what's still in flight — so the count-based wait is the only shape available.

Verified locally: the full e2e suite passes under `-shuffle=on` (`TestHTTPServerDBClient`, `TestHttp`, `TestGrpc`, `TestKafka` all green).

sanjana2505006:
Thanks, @darccio, applied both suggestions (HTTP 200 on the server span matcher, and semconv.DBOperationNameKey for the SQL client span). Also pulled in your WaitForSpans fix.

darccio (approved):

PR #941, opened 2026-08-04
fix(setup): use atomic write for storing matched rules in SetupPhase.store

### LINKED ISSUE #935: fix(setup): use atomic write for storing matched rules in SetupPhase.store
### Description
In `tool/internal/setup/store.go`, `SetupPhase.store` writes the JSON representation of matched rule sets using `os.Create` and `file.Write`, relying on `defer file.Close()`.

If an I/O error or disk flush failure occurs during `file.Close()`, the error is swallowed. Additionally, writing directly to the target file path can leave behind corrupted JSON artifacts if the process is terminated mid-write.

### Location
- `tool/internal/setup/store.go`: lines 84–107

### Proposed Solution
Refactor `SetupPhase.store` to use `util.WriteFileAtomic(f, bs)`, ensuring safe atomic replacements and proper error checking on file close operations.

vedant21-oss:
### Description
In `tool/internal/setup/store.go`, `SetupPhase.store` writes the JSON representation of matched rule sets using `os.Create` and `file.Write`, relying on `defer file.Close()`.

If an I/O error or disk flush failure occurs during `file.Close()`, the error is swallowed. Additionally, writing directly to the target file path can leave behind corrupted JSON artifacts if the process is terminated mid-write.

### Changes
- Refactored `SetupPhase.store` to use `util.WriteFileAtomic(f, bs)`.
- Removed unused `os` package import.

Closes #935

### DIFF
--- a/tool/internal/setup/store.go
+++ b/tool/internal/setup/store.go
@@ -6,7 +6,6 @@ package setup
 import (
 	"context"
 	"encoding/json"
-	"os"
 
 	"go.opentelemetry.io/otelc/tool/ex"
 	"go.opentelemetry.io/otelc/tool/internal/rule"
@@ -86,21 +85,15 @@ func (sp *SetupPhase) store(ctx context.Context, matched []*rule.InstRuleSet, mo
 		return ex.Wrapf(err, "resolving rule paths")
 	}
 
-	f := util.GetMatchedRuleFile()
-	file, err := os.Create(f)
-	if err != nil {
-		return ex.Wrapf(err, "failed to create file %s", f)
-	}
-	defer file.Close()
-
 	bs, err := json.Marshal(matched)
 	if err != nil {
 		return ex.Wrapf(err, "failed to marshal rules to JSON")
 	}
 
-	_, err = file.Write(bs)
+	f := util.GetMatchedRuleFile()
+	err = util.WriteFileAtomic(f, bs)
 	if err != nil {
-		return ex.Wrapf(err, "failed to write JSON to file %s", f)
+		return ex.Wrapf(err, "failed to write matched rules to file %s", f)
 	}
 	sp.Info("Stored matched sets", "path", f)
 	return nil


darccio:
@vedant21-oss Please address the failing CI checks.

tool/internal/setup/store.go
```diff
@@ -86,21 +85,15 @@ func (sp *SetupPhase) store(ctx context.Context, matched []*rule.InstRuleSet, mo
...
-	defer file.Close()
-
 	bs, err := json.Marshal(matched)
 	if err != nil {
 		return ex.Wrapf(err, "failed to marshal rules to JSON")
 	}
 
-	_, err = file.Write(bs)
+	f := util.GetMatchedRuleFile()
+	err = util.WriteFileAtomic(f, bs)
```
darccio, line 94: `WriteFileAtomic` with no explicit `perm` defaults a newly-created file to `0o644`. The code it replaces used `os.Create`, which yields `0o666 & ~umask`. For `matched.json` this is probably harmless, but it's a behavior change worth a one-line callout in case anything downstream expected the old mode.

tool/internal/setup/lock.go
darccio, line 153: This hunk (the two `_ = lock.Close()` calls) is byte-identical to current `main` — it already merged via #927 (same author, merged 2026-08-03). It only shows up here because this branch forked before that PR landed. Rebasing onto current `main` will drop this file from the diff entirely; left as-is it's dead weight that makes the PR harder to scope to the actual fix.

tool/internal/setup/lock_nonwindows_test.go
```diff
@@ -51,22 +50,68 @@ func TestTryAcquireStaleLockCleanup(t *testing.T) {
...
+		if tryErr != nil && !acq && l == nil {
+			hitErrCleanup = true
+			break
+		}
+		if l != nil {
+			_ = l.Unlock()
+			_ = l.Close()
+		}
 	}
+	assert.True(t, hitErrCleanup, "expected tryAcquire to hit error cleanup path")
```
darccio: Two problems with this section:

1. `os.Chmod(permDir, 0o000)` does not block access for a process running as root (common in containerized CI). If the test runs as root, `hitErrCleanup` is never set and `assert.True(t, hitErrCleanup, ...)` fails outright rather than merely flaking.
2. Even when it does trigger, the comment's claim ("exercise err != nil cleanup path inside tryAcquire when lockFileIsCurrent fails") is probably not what's being tested: a 0-permission directory most likely makes `flock.New(path).TryLock()` fail to open the file at all, not `lockFileIsCurrent`'s `os.Stat`. The assertion doesn't pin down which branch actually ran, so it doesn't verify what it claims to.

Recommend dropping this section (the original `TestTryAcquireStaleLockCleanup` on `main` doesn't have it) or rewriting it to directly unit-test `lockFileIsCurrent`'s error path without depending on root-sensitive permission bits.

tool/internal/setup/lock_nonwindows_test.go
```diff
@@ -51,22 +50,68 @@ func TestTryAcquireStaleLockCleanup(t *testing.T) {
...
+		if tryErr == nil && !acq && l == nil {
+			hitStaleCleanup = true
+			break
+		}
+		if l != nil {
+			_ = l.Unlock()
+			_ = l.Close()
+		}
+	}
+	assert.True(t, hitStaleCleanup, "expected tryAcquire to hit stale lock cleanup path")
```
darccio: This replaces a simpler, already-merged version of this test (from #927) with a tight `for { select { default: } }` spin loop (lines 59-68) that has no backoff — it pins a CPU core for the loop's duration. Adding a short sleep in the default case would fix that, but it needs a `"time"` import that this file no longer has (this PR removes it). Since `main` already has a working version of this test, it's worth asking whether this rewrite needs to be part of this PR at all, rather than patching the busy-wait in place.

darccio (changes requested):
Scope mismatch: the title/description/issue (#935) describe only the `store.go` atomic-write change, but 2 of the 3 changed files (`lock.go`, `lock_nonwindows_test.go`) re-touch work already merged to `main` via #927 (same author). `lock.go`'s diff is a no-op against current `main`, and the test file rewrites a test that already exists there — it will conflict on rebase.

Recommend rebasing onto current `main` first: that should shrink this PR down to just the `store.go` fix, which is correct and sufficient to close #935 on its own. If the test hardening in `lock_nonwindows_test.go` is intentional, it'd be cleaner as its own PR with its own rationale, separate from #935's Close-error swallowing fix.

(Unrelated to the diff: `store_test.go`'s `TestSetupPhaseStoreCreateError` has a comment saying "so `os.Create` fails" — after this fix, the failure actually comes from `CreateTemp` inside `WriteFileAtomic`. Not part of this diff, so not blocking, just a heads-up for a future touch of that file.)

vedant21-oss:
@darccio Thanks for the review!

1. **Rebased onto `main`**: Dropped the redundant `lock.go` and `lock_nonwindows_test.go` hunks since they were already merged via #927. The PR is now strictly scoped to `SetupPhase.store`.
2. **Permissions Callout**: Noted that `util.WriteFileAtomic` defaults to `0o644` permissions for `matched.json`.

Ready for re-review!

darccio (approved):

Issue #922, opened 2026-08-02
bug(imports): fingerprint mismatch upon otelc introduction to project that builds fine otherwise


julian45:
### Describe the bug

When attempting to use compile-time instrumentation with the [Venator](https://codeberg.org/matrix-venator/venator) project, builds fail with multiple instances of errors like:

`/usr/local/go/pkg/tool/linux_amd64/link: fingerprint mismatch: go.opentelemetry.io/otelc/pkg/runtime has 5babcad2e54ff244, import from go.opentelemetry.io/otelc/instrumentation/database/sql expecting 8ddc38b67cd3c64b`

The project builds without trouble if compile-time instrumentation is not attempted, so some signs point to the issue originating in the introduction of CTI.

### Environment

 - OS: Fedora 44 x86_64 (also replicated in `golang:1.26-alpine` container)
 - Go Version: 1.26.5
 - Version: 1.0.1

### To Reproduce

1. Clone Venator project from repo
2. Apply the following changes in service of gaining compile-time instrumentation:
```diff
diff --git a/build.sh b/build.sh
index c05ec68..db09f31 100755
--- a/build.sh
+++ b/build.sh
@@ -70,7 +70,7 @@ function build() {
         VENATOR_BUILD_TAGS="$VENATOR_BUILD_TAGS,docs"
         go generate -tags "$VENATOR_BUILD_TAGS" ./internal/embedded_docs/
     fi;
-    go build \
+    otelc go build \
         -v \
         -trimpath \
         -tags "$VENATOR_BUILD_TAGS" \
diff --git a/contrib/docker/release.Dockerfile b/contrib/docker/release.Dockerfile
index 3d9df60..1bf7f64 100644
--- a/contrib/docker/release.Dockerfile
+++ b/contrib/docker/release.Dockerfile
@@ -4,6 +4,7 @@ RUN apk add --no-cache git musl-dev ca-certificates bash
 COPY build.sh .
 COPY go.mod .
 COPY go.sum .
+RUN go install go.opentelemetry.io/otelc/tool/cmd/otelc@latest
 RUN go mod download
 COPY utils utils
 COPY libvenator libvenator
```
3. Attempt to build the project with CTI in place, either by running `./build.sh` locally (with the CTI tool already installed) or in a container, like with `podman build . -f contrib/docker/release.Dockerfile -t venator-otel-cti`
### Expected behavior

The project would successfully build with compile-time instrumentation in place.

[...]

PR #944, opened 2026-08-04
fix(tool): forward trimpath when resolving import archives

ADITYA-CODE-SOURCE:
## Description

Forward `-trimpath` to the nested `go list -export` calls used to resolve archives for imports introduced by instrumentation. This keeps injected archives in the same build context as the outer `go build`.

Without the flag, an instrumented dependency can be compiled with `-trimpath` while an injected transitive archive is resolved without it. The linker then rejects the pair because their export fingerprints differ.

The database/sql integration build now uses `-trimpath` so this archive-resolution path is covered.

## Reproduction

I reproduced the `pkg/runtime` fingerprint mismatch reported at https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/922 on current `main` with Go 1.26.5 on macOS/arm64, in addition to the reported Linux reproductions. Forwarding `-trimpath` removes that fingerprint mismatch.

The full Venator build then reaches a separate duplicate-symbol failure involving log instrumentation. That independent multi-target issue remains outside this PR.

## Testing

- `go test ./internal/setup -run TestExtractBuildFlags`
- `go -C test test -tags integration -run ^TestDBClient -count=1 ./integration`
- `make all` was attempted; build and formatting completed, but full lint is currently blocked by pre-existing `goconst`, `modernize`, and `sloglint` findings on current `main`. A prior full integration invocation also exposed the existing Mongo v1/v2 shared-app lock race; the focused database/sql integration test passes.


darccio:
@ADITYA-CODE-SOURCE Check the linting issues, please.

ADITYA-CODE-SOURCE:
> @ADITYA-CODE-SOURCE Check the linting issues, please.

Thanks for catching that. I applied the formatter, and the lint check is passing now.

tool/internal/setup/setup.go
```diff
@@ -444,14 +444,17 @@ var buildContextFlagsWithValue = map[string]bool{
...
 var buildContextBoolFlags = map[string]bool{
-	"-race":  true, // Race detector
-	"-msan":  true, // Memory sanitizer
-	"-cover": true, // Coverage
-	"-asan":  true, // Address sanitizer
+	"-race":      true, // Race detector
+	"-msan":      true, // Memory sanitizer
+	"-cover":     true, // Coverage
+	"-asan":      true, // Address sanitizer
+	trimpathFlag: true, // Remove file system paths from compiled archives
```
darccio: As we don't have constants for other flags, I'd suggest to follow the same pattern:

```suggestion
// buildContextBoolFlags are go build boolean flags that affect the build context.
//
//nolint:gochecknoglobals // private lookup table
var buildContextBoolFlags = map[string]bool{
	"-race":      true, // Race detector
	"-msan":      true, // Memory sanitizer
	"-cover":     true, // Coverage
	"-asan":      true, // Address sanitizer
	"-trimpath": true, // Remove file system paths from compiled archives
```

tool/internal/setup/setup_test.go
```diff
@@ -410,6 +410,16 @@ func TestExtractBuildFlags(t *testing.T) {
 			args:     []string{"build", "-race", "./..."},
 			expected: []string{"-race"},
 		},
+		{
+			name:     "trimpath flag",
+			args:     []string{"build", trimpathFlag, "./..."},
+			expected: []string{trimpathFlag},
```
darccio: Related to my previous comment:

```suggestion
			expected: []string{"-trimpath"},
```

darccio (approved):
Approving with two small edits.

ADITYA-CODE-SOURCE:
@darccio Done I updated both places to use  -trimpath

darccio (approved):

PR #956, opened 2026-08-05
fix(ci): propagate unit and E2E coverage job failures

### LINKED ISSUE #955: bug(ci): required unit and E2E checks ignore coverage job failures
### Description

The required `Done (Unit Tests)` and `Done (E2E Tests)` jobs depend on their respective coverage jobs, but only inspect the platform test matrix result.

For example, `Done (Unit Tests)` declares:

```yaml
needs: [changes, test-unit-coverage, test-unit]
but checks only:
result="${{ needs.test-unit.result }}"
The E2E workflow follows the same pattern. As a result, an execution failure in a coverage job can leave the required aggregate check green when the platform tests pass.
This includes failures while:
- running make test-unit/coverage or make test-e2e/coverage;
- setting up the coverage job;
- uploading coverage results.
The Integration workflow already handles this correctly by checking change detection, the coverage job, and the platform matrix result.
Expected behavior
The Unit and E2E aggregate checks should:
- fail if change detection fails;
- succeed when no relevant code changed and test jobs were intentionally skipped;
- require both the coverage job and platform matrix to succeed when relevant code changed;
- fail on missing or unexpected change-detection output.
This should follow the existing Done (Integration Tests) implementation.
Scope
Update:
- .github/workflows/test-unit.yaml
- .github/workflows/test-e2e.yaml
This change should not make the informational 70% Codecov threshold blocking. It should only propagate actual coverage-job execution failures to the required aggregate checks.

### PR DESCRIPTION
**Pull request by ADITYA-CODE-SOURCE**
## Description

The required `Done (Unit Tests)` and `Done (E2E Tests)` jobs depend on their respective coverage jobs, but only inspect the platform test matrix result. An execution failure in a coverage job can therefore leave the required aggregate check green when the platform tests pass.

This updates both aggregate checks to follow the existing `Done (Integration Tests)` pattern:

- fail if change detection fails;
- succeed when no relevant code changed and test jobs were intentionally skipped;
- require both the coverage job and platform matrix to succeed when relevant code changed;
- fail on missing or unexpected change-detection output.

This does not make the informational 70% Codecov threshold blocking. It only propagates actual coverage-job execution failures to the required aggregate checks.

Fixes #955

## Testing

- `make format/yaml`
- `make lint/yaml`
- `make lint/action`

### CODE REVIEW

File: .github/workflows/test-unit.yaml
Comment (darccio) [inline .github/workflows/test-unit.yaml:70]: To really close #955, I'd do some more changes.

`codecov-action` defaults `fail_ci_if_error` to `false`, so a failed upload here (network blip, Codecov outage) won't turn this step or the `test-unit-coverage` job red. That leaves one of the three failure modes issue #955 lists — "uploading coverage results" — still invisible to the `Done (Unit Tests)` gate this PR just fixed for the other two modes.

Setting this explicitly closes that gap. One trade-off worth knowing: this job uploads tokenlessly (no `CODECOV_TOKEN`), so it shares Codecov's public rate-limit pool — a transient rate-limit hit would now also block the required check, not just a genuine upload failure.

```suggestion
        with:
          files: coverage-tool.txt
          disable_search: true
          flags: tool
          fail_ci_if_error: true
```

darccio, line 78: Same gap as the tool-coverage upload step above — `fail_ci_if_error` defaults to `false`, so a failed upload here doesn't fail `test-unit-coverage`, leaving the "uploading coverage results" failure mode from #955 unaddressed for pkg coverage too.

```suggestion
        with:
          files: coverage-pkg.txt
          disable_search: true
          flags: pkg
          fail_ci_if_error: true
```

.github/workflows/test-e2e.yaml
darccio: Same gap in the E2E coverage job: `fail_ci_if_error` defaults to `false` here too, so a failed upload leaves `test-e2e-coverage` green and the `Done (E2E Tests)` gate can't catch it even after this PR's fix. This step currently has no `with:` block at all.

```suggestion
      - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7  # v5.5.1
        with:
          fail_ci_if_error: true
```

darccio (changes requested):

ADITYA-CODE-SOURCE:
Updated the Unit and E2E uploads to the immutable Codecov v5.5.5 SHA. The failed E2E job was caused by v5.5.1 using the retired Codecov Keybase signing-key location, so its CLI signature verification could not import the public key. v5.5.5 contains Codecov’s corrected key location. `fail_ci_if_error: true` remains enabled for all three uploads. Please re-review when convenient.

darccio (approved):

PR #957, opened 2026-08-05
fix(tool/internal/ast): avoid panic in typename.matches for unsupported AST node types

### LINKED ISSUE #954: bug(tool/internal/ast): typename.matches panics via util.Unimplemented on slice, map, chan, and func AST nodes
### Description
In `tool/internal/ast/typename.go`, `parsedTypeName.matches` evaluates a `dst.Expr` node to check if it matches a target type string. When encountering composite type AST nodes not explicitly handled in the `switch` statement—such as `*dst.ArrayType` (e.g. `[]byte`), `*dst.MapType`, `*dst.ChanType`, or `*dst.FuncType`—the function executes the `default:` branch and calls `util.Unimplemented(...)`.

`util.Unimplemented` triggers `ex.Fatalf(...)` which panics and terminates the process immediately. As a result, running `otelc` on functions with slice, map, channel, or callback parameters causes `otelc` to crash during compilation.

### Expected Behavior
`typename.matches` should return `false` when an AST type expression does not match the target type, without panicking or aborting compilation.

### Proposed Fix
Modify the `default` case in `typename.matches` to return `false` gracefully:

```go
default:
    // Unsupported AST node types (chan, func, map, slice, array, etc.) cannot match simple type-name filters.
    return false

vedant21-oss:
Fixes #954

### Summary
Fixes `parsedTypeName.matches` in `tool/internal/ast/typename.go` to return `false` instead of calling `util.Unimplemented(...)` (which panics) when evaluating unsupported AST node types like `dst.ArrayType`, `dst.MapType`, `dst.ChanType`, and `dst.FuncType`.

### Changes
- Removed call to `util.Unimplemented(...)` in default branch of `parsedTypeName.matches`.
- - Added unit tests covering unsupported AST types (`ArrayType`, `MapType`, `ChanType`, `FuncType`) in `typename_test.go` to verify they return `false` without panicking.

### DIFF
--- a/tool/internal/ast/typename.go
+++ b/tool/internal/ast/typename.go
@@ -4,13 +4,11 @@
 package ast
 
 import (
-	"fmt"
 	"regexp"
 
 	"github.com/dave/dst"
 
 	"go.opentelemetry.io/otelc/tool/ex"
-	"go.opentelemetry.io/otelc/tool/util"
 )
 
 // typeNameRe parses type-name strings of the form [*][pkg.]Name.
@@ -69,7 +67,6 @@ func (t parsedTypeName) matches(node dst.Expr) bool {
 	default:
 		// Unsupported AST node types (chan, func, map, slice, array, interface
 		// literals) cannot be matched by type-name filters.
-		util.Unimplemented(fmt.Sprintf("signature filter: unsupported type node %T", node))
 		return false
 	}
 }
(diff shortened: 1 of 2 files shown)


darccio (approved):
I think it's a good idea to reduce panicky behaviour.

vedant21-oss:
Thanks @darccio for the review and merge!

PR #967, opened 2026-08-05
fix(tool/internal/setup): add mutex to StateManager to prevent concurrent map access data race

### LINKED ISSUE #962: bug(tool/internal/setup): add mutex to StateManager to prevent concurrent map access data race
### Description
`StateManager` in `tool/internal/setup/state.go` tracks file states during build setup and cleanup phases. While previously marked as not safe for concurrent use in comments, `StateManager` instances are attached to `context.Context` (via `ContextWithStateManager`) and passed across potentially parallel execution flows. Concurrent calls to `s.Track()` mutate `s.files` map without synchronization, leading to potential data races and runtime map concurrent write panics.

### Proposed Solution
Add a `sync.Mutex` field to `StateManager` and synchronize all accesses to `s.files` in `Track()`, `Commit()`, `Discard()`, and `Revert()`.

### Location
- `tool/internal/setup/state.go:33-42`

vedant21-oss:
Fixes #962 
### Description
`StateManager` in `tool/internal/setup/state.go` tracks file states during build setup and cleanup phases. While previously documented as not safe for concurrent use, `StateManager` instances can be attached to `context.Context` (via `ContextWithStateManager`) and passed across concurrent execution flows. Simultaneous calls to `s.Track()` mutate the `s.files` map without synchronization, leading to potential data races and runtime map concurrent write panics.

### Changes
- Added a `sync.Mutex` field (`mu`) to `StateManager`.
- Synchronized all accesses to `s.files` across `Track()`, `Commit()`, `Discard()`, and `Revert()`.
- Added helper `commitLocked()` to handle internal commit calls during tracking without deadlock/double locking.
- Added concurrent unit test `TestStateManagerConcurrentTrack` in `state_test.go`.

### DIFF
--- a/tool/internal/setup/state.go
+++ b/tool/internal/setup/state.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 
 	"go.opentelemetry.io/otelc/tool/ex"
 	"go.opentelemetry.io/otelc/tool/util"
@@ -29,8 +30,9 @@ const (
 // Files that do not exist when tracked are recorded and removed during Revert if
 // they are later created.
 //
-// StateManager is not safe for concurrent use.
+// StateManager is safe for concurrent use.
 type StateManager struct {
+	mu    sync.Mutex
 	files map[string]bool // true = existed when tracked
 }
 
@@ -167,24 +169,32 @@ func (s *StateManager) Track(path string) error {
 	}
 
 	abs = filepath.Clean(abs)
+
+	s.mu.Lock()
 	if _, ok := s.files[abs]; ok {
+		s.mu.Unlock()
 		return nil
 	}
 
 	// If the file doesn't exist, mark it for removal
 	if !util.PathExists(abs) {
 		s.files[abs] = false
-		return s.Commit()
+		err = s.commitLocked()
+		s.mu.Unlock()
+		return err
 	}
 
 	// If the file exists, snapshot it
 	dst := filepath.Join(util.GetBuildTemp(stateDir), stateSnapshotPath(abs))
 	if err = util.CopyFile(abs, dst); err != nil {
+		s.mu.Unlock()
 		return ex.Wrapf(err, "failed to snapshot %s", abs)
 	}
 
 	s.files[abs] = true
-	return s.Commit()
+	err = s.commitLocked()
+	s.mu.Unlock()
+	return err
 }
 
 // Commit persists the tracked state to disk so it can be restored by a future
@@ -194,6 +204,12 @@ func (s *StateManager) Track(path string) error {
 // Track calls Commit automatically; calling it directly is only needed to
 // re-persist after external changes.
 func (s *StateManager) Commit() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.commitLocked()
+}
+
+func (s *StateManager) commitLocked() error {
 	if len(s.files) == 0 {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


anand-242003:
I traced the current call sites for Track()/TrackAll() (AutoPin() and addDeps()), and they appear to execute sequentially. I also noticed StateManager is stored in the context, but I haven't yet identified a code path where the same instance is accessed concurrently. Could you point to the current execution path (or race detector output) that demonstrates the race, or is this intended as a defensive hardening change for future concurrent use?

vedant21-oss:
Hi @anand-242003,

Thanks for taking a look! You're spot on — the current call sites in `AutoPin()` and `addDeps()` execute sequentially, so there isn't an active race in `main` right now.

This change is mainly intended as defensive hardening. Since `StateManager` is attached to `context.Context` via `ContextWithStateManager`, it gets passed around across execution flows. Go contexts are commonly shared across goroutines, so keeping an unsynchronized `map` inside it could easily lead to `concurrent map write` panics if any future setup steps or package loading run concurrently.

Also, I've just pushed a fix to resolve the linter formatting issue that failed in CI earlier. All checks are passing now!

dobbydobap:
I checked whether TestStateManagerConcurrentTrack actually catches the bug, since
@anand-242003's question above means there is no live race to reproduce.

It does. Against main without the mutex, adapting the test only to drop the
s.mu lines, it failed 9 of 10 runs with no -race flag, and 3 of 3 with it. So
this is a genuine regression gate and not just coverage.

Two things I noticed while doing that:

- The tool unit target runs `go test -json -v -shuffle=on -timeout=5m -count=1
  ./tool/...` with no -race. Only the demo app Makefiles use it. So the gate here
  is the runtime's best-effort concurrent map check, which is what let 1 of my 10
  runs through.
- Track holds the mutex across the snapshot copy and the manifest write in
  commitLocked. That is correct, and Commit-per-Track was already the behaviour
  before this PR, so it is not a regression. Just worth knowing the lock now
  serialises that I/O.

I also read through the rest for lock coverage: TrackAll goes through Track, and
the writes in LoadStateManager are on a local instance before it is shared, so
those look fine unlocked.

vedant21-oss:
Hi @dobbydobap,

Thank you so much for taking the time to test and verify TestStateManagerConcurrentTrack against main! I really appreciate you validating that the test reliably catches the race regression both with and without the -race flag. 

Also, thanks for highlighting the lock coverage and I/O serialization details — glad to hear TrackAll and LoadStateManager look solid as well.

Appreciate the thorough review!

amazingakai (changes requested):
Thanks for putting this together.

I don't think we need this right now since there's no actual concurrent usage in `main` (`AutoPin` and `addDeps` are sequential). The struct was documented as unsafe for concurrent use intentionally.

Even if we did merge this, the mutex introduces a huge I/O bottleneck. because the lock wraps the entire `Track()` method, it globally serializes `util.CopyFile` and writing the JSON manifest to disk. If someone actually tried to track 50 files concurrently, they'd be forced to synchronously copy 50 files one-by-one and rewrite the manifest 50 times in a locked queue. 

If we ever actually need to track files concurrently, we should just redesign this to batch the disk writes instead of throwing a mutex over the whole thing. Let's hold off on this until there's an actual use case for it.

darccio:
@amazingakai I agree. I'd suggest closing both the PR and the issue for now.

vedant21-oss:
Thanks @amazingakai and @darccio for the thorough review and architectural context! 

That makes total sense — if concurrent file tracking is introduced in the future, batching disk I/O and manifest writes will be a much cleaner design than serializing through a global mutex. 

Closing this PR for now. Thanks again for taking the time to review!

PR #968, opened 2026-08-05
chore(instrumentation/openai): de-duplicate v2/v3 streaming reader logic

### LINKED ISSUE #958: chore(instrumentation/openai): de-duplicate v1/v2/v3 streaming reader logic
## Context

The OpenAI GenAI streaming instrumentation (`streamingReader` and its helpers) is triplicated verbatim across the three supported module versions:

- `instrumentation/github.com/openai/openai-go/streaming.go`
- `instrumentation/github.com/openai/openai-go/v2/streaming.go`
- `instrumentation/github.com/openai/openai-go/v3/streaming.go`

The three files (and their `streaming_test.go` counterparts) are identical except for the `package` declaration and the `semconv` import path. This predates [PR #828](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/828), which fixed a final-chunk-dropped bug (#827) by adding a `flushRemaining()` method plus two regression tests — and had to hand-apply that same change three times to keep the copies in sync.

## Problem

Every future bug fix or behavior change to the streaming reader has to be manually copy-pasted three times, with nothing enforcing that the copies stay identical. That's easy to get wrong (e.g. forgetting the `v3` copy) and adds reviewer overhead re-verifying three near-identical diffs.

## Suggestion

Investigate extracting the shared `streamingReader` logic (buffering, SSE line parsing, chunk processing) into a common internal package that the v1/v2/v3 packages import, with only the version-specific `semconv` glue left in each version's own package. If the semconv APIs differ enough across versions to make a shared package impractical, consider generating the three copies from a single source (e.g. via `go generate`) instead of hand-maintaining three copies.

## References

- Introduced pattern discussion: review of PR #828
- Related bug affected by this duplication: #827

vizansh:
This de-duplicates the streaming reader implementation between v2 and v3 into a clean shared internal package and closes #958.


Aryanbhargava18:
Hey @vizansh, great initiative. De-duplicating this across v2/v3 is definitely needed to keep the footprint clean.

Just one quick architectural thought: I saw in your latest commit (17ac49d) that you moved streaming_test.go into the internal/streaming package (which is great). However, ParseSSELine is still exported in the source file. Since the test is now in the same package, we can revert that back to the unexported parseSSELine to keep the internal parsing logic properly encapsulated.

vizansh:
Hey @Aryanbhargava18, thanks for the review! 

I've unexported `parseSSELine` in `internal/streaming`, cleaned up package encapsulation, and completed full de-duplication across all streaming readers.

All local unit and integration tests are passing cleanly now! I'm currently monitoring the GitHub Actions CI runs to ensure all remote matrix checks pass as expected.

instrumentation/github.com/openai/openai-go/v3/go.mod
```diff
@@ -70,6 +72,9 @@ require (
 )
 
 replace (
+	go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go/internal/streaming => ../internal/streaming
```
Labreo, line 75: Same issue here as v2/go.mod—since replace directives aren't transitive, target application builds using v3 will fail in CI because Go won't be able to resolve the internal/streaming dependency.

instrumentation/github.com/openai/openai-go/v2/go.mod
```diff
@@ -70,6 +72,9 @@ require (
 )
 
 replace (
+	go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go/internal/streaming => ../internal/streaming
```
Labreo, line 75: Like the root go.mod, this replace will be ignored when building target applications using v2, leading to compilation failures in CI because the compiler won't find the internal/streaming module. Decoupling the modules at the Go dependency level via source-level sync/generation is recommended.

instrumentation/github.com/openai/openai-go/go.mod
```diff
@@ -73,3 +74,5 @@ replace (
 	go.opentelemetry.io/otelc/pkg => ../../../../pkg
 	go.opentelemetry.io/otelc/pkg/runtime => ../../../../pkg/runtime
 )
+
+replace go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go/internal/streaming => ./internal/streaming
```
Labreo, line 78: Making internal/streaming a separate Go module introduces a transitive dependency issue that is currently breaking the integration tests in CI.

Because Go ignores replace directives in dependencies, when a test application (like openaiclient) imports openai-go, the compiler tries to resolve this internal/streaming module path. Since otelc's syncDeps only adds replace directives for direct leaf modules, internal/streaming is not replaced in the test application's environment, causing the build to fail with a module resolution error.

To fix this without complex replace tracking inside otelc, we should consider keeping the code shared via file copying/generation (e.g. keeping internal/streaming under openai-go and copying it to v2/ and v3/ using a Makefile task or check script).

Labreo (commented):
I think the CI failures are happening because `internal/streaming` was set up as a separate Go module. Since Go ignores transitive `replace` directives, `otelc`'s `syncDeps` doesn't replace it in the test apps, causing a module resolution failure.
Maybe we could keep the package local and sync the files via a Makefile check script (similar to `check-api-sync`) to avoid the cross-module dependency. I've left a few inline notes in the diff. Let me know what you think!Other than that mostly everything else looks fine.

instrumentation/github.com/openai/openai-go/go.mod
vizansh: Hi @Labreo,

Thanks a lot for taking the time to review and for explaining the transitive `replace` limitation in Go modules—that's really helpful context!

As noted in issue #958, the maintainers assigned this refactor to @om7057 via PR #990. I left this PR open temporarily for reference, but I won't be pushing further updates here so we don't duplicate effort. 

Appreciate the thorough feedback!

dobbydobap:
Heads up in case you have not seen it: #990 merged on 11 Aug and extracted the
streaming reader for v1, v2 and v3 into a shared internal/streaming module, with
a thin streaming_bridge.go per version.

That looks like it covers what this PR was doing and more, so this one may be
superseded rather than needing a rebase. Worth checking with the maintainers
before spending time on conflicts.

darccio:
Thanks @dobbydobap, you are right. Closing it in favour of #990.

PR #976, opened 2026-08-06
fix(instrumentation): use numeric status for openai error.type

### LINKED ISSUE #965: bug(instrumentation): openai-go middlewares emit full HTTP status line as error.type instead of numeric code
## Description

Per the [OTel HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/), when an HTTP response is received with an error status (4xx/5xx), `error.type` SHOULD be set to the numeric status code as a string (e.g. `"429"`) — it's meant to be a low-cardinality error class, not the full status line.

The OpenAI instrumentation currently sets it to the full status line (e.g. `"429 Too Many Requests"`), which breaks metric/trace grouping in backends that key on `error.type`:

- `instrumentation/github.com/openai/openai-go/middleware.go:175`
- `instrumentation/github.com/openai/openai-go/v2/middleware.go:175`
- `instrumentation/github.com/openai/openai-go/v3/middleware.go:175`

```go
span.SetAttributes(attribute.String("error.type", resp.Status))
```

## Related

This is the same defect fixed for the Anthropic instrumentation in #772. That PR's fix uses `strconv.Itoa(resp.StatusCode)`. Two other instrumentations in this repo already emit `error.type` this way via the official `semconv.ErrorTypeKey` (`go.opentelemetry.io/otel/semconv/v1.37.0`) rather than a raw `attribute.String` or a locally-defined key:

- `instrumentation/net/http/semconv/server.go:270`
- `instrumentation/github.com/linode/linodego/v2/semconv/client.go:129`

## Suggested fix

In each of the three files, replace:
```go
span.SetAttributes(attribute.String("error.type", resp.Status))
```
with:
```go
span.SetAttributes(semconv.ErrorTypeKey.String(strconv.Itoa(resp.StatusCode)))
```
importing `semconv "go.opentelemetry.io/otel/semconv/v1.37.0"` (aliased if it collides with a local `semconv` package import).

Anshumancanrock:
## Summary
openai-go middleware was setting `error.type` to the full HTTP status line (e.g. `"429 Too Many Requests"`). Semconv wants the numeric code as a string (`"429"`) so backends can group on it cleanly.

Updated v1/v2/v3 to use `otelsemconv.ErrorTypeKey.String(strconv.Itoa(resp.StatusCode))`, added a regression assert in `TestOtelMiddleware_HTTPErrorStatus`, and regenerated the bundle.

Fixes #965


ADITYA-CODE-SOURCE (approved):

darccio (approved):
LGTM

darccio:
@Anshumancanrock Regenerate the bundle and it'll be good to go. See #585 about our plan to avoid these conflicts in the short term.

Anshumancanrock:
@darccio done!

PR #977, opened 2026-08-06
test(testutil): increase cross-process span-flush timeout to 15s

### LINKED ISSUE #959: test(e2e): tighten cross-process span-flush budget and evaluate parallelizing E2E tests
## Summary

of the others.

```go
const (
)
```

every new E2E test added.

```go
```

frequent one.

## Suggested approach

singhshresth26:
## Description

This PR increases the `defaultSpanPollTimeout` constant in [test/testutil/readiness.go](file:///d:/github_forked_repo/opentelemetry-go-compile-instrumentation/test/testutil/readiness.go) from `3 * time.Second` to `15 * time.Second`.

## Motivation

During E2E and integration tests, client and server applications are built and run as external OS subprocesses. The telemetry spans they produce are flushed and collected over TCP loopback and standard OS I/O. 

On slower runners (such as Windows-latest in CI, or under heavy CPU load), the telemetry export and retrieval can easily exceed the previous `3 * time.Second` budget, leading to flake/timeout failures. Raising this to `15 * time.Second` gives a safer buffer for subprocess initialization and network/disk write flushes without changing the test logic.

Fixes #959 (Part 1/4)

---

## Checklist

- [x] Tests pass: `make test`
- [x] OpenTelemetry Registry updated (if applicable, see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation))

[^I-know-my-stuff]:


ADITYA-CODE-SOURCE (approved):

darccio (approved):
LGTM

Issue #973, opened 2026-08-05
Our issue and PR intake does not scale


txabman42:
We've started feeling the backlog, let's start discussing ideas. All ideas are welcome.

## The numbers (2026-08-05)

- 72 open issues, 52 open PRs.
- July alone: 85 issues and 193 PRs. June was 15 and 52.
- 38% of open issues carry no label at all.
- 84% of issues opened since June 1 have a PR from the same author within 48 hours (112 of 133).
- 1 issue in this repo's entire history has ever been closed as a duplicate.

## How I read this

- Issues are not a backlog/roadmap we work from.
- Duplicates may not be where the cost is.
- Review capacity is the constraint, not contributor supply.

## Root problems, as I see them

1. **Nothing validates work before it is written.** CONTRIBUTING does not ask for an accepted issue before a PR, and does not define how work is claimed. Contributors are behaving exactly as our process rewards them to, and it would be unfair to frame this as their fault.
2. **We cannot slice the backlog.** There is no component axis in our labels, so "what is already open on kafka?" has no answer.
3. **We have no way to say no early.** By the time we can tell that work is not wanted, someone has already spent their evening on it and we have already spent a review.

txabman42:
Our review capacity is roughly what it should be for a project this size. What is missing is anything that shapes inbound to fit it. The tracker accepts everything, so we process everything, and we do that processing after the code is already written. That is the expensive order to do it in.

## My initial proposal

1. **Close the front door.** Set `blank_issues_enabled: false` and fix the two template labels that do not exist so new issues stop arriving unlabeled. Any other template missing?
2. **Write the intake contract into CONTRIBUTING.** Propose in an issue before writing code, claim by getting assigned, keep PRs small and expect to be asked to split large ones.
3. **Only work on issues a maintainer has marked ready.** Key point.  It moves discussion onto the issue, where it is cheap, and it ties PR volume to our tagging rate instead of to contributor supply. We remove the PR review pressure. However, this would force us to tag issues regularly, since triage becomes the new queue. 
4. **Enforce it gently.** Nobody's PR gets closed for arriving without a tagged issue. Someone fixing a real bug or adding instrumentation we want should not be turned away on procedure. What changes is what they can expect: we triage the underlying issue, tell them honestly that review may be slow, and then either let the PR run or park it until the issue is picked up. The tag decides review order, not whether the work is welcome. Ideally we should only accept PRs tagged as bugs without issue.
[...]

PR #979, opened 2026-08-06
feat(genai): record span errors and finish_reasons on premature SSE stream aborts

### LINKED ISSUE #970: feat(genai): record span errors and finish_reasons on premature SSE stream aborts
Hey, while testing the `openai-go` middleware from #740, I noticed a telemetry gap in how we handle abnormal stream terminations in `streamingReader`.

Currently, if an SSE stream drops mid-generation due to a network timeout (`Read` returns a non-EOF error) or client cancellation (`Close` is called before `[DONE]`), the `finalize` logic is triggered. However, `finalize` simply calls `r.span.End()` without ever recording the error or setting the span status to `codes.Error`. 

To the OTel backend, a network crash halfway through a GenAI stream looks like a perfectly successful 200 OK trace, just with a lower token count. We are silently swallowing telemetry failures for aborted streams.

I'd like to update `streamingReader` to properly detect and record abnormal terminations:
1. **In `Read()` and `Close()`:** Pass the terminating `error` (if any) to `finalize()`.
2. **In `finalize()`:** If the stream was aborted prematurely (e.g., `[DONE]` was never parsed), automatically call `r.span.SetStatus(codes.Error, "stream aborted")`.
3. **Semantic Fallback:** If the connection dropped before we received the final chunk, inject `gen_ai.response.finish_reasons = ["error"]` to align with the GenAI semantic conventions for failed generations.
4. **Error Recording:** Call `r.span.RecordError(err)` if a specific network/context error triggered the teardown.

Relying on HTTP status codes. But since SSE streams usually return a `200 OK` header phase before the streaming chunk loop begins, standard HTTP instrumentation can't detect a teardown that happens 30 seconds later during the body read. The error state must be managed by the chunk reader.

**Additional context**
Ran into this while mapping out the state-machine requirements for the `genai.Observer` architecture (ADR 0007). I'm happy to push a targeted PR to patch this in the OpenAI instrumentation now, or bundle it into the upcoming mentorship timeline as a broader safety net for the generic adapter. Let me know what fits the roadmap best!

theycallmeloki:
feat(genai): record span errors and finish_reasons on premature SSE stream aborts

Fixes #970

Aborted OpenAI streaming responses (mid-stream network errors, client
cancellation before `[DONE]`) previously ended the span as a successful 200
with no error recorded. `streamingReader` now tracks stream completion
(finish_reason or `[DONE]` parsed) and, when a stream is torn down before
completion, marks the span errored with description `"stream aborted"`,
records the terminating error as an exception event, and emits
`finish_reasons=["error"]` to align with GenAI semantic conventions.

Design note: completion is defined as a parsed finish_reason OR the `[DONE]`
marker, not `[DONE]` alone — a caller that stops reading after the final
chunk (or a connection closed right after the last chunk) must not be
flagged as an error. The existing `CloseWithoutDrainingFlushesFinalChunk`
and `FinalChunkWithoutTrailingNewline` tests pin this behavior, and
`Close` still flushes the line buffer so the final chunk's attributes are
never lost.

Tests added:
- `TestStreamingReader_AbortOnReadError`
- `TestStreamingReader_AbortOnCloseBeforeCompletion`
- `TestStreamingReader_CloseAfterFinishReasonIsNormal`

All existing `TestStreamingReader_*` tests continue to pass.

Followup fixes for after #990 : 

fix now in the shared `internal/streaming` package covering v1/v2/v3, superseding the v3-only approach, resolving the conflict with #990's dedup


theycallmeloki:
@Aryanbhargava18 — thanks for the detailed review, and you're right on the race condition.

The hole: `completed := r.completed.Load() || len(r.reasons) > 0` treated a parsed finish_reason as completion even when the stream terminated with a transport error — so an `io.ErrUnexpectedEOF` after the last chunk but before `[DONE]` would have been recorded as a successful 200 while the app surfaced the error. Telemetry has to match application reality.

**Fix (pushed, commit `2fd8d93`):** the terminating error now takes precedence. `finalize` marks the span errored whenever the stream ends with a non-EOF error, regardless of whether a finish_reason was already parsed. Clean EOF or `Close()` after a finish_reason stays a normal completion, so the existing `FinalChunkWithoutTrailingNewline` / `CloseWithoutDrainingFlushesFinalChunk` tests still hold. Added `TestStreamingReader_TruncatedAfterFinishReasonIsError` covering your exact scenario (real finish reason preserved, error status + exception event recorded).

**On sequencing** — agreed. #970 is untriaged, and `streaming.go` is contested by #740 and the Observer design in #892. I don't want to add to the review queue against the direction in #973, so I'm leaving the sequencing to the maintainers: happy to (a) hold this as a draft until they decide, (b) rebase onto #740 when it lands, or (c) fold the abort semantics into the genai.Observer architecture if that's where they want it.

darccio:
@theycallmeloki Please fix the conflicts.

theycallmeloki:
@darccio the dedup moved streaming into `internal/streaming` shared by all versions, I'll port the abort logic there so the fix covers v1/v2/v3

darccio (approved):
LGTM

PR #990, opened 2026-08-06
chore(instrumentation/openai): de-duplicate v1/v2/v3 streaming reader logic

### LINKED ISSUE #958: chore(instrumentation/openai): de-duplicate v1/v2/v3 streaming reader logic

om7057:
## Description

`streamingReader` and its helpers (`streaming.go`, `streaming_test.go`) were triplicated verbatim across the v1, v2, and v3 openai-go instrumentation packages. Confirmed none of the three copies actually depend on the versioned `openai-go` SDK, so extracts the shared logic into `instrumentation/github.com/openai/openai-go/internal/streaming`, a new module that all three now `require` and locally `replace`.

Each version's own package keeps a thin `streaming_bridge.go` (~30 lines) that adapts its local `operationType` into the shared package's `OperationType` and delegates construction there.

## Design notes

- The shared package defines its own small set of GenAI attribute keys (`attrs.go`) rather than importing any version's public `semconv` package. Importing v1's `semconv` from the shared package would create a module dependency cycle, since v1 already depends on the shared package for `streamingReader`.
- `middleware_integration_test.go` in each version still needs the `assertAttribute`/`assertInt64Attribute`/etc. helpers that used to live in `streaming_test.go`. Moved those into a small `testhelpers_test.go` per version rather than trying to share test-only helpers across the module boundary.
- Location was discussed and confirmed in #958: `internal/streaming` under this instrumentation's own tree, not `pkg/` (which currently holds only engine-generic internals, not instrumentation-specific code).

Fixes #958

---

## Checklist

- [x] Tests pass: `make test`


om7057:
CC: @darccio

tool/internal/setup/sync.go
```diff
@@ -117,6 +119,41 @@ func warnVersion(ctx context.Context, goModPath string, before versionSnapshot)
...
+		if d.IsDir() || d.Name() != "go.mod" || path == topGoMod {
+			return nil
+		}
+
+		modFile, parseErr := parseGoMod(path)
+		if parseErr != nil {
+			return ex.Wrapf(parseErr, "loading %s", path)
+		}
+
+		nested[modFile.Module.Mod.Path] = filepath.Dir(path)
```
ADITYA-CODE-SOURCE, line 154: Could we check that modFile.Module is non-nil before using it?  modfile.Parse accepts a go.mod without a module directive  which would make this path panic instead of returning an error  It would be good to add a focused test for a missing module directive as well.

tool/internal/setup/sync.go
```diff
@@ -140,6 +177,21 @@ func syncDeps(ctx context.Context, modPaths map[string]bool, moduleDir string) e
...
 
+	// Some matched instrumentation modules have their own nested local
+	// modules (see discoverNestedModuleReplaces). Add replace directives for
+	// those too, so "go mod tidy" below doesn't try to fetch them remotely.
+	matchedDirs := make([]string, 0, len(replaces))
+	for _, dir := range replaces {
+		matchedDirs = append(matchedDirs, dir)
+	}
+	for _, dir := range matchedDirs {
+		nested, nestedErr := discoverNestedModuleReplaces(dir)
```
ADITYA-CODE-SOURCE, line 195: Since this adds generic nested-module handling for every matched instrumentation  could we add a syncDeps test covering the new behavior?   The test should verify that the nested helper receives a local replacement and that an existing user replacement for it is preserved.

ADITYA-CODE-SOURCE (changes requested):

tool/internal/setup/sync.go
om7057: Fixed. `discoverNestedModuleReplaces` now checks `modFile.Module == nil` and returns `ex.Newf(...)` instead of dereferencing it, plus `TestDiscoverNestedModuleReplaces_MissingModuleDirective` covering a nested go.mod without a module directive.

tool/internal/setup/sync.go
om7057: Added `TestSyncDeps_NestedModuleReplace` (fresh replace directive added for a discovered nested module) and `TestSyncDeps_NestedModuleReplace_ExistingReplacePreserved` (an existing user-authored replace for that nested module is left untouched, same pattern as `TestSyncDeps_ExistingReplace`).

om7057:
@ADITYA-CODE-SOURCE, I've resolved the comments and pushed the changes; please review them!

ADITYA-CODE-SOURCE (approved):

ADITYA-CODE-SOURCE:
@om7057  Only CI is still blocking now: `Verify Bundle` and `codecov/patch`
 Could you please fix those ?

om7057:
@darccio just a heads up on this one!

Aryanbhargava18:
Great execution on this! Extracting `internal/streaming` is definitely the right architectural seam to stop the 3-way code duplication across `v1`, `v2`, and `v3`, and adding `discoverNestedModuleReplaces` cleanly solves nested local module resolution for `syncDeps`.

A couple of subtle edge cases in `discoverNestedModuleReplaces` (`tool/internal/setup/sync.go`):

1. **Skipping `testdata` and `vendor` directories during walk:**
   `filepath.WalkDir` currently traverses all subdirectories under `dir`. If an instrumentation module has integration test fixtures in `testdata/` containing mock `go.mod` files, `WalkDir` will parse those test fixtures and inject unwanted `replace` directives into consumer `go.mod` files during `syncDeps`.

   We can prevent this by returning `filepath.SkipDir` when encountering `testdata`, `vendor`, or hidden directories:
   ```go
   if d.IsDir() {
       name := d.Name()
       if name == "testdata" || name == "vendor" || strings.HasPrefix(name, ".") {
           return filepath.SkipDir
       }
       return nil
   }
   ```

2. **Path comparison safety:**
   Comparing `path == topGoMod` directly can occasionally hit path canonicalization differences depending on how `dir` is passed. Wrapping both in `filepath.Clean(path) == filepath.Clean(topGoMod)` keeps the root skip deterministic across platforms.

Looking forward to rebasing our GenAI content capture work onto `internal/streaming` once this lands!

tool/internal/setup/sync.go
```diff
@@ -117,6 +119,41 @@ func warnVersion(ctx context.Context, goModPath string, before versionSnapshot)
...
+// match as an instrumentation import on their own. Without a replace
+// directive for them here, a downstream consumer's "go mod tidy" tries to
+// fetch them from a real module proxy and fails.
+func discoverNestedModuleReplaces(dir string) (map[string]string, error) {
+	nested := make(map[string]string)
+	topGoMod := filepath.Join(dir, "go.mod")
+
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
```
Aryanbhargava18, line 136: To prevent traversing `testdata`, `vendor`, or hidden directories, we can return `filepath.SkipDir` when encountering them:

```suggestion
		if d.IsDir() {
			name := d.Name()
			if name == "testdata" || name == "vendor" || strings.HasPrefix(name, ".") {
				return filepath.SkipDir
			}
			return nil
		}
		if d.Name() != "go.mod" || filepath.Clean(path) == filepath.Clean(topGoMod) {
			return nil
		}
```

om7057: Added TestDiscoverNestedModuleReplaces_SkipsTestdataVendorAndHiddenDirs, which plants a malformed go.mod under each of testdata/, vendor/, and .git/ and asserts none of them get parsed (a malformed file would error if the walk ever reached it, so this proves the directories are actually pruned, not just incidentally ignored).

om7057:
@darccio since you scoped and assigned #958, wanted to flag this is ready for a look. All review feedback so far is resolved and CI's green.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
```diff
@@ -33,19 +44,19 @@ type streamingReader struct {
 	model         string
 	opName        string
 	provider      string
```
darccio: `model`, `opName`, and `provider` are stored but never read anywhere in
`finalize` or elsewhere in this file — only the chunk-derived `responseModel`
is used. This was already dead in the old private `streamingReader`, but the
move promotes it into a newly-exported struct with no external caller yet, so
now is the moment to drop it rather than carry it forward as public API.

```suggestion
```

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
```diff
@@ -56,7 +67,7 @@ func newStreamingReader(
 	}
 }
```
darccio, line 59: Following on from the struct trim above: `model`, `opName`, `provider` become
dead constructor params once the fields are gone, and the trailing
`_ context.Context` is discarded outright — an unused parameter is a smell in
its own right, and a discarded `context.Context` doubly so (idiomatic Go puts
`ctx` first and only when it's actually threaded through). Trimming both
leaves a signature that says what the reader actually needs.

Also drop `"context"` from the import block (line 13) once this param is
gone, and update the three call sites (`streaming_bridge.go` in v1/v2/v3) and
`streaming_test.go` to match — none of them use the return value of `ctx`
either, so this is a mechanical follow-through, not a design change.

```suggestion
func NewStreamingReader(
	body io.ReadCloser,
	span trace.Span,
	start time.Time,
	op OperationType,
) *StreamingReader {
	return &StreamingReader{
		reader: body,
		start:  start,
		span:   span,
		op:     op,
	}
}
```suggestion

File: tool/internal/setup/sync.go [hunk 6]
```diff
@@ -140,6 +184,21 @@ func syncDeps(ctx context.Context, modPaths map[string]bool, moduleDir string) e
...
+		matchedDirs = append(matchedDirs, dir)
+	}
+	for _, dir := range matchedDirs {
+		nested, nestedErr := discoverNestedModuleReplaces(dir)
+		if nestedErr != nil {
+			return ex.Wrapf(nestedErr, "discovering nested modules under %s", dir)
+		}
+		maps.Copy(replaces, nested)
+	}
+
```
darccio, line 218: This only walks *down* from each matched module's own directory, but the
module it's meant to catch here — a package shared between several versioned
copies of one instrumentation — is a *sibling* of v2/v3, not a descendant of
either. Concretely: `openai-go/internal/streaming` sits under v1's own
directory (`openai-go/`), alongside `v2/` and `v3/`, not inside either. A
consumer depending only on the v2 or v3 openai-go SDK matches only that
version's own `otelc.yaml` target
(`matchInstrumentationImports` in pin.go compares exact target strings), so
`modPaths` never includes v1, `replaces` never includes `openai-go/`, and this
walk never reaches `internal/streaming`. `v2/go.mod` and `v3/go.mod` still
`require` it at the placeholder pseudo-version with no replace to resolve it,
so `go mod tidy` in that consumer's build tries to fetch an unpublished module
from a real proxy and fails.

Walking each matched directory's parent as well as the directory itself
catches this shape too, since sibling versions and their shared internal
module live under a common parent in this repo's layout:

```suggestion
	// Some matched instrumentation modules have their own nested local
	// modules (see discoverNestedModuleReplaces) — either inside the matched
	// module's own directory, or as a sibling shared by several versioned
	// copies of one instrumentation (e.g. openai-go's v1/v2/v3 all sharing
	// openai-go/internal/streaming, which sits under v1's directory, a
	// sibling of v2/ and v3/ rather than a descendant of either). Walking
	// each matched directory's parent in addition to the directory itself
	// catches both shapes. The parent walk can also re-discover sibling
	// version directories or unrelated instrumentations under a shared
	// parent; a replace directive for a module the consumer doesn't
	// actually require is harmless, since "go mod tidy" ignores it.
	walkDirs := make(map[string]bool, len(replaces)*2)
	for _, dir := range replaces {
		walkDirs[dir] = true
		walkDirs[filepath.Dir(dir)] = true
	}
	for dir := range walkDirs {
		nested, nestedErr := discoverNestedModuleReplaces(dir)
		if nestedErr != nil {
			return ex.Wrapf(nestedErr, "discovering nested modules under %s", dir)
		}
		maps.Copy(replaces, nested)
	}
```

Worth a regression test mirroring the sibling shape directly — a matched
`v2/` dir with `internal/streaming` living under its parent rather than
under `v2/` itself — since `TestSyncDeps_NestedModule` only covers the
descendant case.

tool/internal/setup/sync.go
```diff
@@ -117,6 +119,48 @@ func warnVersion(ctx context.Context, goModPath string, before versionSnapshot)
...
 }
 
+// discoverNestedModuleReplaces walks dir for go.mod files nested inside it
+// (excluding dir's own go.mod), returning a map of module path to directory.
+// This picks up local helper modules that live inside an instrumentation
+// module's tree but carry no otelc.yaml of their own — e.g. a package shared
+// between several versioned copies of one instrumentation — so they never
+// match as an instrumentation import on their own. Without a replace
+// directive for them here, a downstream consumer's "go mod tidy" tries to
+// fetch them from a real module proxy and fails.
```
darccio, line 129: Worth documenting explicitly (rather than leaving future readers to work it
out): walking a matched directory that nests sibling version copies inside
it — v1's own dir contains `v2/` and `v3/` — already picks up their module
paths as extra, unused replace directives today. With the parent-directory
walk from the Must Fix comment on this file, that widens further to
whatever else shares a parent with the matched directory. Both are
intentional and harmless — `go mod tidy` ignores a replace directive for a
module that isn't in the build list — but that's worth saying in the doc
comment so it doesn't read as an oversight later.

```suggestion
// discoverNestedModuleReplaces walks dir for go.mod files nested inside it
// (excluding dir's own go.mod), returning a map of module path to directory.
// This picks up local helper modules that live inside an instrumentation
// module's tree but carry no otelc.yaml of their own — e.g. a package shared
// between several versioned copies of one instrumentation — so they never
// match as an instrumentation import on their own. Without a replace
// directive for them here, a downstream consumer's "go mod tidy" tries to
// fetch them from a real module proxy and fails.
//
// The walk can pick up modules the consumer never actually requires — e.g.
// a matched v1 directory that nests v2/v3 copies inside it, or, with the
// parent-directory walk in syncDeps, sibling instrumentations under a
// shared parent. That's intentional and harmless: "go mod tidy" ignores a
// replace directive for a module that isn't in the build list.
```suggestion

darccio (changes requested):
Every consumer app that exercises the built `otelc` tool against the openai
instrumentation imports the bare v1 SDK:

- `test/apps/openaiclient/go.mod:5` → `github.com/openai/openai-go v1.12.0`
- `demo/app/openai/go.mod:5` → `github.com/openai/openai-go v1.12.0`

No `go.mod` anywhere outside `instrumentation/` requires `openai-go/v2` or
`/v3`. `syncDeps`/`discoverNestedModuleReplaces` only run when the built
tool processes a real consumer's `go.mod` — inside this repo, `v2/go.mod`'s
`replace` for `internal/streaming` is hand-authored directly to
`../internal/streaming`, so it never exercises this code path at all.

That means nothing in this repo's CI — `make test`, `go build ./...`, or the
e2e suite — would have caught the sibling-discovery gap flagged elsewhere in
this review, and nothing will catch a regression in whatever fix lands for
it either. Once that fix is in, it's worth adding (or pointing) an
integration app at `openai-go/v2` (or `/v3`) so the tool is actually
exercised end-to-end against the shape this PR introduces — a versioned SDK
whose instrumentation depends on a sibling shared module.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
om7057: Dropped both, and trimmed the bridge functions' own signatures in v1/v2/v3 too since those params became fully unused there once they stopped being forwarded (unparam is enabled in this repo's lint config and would've flagged them). Updated the three call sites and the shared package's tests to match.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming.go
om7057: Same fix as above, covered together.

tool/internal/setup/sync.go
om7057: Confirmed and fixed. Added TestSyncDeps_SiblingNestedModule reproducing this exact shape (matched v2 dir, internal/streaming under its parent), watched it fail on the old code, then applied your suggested parent-walk fix. Also verified end to end: built otelc from a clean checkout and ran it against a real app importing only openai-go/v2, confirmed go mod tidy resolves cleanly and the binary emits a correct span.

tool/internal/setup/sync.go
om7057: Added, thanks.

om7057:
@darccio Added test/apps/openaiclientv2 and TestOpenAIClientV2, building and running it through otelc go build against the real v2 SDK. Verified end to end from a clean checkout that it resolves internal/streaming correctly and the built binary emits a correct span, so this exercises the sibling-module shape the fix above addresses.

darccio (approved):

darccio:
@om7057 There seem to be some linting issues and failing integration tests.

om7057:
> @om7057 There seem to be some linting issues and failing integration tests.

@darccio fixed, please check once again.

PR #991, opened 2026-08-06
test: parallelize E2E tests and optimize build caching

### LINKED ISSUE #959: test(e2e): tighten cross-process span-flush budget and evaluate parallelizing E2E tests

singhshresth26:
## Description

This PR parallelizes the E2E test suite by introducing the following changes:
- Adds `t.Parallel()` to all E2E tests (`grpc_test.go`, `grpcserver_dbclient_test.go`, `http_test.go`, `httpserver_dbclient_test.go`, and `kafka_test.go`).
- Dynamically allocates free ports inside the tests to prevent parallel network address collisions.
- Sandboxes build and execution directories by copying application source packages to a unique `app_TestName` directory.
- Introduces a global `buildMu` serialization lock inside the test infrastructure to prevent concurrent compilation commands from overwriting or colliding on shared `net/http` standard library source directories (within the `GOROOT`).
- Enables support for a shared build cache when the `OTELC_TEST_GOCACHE` environment variable is set.

## Motivation

E2E tests are the most expensive test suite since they involve full process compiles and spawns. Originally, serializing them significantly impacted local and CI wall-clock times. This PR addresses that limitation by enabling concurrent execution with dynamic port allocation and sandbox directory isolation, while using a serialization mutex specifically for the compilation step to prevent shared standard library write conflicts.

Fixes #959 Part(2/4)

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


test/testutil/infra.go
```diff
@@ -109,6 +117,17 @@ func Build(t *testing.T, appsDir, app string, args ...string) {
...
 		require.NoError(t, err)
+
+		sanitized := strings.ReplaceAll(t.Name(), "/", "_")
+		tempAppName := app + "_" + sanitized
+		srcDir := filepath.Join(appsDir, app)
+		dstDir := filepath.Join(appsDir, tempAppName)
+
+		require.NoError(t, copyDir(srcDir, dstDir))
+		t.Cleanup(func() {
+			_ = os.RemoveAll(dstDir)
```
ADITYA-CODE-SOURCE: Could we create one sandbox per test and clean it up only after all started processes have stopped? Each Build currently registers directory cleanup independently so cleanup order may remove an app directory while a server from that sandbox is still running, especially on Windows.

test/testutil/infra.go
```diff
@@ -101,6 +107,8 @@ func Build(t *testing.T, appsDir, app string, args ...string) {
 		}
 	}
 
+	env = setEnv(env, "GOWORK", "off")
```
ADITYA-CODE-SOURCE: Should this be limited to copied standard test apps?   Applying GOWORK=off unconditionally also changes behavior for callers using a custom appsDir.

ADITYA-CODE-SOURCE (changes requested):
Could we create one sandbox per test and clean it up only after all started processes have stopped? Each Build currently registers directory cleanup independently, so cleanup order may remove an app directory while a server from that sandbox is still running, especially on Windows.

test/testutil/infra.go
singhshresth26: Agreed. I have scoped setting GOWORK=off (and the local GOCACHE configuration logic) to standard apps only (standardAppsDir == true). Callers using custom appsDir directories are now unaffected.

test/testutil/infra.go
singhshresth26: I have refactored the test sandbox logic. We now initialize a sandbox directory per top-level test at test/.tmp_<TopLevelTestName>.

Keeping this at a flat depth of 2 (direct subdirectories under the .tmp_ parent) ensures that relative Go module replacements (e.g. ../../shared/testdb in the test apps' go.mod) resolve correctly. The entire sandbox directory is registered for cleanup on the first Build call of the test, guaranteeing that in LIFO cleanup execution, it is deleted only after all started processes in the test have stopped and been cleaned up.

darccio:
@singhshresth26 Can you check the failed Ci jobs? Thanks!

test/testutil/infra.go
```diff
@@ -70,24 +78,58 @@ func appsPath() (string, error) {
...
+	pwd, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
+	tName := t.Name()
+	if idx := strings.Index(tName, "/"); idx != -1 {
+		tName = tName[:idx]
+	}
+	sanitized := strings.ReplaceAll(tName, "/", "_")
+	return filepath.Join(pwd, "..", ".tmp_"+sanitized), nil
```
darccio: tName is truncated to the substring before the first "/" on the lines above, so by the time ReplaceAll runs it can no longer contain a "/" — this call is a no-op. Either drop it, or reorder the logic if multi-level subtest names were meant to collapse into multiple underscores.

```suggestion
	tName := t.Name()
	if idx := strings.Index(tName, "/"); idx != -1 {
		tName = tName[:idx]
	}
	return filepath.Join(pwd, "..", ".tmp_"+tName), nil
```

test/testutil/infra.go
```diff
@@ -178,3 +235,50 @@ func Start(t *testing.T, appsDir, app string, env []string, args ...string) *exe
 
 	return cmd
 }
+
+func copyDir(src, dst string) error {
```
darccio, line 248: copyDir is a bespoke recursive copy: entries are read fully into memory via os.ReadFile rather than streamed, symlinks aren't dereferenced or special-cased (entry.Info() on a symlink won't copy the link target), and new directories are always created with 0o755 regardless of the source directory's permissions. None of the current test-app source trees exercise these edges, but if copyDir gets reused for other fixtures later, these gaps are worth knowing about.

test/testutil/infra.go
```diff
@@ -132,12 +185,14 @@ func Run(t *testing.T, appsDir, app string, env []string, args ...string) string
 	if util.IsWindows() {
 		appName += ".exe"
 	}
+	var appDir string
 	if appsDir == "" {
-		var err error
-		appsDir, err = appsPath()
+		sandboxDir, err := sandboxPath(t)
 		require.NoError(t, err)
+		appDir = filepath.Join(sandboxDir, app)
```
darccio, line 195: Run resolves the sandbox dir independently of Build and assumes a prior Build call already populated it for this same t — there's no check that Build ran first. Start has the identical assumption. Today's usage always pairs BuildAndStart/BuildAndRun so this holds, but if a test ever calls Run standalone for a standard app, the failure surfaces as a generic exec "no such file or directory" from the missing binary rather than a clear message about sandbox setup.

test/testutil/infra.go
```diff
@@ -105,12 +147,23 @@ func Build(t *testing.T, appsDir, app string, args ...string) {
...
+		initSandbox(t, sandboxDir)
+
 		appsDir, err = appsPath()
 		require.NoError(t, err)
+
+		srcDir := filepath.Join(appsDir, app)
+		dstDir := filepath.Join(sandboxDir, app)
+
+		require.NoError(t, copyDir(srcDir, dstDir))
+		appDir = dstDir
```
darccio, line 166: copyDir runs unconditionally on every Build call, even if the same app was already copied into this sandbox by an earlier Build in the same test. initSandbox dedups the cleanup registration but not the copy itself, so a repeat build for the same app re-copies the full source tree. Harmless today since no test does this, but worth a comment or a skip-if-exists check if that usage pattern shows up.

test/testutil/infra.go
```diff
@@ -70,24 +78,58 @@ func appsPath() (string, error) {
...
+
+func initSandbox(t *testing.T, sandboxDir string) {
+	sandboxMu.Lock()
+	defer sandboxMu.Unlock()
+
+	if sandboxes[sandboxDir] {
+		return
+	}
+
+	t.Cleanup(func() {
```
darccio, line 104: Sandbox dirs are only removed via t.Cleanup, so a killed process (CI timeout, OOM, SIGKILL) leaves test/.tmp_* behind. It's gitignored so this doesn't pollute the repo, just accumulates disk across repeated interrupted runs — consistent with how .otelc-build is already handled elsewhere, so likely an accepted tradeoff rather than a gap specific to this PR.

test/testutil/infra.go
```diff
@@ -9,13 +9,21 @@ import (
...
 	"github.com/stretchr/testify/require"
 
 	"go.opentelemetry.io/otelc/tool/util"
 )
 
+var (
+	buildMu   sync.Mutex
+	sandboxMu sync.Mutex
+	sandboxes = make(map[string]bool)
+)
```
darccio, line 28: No comment here explains why buildMu serializes every Build call. The PR description says it's to prevent concurrent compiles from colliding on "shared net/http standard library source directories (within GOROOT)", but there are zero GOROOT references anywhere in tool/ — instrumentation runs via -toolexec and writes modified sources into per-build temp dirs (.otelc-build / compiler tmpdir), never into GOROOT. There's also already a per-work-dir advisory lock (tool/internal/setup/lock.go, .otelc-build.lock) that serializes runs mutating the same module dir, and with this PR's sandboxing every test now has its own module dir — so that lock no longer needs an external backstop for the reason given.

If buildMu is actually a resource throttle (bounding concurrent full `go build` compiles on CI runners) rather than a correctness fix, that's a reasonable thing to want, but it should be named as such — and it's worth knowing, since it means the expensive part of every test (the build) still queues one-at-a-time, so the parallelization win is limited to overlapping the non-build phases rather than free concurrent compiles. Can you confirm the actual hazard this protects against?

test/testutil/infra.go
```diff
@@ -70,24 +78,58 @@ func appsPath() (string, error) {
...
+		sandboxMu.Unlock()
+	})
+	sandboxes[sandboxDir] = true
+}
+
 // Build builds the application with the instrumentation tool. The built binary
 // is registered for cleanup via t.Cleanup. Standard test apps use
 // OTELC_TEST_GOCACHE/<app> as GOCACHE when OTELC_TEST_GOCACHE is set, and drop
 // -a so warm builds can reuse that per-app cache. Custom appsDir builds keep
 // the caller's arguments and environment unchanged.
```
darccio: This doc comment and the inline comment at lines 134-135 both describe the pre-PR per-app cache path (OTELC_TEST_GOCACHE/<app>). The code below now points GOCACHE directly at the shared cacheRoot, so builds across different apps — and different instrumentation configs — share one cache. Go's build cache is content-addressed so this should be safe, but both comments need updating to match: the inline one currently says "reuse this app-local cache without sharing compiled objects across different instrumentation configs," which asserts the opposite of what the code now does.

```suggestion
// Build builds the application with the instrumentation tool. The built binary
// is registered for cleanup via t.Cleanup. Standard test apps use
// OTELC_TEST_GOCACHE as a shared GOCACHE when OTELC_TEST_GOCACHE is set, and
// drop -a so warm builds can reuse that shared cache. Custom appsDir builds
// keep the caller's arguments and environment unchanged.
```

test/testutil/infra.go
```diff
@@ -70,24 +78,58 @@ func appsPath() (string, error) {
...
+	defer buildMu.Unlock()
+
 	otelc, err := OtelcPath()
 	require.NoError(t, err)
 
 	standardAppsDir := appsDir == ""
-	var env []string
+	env := os.Environ()
 	if standardAppsDir {
+		env = setEnv(env, "GOWORK", "off")
```
darccio: No go.work file exists anywhere in the repo today, so this is currently a no-op. Presumably defensive against a contributor's local go.work leaking into the sandboxed build (the sandbox copies source into a new directory that could pick up workspace resolution from a go.work higher up the tree) — reasonable to keep, just flagging that nothing in this PR or CI exercises it, so it's unverified by the test suite itself.

.gitignore
```diff
@@ -41,3 +41,8 @@ profiles
 # Pre-fetched upstream semantic-conventions registry used by weaver to resolve
 # the schemas/otelc/ dependency. Regenerated by `make fetch-upstream-semconv`.
 schemas/otelc/.deps/
+
+# Test build cache and workspaces
+test/testutil/.gocache/
+test/testutil/.tmp/
+test/.tmp_*/
```
darccio, line 47: Two of these three entries don't match anything this PR (or current CI) writes. `test/testutil/.gocache/` and `test/testutil/.tmp/` don't correspond to any path produced by infra.go or the workflows — the actual sandbox dirs are `test/.tmp_<TestName>/` (already covered by the third entry) and the actual shared build cache is `test/.otelc-gocache` (set via `OTELC_TEST_GOCACHE` in `.github/workflows/test-integration.yaml`), which lives one level up from `test/testutil/` and isn't matched by any of these three patterns. Likely leftover from an earlier design where sandboxes/cache lived under `test/testutil/`.

```suggestion
# Test build cache and workspaces
test/.otelc-gocache/
test/.tmp_*/
```

test/testutil/readiness.go
```diff
@@ -14,7 +14,7 @@ import (
 const (
 	defaultReadinessTimeout  = 10 * time.Second
 	defaultReadinessInterval = 100 * time.Millisecond
-	defaultSpanPollTimeout  = 3 * time.Second
+	defaultSpanPollTimeout   = 3 * time.Second
```
darccio: Same conflict is in the other PR.

darccio (changes requested):

darccio:
@singhshresth26 Can you address my feedback on this PR? Thanks!

PR #992, opened 2026-08-06
test(database/sql): run dsnparse/semconv unit tests in CI

Anshumancanrock:
## Summary

`make test-unit` was skipping the entire `instrumentation/database/sql` module because `client.go` references fields that only exist after otelc injects them into `database/sql` types. That also meant `dsnparse` and `semconv` never ran in CI, even though they type-check fine on their own.

This wires those subpackages back into the unit + coverage targets, documents the exception, and moves SQL operation-name extraction into `semconv` so it can be unit-tested without compiling the field-injected root package.

## Changes

- `Makefile`: after the normal instrumentation loop, run `go test ./dsnparse/... ./semconv/...` (and the coverage equivalent) for `database/sql`
- `docs/testing.md`: note how field-injected packages should be unit-tested
- `semconv.OperationName`: extracted from `calOp` in `client.go`, with table tests
- refreshed `tool/data/otelc-bundle.tgz` for the `client.go` change


abhishekkamble12:
Hi! I'd like to take this on.

instrumentation/database/sql/semconv/db.go
```diff
@@ -21,6 +22,20 @@ type DatabaseSqlRequest struct {
...
 
+// OperationName returns the uppercased first token of a SQL statement for use
+// as db.operation.name. Empty or whitespace-only input returns "".
+func OperationName(query string) string {
+	trimmed := strings.TrimSpace(query)
+	if trimmed == "" {
+		return ""
+	}
+	fields := strings.Fields(trimmed)
+	if len(fields) == 0 {
```
Labreo, line 33: Very nitpicky but this is kind of redundant at this point as strings.Fields(trimmed) on a non-empty string will always return at least 1 element, making if len(fields) == 0 unreachable. though this is completely safe and harmless.

darccio: Agreed. I'm approving the PR, but this can be removed.

darccio (approved):

darccio:
@abhishekkamble12 Bundle, sorry.

Anshumancanrock:
@darccio done, regenerated the bundle.

PR #993, opened 2026-08-06
test(testutil): harden FreePort against close/rebind race conditions

### LINKED ISSUE #959: test(e2e): tighten cross-process span-flush budget and evaluate parallelizing E2E tests

singhshresth26:
Track recently allocated ports in a synchronized map to prevent parallel test runs from colliding on the same ephemeral port before the subprocess server binds to it.

## Description

This PR introduces a thread-safe registry to prevent concurrent test runs from colliding on duplicate port allocations:
* Adds a global `allocatedPorts` map guarded by `sync.Mutex` in `test/testutil/readiness.go`.
* Automatically sweeps/evicts registry entries older than 30 seconds to keep the memory footprint minimal.
* Hardens the `FreePort` helper to check incoming OS-assigned ephemeral ports against the registry. If a port was recently allocated (within the last 30 seconds), it discards it, sleeps briefly, and retries.

## Motivation

Now that E2E tests are parallelized (following PR #991), multiple test fixtures run concurrently. This heavily increases the likelihood of a race condition where the OS assigns and closes the same ephemeral port to two parallel tests before either subprocess server can bind to it, causing flaky socket collisions and test timeouts. By tracking and temporarily reserving recently returned ports, we eliminate this class of flaky port collisions.

Fixes #959 part(3/4)

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


test/testutil/readiness.go
```diff
@@ -51,14 +52,41 @@ func pollForSpans(c *Collector, minSpans int, timeout time.Duration) bool {
...
+	allocatedPortsMu.Lock()
+	now := time.Now()
+	for p, allocatedAt := range allocatedPorts {
+		if now.Sub(allocatedAt) > 30*time.Second {
+			delete(allocatedPorts, p)
+		}
+	}
+	allocatedPortsMu.Unlock()
+
+	for {
```
Labreo: Shouldn't for {} need a max-retry cap + t.Fatal to prevent an infinite loop, so it can't it hang forever on constrained systems?

Labreo:
Just a suggestion btw.I am not a maintainer just a contributor.Also having a  a quick parallel test asserting no two concurrent FreePort calls return the same port would be nice aswell.

test/testutil/readiness.go
```diff
@@ -51,14 +52,41 @@ func pollForSpans(c *Collector, minSpans int, timeout time.Duration) bool {
...
+		lis, err := net.Listen("tcp", "localhost:0")
+		require.NoError(t, err)
+		port := lis.Addr().(*net.TCPAddr).Port
+		require.NoError(t, lis.Close())
+
+		allocatedPortsMu.Lock()
+		if _, exists := allocatedPorts[port]; !exists {
```
ADITYA-CODE-SOURCE, line 93: This prevents duplicate values returned by FreePort within this test process, but it does not reserve the port after the listener closes. Another process can still bind before the server starts, so the original close/rebind race remains. Do we have a reproduced failure showing that same-process duplicate allocation is the actual problem?

test/testutil/readiness.go
```diff
@@ -51,14 +52,41 @@ func pollForSpans(c *Collector, minSpans int, timeout time.Duration) bool {
...
+			allocatedPorts[port] = time.Now()
+			allocatedPortsMu.Unlock()
+			return port
+		}
+		allocatedPortsMu.Unlock()
+
+		time.Sleep(10 * time.Millisecond)
```
ADITYA-CODE-SOURCE, line 100: Could this retry be bounded or tied to the test context? As written repeated collisions can keep the helper blocked indefinitely.

ADITYA-CODE-SOURCE (changes requested):

test/testutil/readiness.go
singhshresth26: I've capped the loop attempts at 100, and it will now call t.Fatalf if it fails to allocate a unique free port within that limit to prevent any infinite loops.

test/testutil/readiness.go
singhshresth26: Yes, during parallel test execution on Windows, we hit instances where concurrent test workers/goroutines got assigned the same port immediately after a previous listener closed it but before the actual target server started. The local registry of recently returned ports defends against these same-process concurrent allocation collisions.

test/testutil/readiness.go
singhshresth26: I've updated the retry loop to be bounded (max 100 attempts) and check the test context (t.Context().Done()) on each iteration so it fails fast if the context is cancelled.

ADITYA-CODE-SOURCE (approved):

test/testutil/readiness.go
```diff
@@ -14,7 +15,7 @@ import (
 const (
 	defaultReadinessTimeout  = 10 * time.Second
 	defaultReadinessInterval = 100 * time.Millisecond
-	defaultSpanPollTimeout  = 3 * time.Second
+	defaultSpanPollTimeout   = 3 * time.Second
```
darccio: This is already conflicting after merging #977.

test/testutil/readiness_test.go
```diff
@@ -59,3 +59,27 @@ func TestWaitForSpans_timesOut(t *testing.T) {
...
+		assert.Greater(t, port, 0)
+	})
+
+	t.Run("concurrent", func(t *testing.T) {
+		const numGoroutines = 50
+		ports := make(chan int, numGoroutines)
+		for i := 0; i < numGoroutines; i++ {
+			go func() {
+				ports <- FreePort(t)
+			}()
```
darccio: `FreePort` can call `t.Fatalf`/`require.NoError` (via `FailNow`) from inside this spawned goroutine. Per `testing.T.FailNow` docs: "FailNow must be called from the goroutine running the test... Calling FailNow does not stop those other goroutines." If that path ever fires here, the goroutine `Goexit`s before reaching `ports <- FreePort(t)`. The send never happens, and the receiving loop below blocks on that missing value until the suite's `-timeout` kills it. The real failure (context cancelled, or listen exhausted after 100 attempts) gets replaced by an opaque timeout panic instead of a clear test failure.

Low probability under normal CI load, but worth hardening, e.g. run each goroutine as its own `t.Run(..., func(t *testing.T) { t.Parallel(); ... })` subtest so `FailNow` has a valid per-goroutine `*testing.T` to target, instead of sharing the parent `t` across raw goroutines.

test/testutil/readiness.go
```diff
@@ -51,14 +52,54 @@ func pollForSpans(c *Collector, minSpans int, timeout time.Duration) bool {
...
+		case <-t.Context().Done():
+			t.Fatalf("FreePort: context cancelled: %v", t.Context().Err())
+		default:
+		}
+
+		lis, err := net.Listen("tcp", "localhost:0")
+		if err != nil {
+			time.Sleep(10 * time.Millisecond)
+			continue
+		}
```
darccio, line 89: On `net.Listen` failure the error is discarded entirely; never logged, never surfaced. On `main`, `require.NoError(t, err)` reported the actual OS error (e.g. "too many open files"). Now, if all 100 attempts fail, the final `t.Fatalf` message is generic ("failed to allocate a unique free port after 100 attempts") with no trace of *why*. A real diagnosability regression versus the previous fail-fast behavior.

```suggestion
		lis, err := net.Listen("tcp", "localhost:0")
		if err != nil {
			t.Logf("FreePort: listen attempt %d/%d failed: %v", i+1, maxAttempts, err)
			time.Sleep(10 * time.Millisecond)
			continue
		}
```

darccio (changes requested):

darccio:
Good job. Some comments for adding context:

The registry in `allocatedPorts`/`allocatedPortsMu` is scoped per test binary process, not shared across `test/integration` and `test/e2e`. That looked like a potential gap at first, but checking `.github/workflows/test-integration.yaml` and `test-e2e.yaml` confirms it isn't:

- The two suites run as separate GitHub Actions workflows with separate job graphs, and every matrix cell (`test-integration`: 4 platforms × 3 shards; `test-e2e`: 4 platforms) gets its own GitHub-hosted runner VM — never shared with another job.
- `test/integration` and `test/e2e` are each a single Go package (`package test`, no sub-packages), so `go -C test test ... ./integration/...` and `./e2e/...` (Makefile:575,624) each compile to exactly one test binary per job.

So in this repo's actual CI, "job boundary" always coincides with "VM boundary" and "process boundary." The only place two `FreePort` calls can ever race for the same OS ephemeral-port pool is within one `go test` process, among its `t.Parallel()` subtests — which is exactly what this per-process registry protects. The scope isn't a partial fix with a gap; it covers the entire collision surface this CI setup can actually produce.

darccio (approved):

PR #998, opened 2026-08-07
refactor(tool/internal/ast): simplify tokenize in directive.go by eliminating redundant strings.Builder error checks

### LINKED ISSUE #997: refactor(tool/internal/ast): simplify tokenize in directive.go by eliminating redundant strings.Builder error checks
## Description
In `tool/internal/ast/directive.go`, `tokenize` checks errors returned by `current.WriteRune(ch)` (`strings.Builder`).

Since `strings.Builder.WriteRune` is guaranteed by the Go standard library specification to never return an error, checking `err != nil` 4 times per character in the tokenization loop creates dead branches and redundant code boilerplate.

## Location
- `tool/internal/ast/directive.go:177-212`
## Proposed Fix
Call `current.WriteRune(ch)` directly without checking `err`, making `tokenize` cleaner and eliminating dead branches.

vedant21-oss:
## Description
In `tool/internal/ast/directive.go`, `tokenize` checked errors returned by `current.WriteRune(ch)` (`strings.Builder`).

Since `strings.Builder.WriteRune` is guaranteed by the Go standard library specification to never return an error, checking `err != nil` 4 times per character in the tokenization loop created dead branches and redundant code.

Fixes #997 
## Changes Made
Simplified `tokenize` in `tool/internal/ast/directive.go` by calling `current.WriteRune(ch)` directly without checking `err`.
Cleaned up dead branches in rune tokenization logic.

## Verification
golangci-lint: 0 issues
go test ./tool/internal/ast/...: All tests pass cleanly.

### DIFF
--- a/tool/internal/ast/directive.go
+++ b/tool/internal/ast/directive.go
@@ -169,33 +169,23 @@ func FindFuncsByDirective(file *dst.File, directive string) []*dst.FuncDecl {
 func tokenize(input string) ([]string, error) {
 	var tokens []string
 	var current strings.Builder
-	var err error
 	inQuote := false
 	escaped := false
 
 	for _, ch := range input {
 		if escaped {
-			_, err = current.WriteRune(ch)
-			if err != nil {
-				return nil, ex.Wrapf(err, "failed to write rune")
-			}
+			current.WriteRune(ch)
 			escaped = false
 			continue
 		}
 		if ch == '\\' && inQuote {
-			_, err = current.WriteRune(ch)
-			if err != nil {
-				return nil, ex.Wrapf(err, "failed to write rune")
-			}
+			current.WriteRune(ch)
 			escaped = true
 			continue
 		}
 		if ch == '"' {
 			inQuote = !inQuote
-			_, err = current.WriteRune(ch)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


dobbydobap:
I ran the old and new tokenize side by side rather than eyeballing the diff,
since a refactor of a hand-rolled parser is easy to get subtly wrong.

Corpus was 9366 inputs: a hand-written set covering quotes, escapes, unclosed
quotes, unicode spaces and empty input, plus every string of length 1 to 5 over
a, space, double-quote, backslash, colon and tab. Zero differences in either the
token slice or the error. So this is behaviour-preserving.

The premise holds too: strings.Builder.WriteRune always returns a nil error, so
those four branches were unreachable. ex is still needed for the unclosed quote
path, and control flow is untouched.

One thing I noticed in the same function while testing, not for this PR: scanArgs
splits on the first colon with strings.Cut, but tokenize keeps the quote
characters, so the split is quote-unaware. `"key:value"` parses to Key `"key`
and Value `value"` with no error, and `:v` gives an empty key. I will raise that
separately so it does not widen this refactor.

y1yang0 (approved):

Issue #981, opened 2026-08-06
feat(tool): use instrumentation manifest for pin discovery, remove otelc-bundle.tgz [2/6]


darccio:
# feat(tool): use instrumentation manifest for pin discovery, remove `otelc-bundle.tgz` [2/6]

**Tracking:** part of #585 (decouple built-in instrumentations from `otelc-bundle.tgz`)
**Depends on:** #980
**Blocks:** #983 (should land together with, or before, #983 — see note below)

## Background

`otelc` currently embeds all built-in instrumentation source (`instrumentation/**`, 26 packages) and the shared runtime library (`pkg/`) into the binary as `tool/data/otelc-bundle.tgz` via `//go:embed` in `tool/data/export.go`, extracted at runtime by `tool/internal/setup/extract.go`'s `extractOtelcBundle()`. `otelc pin`'s `generatePinnedProjects()` (`tool/internal/setup/pin.go` lines ~448-511) calls `extractOtelcBundle()` + `loadMinimalRules()` purely to learn which built-in instrumentation modules exist and what they target, in order to decide which blank imports to add to a freshly-generated `otel.instrumentation.go`.

#980 introduced `tool/internal/manifest`, a small JSON manifest holding exactly this metadata (`{modulePath, target, versionRange}` per built-in instrumentation), generated from the real `instrumentation/` tree instead of requiring full source to be embedded — but left it unwired. This issue wires it in and removes the now-redundant bundle machinery.

Everything else in the setup/matching pipeline — `resolveInstrumentationConfigs()` in `tool/internal/setup/config.go`, rule-file discovery in `tool/internal/setup/match.go`, AST-based weaving — already resolves instrumentation packages generically via real Go module mechanics (`golang.org/x/tools/go/packages`) and needs **no changes**; it already works identically for built-in and third-party packages once an import exists in `otel.instrumentation.go`.

### Tasks

- [ ] In `tool/internal/setup/pin.go` `generatePinnedProjects()` (~lines 448-511): replace the `extractOtelcBundle()` + `loadMinimalRules(...)` call with `manifest.Load()`. Adjust `matchInstrumentationImports()` (~line 309+) to consume `manifest.Manifest`/`manifest.Entry` instead of the local `yamlRule` type — the matching logic itself (exact/glob target match, root-target always-match, `util.VersionInRange`) does not need to change.
[...]

PR #999, opened 2026-08-07
feat(tool): add instrumentation manifest generator

### LINKED ISSUE #980: feat(tool): add lightweight instrumentation manifest generator [1/6]
# feat(tool): add lightweight instrumentation manifest generator [1/6]

**Tracking:** part of #585 (decouple built-in instrumentations from `otelc-bundle.tgz`)
**Depends on:** none
**Blocks:** #981

## Background

Today, all built-in instrumentations under `instrumentation/**` (26 packages: `net/http`, `database/sql`, `google.golang.org/grpc`, `github.com/redis/go-redis/v9`, etc.) plus the shared runtime support library `pkg/` are tarred into `tool/data/otelc-bundle.tgz`, embedded into the `otelc` binary via `//go:embed`, and extracted to a temp dir every time `otelc` runs. `otelc pin`'s auto-discovery (`generatePinnedProjects` in `tool/internal/setup/pin.go`) extracts this bundle purely to learn "which built-in instrumentation targets which import path, at which version range, and which module path provides it" — via `loadMinimalRules`/`loadModuleRules` (`tool/internal/setup/pin.go` lines ~169-238), which walks the extracted `instrumentation/` tree, finds each nested `go.mod`, and parses each rule file's `target`/`version` fields.

This issue is step 1 of decoupling: introduce a small, git-friendly **manifest** that captures exactly this metadata (and nothing else — no source code) without changing any runtime behavior yet.

## Scope

Add a new manifest package and generator, **unwired** — `otelc pin` continues using the old bundle-extraction path until #981 lands. This PR should be a pure addition, safe to review in isolation.

### Tasks

- [ ] Add `tool/internal/manifest/manifest.go`:
  - `type Entry struct { ModulePath, Target, VersionRange string }`
  - `type Manifest []Entry`
  - `func Generate(instrumentationRoot string) (Manifest, error)` — port of the existing `loadMinimalRules`/`loadModuleRules` walk logic (`tool/internal/setup/pin.go` ~169-238): walk `instrumentationRoot`, find nested `go.mod` files, parse each nested module's rule files (`otelc.yml`/`otelc.yaml`/`*.otelc.yml`/`*.otelc.yaml`, matched via the existing `isRuleFile()` helper in `tool/internal/setup/match.go`), extract `target`+`version` per rule. Unlike today's runtime version, this walks the **real** `instrumentation/` directory directly — no copying to a `*_temp` dir, no `go mod tidy`.
  - `func Load() (Manifest, error)` — parses the embedded JSON produced by the generator below.
  - Sort entries deterministically by `(ModulePath, Target)` before serialization for stable, minimal diffs.
- [ ] Add `tool/cmd/gen-manifest/main.go` — a dev/release-time-only CLI that calls `manifest.Generate("instrumentation")`, marshals to indented JSON, and writes `tool/data/instrumentation-manifest.json`.
- [ ] Add `tool/data/manifest.go`:
  ```go
  //go:embed instrumentation-manifest.json
  var manifestJSON []byte
  func GetManifestJSON() []byte { return manifestJSON }
  ```
  This coexists with the existing `tool/data/export.go` (`//go:embed otelc-bundle.tgz`) for now — do **not** remove the old bundle embed in this PR, that happens in #981.
- [ ] Generate and commit `tool/data/instrumentation-manifest.json` from the current `instrumentation/` tree (run `gen-manifest` once).
- [ ] Add a `make manifest` Makefile target invoking `gen-manifest`.
- [ ] Add a `verify-manifest` CI check (workflow or script) that runs `go run ./tool/cmd/gen-manifest && git diff --exit-code tool/data/instrumentation-manifest.json` to catch a forgotten regeneration. This is cheap (no `go mod tidy`, no temp dirs) — no PR-comment regenerate-bot is needed.
- [ ] Move `TestLoadMinimalRules_*` (currently in `tool/internal/setup/pin_test.go`) to `tool/internal/manifest/manifest_test.go`, adapted to call `manifest.Generate` directly. Add a round-trip test: `Generate` → marshal → `Load`/unmarshal → equal.

## Out of scope

- Wiring this manifest into `otelc pin`'s actual discovery logic (#981).
- Removing `otelc-bundle.tgz`, `tool/data/export.go`, `.tools/bundle/`, or any bundle-sync CI (#981).
- Any change to `sync.go`'s `replace`-injection behavior (#983).

## Acceptance criteria

- `make manifest` regenerates `tool/data/instrumentation-manifest.json` byte-for-byte identical to the committed version (idempotent).
- `go test ./tool/internal/manifest/...` passes.
- No observable behavior change in `otelc` — it still uses the old bundle-extraction path for auto-pin discovery.

ADITYA-CODE-SOURCE:
## Description

Adds a lightweight JSON manifest generated from the built-in instrumentation tree. The manifest records each instrumentation module path, target, and supported version range without embedding any source code.

This PR adds:

- `tool/internal/manifest` for deterministic manifest generation and loading
- `tool/cmd/gen-manifest` for regenerating the JSON file
- an embedded `tool/data/instrumentation-manifest.json`
- `make manifest` and `make verify-manifest`
- a CI consistency check to detect a stale manifest
- tests for nested modules, rule-file discovery, ordering, malformed input, and embedded loading

This change is intentionally additive. `otelc pin` still uses the existing bundle-based discovery flow; runtime wiring and bundle removal are left for #981.

## Motivation

This is the first step toward decoupling built-in instrumentations from `otelc-bundle.tgz`. The small, text-based manifest provides the metadata needed for instrumentation discovery while producing stable and reviewable diffs.

Fixes #980


tool/internal/manifest/manifest.go
```diff
@@ -0,0 +1,154 @@
...
+		if cmp := strings.Compare(a.ModulePath, b.ModulePath); cmp != 0 {
+			return cmp
+		}
+		if cmp := strings.Compare(a.Target, b.Target); cmp != 0 {
+			return cmp
+		}
+		return strings.Compare(a.VersionRange, b.VersionRange)
+	})
+	return manifest, nil
+}
```
darccio, line 74: `Generate` emits one `Entry` per YAML rule, but `Entry` only carries `(module, target, version)`. Fields shared by every rule that targets the same package at the same version. The generated artifact currently has 515 entries but only 28 unique tuples; the `linodego/v2` rule set alone contributes 450 identical rows with zero distinguishing information. Dedup after sorting so the manifest reflects distinct combinations instead of one row per hook.

```suggestion
	})
	manifest = slices.CompactFunc(manifest, func(a, b Entry) bool {
		return a == b
	})
	return manifest, nil
}
```

tool/internal/manifest/manifest.go
```diff
@@ -0,0 +1,154 @@
...
+
+func loadModuleEntries(moduleDir, modulePath string) (Manifest, error) {
+	entries := make(Manifest, 0)
+	root, err := os.OpenRoot(moduleDir)
+	if err != nil {
+		return nil, ex.Wrapf(err, "opening module root %s", moduleDir)
+	}
+	defer root.Close()
+
+	err = filepath.WalkDir(moduleDir, func(path string, d os.DirEntry, err error) error {
```
darccio: `loadModuleEntries` opens `os.OpenRoot(moduleDir)` and uses it only for the final `root.ReadFile` call. The traversal itself (`filepath.WalkDir` here, and the nested-module check via `os.Stat` at line 107) still walks the real filesystem rather than the root-scoped one, so a symlink inside the module tree could still be followed before the sandboxed read ever runs. Since this only processes the repo's own instrumentation tree today, impact is low, but worth either applying the root scoping consistently (e.g. walking via the root's own FS) or dropping `OpenRoot` if it's not adding real protection here.

tool/internal/manifest/manifest.go
```diff
@@ -0,0 +1,154 @@
...
+	"gopkg.in/yaml.v3"
+
+	"go.opentelemetry.io/otelc/tool/data"
+	"go.opentelemetry.io/otelc/tool/ex"
+)
+
+type Entry struct {
+	ModulePath   string `json:"modulePath"`
+	Target       string `json:"target"`
+	VersionRange string `json:"versionRange"`
```
darccio: Rules without a `version:` key (e.g. `database/sql`) always serialize as `"versionRange": ""`. Adding `omitempty` avoids carrying a redundant empty field for every version-less entry.

```suggestion
	VersionRange string `json:"versionRange,omitempty"`
```

tool/cmd/gen-manifest/main.go
```diff
@@ -0,0 +1,38 @@
...
+const generatedFilePerm = 0o644
+
+func main() {
+	if err := run(); err != nil {
+		_, _ = fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
```
darccio, line 26: `run()` resolves both the instrumentation root and the output path relative to the process's working directory, so it only behaves correctly when invoked from the repo root. That's true in practice because `make manifest` runs it from there, and running it elsewhere fails loudly rather than writing to the wrong place — but the assumption isn't written down anywhere. A doc comment saves the next person from having to trace through `Generate` and `WriteFileAtomic` to figure out why a stray `go run` from a subdirectory errored.

```suggestion
// run must execute with the repository root as its working directory — both
// the instrumentation scan and the output path below are relative to it.
// make manifest guarantees this; invoking it from elsewhere fails loudly
// instead of writing to the wrong location.
func run() error {
```

tool/data/manifest.go
```diff
@@ -0,0 +1,13 @@
...
+package data
+
+import _ "embed"
+
+//go:embed instrumentation-manifest.json
+var manifestJSON []byte
+
+func GetManifestJSON() []byte {
+	return manifestJSON
+}
```
darccio: `GetBundleReader()` three lines away in `export.go` wraps its embedded bytes in a `bytes.Reader` specifically so callers can't mutate the shared underlying array. `GetManifestJSON()` returns the embedded `manifestJSON` slice directly, so a caller writing into the returned slice would corrupt the package-level singleton for every subsequent call, including `manifest.Load()`. Cloning on the way out matches the defensive pattern already established in this file.

```suggestion
import (
	"bytes"
	_ "embed"
)

//go:embed instrumentation-manifest.json
var manifestJSON []byte

func GetManifestJSON() []byte {
	return bytes.Clone(manifestJSON)
}
```

.github/workflows/check-consistency.yaml
```diff
@@ -34,6 +34,13 @@ jobs:
               - 'pkg/hook/context.go'
               - 'tool/internal/instrument/api.tmpl'
               - 'tool/internal/instrument/testdata/**'
+              - 'instrumentation/**/go.mod'
+              - 'instrumentation/**/*.otelc.yml'
+              - 'instrumentation/**/*.otelc.yaml'
```
darccio, line 41: These two patterns only match the prefixed rule-file form (`<name>.otelc.yaml`). `isRuleFile()` in `tool/internal/manifest/manifest.go` also accepts the bare `otelc.yml`/`otelc.yaml` filename — and that's the form 26 of the 27 real rule files in `instrumentation/` actually use (only `linodego/v2/public_methods.otelc.yaml` is prefixed). As written, editing any of those 26 files — the realistic edit path — won't set `changes.outputs.code = true` on its own, so `verify-manifest` (and the other consistency jobs gated on the same output) gets skipped unless the same diff happens to also touch a `.go` or `go.mod` file.

```suggestion
              - 'instrumentation/**/otelc.yml'
              - 'instrumentation/**/otelc.yaml'
              - 'instrumentation/**/*.otelc.yml'
              - 'instrumentation/**/*.otelc.yaml'
```

tool/internal/manifest/manifest_test.go
```diff
@@ -0,0 +1,159 @@
...
+	writeRuleFile(t, root, "parent/otelc.yaml", `
+later:
+  target: example.com/target
+  version: v2.0.0
+earlier:
+  target: example.com/target
+  version: v1.0.0
+empty:
+  version: v3.0.0
+`)
```
darccio, line 32: `TestGenerate`'s fixture never produces two rules that collapse to the same `(module, target, version)` tuple, so nothing here would catch a regression once `Generate` is fixed to dedup. Adding a rule that duplicates `earlier`'s tuple — while leaving the expected `Manifest` at the bottom unchanged — turns this into a regression test: it fails against today's implementation and passes once the dedup fix lands.

```suggestion
	writeRuleFile(t, root, "parent/otelc.yaml", `
later:
  target: example.com/target
  version: v2.0.0
earlier:
  target: example.com/target
  version: v1.0.0
duplicate:
  target: example.com/target
  version: v1.0.0
empty:
  version: v3.0.0
`)
```

tool/internal/manifest/manifest_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+func TestLoad(t *testing.T) {
+	got, err := Load()
+	require.NoError(t, err)
+	require.NotEmpty(t, got)
+
+	var decoded Manifest
+	require.NoError(t, json.Unmarshal(data.GetManifestJSON(), &decoded))
+	require.Equal(t, decoded, got)
+}
```
darccio, line 188: This re-derives `decoded` via the exact same operation `Load()` performs internally (`json.Unmarshal(data.GetManifestJSON(), ...)`), so `require.Equal(t, decoded, got)` doesn't test anything beyond what `require.NoError` + `require.NotEmpty` already cover. Asserting a real invariant — e.g. that every entry has a non-empty `ModulePath`/`Target` — would actually catch a corrupted or hand-edited artifact.

```suggestion
func TestLoad(t *testing.T) {
	got, err := Load()
	require.NoError(t, err)
	require.NotEmpty(t, got)

	for _, entry := range got {
		assert.NotEmpty(t, entry.ModulePath)
		assert.NotEmpty(t, entry.Target)
	}

	var decoded Manifest
	require.NoError(t, json.Unmarshal(data.GetManifestJSON(), &decoded))
	require.Equal(t, decoded, got)
}
```

tool/internal/manifest/manifest.go
```diff
@@ -0,0 +1,154 @@
...
+}
```
darccio, line 29: The struct has no doc comment, and "VersionRange" invites the wrong mental model. It's actually a single minimum-version bound, not a min+max range — that only becomes clear by reading a comment in `instrumentation/github.com/linode/linodego/v2/otelc.yaml` that references `tool/util.VersionInRange`, several files away from this definition. Worth documenting here so the schema is self-explanatory for whatever consumes `Load()` next (#981).

```suggestion
// Entry describes one instrumentation rule: the module that owns the rule,
// the package it targets, and the version bound the rule applies from.
type Entry struct {
	ModulePath string `json:"modulePath"`
	Target     string `json:"target"`
	// VersionRange is a minimum version bound, not a min-max range (see
	// tool/util.VersionInRange). Empty means the rule applies to all versions.
	VersionRange string `json:"versionRange"`
}
```

darccio (changes requested):
This PR builds a three-layer safety net against a stale manifest: generator correctness (`Generate`), a local check (`make verify-manifest`), and a CI gate (the `paths-filter` job). Each layer independently has a gap — `Generate` doesn't dedup identical rules, and the CI path filter's `*.otelc.yml`/`*.otelc.yaml` patterns miss the bare `otelc.yaml`/`otelc.yml` filename that 26 of the 27 real rule files in the tree actually use. None of the three layers would have caught the other's gap: the checked-in artifact went in at 515 entries (only 28 unique) without any of them flagging it. Worth fixing both (see the inline comments on `manifest.go:66-68` and `check-consistency.yaml:38-39`) rather than either alone — a fix to just one still leaves the other's blind spot standing.

ADITYA-CODE-SOURCE:
Thanks for the detailed review. I’ve pushed the fixes: the manifest is now deduplicated  the workflow covers all rule-file names, and the related safety checks and tests have been tightened. I also regenerated the manifest and reran the focused, race, unit, and lint checks.

darccio (approved):

darccio:
@ADITYA-CODE-SOURCE Please resolve the conversations if they are implemented. Comment on those you didn't apply the suggestion or think differently (don't resolve these so I can acknowledge them).

ADITYA-CODE-SOURCE:
> @ADITYA-CODE-SOURCE Please resolve the conversations if they are implemented. Comment on those you didn't apply the suggestion or think differently (don't resolve these so I can acknowledge them).

 @darccio All review suggestions have been addressed, and the conversations are resolved. All checks are now passing.

PR #1003, opened 2026-08-07
test: enable build caching for E2E tests in CI

### LINKED ISSUE #959: test(e2e): tighten cross-process span-flush budget and evaluate parallelizing E2E tests

singhshresth26:
## Description

Configure `actions/cache` and `OTELC_TEST_GOCACHE` environment variables for the E2E tests inside `.github/workflows/test-e2e.yaml` (both coverage and standard runs).

## Motivation

Currently, the E2E workflow (`test-e2e.yaml`) does not specify `OTELC_TEST_GOCACHE` or configure Go caching. This causes each E2E test to compile and build the test applications from scratch, leading to high compile/run times. Enabling build caching matches the configuration in integration tests and speeds up E2E runs in CI.

Fixes #959 (Part 4/4)

---

## Checklist

- [x] Tests pass: `make test`
- [x] OpenTelemetry Registry updated (if applicable, see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation))

[^I-know-my-stuff]:

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -47,7 +47,16 @@ jobs:
       - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c  # v6.1.0
         with:
           go-version-file: go.mod
+          cache-dependency-path: "**/go.mod"
+      - name: Restore otelc Go cache
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830  # v4
+        with:
+          path: test/.otelc-gocache
+          key: ${{ runner.os }}-amd64-otelc-gocache-e2e-coverage-${{ hashFiles('go.mod', '**/go.mod', '**/go.sum', 'Makefile',
+            'tool/**', 'pkg/**', 'instrumentation/**', 'test/apps/**') }}
       - run: make test-e2e/coverage
+        env:
+          OTELC_TEST_GOCACHE: ${{ github.workspace }}/test/.otelc-gocache
       - uses: codecov/codecov-action@0fb7174895f61a3b6b78fc075e0cd60383518dac  # v5.5.5
         with:
           fail_ci_if_error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/test-e2e.yaml
```diff
@@ -47,7 +47,16 @@ jobs:
       - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c  # v6.1.0
         with:
           go-version-file: go.mod
+          cache-dependency-path: "**/go.mod"
+      - name: Restore otelc Go cache
```
darccio, line 51: This step also *saves* the cache at the end of the job (plain `actions/cache`, not `actions/cache/restore`), so "Restore" undersells what it does. Worth a clearer name since this pattern is now duplicated across two workflows.

```suggestion
      - name: Restore/save otelc Go cache
```

.github/workflows/test-e2e.yaml
```diff
@@ -78,9 +87,16 @@ jobs:
         uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b  # ratchet:arduino/setup-protoc@v3
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Restore otelc Go cache
```
darccio, line 90: Same as the coverage job's cache step above — this also restores and saves, not just restores.

```suggestion
      - name: Restore/save otelc Go cache
```

darccio (approved):
Minor: `test-e2e-coverage` (Linux amd64, key `…-otelc-gocache-e2e-coverage-<hash>` at line 55) and the `test-e2e` matrix's ubuntu-latest/amd64 leg (key `…-otelc-gocache-e2e-<hash>` at line 94) build the same set of E2E test apps for the same OS/arch, but under different cache keys — so they never share a cache entry. Each pays its own first-cold-build cost and duplicates cache storage. This mirrors the existing coverage/shard split already accepted in `test-integration.yaml`, so it's consistent with established practice rather than a new inefficiency, but a shared key between these two same-arch legs would avoid the duplication if it's worth the added coupling between the jobs.

PR #1008, opened 2026-08-07
fix(instrumentation/openai): map `max_completion_tokens` to `gen_ai.request.max_tokens`

### LINKED ISSUE #1007: bug(instrumentation/openai): max_completion_tokens not mapped to gen_ai.request.max_tokens
### Describe the bug

For chat completions, `parseChatRequest` only looks at `max_tokens`. Modern OpenAI chat requests (and the official Go SDK) usually send `max_completion_tokens` instead,  `max_tokens` is deprecated and doesn't work on o-series models.

So if an app sets `MaxCompletionTokens`, the JSON on the wire has `max_completion_tokens`, we ignore it, and the span never gets `gen_ai.request.max_tokens`.

Same in v1, v2, and v3.

### To Reproduce

Send a chat completion body like:

```json
{"model":"o1-mini","max_completion_tokens":100}
```

Span is created, but there's no `gen_ai.request.max_tokens` attribute.

With the old field it works:

```json
{"model":"gpt-4","max_tokens":100}
```

### Expected behavior

Either JSON field should set `gen_ai.request.max_tokens`. If both are present, prefer `max_completion_tokens`.

Happy to send a PR for v1/v2/v3 + tests.

Anshumancanrock:
## Description

Chat request parsing only looked at `max_tokens`. OpenAI's Go SDK (and o-series models) use `max_completion_tokens` instead, so those requests never got `gen_ai.request.max_tokens` on the span.

Updated `parseChatRequest` in v1/v2/v3 to read both fields and prefer `max_completion_tokens` when both are set. Added tests for the old field, the new field, and the preference case.

## Motivation

Modern chat calls often only send `max_completion_tokens`. We already emit `gen_ai.request.max_tokens` when the legacy field is present. Just need to handle the current wire format too.

Fixes #1007

---

## Checklist

- [ ] Tests pass: `make test`

[^I-know-my-stuff]:


Aryanbhargava18:
Nice addition! Preferring `max_completion_tokens` over deprecated `max_tokens` is the right move for `o1`/`o3` models.

One small thing in `middleware_test.go` (across v1, v2, v3): 

In the `"max_tokens"` test case, the JSON body sets `"temperature": 0.7`, but the assertion loop only extracts `gen_ai.request.max_tokens`:
```go
if a.Key == "gen_ai.request.max_tokens" {
    v := a.Value.AsInt64()
    gotMax = &v
    break
}
```
Because of that, `temperature` isn't actually being asserted anywhere in the test suite now. Might be good to add an assertion for `gen_ai.request.temperature` in that case so we don't drop coverage on the optional request attributes.

darccio (approved):
LGTM

darccio:
@Anshumancanrock Bundle, please.

Anshumancanrock:
@darccio done!

darccio:
Closed in favour of #1083

PR #1010, opened 2026-08-08
fix(tool/internal/ast): return standard zero-value token.Position for unmapped nodes in FindPosition

### LINKED ISSUE #1009: fix(tool/internal/ast): return standard zero-value token.Position for unmapped nodes in FindPosition
## Description
`AstParser.FindPosition` in `tool/internal/ast/parser.go` returns `token.Position{Filename: "", Line: -1, Column: -1}` when a node is absent from the decoration AST node map.

In standard Go (`go/token`), an invalid position is conventionally represented by the zero value `token.Position{}` (`Line: 0, Column: 0`). `pos.IsValid()` is defined as `pos.Line > 0`, so returning negative line numbers breaks standard position validation and diagnostic error formatting.

## Location
- `tool/internal/ast/parser.go:90`
## Proposed Fix
Return standard `token.Position{}` when `astNode == nil` and update unit tests to assert `token.Position{}` and `!pos.IsValid()`.

vedant21-oss:
## Description
AstParser.FindPosition in tool/internal/ast/parser.go previously returned token.Position{Filename: "", Line: -1, Column: -1} when a node was absent from the decoration AST node map.

In standard Go (go/token), an invalid position is conventionally represented by the zero value token.Position{} (Line: 0, Column: 0). Returning -1 for Line and Column violates standard Go token.Position conventions and causes pos.IsValid() (which checks Line > 0) or diagnostic formatters to produce unexpected negative coordinates.

Fixes #1009 
## Changes Made
Updated AstParser.FindPosition in tool/internal/ast/parser.go to return token.Position{} when astNode == nil.
Updated unit test in tool/internal/ast/parser_extra_test.go to assert token.Position{} and !pos.IsValid().

## Verification
golangci-lint: 0 issues
go test ./tool/internal/ast/...: All tests pass cleanly.

### DIFF
--- a/tool/internal/ast/parser.go
+++ b/tool/internal/ast/parser.go
@@ -84,10 +84,11 @@ func (ap *AstParser) ParseSource(source string) (*dst.File, error) {
 }
 
 // FindPosition finds the source position of a node in the AST.
+// It returns a zero-value token.Position{} when the node is unmapped.
 func (ap *AstParser) FindPosition(node dst.Node) token.Position {
 	astNode := ap.dec.Ast.Nodes[node]
 	if astNode == nil {
-		return token.Position{Filename: "", Line: -1, Column: -1} // Invalid
+		return token.Position{}
 	}
 	return ap.fset.Position(astNode.Pos())
 }
--- a/tool/internal/ast/parser_extra_test.go
+++ b/tool/internal/ast/parser_extra_test.go
@@ -4,6 +4,7 @@
 package ast
 
 import (
+	"go/token"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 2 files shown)


dobbydobap:
Had a look since this has been sitting a while.

The change is correct and behaviour-preserving at both call sites. FindPosition
has two callers, apply_func.go:219 and :227. The one at :227 guards with
pos.IsValid(), which is Line > 0, so -1 and 0 both read as invalid there.

The else branch at :235 does not guard, it formats pos straight into a //line
directive, so I checked what each prints:

    token.Position{Filename: "", Line: -1, Column: -1}.String() == "-"
    token.Position{}.String()                                   == "-"

Identical, so nothing downstream changes. LGTM.

y1yang0 (approved):

PR #1012, opened 2026-08-08
fix(net/http): set `http.request.method=_OTHER` for unknown methods

### LINKED ISSUE #1011: bug: `net/http` records unknown HTTP methods as `GET` instead of `_OTHER`
### Description

Span attributes for unknown / empty HTTP methods fall back to `GET` in both the server and client helpers:

```go
// instrumentation/net/http/semconv/server.go (same pattern in client.go)
return semconv.HTTPRequestMethodGet, orig
```

So a request with method `CUSTOM` ends up as:

- `http.request.method` = `GET`
- `http.request.method_original` = `CUSTOM`

That’s incorrect telemetry as unknown traffic gets mixed into real GETs.

Metrics already handle this correctly via `StandardizeHTTPMethod` (`CUSTOM` / `""` → `_OTHER`). Spans don’t. Upstream `otelhttp` returns `HTTPRequestMethodOther` here.

### Spec

[HTTP spans](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) (v1.37.0, what we pin):

> If the HTTP request method is not known to instrumentation, it MUST set the `http.request.method` attribute to `_OTHER`.

And for span names:

> In other cases (when `{http.request.method}` is set to `_OTHER`), `{method}` MUST be `HTTP`.

We still put the raw method in the span name (`FOOBAR /path`, client span `FOOBAR`), including when gin renames the span after route match.

### Expected

| method | `http.request.method` | `http.request.method_original` | span name |
| --- | --- | --- | --- |
| `CUSTOM` | `_OTHER` | `CUSTOM` | `HTTP` / `HTTP {route}` |
| `""` | `_OTHER` | omitted | `HTTP` |
| `get` | `GET` | `get` | `GET` / `GET {route}` |

Anshumancanrock:
## Summary

Unknown / empty HTTP methods were recorded as `GET` on spans. That breaks the HTTP semconv MUST and mixes custom traffic into real GET counts.

This pr aligns span attributes and span names with the spec and with upstream `otelhttp`:

- unknown / empty → `http.request.method=_OTHER`
- keep `http.request.method_original` when the raw method is non-empty
- span name uses `HTTP` (or `HTTP {route}`) when the method is `_OTHER`
- shared helpers so server/client don’t drift; gin route rename uses the same naming

Fixes #1011 

Metrics were already correct via `StandardizeHTTPMethod`.

## Test plan

- [x] `go test` in `instrumentation/net/http/semconv` (`CUSTOM`, empty, lowercase known method, span names)
- [x] `go test` in `instrumentation/net/http/{client,server}`
- [x] `go test` in `instrumentation/github.com/gin-gonic/gin`


instrumentation/net/http/semconv/util.go
```diff
@@ -135,3 +135,30 @@ func StandardizeHTTPMethod(method string) string {
...
+// http.request.method_original. Unknown and empty methods use _OTHER.
+func requestMethodAttrs(method string) (attribute.KeyValue, attribute.KeyValue) {
+	if method == "" {
+		return upstream.HTTPRequestMethodOther, attribute.KeyValue{}
+	}
+	if attr, ok := MethodLookup[method]; ok {
+		return attr, attribute.KeyValue{}
+	}
+
+	orig := upstream.HTTPRequestMethodOriginal(method)
```
amazingakai: If a caller explicitly sends an HTTP request where the method is exactly `"_OTHER"`, `requestMethodAttrs` returns:
- `http.request.method=_OTHER`
- `http.request.method_original=_OTHER`

This violates the OpenTelemetry Semantic Conventions for [`http.request.method_original`](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/http/http-spans.md), which explicitly states that it is conditionally required:
> "If and only if it's different than `http.request.method`."

Because `_OTHER` == `_OTHER`, `method_original` must be omitted (i.e., return an empty `attribute.KeyValue{}`) when the raw method is `_OTHER`.

instrumentation/net/http/semconv/util.go
```diff
@@ -135,3 +135,30 @@ func StandardizeHTTPMethod(method string) string {
...
+	if attr, ok := MethodLookup[strings.ToUpper(method)]; ok {
+		return attr, orig
+	}
+	return upstream.HTTPRequestMethodOther, orig
```
amazingakai: By coercing unknown methods to `_OTHER` (and `HTTP` for span names), the instrumentation now converts valid, non-standard HTTP request methods (like `PROPFIND` or `PURGE`) to `_OTHER`. The OpenTelemetry HTTP semantic conventions for [`http.request.method`](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/registry/attributes/http.md) mandate:
> "If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it **MUST provide a way to override the list of known HTTP methods.** If this override is done via environment variable, then the environment variable **MUST be named `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS`**..."

This PR introduces the fallback to `_OTHER` but fails to implement the required `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS` environment variable override. Without this, consumers have no way to whitelist their valid custom methods, resulting in them being irreversibly masked as `_OTHER` and their span names mangled to `HTTP`.

instrumentation/net/http/semconv/util.go
```diff
@@ -135,3 +135,30 @@ func StandardizeHTTPMethod(method string) string {
 	}
 	return method
 }
```
amazingakai, line 185: This still uses a hardcoded `switch` block that duplicates the exact same keys present in `MethodLookup`. This is an unnecessary risk for drift (e.g., if a new method is added to the map but forgotten in the switch). Let's use the same lookup method as `requestMethodAttrs` here.

amazingakai (changes requested):
Clean fix! We can do a few improvements here.

instrumentation/net/http/semconv/util.go
Anshumancanrock: good catch, fixed.

instrumentation/net/http/semconv/util.go
Anshumancanrock: done!

instrumentation/net/http/semconv/util.go
Anshumancanrock: agree, that was gonna drift. `StandardizeHTTPMethod` goes through the same known-methods lookup now.

instrumentation/net/http/semconv/util.go
```diff
@@ -116,22 +119,89 @@ var MethodLookup = map[string]attribute.KeyValue{
 	"QUERY":            upstream.HTTPRequestMethodKey.String("QUERY"),
```
amazingakai: The HTTP semantic conventions explicitly limit default known methods to [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-methods) plus `PATCH` from [RFC5789](https://www.rfc-editor.org/rfc/rfc5789.html). Since `QUERY` isn't in those RFCs (which is why upstream doesn't have a constant for it), it shouldn't be hardcoded in our default `MethodLookup`.

Removing it is technically a breaking change, but since this PR introduces `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS`, this is the right time to drop it. Users who need `QUERY` can now officially opt-in using the new environment variable.

Could you drop `"QUERY"` from `MethodLookup` (either in this PR or a follow-up) and add a note in `CHANGELOG.md` explaining how users can restore it via the env var?

instrumentation/net/http/semconv/util.go
```diff
@@ -116,22 +119,89 @@ var MethodLookup = map[string]attribute.KeyValue{
...
-		method = "_OTHER"
+	lookup := knownMethods()
+	if _, ok := lookup[method]; ok {
+		return method
+	}
+	upper := strings.ToUpper(method)
+	if _, ok := lookup[upper]; ok {
+		return upper
+	}
+	return "_OTHER"
```
amazingakai: `_OTHER` is used in a few places in here, it should be a constant.

instrumentation/net/http/semconv/util_test.go
```diff
@@ -102,6 +104,73 @@ func TestStandardizeHTTPMethod(t *testing.T) {
...
+	got := parseKnownMethods("GET, PROPFIND,POST")
+	assert.Equal(t, "GET", got["GET"].Value.AsString())
+	assert.Equal(t, "PROPFIND", got["PROPFIND"].Value.AsString())
+	assert.Equal(t, "POST", got["POST"].Value.AsString())
+	_, hasPut := got["PUT"]
+	assert.False(t, hasPut, "override must fully replace defaults")
+
+	empty := parseKnownMethods(" , , ")
+	assert.Empty(t, empty)
+}
```
amazingakai, line 118: We should have a unit test for `knownMethods`, maybe you can use `sync.Once` instead of `sync.OnceValue` to reset it for unit testing.

amazingakai (approved):

Anshumancanrock:
@amazingakai Added a unit test covering defaults + the env override.

Let me know if you want anything tweaked.

docs/configuration.md
```diff
@@ -46,6 +44,21 @@ OTEL_GO_DISABLED_INSTRUMENTATIONS=nethttp ./myapp
...
+
+To fully override that list (for example to restore `QUERY`), set:
+
+```bash
+OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS=CONNECT,DELETE,GET,HEAD,OPTIONS,PATCH,POST,PUT,TRACE,QUERY
+```
+
+The value is a comma-separated, case-sensitive list and **replaces** the defaults
+entirely (it is not additive).
+
```
amazingakai: Since this is specific to `net/http`, maybe we should document this in `net/http/README.md`? WDYT?

instrumentation/net/http/semconv/util.go
```diff
@@ -113,7 +120,54 @@ var MethodLookup = map[string]attribute.KeyValue{
...
+	})
+	return knownMethodsMap
+}
+
+// resetKnownMethodsForTest clears the cached known-methods set so tests can
+// exercise different OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS values.
+func resetKnownMethodsForTest() {
+	knownMethodsOnce = sync.Once{}
+	knownMethodsMap = nil
+}
```
amazingakai: This should be in `util_test.go` + you can maybe also add `t.Helper()` to this function

CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
+# Changelog
```
amazingakai, line 1: See https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/blob/85b786bf6246466602fc07103eb994dc1930c69d/CHANGELOG.md?plain=1 for the full format.

Anshumancanrock:
hi @amazingakai , addressed all the suggestions. Please take a look now. Thanks!

PR #1016, opened 2026-08-08
fix(pkg/runtime): cache parsed instrumentation lists in Instrumented() to cut hot-path allocations

### LINKED ISSUE #1015: bug(pkg/runtime): process shutdown hijack via os.Exit(0), hot-path allocation churn in Instrumented(), and uncoordinated SDK disablement
### Describe the bug

A technical audit of `pkg/runtime` revealed three interconnected defects in runtime process lifecycle, signal handling, and hot-path configuration evaluation that impact production applications compiled with `otelc`:

1. **Process Shutdown Hijack (`os.Exit(0)`) in `setupSignalHandler()`**:
   In `pkg/runtime/setup.go` (lines 320–344), `setupSignalHandler()` registers a signal channel for `SIGINT` and `SIGTERM`. When a signal is caught, the background goroutine flushes OTel SDK providers and calls `os.Exit(0)`. In Go, `os.Exit` does not run `defer` statements in `main()`, nor does it allow the user application's own `signal.Notify` handlers (e.g., HTTP server `Shutdown()`, gRPC graceful stop, K8s pod termination handlers, or database connection pool drainers) to execute. The auto-instrumentation effectively steals process lifecycle control from the host application.

2. **High-Frequency Allocation Churn & OS Env Lookups on Hot Paths**:
   In `pkg/runtime/otel_setup.go` (lines 75–108), `Instrumented(name)` is called by every instrumentation hook (HTTP handlers, SQL queries, gRPC calls, Redis operations) on every single request. On every call, `Instrumented()` executes `os.Getenv("OTEL_GO_ENABLED_INSTRUMENTATIONS")` and `os.Getenv("OTEL_GO_DISABLED_INSTRUMENTATIONS")`, invoking `parseInstrumentationList()` which allocates a new string slice (`result := append(...)`), iterates via `strings.SplitSeq`, and runs `strings.ToLower` and `strings.TrimSpace`. On high-throughput services executing 50,000+ ops/sec, this produces hundreds of thousands of environment lookups and heap allocations per second on hot execution paths.

3. **Inconsistent `OTEL_SDK_DISABLED` Evaluation**:
   When `OTEL_SDK_DISABLED=true` is set, `SetupOTelSDK()` returns early without initializing SDK providers. However, `Instrumented(name)` does not check `OTEL_SDK_DISABLED`. Hook enablers calling `runtime.Instrumented("sql")` continue to return `true` when `OTEL_SDK_DISABLED=true`. Hooks continue executing, creating `HookContext` objects, parsing parameters, and attempting to obtain tracers from `otel.GetTracerProvider()`, incurring unnecessary overhead despite the SDK being explicitly disabled.

### Environment

 - OS: Linux / All
 - Go Version: 1.25 / 1.26
 - Version: v0.0.0 (main branch)

### To Reproduce

1. **Signal Hijack**: Build an HTTP application instrumented with `otelc` that registers a custom `signal.Notify(ch, syscall.SIGTERM)` handler to log `"App shutting down"` and sleep 2 seconds. Send `SIGTERM`. Observe that `"App shutting down"` is never printed because `otelc` calls `os.Exit(0)` immediately after flushing telemetry.
2. **Allocation Churn**: Run a benchmark invoking `runtime.Instrumented("nethttp")` in a tight loop. Observe non-zero heap allocations per op in `go test -benchmem`.
3. **OTEL_SDK_DISABLED**: Set `OTEL_SDK_DISABLED=true` and invoke `runtime.Instrumented("database")`. Observe that it returns `true`.

### Expected behavior

1. **Non-Destructive Signal Handling**: `setupSignalHandler()` flushes OTel telemetry on `SIGINT`/`SIGTERM` and then un-registers or relays the signal so the host application's signal handlers and `main()` function can complete their own graceful shutdown without forcing `os.Exit(0)`.
2. **Zero-Allocation Hot-Path Enablers**: `Instrumented()` uses an atomic, thread-safe configuration cache (`atomic.Pointer` or `sync.RWMutex`) for parsed `OTEL_GO_ENABLED_INSTRUMENTATIONS` and `OTEL_GO_DISABLED_INSTRUMENTATIONS` to perform O(1) lock-free lookups with 0 heap allocations on hot paths.
3. **Symmetrical `OTEL_SDK_DISABLED` Short-Circuiting**: `Instrumented()` returns `false` immediately whenever `OTEL_SDK_DISABLED` is set to `"true"`.

### Additional context

**Proposed High-Level Implementation Plan**:
1. `pkg/runtime/otel_setup.go`: Introduce an `instrumentationState` struct holding an `atomic.Bool` for `sdkDisabled` and a `map[string]bool` cache of enabled/disabled instrumentation names. Re-evaluate state on startup or environment change, updating an `atomic.Pointer[instrumentationState]`. Update `Instrumented(name)` to read from the atomic pointer.
2. `pkg/runtime/setup.go`: Modify `setupSignalHandler()` to remove `os.Exit(0)`. After calling `Shutdown(ctx)`, call `signal.Stop(sigCh)` so subsequent signals pass directly to user handlers.
3. `pkg/runtime/setup_test.go` & `otel_setup_test.go`: Add benchmark tests (`BenchmarkInstrumented`) and unit tests for signal propagation and SDK disablement.

bhuvan-somisetty:
Fixes #1015

## Description

This PR reduces per-call allocation overhead in \`pkg/runtime.Instrumented()\`, which is on the hot path for every instrumented call (database queries, HTTP/gRPC requests, etc).

The signal-handling portion originally proposed here is no longer part of this PR: #759 landed on \`main\` with a proper fix for the same `os.Exit`-in-signal-handler problem (first-signal shutdown, second-signal escalation, with subprocess tests). This branch is rebased onto that, so \`pkg/runtime/setup.go\` has no diff here.

What's left, scoped to \`pkg/runtime/otel_setup.go\`:

1. **Lock-Free Atomic Caching for `Instrumented()`**:
   - Replaces per-hook `parseInstrumentationList` string slice allocations with an `atomic.Pointer[configCache]`.
   - `Instrumented(name)` still reads `OTEL_SDK_DISABLED`, `OTEL_GO_ENABLED_INSTRUMENTATIONS`, and `OTEL_GO_DISABLED_INSTRUMENTATIONS` via `os.Getenv` on every call (this is not cached — see doc comment). What's cached is the parsed-list -> `map[string]struct{}` conversion, so repeated calls skip re-splitting and re-allocating those lists as long as the env vars haven't changed.
   - Benchmarked: for callers who configure `OTEL_GO_ENABLED_INSTRUMENTATIONS`/`OTEL_GO_DISABLED_INSTRUMENTATIONS`, allocations drop from 4-7 down to 1 per call, and calls get 42-58% faster. The unset (default) case gets ~20ns slower per call due to the extra `OTEL_SDK_DISABLED` lookup and atomic load, which is negligible next to the cost of the instrumented call itself.

2. **Symmetrical `OTEL_SDK_DISABLED` Short-Circuiting**:
   - Ensures `Instrumented()` respects `OTEL_SDK_DISABLED=true` and immediately returns `false`, preventing hook execution and parameter parsing when the SDK is explicitly disabled.

3. **Testing**:
   - Adds `TestInstrumented_OTelSDKDisabled` verifying `OTEL_SDK_DISABLED` behavior.
   - Adds `BenchmarkInstrumented` with separate `unset` and `configured-list` cases so the allocation/latency trade-off is measurable rather than asserted.

---

## Checklist

- [x] Tests pass: `make test`

Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>


pkg/runtime/otel_setup.go
```diff
@@ -63,37 +75,72 @@ func SetupOTelSDK() {
...
+//  2. If OTEL_GO_ENABLED_INSTRUMENTATIONS is set, only those instrumentations are enabled
+//  3. Then OTEL_GO_DISABLED_INSTRUMENTATIONS is applied to disable specific ones
+//  4. If neither is set, all instrumentations are enabled by default
 //
 // The instrumentationName should be lowercase (e.g., "nethttp", "grpc").
 func Instrumented(instrumentationName string) bool {
 	name := strings.ToLower(instrumentationName)
 
-	// Check if specific instrumentations are enabled
+	sdkDisabled := os.Getenv("OTEL_SDK_DISABLED")
```
ADITYA-CODE-SOURCE, line 92: This still performs three environment lookups on every Instrumented call. The cache avoids repeated list parsing, but not the os.Getenv calls. Could the performance claim be narrowed and the benchmark compare base and head for unset and configured-list cases?

pkg/runtime/setup.go
```diff
@@ -337,9 +337,9 @@ func setupSignalHandler() {
 				logger.Info("OpenTelemetry SDK shutdown completed successfully")
 			}
 
-			// After shutdown completes, exit cleanly.
-			signal.Reset(signals...)
-			os.Exit(0)
+			// Stop relaying signals to sigCh so host application signal handlers
+			// can proceed naturally without forcing process exit via os.Exit.
+			signal.Stop(sigCh)
```
ADITYA-CODE-SOURCE, line 342: signal.Stop does not replay the signal already consumed, so an application without its own handler may keep running. Also, while Shutdown is blocked, this goroutine is not receiving a second signal even though the channel remains registered. Could we cover first-signal termination and second-signal escalation with subprocess tests?

ADITYA-CODE-SOURCE (changes requested):

pkg/runtime/setup.go
amazingakai: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/759 fixes this same problem

bhuvan-somisetty:
Thanks for the review! Rebased onto main to pick up #759, which already fixes the signal-handling race you flagged, so I dropped my own os.Exit removal in favor of that. For the Instrumented() caching: you're right that the three os.Getenv calls still happen every time, only the parsed-list-to-map conversion was cached — narrowed the doc comment and commit message to say that plainly, and split the benchmark into unset/configured-list cases so the actual payoff is measurable rather than asserted.

vignesh2027:
Ran the base and head comparison asked for above, `df034c1` against `da85f1c`, `-benchtime=2s -count=3`. Figures were stable to well under a percent across runs, so these are the medians.

| case | base ns/op | head ns/op | base allocs | head allocs |
| --- | --- | --- | --- | --- |
| neither list set | 57.2 | 77.4 | 1 | 1 |
| disable list set | 197.6 | 114.5 | 4 | 1 |
| both lists set | 357.6 | 149.4 | 7 | 1 |

Two things stand out.

The caching does what it claims for anyone who configures the lists. Allocations drop from 4 and 7 down to 1, and the two configured cases get roughly 42% and 58% faster. That is a solid win and the numbers back the claim.

The unset case goes the other way, 57 to 77 ns, about 35% slower. That is the default for anyone who has not configured anything, so it is the path most builds take. It reads like the cost of the third `os.Getenv` for `OTEL_SDK_DISABLED`, plus the atomic load and the three string comparisons, landing on a path that previously did two lookups and returned. Worth deciding explicitly whether that is the trade you want, since it makes the majority case slower to make the configured case faster.

On the remaining allocation, it is `strings.ToLower` on the instrumentation name, and it only fires because most callers pass an uppercase key:

```
Instrumented("NETHTTP")       77.05 ns/op   8 B/op   1 allocs/op
Instrumented("logs/logrus")   55.09 ns/op   0 B/op   0 allocs/op
```

`strings.ToLower` returns its input untouched when there is nothing to lower, so the three `logs/*` instrumentations already pay nothing while the other 18 pay an allocation on every call. Not something this PR needs to solve, and I raised the naming side separately in #1114, but it explains why the count stops at 1 rather than 0.

Happy to share the benchmark file if it is useful.

pkg/runtime/otel_setup.go
bhuvan-somisetty: Good catch — fixed. Narrowed the doc comment to say plainly that all three `os.Getenv` calls still happen on every call, and only the parsed-list → map conversion is cached. Also split `BenchmarkInstrumented` into `unset` and `configured-list` cases so the actual payoff (and the unset-case cost) is measurable instead of asserted — see the numbers @vignesh2027 posted below.

pkg/runtime/setup.go
bhuvan-somisetty: You're both right, and #759 landed on main with a proper fix for this (first-signal shutdown + second-signal escalation, tested). Rebased this branch onto main to pick it up and dropped my own `signal.Stop`/os.Exit change entirely in favor of it — `pkg/runtime/setup.go` no longer has any diff in this PR. What's left here is scoped to `Instrumented()` caching in `pkg/runtime/otel_setup.go` only.

bhuvan-somisetty:
Thanks for running these — that's exactly the comparison I was hoping to see.

On the unset-case regression: I'll keep the trade-off as is. The absolute delta is 20ns (57.2 -> 77.4), which is noise next to the actual cost of the DB/HTTP/gRPC call being instrumented (microseconds-to-milliseconds). The two configured-list cases are the ones that were actually reported as a problem, and those are the ones that see the real win (-42% and -58%, 4/7 allocs down to 1). So this cache is optimizing the path that was slow, at a cost that's immaterial in absolute terms on the path that was already fast. I don't think it's worth adding complexity (e.g. skipping the OTEL_SDK_DISABLED lookup based on some other cached signal) to close a 20ns gap.

On the strings.ToLower allocation: agreed that's out of scope here, thanks for filing #1114 separately rather than blocking this one on it.

ADITYA-CODE-SOURCE (changes requested):
Thanks for updating this. The earlier signal-handling concern is resolved now that this branch relies on #759, and the benchmark results support the configured-list caching improvement.

I found one remaining consistency issue in `pkg/runtime/otel_setup.go`: `Instrumented()` trims `OTEL_SDK_DISABLED` before comparing it with `true`, while `SetupOTelSDK()` uses strict `EqualFold` without trimming. That means `OTEL_SDK_DISABLED=" true "` initializes the SDK but disables every instrumentation hook. The OTel boolean definition permits only the case-insensitive string `true` and says implementations must not extend it.

Could this use the same strict parsing as `SetupOTelSDK()` and add test cases for mixed-case `true`, invalid values, and surrounding whitespace?

Verified locally from the `pkg/runtime` module:
- `go test ./...`
- `go test -race ./...`
- `go test ./... -run '^$' -bench BenchmarkInstrumented -benchmem -benchtime=1s -count=3`

vignesh2027:
Confirmed the `OTEL_SDK_DISABLED` divergence on the rebased head, `81fe7c3`. Comparing what `SetupOTelSDK` decides on line 63 against what a hook gets back from `Instrumented`:

| value | SDK skips init | hook enabled | |
| --- | --- | --- | --- |
| `"true"` | true | false | consistent |
| `"TRUE"` | true | false | consistent |
| `" true "` | false | false | diverges |
| `"true "` | false | false | diverges |
| `"1"` | false | true | consistent |
| `"yes"` | false | true | consistent |

So it is only the whitespace forms. With a padded value the SDK initialises normally while every hook reports disabled, which is the worst shape for the failure since telemetry infrastructure comes up and then produces nothing.

`"1"` and `"yes"` already behave correctly on both sides, so dropping the `TrimSpace` in `buildConfigCache` looks like it lines the two up without changing anything else.

Also re-ran the benchmarks against `81fe7c3` since the rebase moved the head. The picture is unchanged from the numbers above, unset around 75 ns and the two configured cases at roughly 111 and 148 ns, all at one allocation. So the rebase did not disturb the result.

pkg/runtime/otel_setup.go
bhuvan-somisetty: Updated buildConfigCache to drop TrimSpace when evaluating OTEL_SDK_DISABLED, lining it up with SetupOTelSDK()'s strict EqualFold check. Also expanded TestInstrumented_OTelSDKDisabled with test cases for mixed-case true, surrounding whitespace, and invalid values.

vignesh2027:
Verify Bundle is red on `787f92b`, and it is just the archive rather than anything you changed. Checked it out and ran `make package`, and the committed `tool/data/otelc-bundle.tgz` comes back different, so the last edit to `pkg/runtime/otel_setup.go` did not carry a regenerated archive with it. The branch is not behind main, so a rebase will not help here.

```
make package
git add tool/data/otelc-bundle.tgz
```

Two things that saved me time on this, in case they are useful. `./.github/scripts/verify-bundle.sh` is exactly what CI runs, so running it locally tells you the answer before pushing. And `make setup-git` installs a merge driver for the archive, which stops it blocking every rebase and merge from then on, since `pkg/` and `instrumentation/` changes always drag it along.

vignesh2027:
Correction on my last comment. The suggestion itself matches what the job prints, but my reasoning for it did not hold and I should say so.

I compared your branch's archive against a local `make package` and reported that it came back different. Then I ran the same check against clean `main`, where the committed archive is known good, and my local build does not reproduce that either. So the comparison shows something about my machine, not about your branch, and I should have run that control first.

What the failing job actually reports is the useful part:

```
Binary files a/tool/data/otelc-bundle.tgz and b/tool/data/otelc-bundle.tgz differ
archive verification failed.
To fix: rebase your branch on main, run "make package", and commit the updated archive.
```

So the fix is the one the job names, and regenerating in an environment matching CI is the part that matters.

The thing I am now unsure about, and would rather ask than assert: if the archive does not reproduce byte for byte across toolchains, then whether `make package` satisfies the check depends on where a contributor runs it. That would make this a recurring cost for anyone whose Go version or platform differs from the runner, rather than a one-off. Is that a known property of the archive, or worth looking at separately? Happy to dig into what actually differs if it is the latter.

vignesh2027:
Ignore my previous comment, the control I ran was broken rather than the build.

I had done that check inside a git worktree, which does not carry the same build state, so `make package` there produced something that never matched anything. Redone in a full checkout: clean `main` reproduces its committed archive byte for byte, and on `787f92b` the archive still comes back different. So the first comment stands, `make package` plus committing the archive is the fix, and the reproducibility worry I raised was my own tooling rather than a property of the project. Sorry for the noise.

bhuvan-somisetty:
Thanks for the thorough follow-up, both fixed:

- **OTEL_SDK_DISABLED whitespace divergence**: dropped the `strings.TrimSpace` in `buildConfigCache` (787f92b), so `Instrumented()` now uses the same strict `strings.EqualFold(sdkDisabled, "true")` as `SetupOTelSDK`. Padded values like `" true "` no longer disable hooks while the SDK initializes — both sides agree now. Expanded `TestInstrumented_OTelSDKDisabled` to cover mixed-case, whitespace, and invalid values explicitly so this can't silently regress.
- **Verify Bundle failure**: ran `make package` and committed the regenerated `tool/data/otelc-bundle.tgz` (826ab8c). You were right that it hadn't picked up the `pkg/runtime/otel_setup.go` changes. I also reran `./.github/scripts/verify-bundle.sh`-equivalent locally against a fresh clone of `main` first as a control to confirm the toolchain reproduces the committed archive byte-for-byte before trusting the diff on this branch — it does, so this was a genuine staleness issue, not a reproducibility artifact.

Appreciate you running down the reproducibility question on your end too — good to have that confirmed as a one-off rather than a recurring cost.

ADITYA-CODE-SOURCE (approved):
The latest changes resolve my remaining concern. Instrumented now applies the same strict OTEL_SDK_DISABLED parsing as SetupOTelSDK, and the expanded tests cover mixed-case true values, whitespace-padded values, and invalid values. The generated bundle is updated and the full check suite is green. No remaining concerns from my side.

bhuvan-somisetty:
Rebased onto latest `main` and resolved the bundle merge conflict. All unit tests and CI checks are passing green.

pkg/runtime/otel_setup.go
```diff
@@ -63,37 +75,78 @@ func SetupOTelSDK() {
 // Instrumented checks if instrumentation is enabled via environment variables.
 //
 // Environment variables (following OTel JS pattern):
+//   - OTEL_SDK_DISABLED: If set to "true", all instrumentations are disabled
```
darccio: The code uses `strings.EqualFold(sdkDisabled, "true")`, so this is
case-insensitive — same as `OTEL_SDK_DISABLED`'s doc on `SetupOTelSDK`
above (line 30). Worth matching that wording here too.

```suggestion
//   - OTEL_SDK_DISABLED: If set to the case-insensitive string "true", all instrumentations are disabled
```

pkg/runtime/otel_setup.go
```diff
@@ -63,37 +75,78 @@ func SetupOTelSDK() {
...
+	cached := instConfigCache.Load()
+	if cached == nil || cached.sdkDisabled != sdkDisabled || cached.enabledRaw != enabledList ||
+		cached.disabledRaw != disabledList {
+		cached = buildConfigCache(sdkDisabled, enabledList, disabledList)
+		instConfigCache.Store(cached)
+	}
+
+	if cached.isSdkDisabled {
+		return false
+	}
```
darccio: `name := strings.ToLower(instrumentationName)` runs before the
`OTEL_SDK_DISABLED` short-circuit, so the one remaining per-call
allocation (`ToLower` on uppercase instrumentation names) is paid even
when the SDK is disabled and the result is discarded immediately.
Moving it after the short-circuit avoids that allocation on the
disabled path.

```suggestion

	sdkDisabled := os.Getenv("OTEL_SDK_DISABLED")
	enabledList := os.Getenv("OTEL_GO_ENABLED_INSTRUMENTATIONS")
	disabledList := os.Getenv("OTEL_GO_DISABLED_INSTRUMENTATIONS")

	cached := instConfigCache.Load()
	if cached == nil || cached.sdkDisabled != sdkDisabled || cached.enabledRaw != enabledList ||
		cached.disabledRaw != disabledList {
		cached = buildConfigCache(sdkDisabled, enabledList, disabledList)
		instConfigCache.Store(cached)
	}

	if cached.isSdkDisabled {
		return false
	}

	name := strings.ToLower(instrumentationName)
```

pkg/runtime/otel_setup_test.go
```diff
@@ -92,3 +92,54 @@ func TestInstrumented(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run("OTEL_SDK_DISABLED="+tt.value, func(t *testing.T) {
+			t.Setenv("OTEL_SDK_DISABLED", tt.value)
+			assert.Equal(t, tt.expected, Instrumented("nethttp"))
+			assert.Equal(t, tt.expected, Instrumented("grpc"))
+		})
+	}
+}
+
+func BenchmarkInstrumented(b *testing.B) {
```
darccio, line 139: The review thread's benchmark comparison (base vs. head, disable-list-
only case) showed the largest relative win here: 4 allocs -> 1, ~42%
faster. This benchmark only keeps `unset` and `configured-list` (both
lists set), so that case isn't independently exercised going forward —
if disabled-list-only handling regresses without touching the enabled-
list path, this benchmark wouldn't catch it. `TestInstrumented`'s
"explicitly disabled" case still covers correctness for that path,
just not the allocation-count story.

```suggestion
	b.Run("disabled-list", func(b *testing.B) {
		b.Setenv("OTEL_GO_DISABLED_INSTRUMENTATIONS", "grpc,gin")
		b.ReportAllocs()
		for i := 0; i < b.N; i++ {
			_ = Instrumented("nethttp")
		}
	})

	b.Run("configured-list", func(b *testing.B) {
```

darccio (changes requested):

bhuvan-somisetty:
Thanks for the catch. Fixed the doc wording, moved the ToLower after the disabled check, and added the disabled list benchmark case. Also found the enabled list was still triggering on a commas only value so fixed that too. All tests pass locally.

darccio (approved):

PR #1020, opened 2026-08-08
test(integration): assert server span is the trace root in propagation tests

### LINKED ISSUE #1019: test(integration): missing trace-root assertions in propagation tests
## Problem

Three `//go:build integration` propagation tests under `test/integration/` assert
parent/child span relationships but never assert that the *initiating* span is
actually the trace root. This is the same coverage gap that was fixed for the
`test/e2e/` suite in #963 (fixing #960) — it just wasn't fixed for the sibling
`test/integration/` suite, which has near-identical test bodies.

Without this assertion, a bug that gives the root span a bogus non-empty
`ParentSpanID` (e.g. a stray injected parent context, or a propagator that
incorrectly extracts a parent from an unrelated header) would slip through
these tests undetected, since only the *relative* parent/child links are
checked, never that the chain terminates.

The codebase already has precedent for this exact assertion style in
`test/integration/otel_sdk_test.go:46`:
```go
require.True(t, workerSpan.ParentSpanID().IsEmpty(), ...)
```

## Affected tests

- `test/integration/grpcserver_dbclient_test.go` — `grpcServerSpan` (the gRPC
  frontend) is the trace-initiating span (the test calls it directly via an
  uninstrumented `grpc.NewClient`), but nothing asserts it's the root.
- `test/integration/httpserver_dbclient_test.go` — same gap for
  `httpServerSpan`.
- `test/integration/ginserver_httpclient_test.go` — same gap for
  `ginServerSpan`.

(Note: unlike the `test/e2e/` versions, these `test/integration/` tests have
no separate instrumented client app — the test process itself makes the
initial call — so the root span here is the *server* span, not a client
span.)

## Proposed fix

Add one `require.True(t, <span>.ParentSpanID().IsEmpty(), "<span> must be the
trace root")` assertion to each of the three tests, following the exact
pattern used in #963 and in `otel_sdk_test.go`.

SatyamPandey-07:
## test(integration): assert server span is the trace root in propagation tests

Fixes #1019 

### What

Adds a `require.True(t, <span>.ParentSpanID().IsEmpty(), ...)` assertion to
three `test/integration/` propagation tests, mirroring the fix already
applied to the `test/e2e/` suite in #963:

- `test/integration/grpcserver_dbclient_test.go` — asserts `grpcServerSpan`
  (the gRPC frontend) is the trace root.
- `test/integration/httpserver_dbclient_test.go` — asserts `httpServerSpan`
  is the trace root.
- `test/integration/ginserver_httpclient_test.go` — asserts `ginServerSpan`
  is the trace root.

### Why

These tests previously only asserted *relative* parent/child span links
(e.g. "SQL client's parent is the gRPC server"), never that the chain
actually terminates at the initiating span. A bug that attached a bogus
non-empty `ParentSpanID` to the root span (e.g. an incorrectly extracted
parent from an unrelated context/header) would pass unnoticed. The assertion
style matches existing precedent in `test/integration/otel_sdk_test.go:46`
and the `test/e2e/` fix from #963.

Unlike the `test/e2e/` versions of these tests, `test/integration/` has no
separate instrumented client app driving the call — the test process itself
is the (uninstrumented) client — so the root span here is the *server* span
rather than a client span.

### Test plan

- [x] `go build -tags=integration ./integration/...` (from `test/` module)
- [x] `go vet -tags=integration ./integration/...`
- [x] `gofmt -l` on changed files (clean)
- [ ] `go test -tags=integration ./integration/...` — requires Docker
      (Postgres/Kafka fixtures) not available in this environment; relies on
      CI's integration job to execute.

### Scope

Test-only change, 3 lines added across 3 files. No production code touched.


SatyamPandey-07:
@darccio pls review the pr

darccio (approved):

PR #1018, opened 2026-08-08
fix(genai): set error.type on transport failures

Anshumancanrock:
## Description

On transport failures (`next(req)` returns `err != nil`), OpenAI (v1/v2/v3) and Anthropic GenAI middleware already set span status and call `RecordError`, but never set `error.type`. HTTP 4xx/5xx paths already do.

This sets `error.type` via `otelsemconv.ErrorType(err)` on that path and asserts it in the existing transport-error tests. Bundle regenerated.

## Motivation

GenAI span semconv marks `error.type` as conditionally required when the operation ends in an error. `RecordError` only adds an exception event , it does not set the span attribute and GenAI middleware suppresses nested HTTP client instrumentation, so nothing else fills this in.

Without it, transport/dial failures can't be grouped the same way as HTTP status errors.

Fixes #1017

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [ ] Code formatted: `make format`
- [ ] Linters pass: `make lint`
- [ ] Tests pass: `make test`
- [x] Tests added for new functionality
- [x] Tests follow [testing guidelines](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/testing.md)
- [ ] Documentation updated (if applicable)
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation))
- [ ] This PR has content that I did not fully write myself.
- [ ] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

PR #1021, opened 2026-08-08
fix(instrumentation): set error.type on GenAI transport failures

### LINKED ISSUE #1017: bug(instrumentation): GenAI middleware omits error.type on transport failures
### Describe the bug

OpenAI and Anthropic GenAI middleware set span status + `RecordError` on transport failures (`next(req)` returns `err != nil`), but never set `error.type`.

HTTP 4xx/5xx responses do set `error.type`. Transport / dial failures don't. Same middleware, inconsistent error paths.

```go
resp, err := next(req)
if err != nil {
    span.RecordError(err)
    // missing: error.type
    span.End()
    return resp, err
}
```

Affects:

- `instrumentation/github.com/openai/openai-go` (v1 / v2 / v3)
- `instrumentation/github.com/anthropics/anthropic-sdk-go`

GenAI span semconv marks `error.type` as conditionally required when the operation ends in an error. `RecordError` only adds an exception event, it does not set the span attribute. GenAI middleware also suppresses nested HTTP client instrumentation, so nothing else fills this in.

### To Reproduce

1. Hit OpenAI or Anthropic middleware with a chat/messages request that starts a GenAI span (valid model in the body).
2. Make `next(req)` fail with a transport error (e.g. connection refused / dial failure) instead of returning an HTTP response.
3. Inspect the ended GenAI span.

See error: span status is `Error` and there is an `exception` event, but no `error.type` attribute.

Existing coverage already exercises this path without asserting the attribute:

- OpenAI: `TestOtelMiddleware_NextError`
- Anthropic: `TestOtelMiddleware_TransportError`

### Expected behavior

Transport failures should set `error.type` on the GenAI span (low-cardinality type string, same idea as `HTTPClientErrorType` in `instrumentation/net/http/semconv/client.go`), consistent with the HTTP error path and GenAI semconv.

vignesh2027:
Fixes #1017

## Problem

OpenAI and Anthropic GenAI middleware set span status and call `RecordError` when `next(req)` returns an error, but never set `error.type`. The HTTP 4xx/5xx path in the same middleware does set it, so the two error paths were inconsistent.

GenAI semantic conventions make `error.type` conditionally required when an operation ends in an error. `RecordError` only adds an exception event, not the span attribute, and the middleware suppresses nested HTTP client instrumentation, so nothing else fills it in.

## Change

Adds an `ErrorType` helper to each GenAI semconv package that derives the fully qualified error type name and falls back to `_OTHER`. This mirrors the existing convention in `instrumentation/net/http/semconv`, so `error.type` stays consistent across instrumentations. Each instrumentation is its own Go module, so the helper cannot be shared.

Wired into the transport-error path of all four middlewares:

- `instrumentation/github.com/openai/openai-go` (v1, v2, v3)
- `instrumentation/github.com/anthropics/anthropic-sdk-go`

## Tests

- `TestErrorType` in each semconv package covers the nil, builtin-pointer, named-type and wrapped-error cases, plus a typed-nil case that would otherwise panic.
- The three existing `TestOtelMiddleware_NextError` tests now assert `error.type` and the error status code.
- New `TestOtelMiddleware_TransportErrorSetsErrorType` for anthropic-sdk-go.

I confirmed each test fails without the middleware change and passes with it, so they exercise the fix rather than the surrounding behaviour.

`go test ./...` passes in all four modules, and `golangci-lint` (v2.11.3, repo config) reports no new findings.


Anshumancanrock:
hii @vignesh2027 , this PR is a duplicate of this #1018

vignesh2027:
You're right, thanks for the pointer — I missed #1018 when I picked this up. Yours was first and covers the same ground, so I'll close mine.

One thing from mine you might find useful: a typed-nil error (`var e *net.OpError; var err error = e`) is non-nil as an interface but has no usable dynamic value, so `reflect.TypeOf` needs a guard before it's dereferenced. Worth a test case either way. Your call entirely.

PR #904, opened 2026-08-01
fix(tool/instrument): resolve undefined idx in collectArguments blank-receiver branch

bhuvan-somisetty:
## Description

This PR originally fixed unguarded type assertions in `GetKeyData`/`SetKeyData`/`HasKeyData`:

> In both `tool/internal/instrument/impl.tmpl` (`HookContextImpl`) and `pkg/hook/hooktest/mock_hook_context.go` (`MockHookContext`), the `HookContext` methods `GetKeyData`, `SetKeyData`, and `HasKeyData` performed direct, unguarded type assertions `c.data.(map[string]interface{})`, causing an unhandled interface conversion panic when a `Before` hook stored a non-map value via `SetData`.

That fix has since landed on `main` independently via #911 (merged 2026-08-07, commit 884eb3cd4) — a different PR by the same author whose diff happened to also touch `impl.tmpl` and `mock_hook_context.go`, applying the equivalent safe-assertion pattern (condition inverted, but behaviorally identical) plus an equivalent test. After merging `main` into this branch, every file originally touched here (`mock_hook_context.go`, `mock_hook_context_test.go`, `impl.tmpl`, and the 31 golden files generated from it) is now byte-identical to `main`.

While resolving the merge conflicts, I found `main` currently fails to build: `collectArguments`'s blank-receiver branch in `tool/internal/instrument/apply_func.go` referenced an undeclared `idx`, left over from before #1035 refactored the rest of the function to use a `next(...)` synthetic namer. This PR now carries that fix instead:

1. `tool/internal/instrument/apply_func.go`: use `next(ignoredParam)` in the blank-receiver branch, consistent with the rest of `collectArguments`.
2. `tool/internal/instrument/apply_func_test.go`: update a stale `"_ignoredParam0"` expectation (predating the #1035 refactor) to `syntheticParam(0)`, matching the sibling test cases.

## Motivation

Originally filed to fix #903. That issue is still open, but its underlying bug is already resolved on `main` via #911's incidental changes — may be worth closing separately. This PR's remaining value is the `main` build-breakage fix above, found as a byproduct of resolving these conflicts. See [this comment](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/904#issuecomment-5256833963) for the full investigation.

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format`
- [x] Linters pass: `make lint`
- [x] Tests pass: `make test`
[...]

PR #1023, opened 2026-08-08
feat(instrumentation/http): enable metrics collection for HTTP client…

ipsitapp8:
## Description

`instrumentation/net/http/semconv/client.go` already defines the metric instruments for the HTTP client, but nothing ever used them. The hook called the package-level helpers (`HTTPClientRequestTraceAttrs`, `HTTPClientStatus`, and so on), and those go through `defaultHTTPClient`, which is built with a nil meter. So the instruments existed and were never recorded to.

This PR wires them up:

- `initInstrumentation` now builds a `semconv.HTTPClient` from the global meter provider, the same way `google.golang.org/grpc/client` does.
- `BeforeRoundTrip` stores `req.ContentLength` alongside the span and start time.
- `AfterRoundTrip` calls `RecordMetrics`, so `http.client.request.duration`, `http.client.request.body.size` and `http.client.response.body.size` get recorded.
- The trace attribute calls move from the package-level helpers to the instance methods, since we now have an instance.

Only three of the six instruments are recorded. `active_requests`, `open_connections` and `connection.duration` need connection-level information that isn't available from inside `RoundTrip` — that would need `httptrace` hooks, which felt like a separate change. `NewHTTPClient` still creates all six, which is existing behaviour I didn't touch, but I'm happy to gate the unused three if you'd rather not register instruments that are never written to.

## Motivation

HTTP is the most used instrumentation here and it only emits traces today, while gRPC already emits both. The metric side was already written, just not connected.

No existing issue covers this — @Labreo asked about that above. Happy to open one if you'd prefer it tracked there first.

## Notes

Three tests added: metrics recorded on success, behaviour when the request fails, and graceful degradation when there is no meter (so a nil meter can't panic in a user's app).

CI is currently red on this branch because `main` doesn't compile (`undefined: idx` in `tool/internal/instrument/apply_func.go`). That's unrelated to this PR — #904 fixes it, and I'll rebase once that lands.

---

## Checklist

- [x] Tests pass: `make test`
- [ ] Documentation updated (if applicable) — n/a, no user-facing config change
- [ ] OpenTelemetry Registry updated (if applicable) — n/a, no new instrumentation

[^I-know-my-stuff]:


Labreo:
Hello there @ipsitapp8 .You need to sign in via easyCLA otherwise your PR wont be able to be merged even if it is right.I see that you have already done so but you need to link your Github to your account.You can do this via getting in touch with the support at https://easycla.lfx.linuxfoundation.org/#/ and by chatting to their support bot.

Labreo:
Please also link the issue No in question as it helps the maintainers and contributors to better judge a PR.

amazingakai:
Hello @ipsitapp8, thank you for putting this together!                                                                                                                                        
                                                                                                                                                                                             
There is actually another in-progress PR #755 that completely deletes the manual metric instruments you are wiring up here, and replaces them with the official upstream typed builders from `semconv/v1.37.0/httpconv`. Because of that, this PR will likely be completely superseded.  
                                                                                                                                                                                             
Also, it looks like your Git client or editor is configured to use Windows line endings (CRLF / `\r\n`). If you look at the "Files changed" tab, it shows almost the entire file being deleted and re-added because Git sees the line endings changing from standard Unix LF to CRLF.

ipsitapp8:
@amazingakai I looked at #755. It does everything this PR does and more: same `NewHTTPClient(meter)` wiring in `initInstrumentation`, plus `AddActiveRequests(+1/-1)` around the round trip, recording on the error path when `res == nil`, and `error.type` attributes for 4xx/5xx. It also refactors the semconv layer, which I didn't touch. I thought active_requests needed httptrace hooks, but stashing an `attribute.Set` in the hook context does it.

I'll leave this open since #755 is conflicting and unreviewed, in case that one stalls. Close it whenever.

The three tests here (`TestHTTPClientMetrics`, `_ErrorCase`, `_NilMeter`) are useful whichever version lands. The nil-meter one catches a panic when no `MeterProvider` is configured. I can port them to #755 if @tim48-robot wants.

The CRLF came from an earlier force-push. Fixed now, diff is +198/-13 instead of +741/-556 and `git ls-files --eol` shows `i/lf` on all three text files. Content is the same. I also squashed to one commit.

The CI note about `main` not compiling is out of date. #904 landed and this is rebased on top of it.

instrumentation/net/http/client/client_hook.go
```diff
@@ -143,7 +168,7 @@ func AfterRoundTrip(ictx hook.HookContext, res *http.Response, err error) {
 	if err != nil {
 		span.RecordError(err)
 		span.SetStatus(codes.Error, err.Error())
-		span.SetAttributes(semconv.HTTPClientErrorType(err))
+		span.SetAttributes(httpClient.ErrorType(err))
 		logger.Debug("AfterRoundTrip called with error", "error", err)
 	}
```
darccio: `RecordMetrics` is only called inside the `if res != nil` branch above (line 154). When the
round trip fails outright (`res == nil`, `err != nil`), no metric is recorded at all — not even
`http.client.request.duration`, even though `start` is already available in `ictx`. Failed
requests currently contribute zero duration samples, which skews any latency dashboard built on
this histogram. Consider recording duration (and `error.type`) here too, mirroring what upstream
`otelhttp` does on the error path.

instrumentation/net/http/client/client_hook_test.go
```diff
@@ -401,3 +405,159 @@ func TestClientEnabler(t *testing.T) {
...
+	// Setup meter with manual reader
+	reader := sdkmetric.NewManualReader()
+	mp := sdkmetric.NewMeterProvider(sdkmetric.WithReader(reader))
+	otel.SetMeterProvider(mp)
+	t.Cleanup(func() { _ = mp.Shutdown(context.Background()) })
+
+	// Create test request
+	req, _ := http.NewRequest("POST", "http://example.com/api/data", nil)
+	req.ContentLength = 100 // Request size
+
```
darccio: This line is a tab-only "blank" line — `gofmt -l client_hook_test.go` flags the file. The PR
checklist claims `make format` was run; CI's format check will fail on this file as committed.

```suggestion

```

instrumentation/net/http/client/client_hook_test.go
```diff
@@ -401,3 +405,159 @@ func TestClientEnabler(t *testing.T) {
...
+	sr, _ := setupTestTracer(t)
+
+	reader := sdkmetric.NewManualReader()
+	mp := sdkmetric.NewMeterProvider(sdkmetric.WithReader(reader))
+	otel.SetMeterProvider(mp)
+	t.Cleanup(func() { _ = mp.Shutdown(context.Background()) })
+
+	// Create test request
+	req, _ := http.NewRequest("GET", "http://example.com/path", nil)
+
```
darccio: Same tab-only blank line as above; same `gofmt` failure.

```suggestion

```

instrumentation/net/http/client/client_hook_test.go
```diff
@@ -401,3 +405,159 @@ func TestClientEnabler(t *testing.T) {
...
+	assert.Equal(t, codes.Error, spans[0].Status().Code)
+
+	// Collect metrics - should have duration but not sizes
+	rm := &metricdata.ResourceMetrics{}
+	collectErr := reader.Collect(context.Background(), rm)
+	require.NoError(t, collectErr)
+
+	// With error and no response, we should not record metrics
+	// (no response means no status code, size, etc.)
+	// This is expected behavior
```
darccio: This block collects `rm` but never asserts on it — the test name and surrounding comments claim
"should have duration but not sizes" / "we should not record metrics," but nothing here checks
metric names or absence of the size histograms. As written, this test passes identically whether
or not `RecordMetrics` runs on the error path, so it can't catch a regression here (see the
related note on `client_hook.go:167-173` about metrics being skipped entirely when `res == nil`).

Suggest asserting explicitly, e.g.:

    metricNames := make(map[string]bool)
    for _, sm := range rm.ScopeMetrics {
        for _, m := range sm.Metrics {
            metricNames[m.Name] = true
        }
    }
    assert.False(t, metricNames["http.client.request.body.size"])
    assert.False(t, metricNames["http.client.response.body.size"])

That would document today's actual behavior and turn amber the moment someone adds duration
recording to the error path without also handling sizes correctly.

darccio:
Closing this PR in favour of #755. This PR has `gofmt` issues and other functionality issues that are already covered in the former.

PR #1026, opened 2026-08-09
fix(instrumentation): add endpoint attributes to GenAI client spans

### LINKED ISSUE #1024: GenAI spans miss server.address and server.port when HTTP instrumentation is suppressed
### Description

OpenAI (v1/v2/v3) and Anthropic create a GenAI client span and suppress the nested `net/http` client span to avoid duplicates. That HTTP span is where `server.address` / `server.port` would normally come from, and the GenAI span does not set them either.

So for instrumented GenAI requests, endpoint information is missing. The request URL is already available in the middleware (it is used for provider detection).

### Expected

Keep HTTP suppression, and put the endpoint attrs on the GenAI span:

- `server.address` from the parsed hostname (`req.URL.Hostname()`)
- `server.port` when required by GenAI SemConv

Cover OpenAI v1/v2/v3 + Anthropic, including custom base URLs and non-default ports.

abhishekkamble12:
Fixes #1024

### Description
Extracted endpoint attributes from `req.URL` to ensure `server.address` and `server.port` are populated on GenAI client spans prior to HTTP instrumentation suppression. 

- Added `server.address` using `req.URL.Hostname()`.
- Added `server.port` utilizing `req.URL.Port()`, converting to `int`. Standard default ports (80 for HTTP, 443 for HTTPS) are safely omitted in accordance with OpenTelemetry Server Semantic Conventions.
- Applies to OpenAI (v1, v2, v3) and Anthropic middlewares.
- Added regression tests covering default endpoints and custom base URLs with non-default ports.

### Checklist
- [x] Code compiles and tests pass locally (`make test`)
- [x] Follows the repository's contributing guidelines


darccio:
@abhishekkamble12 Please check the failed CI checks.

abhishekkamble12:
@darccio I've resolved the merge conflicts and updated the generated bundles. All CI checks are green now and it's ready for review

darccio:
@abhishekkamble12 Thaks, unfortunately the high traffic of merged PRs caused a new conflict. We are working on improving the situation here: #585

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -71,6 +72,53 @@ func getProviderName(host string) string {
...
+func isDefaultPort(scheme string, port int) bool {
+	switch scheme {
+	case "https":
+		return port == 443
+	case "http":
+		return port == 80
+	default:
+		return port == 443 || port == 80
+	}
+}
```
darccio, line 120: These ~45 lines (`endpointAttributes` + `isDefaultPort`) are byte-identical in `openai-go`, `openai-go/v2`, and `openai-go/v3`, and the 8-case `TestEndpointAttributes` table is duplicated alongside them.

Unlike `providerMapping`/`getProviderName` — which are version-coupled lookup tables — this is pure, version-agnostic URL parsing with no dependency on the OpenAI SDK surface. And there is already in-repo precedent for sharing across the three versions: `internal/streaming` is its own module, consumed by all three via `replace` directives:

```
openai-go/go.mod:74     => ./internal/streaming
openai-go/v2/go.mod:74  => ../internal/streaming
openai-go/v3/go.mod:74  => ../internal/streaming
```

Consider moving these two helpers (and their test table) into a shared `internal/` module the same way, so a future semconv correction — say, the default-port policy below — is a one-line change instead of three. Anthropic reasonably stays separate, since it is a different vendor tree with no dependency on the openai internal module.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -71,6 +72,53 @@ func getProviderName(host string) string {
...
+		portStr = req.URL.Port()
+	}
+	if host == "" && req.Host != "" {
+		if h, p, err := net.SplitHostPort(req.Host); err == nil {
+			host = h
+			portStr = p
+		} else {
+			host = req.Host
+		}
+	}
```
darccio, line 89: Neither the `req.URL != nil` guard nor the `req.Host` fallback can trigger on the middleware path: `classifyOperation(req.URL.Path)` (line 165) and `getProviderName(req.URL.Host)` (line 171) both dereference `req.URL` unconditionally before this helper runs, so a nil `req.URL` would already have panicked upstream.

That is fine for a standalone helper, but as written it reads as though `req.Host` is a real production source of the address, and two of the eight test cases assert behavior the middleware can never reach. A short comment makes the intent explicit:

```suggestion
	// req.URL is always non-nil on the middleware path (classifyOperation and
	// getProviderName dereference it first). The nil guard and req.Host fallback
	// below keep this helper safe for direct callers and unit tests.
	if req.URL != nil {
		host = req.URL.Hostname()
		portStr = req.URL.Port()
	}
	if host == "" && req.Host != "" {
		if h, p, err := net.SplitHostPort(req.Host); err == nil {
			host = h
			portStr = p
		} else {
			host = req.Host
		}
	}
```

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -71,6 +72,53 @@ func getProviderName(host string) string {
...
+}
+
+func isDefaultPort(scheme string, port int) bool {
+	switch scheme {
+	case "https":
+		return port == 443
+	case "http":
+		return port == 80
+	default:
+		return port == 443 || port == 80
```
darccio, line 118: The `default` branch treats both 80 and 443 as default, which is a different rule from the two scheme-specific cases — under `https` a port of 80 is *not* default, and under `http` 443 is not. Worth a note on why the looser rule is acceptable here, namely that the branch is only reachable when the scheme is unknown (the `req.Host` fallback), so http and https cannot be distinguished:

```suggestion
	default:
		// Scheme is only empty on the req.Host fallback path, where http and
		// https are indistinguishable - treat both well-known ports as default.
		return port == 443 || port == 80
```

darccio (changes requested):
Confirmed this fixes #1024: the endpoint attributes are appended to `baseAttrs` before `tracer.Start`, hence before `runtime.SuppressHTTPClientInstrumentation`, so they land on the GenAI span that survives.

Test coverage is solid — an 8-case table per module plus an integration test asserting both attributes end up on a real span. Requesting some changes to improve quality.

PR #1027, opened 2026-08-09
fix(instrumentation): record penalties on openai completion requests

Priyanshu-byte-coder:
## Description

`parseCompletionRequest` parsed only `model`, `max_tokens`, `temperature` and `top_p`. `CompletionNewParams` also carries `FrequencyPenalty` and `PresencePenalty`, serialised as `frequency_penalty` and `presence_penalty`:

```go
// openai-go/v3@v3.49.0 completion.go
FrequencyPenalty param.Opt[float64] `json:"frequency_penalty,omitzero"`
PresencePenalty  param.Opt[float64] `json:"presence_penalty,omitzero"`
```

`parseChatRequest` already records both. So for two otherwise identical requests, the chat span reports the sampling penalties and the legacy completions span silently does not.

Both attributes are already defined in `semconv` (`GenAIRequestFrequencyPenalty`, `GenAIRequestPresencePenalty`) and already emitted on the chat path, so this introduces no new attribute — the completions path just stopped short of the fields it was already shaped to read.

Parsed as `*float64` to match the surrounding fields, so an omitted penalty stays absent rather than being reported as `0` — `0` is a meaningful value for both parameters, and `omitempty` on a non-pointer would make "explicitly zero" and "not sent" indistinguishable.

Applied to v1, v2 and v3, which all had the same gap.

## Motivation

Anyone comparing chat and completions traffic on `gen_ai.request.frequency_penalty` or `gen_ai.request.presence_penalty` sees completions requests as if the penalties were never set. The gap is invisible in the span — nothing signals the attribute was dropped rather than unset.

I found this while reading the request-parsing paths, after building a benchmark harness that measures how tool and instrumentation shape affects an LLM agent's behaviour, where request-side sampling parameters are exactly the sort of thing you need recorded to compare runs.

## Testing

Two tests per version:

- `TestParseCompletionRequest_Penalties` — a completions body with `frequency_penalty: 0.5` and `presence_penalty: -0.25` produces both attributes with those values.
- `TestParseCompletionRequest_PenaltiesOmitted` — a body without them produces neither, so absent stays absent.

Confirmed these are real regression tests. Against unpatched `parseCompletionRequest`:

```
--- FAIL: TestParseCompletionRequest_Penalties (0.00s)
    Error: map[attribute.Key]attribute.Value{} does not contain "gen_ai.request.frequency_penalty"
```

With the change, all three modules pass:

```
ok  go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go      7.111s
ok  go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go/v2   6.387s
ok  go.opentelemetry.io/otelc/instrumentation/github.com/openai/openai-go/v3   6.950s
```

`make` is not available on my machine (Windows), so I ran `go test ./...` per module and verified every changed file is `gofmt` clean rather than running `make format` / `make test`. Worth a second pair of eyes on the lint job.

### DIFF
--- a/instrumentation/github.com/openai/openai-go/middleware.go
+++ b/instrumentation/github.com/openai/openai-go/middleware.go
@@ -266,10 +266,12 @@ func parseChatRequest(body []byte) (string, []attribute.KeyValue) {
 
 func parseCompletionRequest(body []byte) (string, []attribute.KeyValue) {
 	var req struct {
-		Model       string   `json:"model"`
-		MaxTokens   *int64   `json:"max_tokens,omitempty"`
-		Temperature *float64 `json:"temperature,omitempty"`
-		TopP        *float64 `json:"top_p,omitempty"`
+		Model            string   `json:"model"`
+		MaxTokens        *int64   `json:"max_tokens,omitempty"`
+		Temperature      *float64 `json:"temperature,omitempty"`
+		TopP             *float64 `json:"top_p,omitempty"`
+		FrequencyPenalty *float64 `json:"frequency_penalty,omitempty"`
+		PresencePenalty  *float64 `json:"presence_penalty,omitempty"`
 	}
 	if err := json.Unmarshal(body, &req); err != nil {
 		return "", nil
@@ -284,6 +286,12 @@ func parseCompletionRequest(body []byte) (string, []attribute.KeyValue) {
 	}
 	if req.TopP != nil {
 		attrs = append(attrs, semconv.GenAIRequestTopP(*req.TopP))
+	}
+	if req.FrequencyPenalty != nil {
+		attrs = append(attrs, semconv.GenAIRequestFrequencyPenalty(*req.FrequencyPenalty))
+	}
+	if req.PresencePenalty != nil {
+		attrs = append(attrs, semconv.GenAIRequestPresencePenalty(*req.PresencePenalty))
 	}
 	return req.Model, attrs
 }
--- a/instrumentation/github.com/openai/openai-go/middleware.go
+++ b/instrumentation/github.com/openai/openai-go/middleware.go
@@ -294,10 +294,12 @@ func parseChatRequest(body []byte) (string, []attribute.KeyValue) {
 
 func parseCompletionRequest(body []byte) (string, []attribute.KeyValue) {
 	var req struct {
-		Model       string   `json:"model"`
-		MaxTokens   *int64   `json:"max_tokens,omitempty"`
-		Temperature *float64 `json:"temperature,omitempty"`
-		TopP        *float64 `json:"top_p,omitempty"`
+		Model            string   `json:"model"`
+		MaxTokens        *int64   `json:"max_tokens,omitempty"`
+		Temperature      *float64 `json:"temperature,omitempty"`
+		TopP             *float64 `json:"top_p,omitempty"`
+		FrequencyPenalty *float64 `json:"frequency_penalty,omitempty"`
+		PresencePenalty  *float64 `json:"presence_penalty,omitempty"`
 	}
 	if err := json.Unmarshal(body, &req); err != nil {
 		return "", nil
@@ -312,6 +314,12 @@ func parseCompletionRequest(body []byte) (string, []attribute.KeyValue) {
 	}
 	if req.TopP != nil {
 		attrs = append(attrs, semconv.GenAIRequestTopP(*req.TopP))
+	}
+	if req.FrequencyPenalty != nil {
+		attrs = append(attrs, semconv.GenAIRequestFrequencyPenalty(*req.FrequencyPenalty))
+	}
+	if req.PresencePenalty != nil {
+		attrs = append(attrs, semconv.GenAIRequestPresencePenalty(*req.PresencePenalty))
 	}
 	return req.Model, attrs
 }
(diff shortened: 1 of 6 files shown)


NameHaibinZhang (approved):

NameHaibinZhang:
LGTM!

Priyanshu-byte-coder:
Thanks for the review.

Rebased onto main — the branch was 25 commits behind and main has since restructured this instrumentation (the new `internal/streaming/` package, `go.mod` changes, and the anthropic-sdk-go addition). Re-checked that the change is still needed: `parseCompletionRequest` on current main still reads only `model`, `max_tokens`, `temperature` and `top_p` in all three versions, so the gap is unchanged. Tests pass against the rebased tree:

```
v1  ok  .../instrumentation/github.com/openai/openai-go       9.110s
v2  ok  .../instrumentation/github.com/openai/openai-go/v2    9.186s
v3  ok  .../instrumentation/github.com/openai/openai-go/v3    9.247s
```

On the **Verify Bundle** failure: that is expected rather than a defect in the change — the job triggers on any `instrumentation/**` edit and `tool/data/otelc-bundle.tgz` has not been regenerated to match.

I would rather not regenerate it myself. `make package` runs `go mod tidy` inside the temp trees before building the archive, so the bytes depend on the toolchain; I am on Windows without `make`, and the Go version I have locally is not the 1.25.0 that `go-version-file: go.mod` pins in CI. Producing the archive under a different toolchain risks committing a binary artefact that differs from what CI would generate, which seems worse than leaving it stale.

`verify-bundle.sh` mentions maintainers can comment `/regenerate-bundle` to auto-fix. @NameHaibinZhang would you mind triggering that? If you would rather I handle it, I can set up a matching Go 1.25 toolchain and push the regenerated archive — just say which you prefer.

Disclosure per the [GenAI policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md): I used an AI assistant for the code and for drafting this comment. The behaviour is verified by hand — I read `FrequencyPenalty` and `PresencePenalty` out of `CompletionNewParams` in `openai-go/v3@v3.49.0` (`completion.go`, lines 288 and 315), and confirmed the added tests fail against the unpatched parser before passing with it.

Priyanshu-byte-coder:
Post-rebase CI went red on three jobs. None of them traces to this diff, which touches only `parseCompletionRequest` and its tests in the three openai modules.

**`LatestLibBuild Tests (go ubuntu-latest amd64 shard 0)`** — module proxy failure, not a build or test error:

```
go.opentelemetry.io/otel/sdk/metric@v1.44.1-0.20260625150014-c84013202f01:
read "https://proxy.golang.org/.../@v/....zip":
stream error: stream ID 11; INTERNAL_ERROR; received from peer
```

**`Coverage Integration Tests`** — `TestGRPCServerGRPCClient` lost a port race:

```
failed to listen on frontPort: listen tcp :35343: bind: address already in use
readiness.go:24: Condition never satisfied
--- FAIL: TestGRPCServerGRPCClient (120.24s)
```

That is the E2E port-allocation fragility #959 is already about. This PR does not touch gRPC or the integration harness.

**`Verify Bundle`** — expected, as covered above: the job fires on any `instrumentation/**` edit and `tool/data/otelc-bundle.tgz` has not been regenerated. #1074 is currently red on the same job for the same reason.

For what it is worth, the repo looks broadly flaky at the moment — #1078 is red on `Unit Tests (go macos-latest arm64)` and #1077 on `Unit Tests (go ubuntu-22.04-arm arm64)`, both unrelated to their diffs.

So the only thing genuinely blocking here is the bundle. A re-run should clear the other two, and `/regenerate-bundle` handles the third whenever you have a moment @NameHaibinZhang. Still happy to install Go 1.25 and regenerate it myself if you would prefer that over the bot.

darccio:
@Priyanshu-byte-coder Please fix the conflicts.

Priyanshu-byte-coder:
Rebased onto main and pushed — conflicts resolved.

The branch was 89 commits behind. The only conflicts were the import blocks in the three `middleware_test.go` files, caused by #1074 adding the `gen_ai.client.operation.duration` metric and pulling in `otel`, `otel/metric`, `sdk/metric` and `metricdata` alongside the `semconv` import this PR adds. Kept both sides, in the grouping the rest of the file already uses.

The three `middleware.go` files merged cleanly. The diff against current main is unchanged in substance — 165 insertions, 12 deletions, and the 12 deletions are just the struct field block being realigned when the two penalty fields were added:

```
$ git diff --stat origin/main
 .../openai/openai-go/middleware.go         | 16 ++++++--
 .../openai/openai-go/middleware_test.go    | 43 +++++++++++++++++++
 .../openai/openai-go/v2/middleware.go      | 16 ++++++--
 .../openai/openai-go/v2/middleware_test.go | 43 +++++++++++++++++++
 .../openai/openai-go/v3/middleware.go      | 16 ++++++--
 .../openai/openai-go/v3/middleware_test.go | 43 +++++++++++++++++++
```

Re-verified on the rebased tree that the gap is still there — `parseCompletionRequest` on current main still reads only `model`, `max_tokens`, `temperature` and `top_p` in all three versions, while `parseChatRequest` records both penalties.

All six packages pass locally:

```
ok  .../openai-go            9.001s
ok  .../openai-go/semconv    1.756s
ok  .../openai-go/v2         6.973s
ok  .../openai-go/v2/semconv 1.770s
ok  .../openai-go/v3         7.066s
ok  .../openai-go/v3/semconv 1.767s
```

The workflows are sitting at `action_required`, so they will need approving before CI reports. `Verify Bundle` will likely still need a `/regenerate-bundle` — I did not regenerate it locally because `make package` runs `go mod tidy` and my toolchain is newer than the pinned CI version, so the bundle I produced would differ for reasons unrelated to this change.

PR #1031, opened 2026-08-10
feat(tool): add struct_literal selector and set_fields modifier

### LINKED ISSUE #1001: feat(tool/rule): no selector matches a composite literal

There is no `where` selector that matches a composite (struct) literal, so a value built inline as `&T{...}` cannot be instrumented.

```go
// otelc has no selector that can reach this literal:
// `struct` matches the type declaration, `function_call` matches call
// expressions, and `identifier` matches package-level var/const.
t := &http.Transport{MaxIdleConns: 100}
```

A selector that matches composite literals of a given type, and a way to set or override fields on the matched literal.

`inject_code` with a `pattern` regex, one rule per construction site. It only works when the literal is assigned to a variable, so a literal in a `return`, an argument, or a nested field stays unreachable.

**Additional context**

Instrumentation sometimes needs to distinguish values by *who created them*, and the only place that information exists is the construction site.

Concrete case: a tracer's own HTTP traffic must stay out of the trace, otherwise the request that exports a span produces the next span and an idle application traces itself forever. dd-trace-go handles this with Orchestrion by [adding an `internal` marker field to the `net/http.Transport` struct declaration and setting it to `true` on every `&http.Transport{...}` literal the tracer builds for itself](https://github.com/DataDog/dd-trace-go/blob/61f964796777d56288cf506113ae3b22877a8aca/contrib/net/http/orchestrion.client.yml#L12-L46). The client instrumentation then skips transports carrying the marker.

otelc can add the field but cannot set it, because there is no `http.NewTransport` to hook and no way to match the literals. The field must be set *only* on the tracer's own transports, so it cannot be done from inside `net/http` where the instrumentation runs, and the transports there are indistinguishable from application ones.

Not covered by the Orchestrion parity checklist in #164 — OP3 lists five join points and struct literals is not one of them.

rarguelloF:
## Description

Adds a `struct_literal` selector that matches composite literals of a given type, and a `set_fields` modifier that sets fields on the matched literal.

A field takes a `value`, a `wrap` template that transforms whatever the literal already assigns, or both. Both is useful because a field is often present in some literals and absent in others:

```yaml
instrument_vault_config:
  target: example.com/app
  where:
    struct_literal: github.com/hashicorp/vault/api.Config
  do:
    - set_fields:
        field:
          - name: HttpClient
            wrap: "vaulttrace.WrapHTTPClient({{ . }})"
            value: "vaulttrace.NewHTTPClient()"
```

Two limits worth calling out: it matches only qualified literals (`pkg.Type{...}`), following the same restriction `function_call` already has, and it sets named fields rather than rewriting a literal as a whole, so it cannot reorder or drop elements.

## Motivation

Some libraries offer no constructor to hook. Their configuration type is written as a literal at each use site, so there is nothing for `inject_hooks` or `wrap_call` to match and the value cannot be reached at all today. `hashicorp/vault`'s `api.Config` is one example: the HTTP client it holds is the only place tracing can be injected, and the type is commonly built inline.

Opening as a draft: happy to change the selector name, the modifier name, or the `value`/`wrap` shape before this is reviewed in earnest.

Fixes #1001

---

## Checklist

- [x] Tests pass: `make test`
- [ ] OpenTelemetry Registry updated (not applicable, no new instrumentation)


Labreo:
Hi @rarguelloF!Thank you so much for taking the time to open this draft pr.Just wanted to let you know that due to lfx rules you need to sign the agreement above.It can be done via clicking the prompt in the above message by the above bot.Or it would restrict the CI/CD pipeline to occur.

tool/internal/instrument/apply_lit.go
```diff
@@ -0,0 +1,203 @@
...
+			return true
+		}
+		if matchesLitRule(lit, r, importAliases) {
+			matched = append(matched, lit)
+		}
+		return true
+	})
+
+	modified := false
+	for _, lit := range matched {
```
darccio: A matched literal nested inside another matched literal's `wrap`-ed field is silently skipped — including its own `value`-set fields — with no error or warning.

`matched` is collected once via a single top-down `dst.Inspect` pass before any mutation runs (lines 26-36), so it lists parents before their nested literals. When a parent's field is wrapped, `compileExpression` splices the *original* subtree into the template placeholder and the result is `dst.Clone`d (line 166) — a deep copy that installs a brand-new subtree in the tree. Any nested literal already captured in `matched` now points at an orphan: the loop still "mutates" it, but the mutation lands on a node the tree no longer contains.

Confirmed through the golden-test pipeline (rule wraps `R`, sets `N` on `io.LimitedReader`):

Input:
```go
&io.LimitedReader{R: &io.LimitedReader{R: base, N: 5}, N: 10}
```
Compiles to:
```go
&io.LimitedReader{R: counting(&io.LimitedReader{R: base, N: 5}), N: 99}
```

The outer literal is instrumented correctly; the inner one is completely untouched and the build succeeds silently. This is an ordinary reader/handler wrapper-chain shape, not a contrived case.

Processing `matched` innermost-first fixes it: every descendant is guaranteed to appear before its ancestors in the reverse of a pre-order walk, so by the time an ancestor's field is cloned, any nested literal has already been mutated in place and the clone carries the up-to-date subtree forward.

```suggestion
	for i := len(matched) - 1; i >= 0; i-- {
		lit := matched[i]
```

tool/internal/instrument/apply_lit_test.go
```diff
@@ -0,0 +1,348 @@
...
+	})
+
+	ip := newTestPhase()
+	require.NoError(t, ip.applyLitRule(context.Background(), r, file))
+
+	require.Len(t, lit.Elts, 1)
+	kv, ok := lit.Elts[0].(*dst.KeyValueExpr)
+	require.True(t, ok)
+	assert.Equal(t, "wrapProxy(myProxy)", wrapExprString(t, kv.Value))
+}
```
darccio, line 329: No existing test — unit or golden — seems to exercise a literal of the rule's own target type nested inside another matched literal's `wrap`-targeted field. `lit-rule-only`'s "Nested" fixture comes closest, but it nests the matched literal inside a plain, non-matched `holder` struct field rather than inside a wrap template's splice point, so it never reaches the `dst.Clone`-orphaning path.

That gap is exactly what let the nested-literal bug ship undetected (see the `apply_lit.go` comment above). Worth adding a regression test alongside the fix, close to the existing wrap tests in this file:

```go
func TestApplyLitRule_WrapPreservesNestedMatchedLiteral(t *testing.T) {
	inner := transportLit(keyed("MaxIdleConns", "100"))
	outer := transportLit(&dst.KeyValueExpr{
		Key:   &dst.Ident{Name: "Proxy"},
		Value: &dst.CallExpr{Fun: &dst.Ident{Name: "identity"}, Args: []dst.Expr{inner}},
	})
	file := makeLitFile(outer)
	r := transportRule(
		&rule.InstLitField{Name: "Proxy", Wrap: "wrap({{ . }})"},
		&rule.InstLitField{Name: "Internal", Value: "true"},
	)

	ip := newTestPhase()
	require.NoError(t, ip.applyLitRule(context.Background(), r, file))

	// Both the outer and the nested literal are of the rule's target type;
	// both must end up instrumented.
}
```

A golden fixture pairing an `io.LimitedReader` nested inside another `io.LimitedReader`'s wrapped `R` field (a realistic reader-chaining pattern) would catch this at the integration level too.

tool/internal/instrument/testdata/golden/lit-rule-only/source.go
```diff
@@ -0,0 +1,35 @@
...
+import "io"
+
+// Empty literal takes both fields.
+func Bare() *io.LimitedReader {
+	return &io.LimitedReader{}
+}
+
+// Existing keys are kept and overridden in place.
+func Configured(r io.Reader) *io.LimitedReader {
+	return &io.LimitedReader{R: r, N: 100}
```
darccio, line 15: `Configured` demonstrates that `value` overrides an existing field in place, but the existing expression it discards (`r`, a bare identifier) can't show what's actually at stake: `value` replaces whatever the literal already assigns, including a call with a side effect, and the discarded expression simply vanishes rather than being preserved or composed. The docs call this out explicitly ("differs from `assign_value`... a declaration either has an initializer or does not"), so a fixture where the overridden field starts as something like `R: mustOpen(path)` — not just `R: r` — would make that discard behavior self-evident from the golden diff rather than only documented in prose.

Not blocking; the current fixture does correctly verify the override-in-place mechanics (field kept at its original position, value replaced).

darccio (changes requested):

tool/internal/instrument/testdata/golden/lit-rule-only/source.go
rarguelloF: > a fixture where the overridden field starts as something like R: mustOpen(path) — not just R: r — would make that discard behavior self-evident from the golden diff rather than only documented in prose.

Sorry I don't see why those are different documentation-wise. Would you mind elaborating a little bit more?

tool/internal/instrument/apply_lit.go
rarguelloF: nice catch, addressing it now!

tool/internal/instrument/testdata/golden/lit-rule-only/source.go
darccio: The difference isn't about the docs saying something different. It's about what the golden diff shows.

Right now `Configured` goes from `R: r` to `R: nil`. But `r` is just a variable name, so replacing it doesn't look like much happened, it could pass for the tool picking a different name.

If the original were `R: mustOpen(path)` instead, the diff would show that whole function call disappearing and getting replaced by `nil`. That makes it obvious that `value` throws away the original expression completely, even one that does real work, instead of reusing it (which is what `wrap` does with `{{ . }}`).

So it's not about clearer docs, it's about picking a fixture value that actually demonstrates the discard, instead of one that's too simple to show it.

darccio (approved):
LGTM

PR #1035, opened 2026-08-10
fix: prevent collisions in synthetic params/returns

### LINKED ISSUE #1014: bug(tool): synthetic parameter/return names can collide with existing vars or shadow global identifiers
### Describe the bug

When the tool generates synthetic names for unnamed or blank (`_`) parameters and return values, the generated names (e.g. `_ignoredParam0`, `_unnamedRetVal0`) can currently collide with existing names in the function signature, as well as shadow identifiers declared at package scope or in the function body.

PR #729 attempts to address this with `syntheticNamer`, but it only checks for collisions against names in the function signature (receiver, params, results). It doesn't account for:

- Package-level variables/constants/functions
- Local variables declared in the function body
- Imported package names

This was previously discussed in [#799 (comment)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/799#discussion_r3689788318). The `syntheticNamer` approach in #729 is an incomplete fix for the same problem.

### Environment

- OS: any
- Go Version: any
- Version: current main

### To Reproduce

1. Have a Go file with a package-level variable or local variable whose name matches a synthetic name, or a function signature with a colliding parameter name:

```go
package main

var _ignoredParam0 = "important"

func foo(_ int, _ignoredParam1 string) {
    println(_ignoredParam0) // expects the global
}
```

2. Apply a func rule that targets `foo`. The tool renames the blank param to `_ignoredParam0`, shadowing the global. It might also generate `_ignoredParam1` for another param, causing a direct signature collision.

3. The instrumented code now prints the parameter value instead of the global.

### Expected behavior

Synthetic names should not collide with any identifier visible in the function's scope. The generated name should be unique enough that shadowing is not a realistic concern.

### Additional context

One approach: use a hash (see `InstFuncRule.Identity`) in the synthetic identifier names, like `_param_a1b2c3d4_0`. This makes collisions nearly impossible without needing to resolve every identifier in scope.

Note that raw rules currently hardcode synthetic names like `_unnamedRetVal0` in YAML (see `instrumentation/runtime/otelc.yaml`). If the naming scheme changes, those rules break. However, there is already a planned followup PR to expand func template variables to raw rules ([hannahkm/opentelemetry-go-compile-instrumentation#1](https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/pull/1)) which would solve this by allowing rule authors to use `{{.FuncReturn 0}}` instead, and the internal naming scheme can safely change.

Related:
- #729 -- introduces `syntheticNamer` (incomplete fix)
- [hannahkm/opentelemetry-go-compile-instrumentation#1](https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/pull/1) -- planned PR to add template vars to raw rules (prerequisite for changing the naming scheme)
- [#799 (comment)](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/799#discussion_r3689788318) -- original discussion about this problem

hannahkm:
## Description

Builds generated names/synthetic names using a CRC32 hash that makes it nearly impossible for two generated names to accidentally be the same. Further ensures no collision by checking that no variable exists with the same name.

## Motivation

Generated names (e.g. _ignoredParam0, _unnamedRetVal0) could collide with existing names in the function signature, shadow identifiers declared at package scope or in the function body, or imported package names.

Fixes #1014

---

## Checklist

- [ ] Tests pass: `make test`

[^I-know-my-stuff]:


darccio (approved):
LGTM

Issue #817, opened 2026-07-24
isSetup() is unimplemented and always returns false


singhshresth26:
### Todo task

In `tool/internal/setup/setup.go`, the helper function `isSetup()` is unimplemented and always returns `false`. It is currently marked with a `TODO: Implement Task`.

Because it always returns `false`, the setup phase cannot be skipped even if it has already completed successfully.

### Location

- **File**: [`tool/internal/setup/setup.go`](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/tool/internal/setup/setup.go)
- **Lines**:
```go
// This function can be used to check if the setup has been completed.
func isSetup() bool {
	// TODO: Implement Task
	return false
}

dobbydobap:
@darccio small bookkeeping thing: #659 and #818 were both closed with "Closing
in favour of #854", but #854 is the selector/modifier registry and only touches
tool/internal/rule. It does not touch setup.go or isSetup, so this issue is now
open with both its PRs closed for a reason that does not seem to apply.

Reading back, @amazingakai's review on #659 looks like the real state: matched.json
and LoadStateManager were both rejected as signals, and the four invalidation
inputs listed there are the hard part. Is the intent to keep this open, or to
fold it into #552? Either is fine, it would just save the next person walking
into the same three attempts.

darccio:
@dobbydobap We'll keep this open, yes. If you want to address it, we're ok with that.

dobbydobap:
Thanks @darccio, happy to take it.

Before writing code I want to check the approach, since #513, #659 and #818 all
took the existence-check route and were closed.

Reading back through #161, which is the same problem and still open, the design
already has agreement. @txabman42 there: "We don't just need to check if it
exist, but compare the hash to know if something has change. For hash, we can
take into account rules, go.mod, envs, etc." and @amazingakai agreed. That is
what the closed PRs were missing.

So concretely: compute a fingerprint of the inputs setup consumes, store it with
the setup state, and have isSetup() return true only on an exact match. Mapping
that to the four invalidation inputs @amazingakai listed on #659:

- dependency graph: hash go.mod content
- module set: include go.work when present
- build environment: the flags and env vars setup reads
- user source: the set of files rule matching walked, plus their content hash

Source is the expensive one, and it is the reason a naive check cannot be
correct. I would take the conservative rule that a stale fingerprint always
[...]

PR #818, opened 2026-07-24
fix(tool): implement isSetup check to skip redundant setup

singhshresth26:
Implement the isSetup() helper in tool/internal/setup/setup.go to check if the setup phase has already been run and completed, by verifying if matched.json exists in the build temporary directory. Also, add a TestIsSetup unit test in setup_test.go to cover this.

Fixes #817 

## Description

- Implements `isSetup()` in `tool/internal/setup/setup.go` by verifying the existence of `matched.json` via `util.PathExists()`.
- Appends `TestIsSetup` to `tool/internal/setup/setup_test.go` to validate both initial and completed setup states.

## Motivation

The `isSetup()` check was previously unimplemented (returning `false` by default). This change allows the CLI to detect when setup has already completed successfully (indicated by the presence of the matched rules output file `matched.json`) and skip the setup phase on subsequent runs.

Fixes # 

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format` (All Go code formatted successfully)
- [x] Linters pass: `make lint`
- [x] Tests pass: `make test` (All local unit tests build and pass)
- [x] Tests added for new functionality
- [x] Tests follow [testing guidelines](docs/testing.md)
- [ ] Documentation updated (if applicable)
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](docs/instrument-guide.md#4-register-the-instrumentation))

PR #1038, opened 2026-08-10
fix(tool): implement isSetup logic to correctly track setup completion

kehach07:
## Description

This PR implements the missing logic for the `isSetup()` function in `tool/internal/setup/setup.go`. 

Previously, this function always returned `false` with a `// TODO: Implement Task` stub. The stub has been removed, and `isSetup()` now correctly determines setup completion by checking for the existence of `matched.json` via `util.PathExists(util.GetMatchedRuleFile())`. 

Additionally, `TestIsSetup` has been updated in `pin_helpers_test.go` to mock the environment and properly assert the `true` and `false` states of this function.

## Motivation

The `isSetup()` function is intended to prevent redundant executions of the setup phase by checking if `otelc setup` has already finished successfully. Without this logic, the tool lacked an accurate mechanism to short-circuit the setup routines, relying entirely on downstream processes. Implementing this speeds up repeated executions and properly completes the intended setup architecture.

Fixes # (insert issue number if applicable, or leave blank)

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### DIFF
--- a/tool/internal/setup/pin_helpers_test.go
+++ b/tool/internal/setup/pin_helpers_test.go
@@ -16,8 +16,20 @@ import (
 )
 
 func TestIsSetup(t *testing.T) {
-	// isSetup is currently a stub that always reports false.
+	// By default, no setup file exists.
 	assert.False(t, isSetup())
+
+	// Set up a mock environment
+	dir := newModuleDir(t)
+	t.Setenv(util.EnvOtelcWorkDir, dir)
+	
+	// Create the build temp dir and the matched file
+	matchedFile := util.GetMatchedRuleFile()
+	require.NoError(t, os.MkdirAll(filepath.Dir(matchedFile), 0o755))
+	require.NoError(t, os.WriteFile(matchedFile, []byte("{}"), 0o644))
+
+	// isSetup should now report true
+	assert.True(t, isSetup())
 }
 
 // newModuleDir creates a minimal Go module in a fresh temp directory and
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -62,8 +62,7 @@ func keepForDebug(ctx context.Context, srcPath string) {
 
 // This function can be used to check if the setup has been completed.
 func isSetup() bool {
-	// TODO: Implement Task
-	return false
+	return util.PathExists(util.GetMatchedRuleFile())
 }
 
 // flagsWithPathValues contains flags that accept a value from "go build" command.


dobbydobap:
Heads up before you put more time into this, since the history is spread across
a few closed PRs and is easy to miss.

The matched.json existence check has been proposed twice already and turned down
both times. On #513 @txabman42 said "Checking the file existence is not enough.
The setup file might be related to a previous configuration, making not possible
to do the setup again." #818 took the same approach and was also closed. #659
tried LoadStateManager instead, and @amazingakai explained there that it only
reflects persisted state, so for otelc go build the state manager can exist only
in memory and never be committed.

The direction agreed on #161 is a fingerprint rather than an existence check.
@txabman42 there: "We don't just need to check if it exist, but compare the hash
to know if something has change. For hash, we can take into account rules,
go.mod, envs, etc."

More to the point, @amazingakai closed #817 two days ago with "We can rely on
Go's incremental caching instead of implementing an isSetup", pointing at #552.

So it is probably worth asking whether isSetup is wanted at all before going
further. I was assigned #817 last week and this is what I found before it got
closed.

amazingakai:
I closed #817 because it was a duplicate of #161, not because I think we don't need an `isSetup` check.

#552 is only a proposal and needs to be well thought out, and have consensus from the maintainers, before implementation begins.

In the meantime, if we can have a reasonable `isSetup` check (that I can't really think of right now :P), that'd be a good idea. Simply checking for the existence of `matched.json`/`state.json` isn't it though.

kehach07:
Thanks @dobbydobap for the historical context, and @amazingakai for the clarification! 

I completely understand why simply checking for `matched.json` is unsafe, since it could belong to a stale configuration from a previous run. 

Since a reliable `isSetup` check is still valuable, would you like me to pivot this PR to implement the fingerprint/hashing approach discussed in #161? 

I can update the logic to generate a hash of the current configuration (e.g., rules, `go.mod`, environment variables) and store it during setup. Then, `isSetup()` can simply compare the current hash against the stored hash. Let me know if you'd like me to proceed in that direction!

darccio:
@kehach07 I think it's a sensible approach. Please proceed accordingly.

om7057:
@kehach07 following up on the fingerprint direction, since it's landed on `go.mod`, rules, and "environment variables" as inputs, worth being explicit that the fingerprint needs to include `GOOS`/`GOARCH` and build tags (`-tags`) specifically, not just env vars generically.

`listBuildPlan` in `tool/internal/setup/find.go` runs `go build -a -x -n` inheriting whatever build target the current invocation has, and the `Sources` list AST-matching runs against is only valid for that target. If the fingerprint doesn't key on `GOOS`/`GOARCH`/tags explicitly, `isSetup()` could still return true for a `GOOS=windows` build reusing a `matched.json` computed under `GOOS=linux`, since neither `go.mod` nor most "environment variables" in the OTel-config sense would differ between those two runs. Same failure mode as the existence check, just one layer more hidden.

Can help if needed with the build!
cc: @darccio

PR #1040, opened 2026-08-11
fix(tool/internal/instrument): resolve file rule path directly to prevent filename suffix collision

### LINKED ISSUE #1039: bug(tool/internal/instrument): applyFileRule uses strings.HasSuffix causing filename collision
### Describe the bug

In `tool/internal/instrument/apply_file.go`, `applyFileRule` scans all files in `rule.ResolvedPath` and matches `rule.File` using `strings.HasSuffix(file, rule.File)`:

```go
// Find the new file we want to introduce
index := slices.IndexFunc(files, func(file string) bool {
    return strings.HasSuffix(file, rule.File)
})
if index == -1 {
    return ex.Newf("file %s not found", rule.File)
}
file := files[index]

vedant21-oss:
## Description

This PR updates `applyFileRule` in `tool/internal/instrument/apply_file.go` to directly resolve the target file via `filepath.Join(rule.ResolvedPath, rule.File)` and verify its existence with `util.PathExists(file)`, instead of listing all directory files with `util.ListFiles` and matching with `strings.HasSuffix`.

Unit tests have been added to `tool/internal/instrument/apply_file_test.go` covering:
- Successful file rule application and package renaming.
- Verifying that files sharing common suffixes (e.g. `custom_helper.go` vs `helper.go`) do not cause accidental collisions.
- Clear error handling for missing files.
- Subdirectory relative path support.

## Motivation

Previously, `applyFileRule` scanned all files in `rule.ResolvedPath` with `util.ListFiles` and found the target file using `strings.HasSuffix(file, rule.File)`. Because `strings.HasSuffix` does not respect path segment boundaries, any file sharing a suffix (e.g., `my_hook.go` when `hook.go` is specified) could be chosen if enumerated first during directory traversal. Directly constructing and validating `filepath.Join(rule.ResolvedPath, rule.File)` eliminates false suffix collisions and avoids unnecessary directory walks.

Fixes #1039

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/instrument/apply_file.go
+++ b/tool/internal/instrument/apply_file.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"slices"
 	"strings"
 
 	"go.opentelemetry.io/otelc/tool/ex"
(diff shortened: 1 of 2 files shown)


y1yang0 (approved):
This actually changes the semantics, but even so, the current version still seems reasonable. Let's go ahead and approve it for now.

vedant21-oss:
Thanks @y1yang0 for the review and merge! Understood regarding the semantics — direct relative paths make file resolution explicit and avoid unintended matches.

PR #1045, opened 2026-08-11
fix(tool/internal/instrument): skip directive rule injection when no function matches

### LINKED ISSUE #1043: bug(tool/internal/instrument): directive rule injects its imports even when no function matches
### Describe the bug

`applyDirectiveRule` in `tool/internal/instrument/apply_directive.go` adds the rule's
`imports:` to the target file *before* it looks for the functions the directive annotates:

```go
func (ip *InstrumentPhase) applyDirectiveRule(ctx context.Context, r *rule.InstDirectiveRule, root *dst.File) error {
	if err := ip.addRuleImports(ctx, root, r.Imports, r.Name); err != nil {   // <- import added here
		return err
	}
	tmpl, err := fasttemplate.NewTemplate(r.Template, "{{", "}}")
	...
	funcs := ast.FindFuncsByDirective(root, r.Directive)                       // <- match computed here
```

`ast.FindFuncsByDirective` only matches a directive that is a *leading* comment on a
top-level `FuncDecl`. If the same comment appears anywhere else in the file — inside a
function body, on a `GenDecl`, on a struct field — `funcs` is empty and nothing is
injected. The import is already in the AST at that point, and `instrument()` writes the
file out regardless, so the user's package ends up with:

```
main.go:6:8: "fmt" imported and not used
```

in code they did not write and cannot easily trace back to a rule.

The other rule types already guard against this. `apply_call.go` has an explicit
`if !appendModified && !replaceModified { return nil }` before its `addRuleImports` call;
`apply_struct.go` and `apply_decl.go` both locate their target and return early when it is
missing, before touching imports. `apply_directive.go` is the only one without the guard.

There is a second, quieter symptom in `applyOneRule` (`tool/internal/instrument/instrument.go`):

```go
case *rule.InstDirectiveRule:
	return true, ip.applyDirectiveRule(ctx, rt, root)
```

It reports `true` ("a globals file is needed") unconditionally, so a directive rule that
matched nothing still makes `instrument()` call `writeGlobals` and emit the `HookContext`
interface into the target package.

### Environment

- OS: Linux / any
- Go Version: 1.25
- Version: main (efb163d)

### To Reproduce

1. Write a source file where the directive is not a leading comment on a top-level func:

   ```go
   package main

   func main() {
   	//otelc:span
   	value := 1
   	println(value)
   }
   ```

2. Apply a directive rule that carries a non-empty `imports:` map:

   ```yaml
   span_directive_no_match:
     target: main
     where:
       directive: "otelc:span"
     do:
       - expand_directive:
           template: |-
             fmt.Println("span start: {{FuncName}}")
     imports:
       fmt: "fmt"
   ```

3. Build. The compile fails with `"fmt" imported and not used`.

### Expected behavior

A directive rule that matches no function should be a no-op: no imports injected, no
globals file emitted, and the target file unchanged.

darkraider01:
## Description

`applyDirectiveRule` added the rule's `imports:` to the target file before it computed
`ast.FindFuncsByDirective`. When the directive comment exists in the file but is not a
leading comment on a top-level func, the match set is empty and nothing is injected — but
the import is already in the AST, and `instrument()` writes the file out regardless. The
result is an `"fmt" imported and not used` compile error in code the user did not write.

This reorders the function so the match is computed first and returns early when nothing is
annotated, which is the same guard `apply_call.go` (`if !appendModified && !replaceModified`)
and `apply_struct.go` / `apply_decl.go` (find target, return early) already have.

`applyDirectiveRule` now also reports whether it instrumented anything, so `applyOneRule`
no longer returns `true` unconditionally for `*rule.InstDirectiveRule`. Previously a
zero-match directive rule still made `instrument()` call `writeGlobals` and emit the
`HookContext` interface into the target package.

Two tests cover this:

- A golden fixture, `testdata/golden/directive-rule-no-match/`, applies a directive rule
  carrying `imports: {fmt: fmt}` to a file whose directive sits on an assignment inside a
  function body. Against the unfixed code the fixture fails to compile with
  `"fmt" imported and not used`; with the fix the output is byte-identical to the input.
- `TestApplyDirectiveRule_NoMatchingFuncs` calls `applyDirectiveRule` directly and asserts
  that no import spec was added and that the call reports "nothing instrumented". Note it
  counts import specs in `root.Decls` rather than checking `len(root.Imports)`:
  `imports.AddToFile` appends to `root.Decls` and never populates `root.Imports`, so an
  assertion on the latter would pass with or without this fix.
  `TestApplyDirectiveRule_MatchingFunc` covers the positive path, and
  `TestApplyDirectiveRule_TemplateErrors` covers the three template failure
  modes — an unterminated tag, an unknown tag, and a template that renders to
  invalid Go — each of which must be reported rather than silently skipped.

## Motivation

A directive rule that carries imports and matches nothing currently breaks the user's
build with an error that points at their own source file and gives no hint that a rule
caused it.

Fixes #1043

## Verification

`make format` and `make test-unit` pass locally. The integration and e2e halves of
`make test` were verified through this PR's CI jobs — `Test (go …)`, the `Integration
Tests` shards on ubuntu/macOS/Windows and arm64, and `Coverage E2E Tests` — rather than a
local run, since the testcontainers-based suites had no usable Docker host here.

`.github/workflows/lint-golang.yaml` pins — 0 issues. Worth flagging separately:
`.tools/go.mod` pins **v2.12.2**, so `make lint/go` locally builds a newer linter than CI
runs, and under v2.12.2 pristine `main` already reports 37 findings in files this PR does
not touch.

---

## Checklist

- [x] Tests pass: `make test`


amazingakai:
@darkraider01 needs a rebase

darkraider01:
Done — rebased. Wasn't quite routine though: the directive-templating system got rewritten on main since this branch, so this needed actual reconciliation, not just resolving text conflicts. `main`'s new version had picked up the args/import-alias features but never picked up this fix, so `addRuleImports` was still running unconditionally there too — re-applied the ordering fix against the new function shape and merged the test coverage from both sides.

Bonus find along the way: my own golden fixture (`directive-rule-no-match`) still used the old template syntax and was silently producing a nil rule under the new engine's validation, which crashed. Fixed that too.

tool/internal/instrument/apply_directive.go
```diff
@@ -16,18 +16,29 @@ import (
...
+		return false, nil
+	}
+	if importErr := ip.addRuleImports(ctx, root, r.Imports, r.Name); importErr != nil {
+		return false, importErr
+	}
+	tmpl, err := rule.ParseDirectiveTemplate(r.Template)
 	if err != nil {
-		return ex.Wrap(err)
+		return false, ex.Wrap(err)
 	}
```
darccio: Now that matching happens before any mutation, this could go one step further:
`ParseDirectiveTemplate` doesn't depend on anything `addRuleImports` produces, so
validating the template before adding imports closes the remaining window where a
matched-but-invalid-template rule leaves an import spec in `root.Decls` before failing.
Currently harmless — `instrument()` aborts the whole file on any rule error, so the AST
mutation never reaches disk — but it keeps the "validate before mutate" invariant this fix
establishes fully intact, rather than partially.

```suggestion
	tmpl, err := rule.ParseDirectiveTemplate(r.Template)
	if err != nil {
		return false, ex.Wrap(err)
	}
	if importErr := ip.addRuleImports(ctx, root, r.Imports, r.Name); importErr != nil {
		return false, importErr
	}
```

tool/internal/instrument/apply_directive_test.go
```diff
@@ -216,11 +290,26 @@ func TestApplyDirectiveRule(t *testing.T) {
...
+			},
+		}
+		root := &dst.File{Decls: []dst.Decl{funcDecl}}
 
 		ip := &InstrumentPhase{logger: slog.Default()}
-		err := ip.applyDirectiveRule(context.Background(), r, root)
+		modified, err := ip.applyDirectiveRule(context.Background(), r, root)
 		require.Error(t, err)
+		assert.False(t, modified)
 	})
```
darccio, line 313: No subtest covers the combination of a matching func + configured `Imports` + a template
that fails to parse. Every current error subtest either has no `Imports` set or fails
before reaching the template-parse step, so the "imports added, then template parse
fails" ordering introduced by this PR's reorder (see the sibling comment on
apply_directive.go) isn't exercised by any assertion. Worth a small addition here even
though the outcome is currently masked by the caller discarding the file on error — it
documents the invariant explicitly instead of leaving it implicit.

darccio (changes requested):

darkraider01:
@darccio addressed both:

- Reordered so the template is parsed before imports are added — a broken template now fails before touching `root.Decls` at all, closing that last "validate before mutate" gap you called out.
- Added the missing subtest: matching func + configured `Imports` + a template that fails to parse, asserting zero import specs get added.

Also updated the branch against main — picked up the `DirectiveTemplate`→`FuncTemplate` rename from #1140 along the way, verified the reorder survived that cleanly.

darccio (approved):

PR #1046, opened 2026-08-11
fix(tool/internal/instrument): rewrite only the targeted name of a multi-name declaration

### LINKED ISSUE #1044: bug(tool/internal/instrument): decl rule replace/wrap rewrites every name in a multi-name var/const declaration
### Describe the bug

A decl rule targets one identifier, but `applyDeclRule` in
`tool/internal/instrument/apply_decl.go` rewrites every name declared in the same
`ValueSpec`.

`ast.FindNamedDecl` / `findValueDecl` return the whole `*dst.ValueSpec` that contains the
matched name and discard which name it was. `applyDeclRule` then has no way to tell the
targeted name from its siblings, and the replace path assigns to all of them:

```go
// Assign the expression to all names in the spec.
spec.Values = make([]dst.Expr, len(spec.Names))
for i := range spec.Values {
	spec.Values[i] = util.AssertType[dst.Expr](dst.Clone(expr))
}
```

`wrapDeclValues` has the same shape — it loops over every entry in `spec.Values`.

So given

```go
var readTimeout, writeTimeout = 5 * time.Second, 30 * time.Second
```

a rule targeting only `readTimeout` silently rewrites `writeTimeout` as well. The failure
is silent: the build succeeds and the instrumented binary just has the wrong value for a
variable no rule mentioned.

There is also an arity hazard specific to the replace path. For a tuple-valued
initializer:

```go
var alpha, beta = pair()   // 2 names, 1 initializer
```

`make([]dst.Expr, len(spec.Names))` produces 2 values from a spec that had 1, emitting
`var alpha, beta = expr, expr` — a structurally different declaration from the one the
user wrote.

### Environment

- OS: Linux / any
- Go Version: 1.25
- Version: main (efb163d)

### To Reproduce

Apply a decl rule with `kind: var`, `identifier: alpha`, `replace: "99"` to:

```go
var alpha, beta = 1, 2
```

Observed output: `var alpha, beta = 99, 99`

The wrap advice behaves the same way: wrapping `alpha` with `double({{ . }})` produces
`var alpha, beta = double(1), double(2)`.

### Expected behavior

Only the initializer belonging to the targeted name is rewritten:
`var alpha, beta = 99, 2`.

For the tuple-valued case, where no single initializer belongs to the targeted name,
the tool should report the situation rather than silently reshape the declaration.

darkraider01:
## Description

`ast.FindNamedDecl` returns the whole `*dst.ValueSpec` that contains a matched name and
discards which name matched, so `applyDeclRule` rewrote every name in the spec: the replace
path reassigned all of `spec.Values`, and `wrapDeclValues` wrapped all of them. A rule
targeting `alpha` in `var alpha, beta = 1, 2` produced `var alpha, beta = 99, 99`.

The fix locates the targeted name's position in `spec.Names` and rewrites only the
initializer at that index. `wrapDeclValues` becomes `wrapDeclValue` and the replace path
moves into `replaceDeclValue`, both taking that index.

### Arity handling — please weigh in

For a tuple-valued initializer such as `var alpha, beta = pair()` there are two names and
one initializer, so no single value belongs to the targeted name. The old code turned that
into `var alpha, beta = expr, expr`, a structurally different declaration. I chose to
**report an error** for both `replace` and `wrap` in that case rather than guess, since
producing correct code would mean splitting the declaration, and the alternatives all
change the user's declaration shape. I'd like reviewer input on whether erroring is the
behaviour you want, or whether a skip-with-warning would fit the tool better.

One existing behaviour is deliberately preserved: a **single** name with no initializer
still gains one on the replace path (`var X T` becomes `var X T = expr`). A *multi*-name
declaration with no initializer now errors instead of gaining one initializer per name.

### Tests

`TestDeclRuleDoesNotClobberSiblingNames` and `TestDeclRuleWrapDoesNotClobberSiblingNames`
render the rewritten file back to source and assert `var alpha, beta = 99, 2` and
`var alpha, beta = 1, double(2)`. Against the unfixed code they produce `99, 99` and
`double(1), double(2)`. `TestDeclRuleReplaceRejectsTupleValuedInitializer` and its wrap
companion cover the arity case and assert the declaration is left untouched. The existing
`wrapDeclValues` unit tests are renamed and `TestWrapDeclValue_MultipleValues` now asserts
the sibling is left alone rather than that both values are wrapped.

## Motivation

The failure is silent — the build succeeds and the instrumented binary just carries the
wrong value for a variable no rule mentioned. `var readTimeout, writeTimeout = ...` is a
realistic shape in the kind of library code this tool instruments.

Fixes #1044

## Verification

`make format` and `make test-unit` pass locally. The integration and e2e halves of
`make test` were verified through this PR's CI jobs — `Test (go …)`, the `Integration
Tests` shards on ubuntu/macOS/Windows and arm64, and `Coverage E2E Tests` — rather than a
local run, since the testcontainers-based suites had no usable Docker host here.

`.github/workflows/lint-golang.yaml` pins — 0 issues. Worth flagging separately:
`.tools/go.mod` pins **v2.12.2**, so `make lint/go` locally builds a newer linter than CI
runs, and under v2.12.2 pristine `main` already reports 37 findings in files this PR does
not touch.

---

## Checklist

- [x] Tests pass: `make test`


amazingakai:
@darkraider01 needs a rebase

darkraider01:
@amazingakai done — rebased against main. Only real conflict was in `apply_decl_test.go`, and it was just two unrelated test additions landing in the same spot (my multi-name ValueSpec tests vs main's new `TestParseValueExpr`), so I kept both.

darkraider01:
@darccio mind giving this one a look too when you get a chance? Same area as [#1045](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1045) — decl rule fix

darkraider01:
Heads up — while updating this branch I noticed the same ordering gap darccio flagged on [#1045](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1045): addRuleImports ran before the decl rule was actually confirmed valid, so a rule that fails its type check, wrap/replace parse, or tuple-valued-initializer arity check could still leave an import spec behind before erroring out. Harmless today (same reason as [#1045](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1045) — a failed rule discards the whole file), but fixed it proactively rather than wait for review to catch it. Also added the two import-interaction test cases that had zero coverage before.

darccio:
@darkraider01 We are aware there are a lot of pending reviews. Please refrain from pinging a specific maintainer unless they have participated already in the PR.

darkraider01:
Sorry about that, didn't mean to add to the noise. I'll keep it in mind going forward.

tool/internal/instrument/apply_decl_test.go
```diff
@@ -199,6 +236,102 @@ func TestApplyDeclRule_EmptyKind_FunctionTarget(t *testing.T) {
...
+		Identifier:   "alpha",
+		Replace:      "99",
+	}
+
+	err := newTestPhase().applyDeclRule(context.Background(), r, file)
+
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "declares 2 names but has 1 initializers")
+	assert.Contains(t, renderFile(t, file), "var alpha, beta = pair()", "declaration must be left untouched")
+}
```
darccio, line 319: The PR body calls out that multi-name declarations with no initializer now error instead of silently gaining one initializer per name — a real behavior change from the old code. Only the tuple-valued arity case has a direct regression test; this branch doesn't. Worth adding a sibling test so the new behavior stays pinned.

```suggestion
}

func TestDeclRuleReplaceRejectsMultiNameNoInitializer(t *testing.T) {
	file := parseTestFile(t, "package main\n\nvar alpha, beta int\n")
	r := &rule.InstDeclRule{
		InstBaseRule: rule.InstBaseRule{Name: "replace_alpha"},
		Kind:         "var",
		Identifier:   "alpha",
		Replace:      "99",
	}

	err := newTestPhase().applyDeclRule(context.Background(), r, file)

	require.Error(t, err)
	assert.Contains(t, err.Error(), "declares 2 names but has 0 initializers")
	assert.Contains(t, renderFile(t, file), "var alpha, beta int", "declaration must be left untouched")
}
```

tool/internal/instrument/apply_decl.go
```diff
@@ -42,61 +43,95 @@ func (ip *InstrumentPhase) applyDeclRule(ctx context.Context, r *rule.InstDeclRu
...
+// name there would change the shape of the declaration, so it is refused
+// instead of guessed at.
+func declArityError(spec *dst.ValueSpec, nameIdx int) error {
+	return ex.Newf(
+		"declaration %q declares %d names but has %d initializers; "+
+			"rewriting a single name requires one initializer per name "+
+			"(tuple-valued declarations such as `var a, b = f()` are not supported)",
+		spec.Names[nameIdx].Name, len(spec.Names), len(spec.Values),
+	)
+}
```
darccio, line 141: Grammar: "declares 2 names but has 1 initializers" should be singular for count 1. Using "initializer(s)" avoids implementing real pluralization for a one-off error message.

```suggestion
func declArityError(spec *dst.ValueSpec, nameIdx int) error {
	return ex.Newf(
		"declaration %q declares %d names but has %d initializer(s); "+
			"rewriting a single name requires one initializer per name "+
			"(tuple-valued declarations such as `var a, b = f()` are not supported)",
		spec.Names[nameIdx].Name, len(spec.Names), len(spec.Values),
	)
}
```

tool/internal/instrument/apply_decl_test.go
```diff
@@ -176,6 +193,26 @@ func TestApplyDeclRule_WrapExpression_NoInitializer(t *testing.T) {
...
+		Kind:       "var",
+		Identifier: "X",
+		Wrap:       "func {{ . }}",
+	}
+
+	err := newTestPhase().applyDeclRule(context.Background(), r, file)
+
+	require.Error(t, err)
+	assert.Zero(t, countImportSpecs(file))
+}
```
darccio, line 214: Both new import-related tests only cover the skip-on-failure direction (imports absent after a rewrite fails). There's no test confirming imports are still added when the rewrite succeeds, which is exactly the behavior this reorder is supposed to preserve.

```suggestion
}

func TestApplyDeclRule_ReplaceSuccess_AddsImports(t *testing.T) {
	file := makeVarFile("X", &dst.BasicLit{Kind: token.INT, Value: "1"})
	r := &rule.InstDeclRule{
		InstBaseRule: rule.InstBaseRule{
			Name:    "replace_x",
			Imports: map[string]string{"fmt": "fmt"},
		},
		Kind:       "var",
		Identifier: "X",
		Replace:    "99",
	}

	err := newTestPhase().applyDeclRule(context.Background(), r, file)

	require.NoError(t, err)
	assert.Equal(t, 1, countImportSpecs(file), "a successful rewrite must still add the rule's imports")
}

```

tool/internal/instrument/apply_decl_test.go
```diff
@@ -199,6 +236,102 @@ func TestApplyDeclRule_EmptyKind_FunctionTarget(t *testing.T) {
 	assert.Contains(t, err.Error(), "is not a var or const declaration")
 }
```
darccio, line 259: This test exercises the non-ValueSpec (func target) failure path but doesn't set Imports on the rule, so it never confirms imports are skipped on that branch specifically — only the wrap/tuple-arity failure branches are covered for that.

```suggestion
func TestApplyDeclRule_EmptyKind_FunctionTarget(t *testing.T) {
	file := &dst.File{
		Name: &dst.Ident{Name: "main"},
		Decls: []dst.Decl{
			&dst.FuncDecl{
				Name: &dst.Ident{Name: "DefaultHandler"},
				Type: &dst.FuncType{},
			},
		},
	}
	r := &rule.InstDeclRule{
		InstBaseRule: rule.InstBaseRule{
			Name:    "replace_handler",
			Imports: map[string]string{"fmt": "fmt"},
		},
		Kind:       "",
		Identifier: "DefaultHandler",
		Replace:    "CustomHandler",
	}

	err := newTestPhase().applyDeclRule(context.Background(), r, file)

	require.Error(t, err)
	assert.Contains(t, err.Error(), "is not a var or const declaration")
	assert.Zero(t, countImportSpecs(file), "imports must not be injected when the target isn't a var/const declaration")
}
```

tool/internal/instrument/apply_decl_test.go
```diff
@@ -199,6 +236,102 @@ func TestApplyDeclRule_EmptyKind_FunctionTarget(t *testing.T) {
...
+		Identifier: "alpha",
+		Replace:    "99",
+	}
+
+	err := newTestPhase().applyDeclRule(context.Background(), r, file)
+
+	require.Error(t, err)
+	assert.Zero(t, countImportSpecs(file))
+}
+
```
darccio, line 372: The single-name-no-initializer branch of `replaceDeclValue` is only tested against a plain `var X T`. It also applies to a `const` spec whose value comes from `iota`-implicit-repetition (e.g. `B` in `const (A = iota; B)`, which parses as `Names=[B]`, `Values=[]`). Worth a regression test confirming that shape is handled the same way rather than assumed.

```suggestion
}

func TestDeclRuleReplaceHandlesConstIotaImplicitRepeat(t *testing.T) {
	file := parseTestFile(t, "package main\n\nconst (\n\tA = iota\n\tB\n)\n")
	r := &rule.InstDeclRule{
		InstBaseRule: rule.InstBaseRule{Name: "replace_b"},
		Kind:         "const",
		Identifier:   "B",
		Replace:      "99",
	}

	err := newTestPhase().applyDeclRule(context.Background(), r, file)

	require.NoError(t, err)
	assert.Contains(t, renderFile(t, file), "B = 99")
}

```

tool/internal/instrument/apply_decl.go
```diff
@@ -42,61 +43,95 @@ func (ip *InstrumentPhase) applyDeclRule(ctx context.Context, r *rule.InstDeclRu
...
-
 	spec, ok := node.(*dst.ValueSpec)
 	if !ok {
 		return ex.Newf("declaration %q (kind: %q) is not a var or const declaration", r.Identifier, r.Kind)
 	}
 
+	// One ValueSpec can declare several names (var a, b = 1, 2), and only the
+	// name the rule targets may be rewritten.
+	nameIdx := slices.IndexFunc(spec.Names, func(name *dst.Ident) bool { return name.Name == r.Identifier })
+	util.Assert(nameIdx >= 0, "matched spec must declare the targeted identifier")
```
darccio, line 58: `util.Assert` here is process-fatal (`ex.Fatalf`), not a returned error, so it's worth spelling out why this specific condition can never trigger — the re-search uses the identical `name.Name == r.Identifier` comparison the caller already used to select this spec, so the identifier is guaranteed present. A short comment saves the next reader from re-deriving that.

```suggestion
	// nameIdx is guaranteed non-negative: FindNamedDecl already selected this
	// spec via the identical name.Name == r.Identifier comparison, so the
	// targeted identifier is present in spec.Names by construction.
	nameIdx := slices.IndexFunc(spec.Names, func(name *dst.Ident) bool { return name.Name == r.Identifier })
	util.Assert(nameIdx >= 0, "matched spec must declare the targeted identifier")
```

darccio (changes requested):
**Requesting changes:** `docs/rules.md` (lines 1202-1203, "Notes" under the decl-rule section) still documents the pre-fix behavior as the intended contract:

> - If `replace` matches multiple names in a single declaration (e.g., `var a, b = ...`), the replacement expression is cloned and assigned to each name.
> - If `wrap` matches multiple initialized values in a single declaration, each initializer is wrapped independently.

This is exactly the bug #1044 reports, described as correct behavior. Please update these two bullets before merging — a rule author reading the docs after this lands would otherwise expect fan-out semantics that no longer exist, and would be surprised by the new `declArityError` on tuple-valued declarations instead. Suggested replacement text:

```
- If `replace` targets one name in a multi-name declaration (e.g., `var a, b = ...`), only that name's initializer is replaced; sibling names are left untouched. If the declaration's names and initializers don't correspond one-to-one (e.g., a tuple-valued `var a, b = f()`), the rule reports an error instead of guessing.
- If `wrap` targets one name in a multi-name declaration, only that name's initializer is wrapped; sibling names are left untouched. The same one-to-one requirement applies as for `replace`.
```

darkraider01:
Appreciate the thoroughness here, honestly learning a lot from this review , the "validate before mutate" framing especially, and the habit of asking "does the success path have the same coverage as the failure path" for stuff like the import injection. Applied all six, should be good for another look whenever you get a chance.

darccio (approved):

PR #1047, opened 2026-08-11
fix(tool): allow profile paths sharing build temp prefix

### LINKED ISSUE #1042: bug(tool/cmd/otelc): --profile-path rejects sibling directories sharing the build-temp prefix
### Describe the bug

  `--profile-path` rejects directories whose names begin with `.otelc-build`, even when they are siblings of the
  build directory rather than inside it.

  For example, `.otelc-build-profiles` is rejected as being inside `.otelc-build`.

  ### Environment

  - OS: Windows
  - Go Version: 1.25.5
  - Version: main at `efb163d`

  ### To Reproduce

  From the repository root, run:

  `go run -buildvcs=false ./tool/cmd/otelc --work-dir . --profile-path .otelc-build-profiles --profile heap version`

  The command fails with:

  `profile-path "...\.otelc-build-profiles" must not be inside the build temp directory "...\.otelc-build"`

  ### Expected behavior

  `.otelc-build` itself and its descendants should be rejected, but sibling directories such as `.otelc-build-
  profiles` should be accepted.

  ### Additional context

  `initProfiling` currently checks containment with:

  ```go
  strings.HasPrefix(profilePath, buildTemp)
  ```

ADITYA-CODE-SOURCE:
## Description

Replace the textual prefix check for `--profile-path` with path-aware directory containment using `filepath.Rel`.

The `.otelc-build` directory itself and its descendants remain rejected, while sibling directories sharing the same prefix, such as `.otelc-build-profiles`, are now accepted.

Regression tests cover both the exact build directory and a prefix-sharing sibling.

## Motivation

The previous `strings.HasPrefix` check treated sibling paths beginning with `.otelc-build` as though they were inside the build temporary directory.

Fixes #1042

### DIFF
--- a/tool/cmd/otelc/profiling.go
+++ b/tool/cmd/otelc/profiling.go
@@ -46,7 +46,11 @@ func initProfiling(ctx context.Context, cmd *cli.Command) (context.Context, erro
 
 	// Guard against placing profiles inside .otelc-build/ which Cleanup removes.
 	buildTemp := util.GetBuildTempDir()
-	if strings.HasPrefix(profilePath, buildTemp) {
+	rel, err := filepath.Rel(buildTemp, profilePath)
+	if err != nil {
+		return ctx, ex.Wrapf(err, "resolve profile path relative to build temp directory")
+	}
+	if rel == "." || (rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator))) {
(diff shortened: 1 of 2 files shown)


ADITYA-CODE-SOURCE:
Closing because #1052 merged the same fix and resolved #1042.

PR #1049, opened 2026-08-11
fix(instrumentation/k8s.io/client-go): forward original object on OnDelete

### LINKED ISSUE #1048: bug(instrumentation/k8s.io/client-go): OnDelete forwards unwrapped object, stripping DeletedFinalStateUnknown from the user's handler
### Describe the bug

`k8SOtelEventHandler.OnDelete` in `instrumentation/k8s.io/client-go/k8s_otel_handler.go` unwraps `cache.DeletedFinalStateUnknown` to extract object metadata for span attributes, but then forwards that same *unwrapped* value to the wrapped user handler instead of the original object:

```go
func (h k8SOtelEventHandler) OnDelete(obj any) {
	if o, ok := obj.(cache.DeletedFinalStateUnknown); ok {
		obj = o.Obj
	}

	objInfo := getObjectInfo(obj)
	...
	h.handler.OnDelete(obj)
}
```

`h.handler` is the application's real `cache.ResourceEventHandler` - confirmed via `beforeProcessDeltas` in `hook.go`, which fully replaces the informer's handler parameter with this wrapper (`ictx.SetParam(0, newK8SOtelEventHandler(handler, ctx))`). So every application using k8s client-go informers with `otelc` compile-time instrumentation loses the `DeletedFinalStateUnknown` wrapper on every delete event, regardless of whether it was actually a tombstone.

`DeletedFinalStateUnknown` signals that client-go's local cache lost track of the object (e.g. a missed watch event during a resync), so the delete is inferred rather than confirmed, and the wrapped object may be stale. Kubernetes controllers commonly type-assert on this to decide how to handle the delete (e.g. skip finalizer logic on a possibly-stale object). Stripping it silently changes application-observable behavior based on whether instrumentation is applied, not just what gets exported as telemetry.

### Environment

 - OS: Linux (not platform-specific)
 - Go Version: 1.25
 - Version: main (450a2b6)

### To Reproduce

1. Instrument an app with `k8s.io/client-go` that registers an informer with an `OnDelete(obj any)` handler checking `if _, ok := obj.(cache.DeletedFinalStateUnknown); ok { ... }`.
2. Trigger a delete event that arrives as `DeletedFinalStateUnknown` (e.g. during a cache resync after a missed watch event).
3. The type assertion in the user's handler never succeeds under `otelc` instrumentation, even though it would without it.

### Expected behavior

The instrumentation should pass through the original `obj` (wrapped or not) to `h.handler.OnDelete`, using a separate variable only for the span-attribute extraction.

### Additional context

`TestOnDelete` in `k8s_otel_handler_test.go` only exercises a plain `*corev1.Pod`, never a `cache.DeletedFinalStateUnknown`-wrapped object, which is why this wasn't caught.

Relevant code:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/450a2b6abdd37a46624a24f825e7027c219928fd/instrumentation/k8s.io/client-go/k8s_otel_handler.go#L59-L76
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/450a2b6abdd37a46624a24f825e7027c219928fd/instrumentation/k8s.io/client-go/hook.go#L50-L78

om7057:
## Description

`k8SOtelEventHandler.OnDelete` unwraps a `cache.DeletedFinalStateUnknown` tombstone to build span attributes, but then forwards that unwrapped value to the wrapped user handler instead of the original object. `beforeProcessDeltas` fully replaces the informer's `cache.ResourceEventHandler` with this wrapper, so the wrapped handler is the application's real `OnDelete` callback.

Kubernetes controllers commonly type-assert on `DeletedFinalStateUnknown` to detect a possibly-stale, inferred delete (a missed watch event surfaced during a resync) versus a confirmed one. Under `otelc` instrumentation, that assertion always fails, since the wrapper never reaches the application. Compile-time instrumentation should only add telemetry, not change observable application behavior.

## Changes

- Keep the unwrap local to metadata extraction (`getObjectInfo`); forward the original, possibly-wrapped `obj` unchanged to `h.handler.OnDelete`.
- Added `TestOnDelete_TombstoneForwardsWrappedObject`, which wraps a pod in `cache.DeletedFinalStateUnknown` and asserts the inner handler receives the wrapper (not the unwrapped pod), while span attributes still reflect the unwrapped object's metadata. Verified this test fails against the old code (`got *v1.Pod` instead of the tombstone) and passes against the fix.

Fixes #1048

---

## Checklist

- [x] Tests pass: `make test`


amazingakai (approved):

amazingakai:
@om7057 needs a rebase

om7057:
> @om7057 needs a rebase

Rebased; bundle regenerated too. 
Please check once more @amazingakai !

amazingakai:
Looks good, good catch!

om7057:
> Looks good, good catch!

Thanks!

om7057:
@darccio, could you please check once and merge, since it's approved by @amazingakai!

om7057:
@y1yang0 @darccio this PR is approved, could you check once please!

PR #1050, opened 2026-08-11
fix(instrumentation/net/http): http.ServeFile loses the sendfile fast path on instrumented servers

ipsitapp8:
## Description

`BeforeServeHTTP` swaps the handler's `http.ResponseWriter` for `writerWrapper` so the status code can be read back afterwards. net/http's own `*response` implements `io.ReaderFrom`, and that is the method `io.Copy` looks for on the destination — it's what ends up handing the body to the connection and reaching `sendfile(2)`.

`writerWrapper` didn't implement it. So once the wrapper is in place, `io.Copy` no longer finds a `ReaderFrom` and falls back to its generic 32KiB copy loop. `http.ServeFile` and `http.ServeContent` both end in `io.CopyN(w, sendContent, sendSize)`, so every file served by an instrumented server takes the slow path — not because of anything the application did, but because it's instrumented. The response is still correct, which is presumably why it went unnoticed.

This is the same shape as #795, which added `http.Pusher` to this wrapper: it shouldn't hide a capability of the writer it replaces.

## Fix

Implement `io.ReaderFrom` on `writerWrapper`. If the underlying writer has one (the real net/http case), delegate to it. If it doesn't, copy through `Write` behind a small `writeOnly` shim so `io.Copy` can't pick this method up again and recurse. Both paths record the implicit `200` first, the same as `Write` already does, so status capture doesn't change.

The second commit implements `io.StringWriter` for the same reason: net/http's `*response` implements it so `io.WriteString` can skip a `[]byte` conversion, and hiding it forces that conversion back on every string write. Same delegate-or-fall-back structure, same implicit-200 handling.

Both interfaces are added to the compile-time assertion block that already lists `Hijacker`, `Flusher` and `Pusher`.

## Performance

Measured on Linux (WSL), 1MB file over loopback, comparing against the pre-fix wrapper forced onto the copy loop:

| | latency | alloc/request |
|---|---|---|
| pre-fix wrapper (copy loop) | 24.7ms | 41KB |
| with this fix | 16.9ms | 10KB |
| unwrapped (no instrumentation) | 14.2ms | — |

The ~31KB allocation difference is the 32KiB buffer `io.Copy` allocated on every request while `ReadFrom` was hidden. The fix lands close to the unwrapped ceiling, so the remaining gap is wrapper overhead unrelated to the copy path.

These numbers come from a local benchmark comparing the copy loop against the fast path. Per review feedback the benchmark isn't part of the diff — it needed a duplicate of the old wrapper to measure against, which isn't worth carrying in the tree.

## Motivation

Instrumentation shouldn't make the application slower in ways the application can't see or control. No existing issue covers this — I found it reading the wrapper while looking at something else. Happy to open one if you'd rather it were tracked.

To be clear about the scope: this restores performance, it does not fix incorrect output. Responses were always byte-correct, and the improvement is zero for handlers that don't serve files or write strings.

## Notes on the tests

The ones that matter are `TestWriterWrapper_ReadFromUsesUnderlyingFastPath` and `TestWriterWrapper_WriteStringUsesUnderlyingFastPath`, which fail on `main` — I checked by stashing the fix and re-running.

One thing that caught me out and is worth knowing if you touch these tests: `io.Copy` prefers a `WriterTo` on the **source** over a `ReaderFrom` on the destination. My first version used `strings.NewReader` as the body and passed without the fix, because the copy never consulted the writer at all. The tests use an `*io.LimitedReader` instead — no `WriterTo`, and it's the exact shape `http.ServeContent` produces via `io.CopyN`. There's a comment on the helper so it doesn't get simplified back later.

A related trap, caught in review: `httptest.ResponseRecorder` **does** implement `io.StringWriter`. The `WriteString` fallback test can't be built on the recorder, because it would take the fast path and never exercise the fallback — it passed for the wrong reason, since both paths emit identical bytes. That test now uses a minimal writer implementing exactly `http.ResponseWriter`, asserts the fallback actually goes through `Write`, and guards the premise so it can't silently drift back onto the fast path. (`ResponseRecorder` has no `ReadFrom`, so the `ReadFrom` fallback test is unaffected.)

`tool/data/otelc-bundle.tgz` is regenerated since `instrumentation/` changed.

---

## Checklist

- [x] Tests pass: `make test`
- [ ] Documentation updated (if applicable) — n/a
- [ ] OpenTelemetry Registry updated (if applicable) — n/a, no new instrumentation

[^I-know-my-stuff]:


ipsitapp8:
@amazingakai can you review this pr?

instrumentation/net/http/server/response_writer.go
```diff
@@ -46,6 +48,34 @@ func (w *writerWrapper) Write(b []byte) (int, error) {
...
+	io.Writer
+}
+
+// ReadFrom implements io.ReaderFrom so that io.Copy keeps reaching the
+// underlying ResponseWriter's own ReadFrom. net/http's *response implements
+// io.ReaderFrom and hands the body to the connection, which lets the kernel
+// serve it with sendfile(2). Without this method the wrapper hides that
+// capability from io.Copy, and http.ServeFile/http.ServeContent silently fall
+// back to a 32KiB user-space copy loop on every instrumented server.
+func (w *writerWrapper) ReadFrom(src io.Reader) (int64, error) {
```
amazingakai, line 64: Should we also implement `WriteString` (`io.StringWriter`): https://github.com/golang/go/blob/master/src/net/http/server.go#L1654?

amazingakai (approved):
Thanks for this PR! I don't like this whack-a-mole game we have to play, but I guess we don't have a choice here.

It'd be great if you can include the performance difference, but I think we should be implementing these regardless.

instrumentation/net/http/server/response_writer.go
ipsitapp8: io.WriteString checks for io.StringWriter the same way io.Copy checks for io.ReaderFrom, so it's the same problem, without it we fall back to converting to [ ]byte and losing the fast path. I'll add it, same pattern as ReaderFrom, delegate if the underlying writer has it.

ipsitapp8:
> Thanks for this PR! I don't like this whack-a-mole game we have to play, but I guess we don't have a choice here.
> 
> It'd be great if you can include the performance difference, but I think we should be implementing these regardless.

 this wrapper keeps growing every time net/http adds an optional interface. I don't have a benchmark yet, I'll add one comparing the copy loop vs sendfile for a reasonably sized file and post the numbers here before merge.

ipsitapp8:
Added `WriteString` and ran the benchmark. On Linux, comparing against the old wrapper (no `ReadFrom`, forced onto the copy loop), latency drops from 24.7ms to 16.9ms and allocation drops from 41KB to 10KB per request for a 1MB file, serving it unwrapped is 14.2ms so the fix gets close to that. The 31KB difference is the 32KiB copy buffer `io.Copy` was allocating every request when `ReadFrom` was hidden.

I benchmarked on WSL/Linux, not Windows, sendfile only kicks in on Linux in this code path so that's the number that matters.

ipsitapp8:
@amazingakai can you check the pr if its okay now?

instrumentation/net/http/server/response_writer_bench_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright The OpenTelemetry Authors
```
amazingakai: Is this needed?  You can put the improvement in the PR description

instrumentation/net/http/server/response_writer_test.go
```diff
@@ -186,6 +190,187 @@ func TestWriterWrapper_Push_NotSupported(t *testing.T) {
...
+
+func TestWriterWrapper_WriteString_NotSupported(t *testing.T) {
+	recorder := httptest.NewRecorder()
+	wrapper := &writerWrapper{
+		ResponseWriter: recorder,
+		statusCode:     http.StatusOK,
+	}
+
+	// The recorder has no WriteString, so the wrapper falls back to Write.
+	n, err := io.WriteString(wrapper, "payload")
```
amazingakai: `httptrest.ResponseRecorder` does implement `StringWriter`

amazingakai (commented):
Mostly done. The bundle needs to be regenerated

instrumentation/net/http/server/response_writer_bench_test.go
ipsitapp8: Removed. The benchmark needed a copy of the old wrapper to measure against, which isn't worth carrying in the tree just for that. Numbers are in the PR description now: 24.7ms -> 16.9ms and 41KB -> 10KB per request for a 1MB file on Linux, against an unwrapped baseline of 14.2ms.

instrumentation/net/http/server/response_writer_test.go
ipsitapp8: Good catch, and it was worse than a wrong comment. Since `ResponseRecorder` implements `StringWriter`, that test was taking the fast path and never exercising the fallback at all. It only passed because both paths emit identical bytes, so the fallback branch had no coverage.

Replaced the recorder with a minimal writer implementing exactly `http.ResponseWriter`, and the test now asserts the fallback really goes through `Write` rather than just checking the output. Also added a guard on the premise so it fails loudly if that type ever grows a `WriteString` and silently drifts back onto the fast path.

`ReadFrom` is unaffected, `ResponseRecorder` genuinely has no `ReadFrom`, so that fallback test was testing what it claimed.

instrumentation/net/http/server/response_writer.go
ipsitapp8: Implemented in the second commit, delegating to the underlying `io.StringWriter` when there is one and falling back to `Write([]byte(s))` otherwise, with the same implicit-200 handling as `ReadFrom`. Both interfaces are in the compile-time assertion block now.

ipsitapp8:
Rebased on main and regenerated the bundle, verify-bundle passes locally now. @amazingakai should be ready for another look.

amazingakai (approved):

amazingakai:
Please regenerate the bundle

ipsitapp8:
@amazingakai  regenerated the bundles and the ci is green now

PR #1052, opened 2026-08-11
fix(tool/cmd/otelc): allow profile-path sibling directories sharing build-temp prefix

### LINKED ISSUE #1042: bug(tool/cmd/otelc): --profile-path rejects sibling directories sharing the build-temp prefix

SatyamPandey-07:
Fixes #1042

### Description
In `tool/cmd/otelc/profiling.go`, `initProfiling` previously checked if `--profile-path` was inside the build temp directory (`.otelc-build`) using `strings.HasPrefix(profilePath, buildTemp)`.

Because string prefix matching is path-boundary unaware, sibling directories sharing the prefix (e.g. `.otelc-build-profiles` or `.otelc-build-123`) were incorrectly matched and rejected as being inside `.otelc-build`.

This PR replaces `strings.HasPrefix` with a path-boundary aware `isSubPath(target, base)` helper leveraging `filepath.Rel`. It also adds unit tests covering sibling directory paths and exact build temp directory paths.

### Checklist
- [x] Tests pass: `make test`

### DIFF
--- a/tool/cmd/otelc/profiling.go
+++ b/tool/cmd/otelc/profiling.go
@@ -46,7 +46,7 @@ func initProfiling(ctx context.Context, cmd *cli.Command) (context.Context, erro
 
 	// Guard against placing profiles inside .otelc-build/ which Cleanup removes.
 	buildTemp := util.GetBuildTempDir()
-	if strings.HasPrefix(profilePath, buildTemp) {
+	if isSubPath(profilePath, buildTemp) {
 		return ctx, ex.Newf(
 			"profile-path %q must not be inside the build temp directory %q",
 			profilePath, buildTemp,
@@ -116,4 +116,13 @@ func stopProfiling(ctx context.Context, cmd *cli.Command) error {
 	logger.InfoContext(ctx, "merging profile files", "dir", profileDir)
 	mergeErr := profile.Merge(ctx, profileDir, types)
 	return ex.Join(stopErr, mergeErr)
+}
+
+// isSubPath reports whether target is equal to base or located inside base.
+func isSubPath(target, base string) bool {
+	rel, err := filepath.Rel(base, target)
+	if err != nil {
+		return false
+	}
+	return rel == "." || !strings.HasPrefix(rel, "..")
 }
--- a/tool/cmd/otelc/profiling_test.go
+++ b/tool/cmd/otelc/profiling_test.go
@@ -68,6 +68,31 @@ func TestInitProfiling(t *testing.T) {
 		assert.Contains(t, err.Error(), "build temp")
 	})
 
+	t.Run("profile-path equal to build temp errors", func(t *testing.T) {
+		activeSession = nil
+		workDir := t.TempDir()
+		t.Setenv(util.EnvOtelcWorkDir, workDir)
+		buildTemp := util.GetBuildTempDir()
+		err := runInitProfiling(t, "--profile", "cpu", "--profile-path", buildTemp)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "build temp")
+	})
+
+	t.Run("profile-path sharing prefix with build temp is allowed", func(t *testing.T) {
+		activeSession = nil
+		workDir := t.TempDir()
+		t.Setenv(util.EnvOtelcWorkDir, workDir)
+		t.Setenv(profile.EnvProfilePath, "")
+		t.Setenv(profile.EnvEnabledProfiles, "")
+		sibling := util.GetBuildTempDir() + "-profiles"
+
+		require.NoError(t, runInitProfiling(t, "--profile", "cpu", "--profile-path", sibling))
+		require.NotNil(t, activeSession)
+
+		require.NoError(t, activeSession.Stop())
+		activeSession = nil
+	})
+
 	t.Run("valid profile starts a session and sets env", func(t *testing.T) {
 		activeSession = nil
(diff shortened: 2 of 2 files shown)


SatyamPandey-07:
@darccio review the pr when u are free

tool/cmd/otelc/profiling.go
```diff
@@ -117,3 +117,12 @@ func stopProfiling(ctx context.Context, cmd *cli.Command) error {
...
 	return ex.Join(stopErr, mergeErr)
 }
+
+// isSubPath reports whether target is equal to base or located inside base.
+func isSubPath(target, base string) bool {
+	rel, err := filepath.Rel(base, target)
+	if err != nil {
+		return false
+	}
+	return rel == "." || (!strings.HasPrefix(rel, "..") && rel != "..")
```
darccio: `rel != ".."` can never affect the result here: when `rel == ".."`, `strings.HasPrefix(rel, "..")` is already `true`, so the `!strings.HasPrefix(...)` half of the `&&` is already `false` regardless of the second operand. Simplifying avoids a reader having to work out that it's dead.

```suggestion
	return rel == "." || !strings.HasPrefix(rel, "..")
```

tool/cmd/otelc/profiling.go
```diff
@@ -117,3 +117,12 @@ func stopProfiling(ctx context.Context, cmd *cli.Command) error {
...
+}
```
darccio, line 128: `tool/internal/setup/extract.go:130-131` has an equivalent `filepath.Rel`-based "is target inside base" check (used there for zip-slip protection during archive extraction). Not asking for a change in this PR, but as this is now the second occurrence of the same pattern, a shared `pathutil.IsWithin`-style helper might be worth it if a third call site ever shows up.

tool/cmd/otelc/profiling_test.go
```diff
@@ -124,3 +149,57 @@ func TestStopProfiling(t *testing.T) {
...
+			target:   filepath.Join(base, "sub", "dir"),
+			base:     base,
+			expected: true,
+		},
+		{
+			name:     "sibling path with shared prefix",
+			target:   base + "-profiles",
+			base:     base,
+			expected: false,
+		},
```
darccio, line 179: Coverage looks solid for the reported bug, but two boundary cases aren't exercised: a path nested under a sibling directory (e.g. `.otelc-build-profiles/sub/dir`, which should also be allowed) and a `base` input with a trailing separator. Both should already pass given how `filepath.Rel` normalizes internally — this is a "nice to have" for future-proofing, not a gap in the current fix.

```suggestion
		{
			name:     "sibling path with shared prefix",
			target:   base + "-profiles",
			base:     base,
			expected: false,
		},
		{
			name:     "nested path under sibling with shared prefix",
			target:   filepath.Join(base+"-profiles", "sub", "dir"),
			base:     base,
			expected: false,
		},
		{
			name:     "target with trailing separator equal to base",
			target:   base + string(filepath.Separator),
			base:     base,
			expected: true,
		},
```

darccio (changes requested):
This branch is 3 commits behind `main` (missing `0702029d`, `66d4af33`, `11106215`). That matters here specifically because the `apply_func.go` / `apply_func_test.go` changes in this PR (the `collectArguments` blank-receiver `idx` → `next(ignoredParam)` fix, and the matching test expectation update) are byte-for-byte identical to a fix already merged independently via #904 (commit `66d4af33`).

Could you rebase onto current `main`? After rebasing, these two hunks should collapse to no-ops since `main` already contains the "after" state they're patching toward — worth confirming that's clean rather than assuming. Once rebased, this PR would be scoped to just the profile-path fix described above, which makes it much easier to review in isolation.

SatyamPandey-07:
Thanks @darccio — all addressed:

Rebased onto current main. You were right: the apply_func.go/apply_func_test.go hunks collapsed to no-ops since #904 already merged the identical fix. Dropped cleanly — this PR is now scoped to just profiling.go + profiling_test.go (101 lines, 2 files).
Simplified isSubPath: rel == "." || (!strings.HasPrefix(rel, "..") && rel != "..") → rel == "." || !strings.HasPrefix(rel, ".."), per your point that rel != ".." was dead.
Added both boundary cases you suggested: nested path under a sibling directory, and a base with a trailing separator. Both pass as expected.
pathutil.IsWithin helper: leaving that for if/when a third call site shows up, per your note.
Force-pushed the rebase + fixups as a single clean commit. Ready for re-review whenever you get a chance.

darccio (approved):

PR #1055, opened 2026-08-11
fix(tool): use collision-safe names for blank receivers

### LINKED ISSUE #1054: bug(tool): blank receivers bypass collision-safe synthetic naming
### Describe the bug

PR #1035 introduced collision-safe synthetic names for unnamed and blank parameters and return values. However, the blank receiver branch in `collectArguments` still uses the old `_ignoredParamN` naming scheme instead of `syntheticNamer`.

In `tool/internal/instrument/apply_func.go`, blank receivers are currently renamed using logic equivalent to:

```go
receiver = fmt.Sprintf("%s%d", ignoredParam, idx)
```

Unnamed receivers and blank parameters instead use:

```go
next(ignoredParam)
```

As a result, blank receivers can still collide with identifiers already declared in the function signature.

### Environment

- OS: any
- Go version: any supported version
- Version: current `main`, after #1035

### To Reproduce

Instrument a method with a blank receiver and a parameter using the legacy generated receiver name:

```go
package example

type T struct{}

func (_ T) Method(_ignoredParam0 int) {}
```

Apply a function rule targeting `Method`.

The blank receiver may be renamed to `_ignoredParam0`, producing a signature equivalent to:

```go
func (_ignoredParam0 T) Method(_ignoredParam0 int) {}
```

The generated source fails to compile because `_ignoredParam0` is redeclared in the same scope.

### Expected behavior

Blank receivers should use the same rule-identity-based, collision-safe synthetic name generator used for unnamed receivers, blank parameters, and return values.

### Proposed fix

Use `next(ignoredParam)` in the blank receiver branch of `collectArguments` and add regression coverage for a blank receiver whose legacy generated name collides with an existing parameter.

ADITYA-CODE-SOURCE:
## Description

Use `syntheticNamer` when remapping blank method receivers.

PR #1035 migrated generated parameter and return names to collision-safe, rule-identity-based names, but the blank-receiver branch retained the old counter-based implementation. Because the local counter was removed in the same migration, current `main` also fails to compile with `undefined: idx`.

This change routes blank receivers through the existing `next(ignoredParam)` generator.

## Motivation

Blank receivers should follow the same naming path as unnamed receivers and blank parameters. This prevents collisions with existing signature identifiers and restores compilation on current `main`.

Fixes #1054

### DIFF
--- a/tool/internal/instrument/apply_func.go
+++ b/tool/internal/instrument/apply_func.go
@@ -123,8 +123,7 @@ func collectArguments(funcDecl *dst.FuncDecl, hash string) []string {
 				// so it falls into this branch, but "_" cannot have its
 				// address taken during trampoline generation the way a named
 				// receiver can. Assign it a generated name instead.
-				receiver = fmt.Sprintf("%s%d", ignoredParam, idx)
-				idx++
+				receiver = next(ignoredParam)
 				recv.Names[0].Name = receiver
 			}
 			args = append(args, receiver)
--- a/tool/internal/instrument/apply_func_test.go
+++ b/tool/internal/instrument/apply_func_test.go
@@ -104,7 +104,12 @@ func TestCollectArguments(t *testing.T) {
 		{
 			name:     "underscore receiver",
 			src:      "package main\ntype T struct{}\nfunc (_ T) F() {}",
-			expected: []string{"_ignoredParam0"},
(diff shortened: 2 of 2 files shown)


ADITYA-CODE-SOURCE:
Closing because #904 merged the underlying blank-receiver fix. The remaining additional tests can be proposed separately if maintainers want them.

Issue #985, opened 2026-08-06
docs: remove bundle-sync instructions, describe manifest-driven instrumentation versioning [6/6]


darccio:
> [!WARNING]
> This is part of an on-going discussion. Don't go ahead and implement it. Each issue will be assigned to an owner after each proposal is approved by the maintainers and only the owner will implement the functionality in one or more PRs.

# docs: remove bundle-sync instructions, describe manifest-driven instrumentation versioning [6/6]

**Tracking:** part of #585 (decouple built-in instrumentations from `otelc-bundle.tgz`)
**Depends on:** soft dependency on #981 and #983 for wording accuracy (this issue describes the post-change state) — can be drafted early, but should be merged after those land
**Blocks:** none

## Background

Prior to #585, contributors had to run a git merge driver setup step (`make setup-git`) and manually regenerate `tool/data/otelc-bundle.tgz` (`make package`) after every rebase touching `pkg/` or `instrumentation/`, documented in `CONTRIBUTING.md`'s "Keeping the bundle in sync" section. Several docs pages also describe built-in instrumentation discovery in terms of an "embedded rule bundle."

Once #981 removes the bundle mechanism entirely (replaced by a small committed JSON manifest that merges/diffs like ordinary text, needing no custom merge driver) and #983 makes built-in instrumentation versions genuinely independently pinnable via `go.mod`, this documentation needs a pass to match reality and to explain the new, more flexible versioning model to users.

### Tasks

- [ ] Remove the "Keeping the bundle in sync" section from `CONTRIBUTING.md` (the part describing `make package`, the git merge driver, and `git commit --amend` after every rebase).
- [ ] Remove the `make setup-git` step from `CONTRIBUTING.md`'s initial setup instructions.
- [ ] Confirm the `tool/data/otelc-bundle.tgz` merge-driver line in `.gitattributes` is gone (should already be removed in #981 — double check here).
- [ ] Update `docs/configuration.md`, `docs/external-configuration.md`, and `docs/troubleshooting.md`: replace references to the "embedded rule bundle" with language describing the manifest-driven auto-pin discovery mechanism.
- [ ] Update `docs/getting-started.md`: add a short note that built-in instrumentation module versions are now independently pinnable via `go.mod` (the direct user-facing payoff of #585) — e.g. "you can pin an older version of a specific built-in instrumentation independently of the `otelc` tool version, the same way you'd manage any other Go dependency."
[...]

Issue #983, opened 2026-08-06
feat(tool): pin real instrumentation module versions instead of local bundle replace [4/6]


darccio:
> [!WARNING]
> This is part of an on-going discussion. Don't go ahead and implement it. Each issue will be assigned to an owner after each proposal is approved by the maintainers and only the owner will implement the functionality in one or more PRs.

# feat(tool): pin real instrumentation module versions instead of local bundle `replace` [4/6]

**Tracking:** part of #585 (decouple built-in instrumentations from `otelc-bundle.tgz`) — this is the semantic core of the issue
**Depends on:** #981, #982
**Blocks:** none directly (but #984's value only materializes after this lands)

## Background

This is the change that actually decouples instrumentation versions from the `otelc` tool version — everything in #980/#981/#982 is groundwork for this.

`tool/internal/setup/sync.go`'s `syncDeps()` (lines ~120-194) currently contains this, at lines ~135-157:

```go
// Add replace directives for modules imported to otel.instrumentation.go
replaces := make(map[string]string, len(modPaths))
for m := range modPaths {
    if path, isEmbedded := strings.CutPrefix(m, util.OtelcInstRoot+"/"); isEmbedded {
        replaces[m] = filepath.Join(util.GetBuildTempDir(), unzippedInstDir, path)
    }
}

// Add replace directive for special pkg module
// TODO: Since we haven't published the instrumentation packages yet,
// we need to add the replace directive to the local path.
// Once the instrumentation packages are published, we can remove this.
replaces[util.OtelcPkgRoot] = filepath.Join(util.GetBuildTempDir(), unzippedPkgDir)
replaces[util.OtelcPkgRoot+"/runtime"] = filepath.Join(util.GetBuildTempDir(), unzippedPkgDir, "runtime")
replaces[util.OtelcInstRoot] = filepath.Join(util.GetBuildTempDir(), unzippedInstDir)
```

This unconditionally rewrites the **user's own `go.mod`** to redirect every built-in instrumentation module, `pkg`, `pkg/runtime`, and the shared `instrumentation` anchor module to a local, bundle-extracted copy — which is precisely why instrumentation versions can't be independently pinned today: no matter what a user's `go.mod` says, the tool silently overrides it with a `replace` pointing at whatever source shipped inside the running `otelc` binary. The maintainers' own TODO comment above already flags this as the thing to remove once instrumentation packages are independently resolvable (which #981 + #982 make true).

[...]

PR #1057, opened 2026-08-11
feat(tool): discover instrumentation pins from manifest

### LINKED ISSUE #981: feat(tool): use instrumentation manifest for pin discovery, remove otelc-bundle.tgz [2/6]

ADITYA-CODE-SOURCE:
## Description

Replace bundle-based built-in instrumentation discovery with the generated instrumentation manifest.

This change:

- loads built-in instrumentation metadata through `manifest.Load()`
- preserves exact, glob, `$root`, and version-range matching behavior
- removes bundle extraction, embedding, packaging, verification workflows, scripts, and merge-driver configuration
- updates build and install targets to generate the manifest instead of packaging `otelc-bundle.tgz`
- removes contributor instructions for the deleted bundle workflow; the broader documentation update remains tracked by #985
- adds the maintainer-approved temporary source-checkout bridge for local `replace` directives
- exports the checkout root to all Make targets so unit, integration, E2E, latest-library, benchmark, and coverage subprocesses receive it consistently

The bridge resolves the repository root from `OTELC_SOURCE_ROOT` when set, or by walking upward from the current directory. It will be removed by #983 when built-in modules receive explicit versioned requirements.

## Motivation

`otelc pin` previously extracted the complete embedded source bundle only to discover which built-in instrumentation modules matched application dependencies. The generated manifest introduced by #999 contains that metadata directly, making extraction and the embedded archive unnecessary.

The temporary source-checkout bridge keeps built-in instrumentation builds functional while #983 is developed separately.

Fixes #981


tool/internal/setup/sync.go
```diff
@@ -20,6 +21,35 @@ import (
...
+			util.PathExists(filepath.Join(root, "instrumentation", "go.mod")) {
+			return root, nil
+		}
+		parent := filepath.Dir(root)
+		if parent == root {
+			break
+		}
+		root = parent
+	}
+	return "", ex.New("otelc source checkout not found")
```
darccio: The failure here ("otelc source checkout not found") is a new hard-failure
mode for `otelc pin` introduced by removing the bundle — previously
`syncDeps` always succeeded because the bundle was self-contained.
Mentioning `OTELC_SOURCE_ROOT` in the error message would help consumers
understand the requirement without reading source, since this is a
temporary limitation until #983 removes it.

```suggestion
	return "", ex.New("otelc source checkout not found; set OTELC_SOURCE_ROOT " +
		"or run from inside a full opentelemetry-go-compile-instrumentation " +
		"checkout (temporary requirement, see #983)")
```

tool/internal/setup/sync.go
```diff
@@ -20,6 +21,35 @@ import (
...
+			util.PathExists(filepath.Join(root, "pkg", "go.mod")) &&
+			util.PathExists(filepath.Join(root, "instrumentation", "go.mod")) {
+			return root, nil
+		}
+		parent := filepath.Dir(root)
+		if parent == root {
+			break
+		}
+		root = parent
+	}
```
darccio, line 49: The walk-up has no bound besides the filesystem root. In practice the
three-go.mod fingerprint (root + pkg/go.mod + instrumentation/go.mod) is
specific enough that a false-positive match is unlikely, but a short
comment would make the intentional lack of a stopping condition clear to
future readers instead of looking like an oversight.

```suggestion
	// No depth limit beyond the filesystem root: the combination of a root
	// go.mod plus pkg/go.mod and instrumentation/go.mod nested under it is
	// specific enough that false positives are not expected in practice.
	for {
		if util.PathExists(filepath.Join(root, "go.mod")) &&
			util.PathExists(filepath.Join(root, "pkg", "go.mod")) &&
			util.PathExists(filepath.Join(root, "instrumentation", "go.mod")) {
			return root, nil
		}
		parent := filepath.Dir(root)
		if parent == root {
			break
		}
		root = parent
	}
```

darccio (changes requested):
CONTRIBUTING.md still documents the bundle-based workflow this PR removes,
and it's not just historical color — it's an active Getting Started step.
Step 2 tells new contributors to run `make setup-git`, a target deleted
here; that fails immediately with `make: *** No rule to make target
'setup-git'` before a contributor gets anywhere. The "Keeping the bundle
in sync" section similarly instructs `make package` (deleted) and
references a `/regenerate-bundle` PR-comment command (workflow deleted).

The PR body defers the full doc rewrite to #985, which is reasonable
scope-wise, but leaves a broken onboarding flow in the interim. Consider
a minimal patch now — just removing the two dead instructions (`make
setup-git` step, and the "Keeping the bundle in sync" section) — rather
than leaving them pointing at commands that no longer exist until #985
lands.

tool/internal/setup/sync.go
ADITYA-CODE-SOURCE: Updated the error to name `OTELC_SOURCE_ROOT`, explain that a full checkout is required, and point to #983. I also added assertions for this guidance in both missing-source-root paths.

tool/internal/setup/sync.go
ADITYA-CODE-SOURCE: Added a comment documenting that the walk intentionally continues to the filesystem root because the three-module fingerprint is specific to this repository checkout.

ADITYA-CODE-SOURCE:
@darccio  Addressed in 74db22f: removed the obsolete bundle instructions and rebased onto current main. Local tests and lint checks pass. I’ll request re-review once CI is green.

darccio:
On hold waiting for https://github.com/open-telemetry/community/issues/3639

ADITYA-CODE-SOURCE:
@amazingakai @darccio  Now that #3639 is complete and the contrib repo is available, should I keep #1057 on hold until contrib #5 establishes the migrated module paths, or refresh it on current main now? The current temporary source-checkout bridge predates the contrib design.

darccio:
@ADITYA-CODE-SOURCE Refresh it, we'll tackle the migrated module paths later.

PR #1061, opened 2026-08-12
refactor(tool/internal/setup): use embedded manifest in pin instead of extracting tarball

### LINKED ISSUE #1060: refactor(tool/internal/setup): use embedded manifest in pin instead of extracting tarball
### Description
In `tool/internal/setup/pin.go`, `generatePinnedProjects` currently decompresses the entire `otelc-bundle.tgz` tarball to disk and walks filesystem directories to parse `otelc.yaml` rules via `loadMinimalRules` and `loadModuleRules`.

### Problem
1. **Performance Overhead**: Extracting the tarball and recursively parsing YAML files on every auto-pin invocation adds unnecessary disk I/O and latency.
2. **Code Duplication**: `pin.go` defines its own `yamlRule` struct and `loadMinimalRules` / `loadModuleRules` functions, which duplicate the functionality in `tool/internal/manifest`.
3. **Eager Extraction**: The bundle is decompressed even when zero dependencies match instrumentation targets.

### Proposed Solution
1. Replace `loadMinimalRules` in `pin.go` with `manifest.Load()` from `tool/internal/manifest`.
2. Update `matchInstrumentationImports` to operate directly on `manifest.Manifest`.
3. Extract `otelc-bundle.tgz` only if matching instrumentations were found (`len(imports) > 0`) prior to running `syncDeps`.
4. Remove redundant `yamlRule`, `loadModuleRules`, and `loadMinimalRules` helpers.

vedant21-oss:
## Description

Refactors `generatePinnedProjects` in `tool/internal/setup/pin.go` to consume `manifest.Load()` from `tool/internal/manifest` instead of decompressing `otelc-bundle.tgz` and walking directories to parse YAML rules.

- Removes redundant `yamlRule`, `loadModuleRules`, and `loadMinimalRules` functions.
- Updates `matchInstrumentationImports` to operate directly on `manifest.Manifest`.
- Defers bundle extraction to only run when matching instrumentation packages exist (`len(imports) > 0`).

Closes #1060

## Type of Change

- [x] Refactoring / Code cleanup
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## Checklist

- [x] Follows conventional commit format for PR title (`refactor(tool/internal/setup): ...`)
- [x] Tests added / updated and passing (`go test -race ./...`)
- [x] Code adheres to the repository's coding style

### DIFF
--- a/tool/internal/setup/pin.go
+++ b/tool/internal/setup/pin.go
@@ -166,8 +166,6 @@ func generateOtelInstrumentationGo(imports map[string]bool, opts PinOptions) *ds
 	}
 }
 
-
-
 func ensureOtelcRequireVersion(f *modfile.File, version string) (bool, error) {
 	if !semver.IsValid(version) || version == "v0.0.0" || module.IsPseudoVersion(version) {
 		return false, nil
--- a/tool/internal/setup/pin.go
+++ b/tool/internal/setup/pin.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"go/parser"
 	"go/token"
-	"io/fs"
 	"log/slog"
 	"maps"
 	"os"
@@ -22,10 +21,10 @@ import (
 	"golang.org/x/mod/modfile"
 	"golang.org/x/mod/module"
 	"golang.org/x/mod/semver"
-	"gopkg.in/yaml.v3"
 
 	"go.opentelemetry.io/otelc/tool/ex"
 	"go.opentelemetry.io/otelc/tool/internal/ast"
+	"go.opentelemetry.io/otelc/tool/internal/manifest"
 	"go.opentelemetry.io/otelc/tool/internal/pkgload"
 	"go.opentelemetry.io/otelc/tool/internal/rule"
 	"go.opentelemetry.io/otelc/tool/util"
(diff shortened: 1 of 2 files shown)


darccio:
Closing in favour of #1057

tool/internal/setup/pin.go
```diff
@@ -525,18 +451,13 @@ func generatePinnedProjects(ctx context.Context, moduleDirs map[string]bool, opt
 		return nil, ex.Wrapf(findDepsErr, "finding dependencies")
 	}
 
-	extractErr := extractOtelcBundle()
```
Labreo, line 528: One thing worth noting somewhere: this now assumes the embedded manifest and the bundle are always in sync. A short comment there would save the next person from wondering.

tool/internal/setup/pin.go
```diff
@@ -309,7 +237,7 @@ func ensureOtelcRequire(moduleDir, version string) (bool, error) {
 
 func matchInstrumentationImports(
 	deps []*Dependency,
-	ruleset map[string][]yamlRule,
+	manifestEntries manifest.Manifest,
```
Labreo, line 240: Much cleaner. The flat []Entry with ModulePath on each entry is easier to follow than the old map[string][]yamlRule.

Labreo (commented):
Really nice cleanup, routing through `manifest.Load()` instead of the manual YAML walk simplifies things a lot.

One thing I was wondering about with the deferred extraction: now that `extractOtelcBundle()` only runs when there are matching imports, it might be worth dropping a comment there noting that the embedded manifest and the bundle are expected to stay in sync. CI probably catches any drift, but future readers might not immediately know that.

The lint failure on the first commit looked like it was caused by the double blank lines left behind after removing `yamlRule` and `loadModuleRules`, so good that it's already fixed. On coverage, Codecov seems to be flagging the error branches in `generatePinnedProjects` rather than the no-match path, so those might need error-injection tests to actually close the gap. That said, a test for the "no matching imports, skip extraction" case could still be worth adding just to document the new behaviour, even if it doesn't directly help the Codecov number.

Also noticed that `manifest.go` and `match.go` both have an identical private `isRuleFile`. The duplication predates this PR and removing it would probably need a shared exported helper, so not a blocker at all, maybe just something to track as a follow-up.

Labreo:
Oops really sorry for opening a review on a closed pr.Github showed it as an active pr to me.

PR #1066, opened 2026-08-12
refactor(tool): name the GOWORK off sentinel

### LINKED ISSUE #1065: bug(tool/setup): GOWORK=off tracks an unrelated go.work.sum
### Describe the bug

When workspace mode is explicitly disabled with `GOWORK=off`, auto-pin still tracks a `go.work.sum` file from the current working directory.

`getBackupFiles` runs `go env GOWORK` and treats every non-empty result as a workspace path:

```go
goWorkPath := strings.TrimSpace(string(goWorkOutput))
if goWorkPath != "" {
    goWorkSumPath := filepath.Join(filepath.Dir(goWorkPath), "go.work.sum")
    files = append(files, goWorkSumPath)
}
```

However, `go env GOWORK` returns the literal value `off` when workspace mode is disabled. Because `filepath.Dir("off")` evaluates to `.`, the code adds `./go.work.sum` to the backup list even though no workspace is active.

`AutoPin` then passes this path to `StateManager.TrackAll`:

- If the file exists, it is unnecessarily snapshotted and later restored.
- If the file does not exist, it is recorded for removal if it appears before cleanup.

As a result, a concurrent edit could be overwritten, or a file created by another process in the same directory could be removed during cleanup.

The repository already handles the `off` sentinel correctly in `pkgload.ModuleAndWorkspace`:

```go
workspace := goWork != "" && goWork != "off"
```

### To reproduce

1. Create a temporary module containing a `go.mod`.
2. Set `GOWORK=off`.
3. Call `getBackupFiles` from that directory.
4. Observe that the returned backup list includes `./go.work.sum`.

### Expected behavior

No `go.work.sum` should be tracked when `GOWORK=off`, because workspace mode is explicitly disabled.

### Proposed fix

Exclude the documented `off` sentinel before constructing the workspace sum path:

```go
if goWorkPath != "" && goWorkPath != "off" {
    goWorkSumPath := filepath.Join(filepath.Dir(goWorkPath), "go.work.sum")
    files = append(files, goWorkSumPath)
}
```

A regression test in `TestGetBackupFiles` can set `GOWORK=off` and verify that the result contains the expected module files but no `go.work.sum`.

ADITYA-CODE-SOURCE:
## Description

Define `goWorkOff` for Go's `GOWORK=off` sentinel and use it in `getBackupFiles`.

The behavioral fix and regression test originally included in this PR were merged in #1117. This PR now retains only the small constant improvement requested in review.

## Motivation

Using a named constant documents that `"off"` is a special value returned by `go env GOWORK`, rather than an ordinary workspace path.

Related to #1065.
Follow-up to #1117.

### DIFF
--- a/tool/internal/setup/state.go
+++ b/tool/internal/setup/state.go
@@ -23,6 +23,7 @@ import (
 const (
 	stateDir      = "state"
 	stateFileName = "state.json"
+	goWorkOff     = "off"
 )
 
 // StateManager tracks the original state of files so they can later be restored.
@@ -129,7 +130,7 @@ func getBackupFiles(ctx context.Context, moduleDirs map[string]bool) ([]string, 
 		return nil, ex.Wrapf(err, "failed to get GOWORK environment variable")
 	}
 	goWorkPath := strings.TrimSpace(string(goWorkOutput))
-	if goWorkPath != "" && goWorkPath != "off" {
+	if goWorkPath != "" && goWorkPath != goWorkOff {
 		goWorkSumPath := filepath.Join(filepath.Dir(goWorkPath), "go.work.sum")
 		files = append(files, goWorkSumPath)
 	}


amazingakai:
#1117 was merged with the same fix, should we close this?

Since this was created first and there's a slight improvement here you can keep the `const goWorkOff = "off"` change in this PR to also absorb your contribution

ADITYA-CODE-SOURCE:
Thanks @amazingakai  I’ve reduced the PR to the `goWorkOff` constant change and rebased it onto `main`. Could you please take another look when you have time?

darccio (approved):

Issue #971, opened 2026-08-05
bug(tool/internal/setup): sort moduleDirs in resolveRulePaths for deterministic package resolution


vedant21-oss:
### Description
`resolveRulePaths` in `tool/internal/setup/store.go` iterates over `moduleDirs map[string]bool` directly using `for moduleDir := range moduleDirs`.

Because Go map iteration order is randomized by design, package resolution attempts across candidate module directories occur in arbitrary order. If a package import path exists or resolves differently across multiple candidate module directories, the resolution result or error returned depends on random map order. This introduces non-reproducible build behavior and non-deterministic build artifacts.

### Proposed Solution
Collect and sort the keys of `moduleDirs` using `slices.Sort` prior to iteration to guarantee deterministic path resolution across runs.

### Location
- `tool/internal/setup/store.go:31`

Labreo:
The issue has been fixed and should be closed via #972

vignesh2027:
This looks resolved on main. `resolveRulePaths` now sorts before iterating:

```go
dirs := slices.Sorted(maps.Keys(moduleDirs))
```

so package resolution no longer depends on Go map iteration order. That matches @Labreo's note above that #972 fixed it. Probably safe to close.

PR #1072, opened 2026-08-12
fix(tool/internal/instrument): sort file keys in instrument() for deterministic error reporting

### LINKED ISSUE #1070: bug(tool/internal/instrument): sort file keys in instrument() for deterministic error reporting
### Describe the bug

`instrument()` in `tool/internal/instrument/instrument.go` iterates the file-to-rules map returned by `groupRules` directly:

```go
for file, rules := range groupRules(ip.workDir, rset) {
	root, err := ip.parseFile(file)
	...
	for _, r := range rules {
		funcRule, err1 := ip.applyOneRule(ctx, r, root)
		if err1 != nil {
			return ex.Wrapf(err1, "applying rule %s", r.GetName())
		}
	}
	...
}
```

Go randomizes map iteration order per range statement, and the loop returns on the first error. When more than one file in a package has a failing rule, **which failure is reported is nondeterministic across otherwise-identical runs of the same input**. The `Info`/`Debug` lines emitted per file are likewise interleaved in a varying order.

### Environment

 - OS: Linux (WSL2 / Ubuntu); not platform-specific
 - Go Version: 1.25.7
 - Version: `1110621` (`main`)

### To Reproduce

1. Add the following test to `tool/internal/instrument/`:

```go
func TestGroupRulesIterationOrderVaries(t *testing.T) {
	rset := rule.NewInstRuleSet("m")
	for i := range 8 {
		f := fmt.Sprintf("/abs/file%d.go", i)
		rset.AddDeclRule(f, &rule.InstDeclRule{
			InstBaseRule: rule.InstBaseRule{Name: fmt.Sprintf("r%d", i)},
			Kind:         "var",
			Identifier:   "X",
			Replace:      "1",
		})
	}

	grouped := groupRules("/work", rset)

	seen := map[string]struct{}{}
	for range 20 {
		order := ""
		for f := range grouped {
			order += f + " "
		}
		seen[order] = struct{}{}
	}
	t.Logf("distinct iteration orders over 20 passes: %d", len(seen))
}
```

2. Run `go test -run TestGroupRulesIterationOrderVaries -v ./tool/internal/instrument/`
3. Observe multiple distinct orders — on my run:

```text
distinct iteration orders over 20 passes: 7
```

Since `instrument()` returns on the first error it encounters while walking this map, the order directly determines which of several failing files gets reported.

### Expected behavior

Processing order is stable across runs, so a given broken input always produces the same error message and the same log sequence.

### Additional context

**Severity is limited — generated output is not affected.** I checked the paths that could have made this worse and they are all safe:

- `ip.varDecls`, `ip.hookCtxDecl`, and `ip.hookCtxMethods` are reset and re-derived from the static embedded template on every `materializeTemplate()` call, so the globals file content is identical regardless of which file is processed last.
- `ip.appliedFuncIdentities` persists across files, but `InstFuncRule.Identity()` includes both `Func` and `Recv`, and `findFuncDecl` requires `!HasReceiver` when `recv` is empty. Go forbids duplicate `(Func, Recv)` pairs in one package, so two files cannot produce a colliding identity and no file can be silently skipped.
- `groupRules`' CGO remap is bijective (`resolveCgoFile` maps `foo.cgo1.go` to `foo.go`), so no two source files merge into one key with a nondeterministic rule order.

So this is about error and log reproducibility, not build reproducibility. Filing it because the project has fixed this same class of ordering issue before — #986 (sorting `moduleDirs` in `generatePinnedProjects`/`getBackupFiles`) and #971 (sorting `moduleDirs` in `resolveRulePaths`) — and a non-reproducible error message is a real cost when triaging a failing build or a flaky CI run. Happy to close as not-worth-it if maintainers disagree.

**Suggested fix.** Collect the map keys, `slices.Sort` them, and iterate the sorted slice — matching the approach already used in #986 and #971.

darkraider01:
## Description

`instrument()` iterates the file-to-rules map returned by `groupRules` directly with a bare
`range`. Go randomizes map iteration order per range statement, and the loop returns on the
first error it hits. When a package's rule set spans multiple files and more than one has a
failing rule, which failure gets reported — and the order the per-file `Info`/`Debug` log
lines are emitted in — varies between otherwise-identical runs of the same broken input.

The fix extracts a `sortedFiles` helper that returns `file2rules`'s keys via
`slices.Sorted(maps.Keys(...))` and has `instrument()` range over that instead of the map,
matching the pattern already used in #986 and #971 for the same class of ordering issue.

### Severity

Generated output is unaffected — I checked the three paths that could have made this
worse and confirmed none of them are:

- `ip.varDecls`, `ip.hookCtxDecl`, and `ip.hookCtxMethods` are reset and re-derived from the
  static embedded template on every `materializeTemplate()` call, so the globals file
  content is identical regardless of which file is processed last.
- `ip.appliedFuncIdentities` persists across files, but `InstFuncRule.Identity()` includes
  both `Func` and `Recv`, and `findFuncDecl` requires `!HasReceiver` when `recv` is empty.
  Go forbids duplicate `(Func, Recv)` pairs in one package, so two files can't produce a
  colliding identity and no file can be silently skipped.
- `groupRules`'s CGO remap (`resolveCgoFile`) is bijective (`foo.cgo1.go` → `foo.go`), so no
  two source files merge into one key with a nondeterministic rule order.

So this is about error and log reproducibility during a failing build, not correctness of a
successful one.

### Tests

`TestSortedFiles` builds a rule set with a deliberately unsorted file insertion order and
asserts `sortedFiles` returns them alphabetically, calling it 20 times against the same map
to pin that the result is stable. I verified the test actually catches the regression it's
meant to catch: temporarily reverting `sortedFiles` to a bare `for f := range m` loop made
the test fail with a real order mismatch, confirming it isn't just re-testing that
`slices.Sorted` returns sorted output.

## Motivation

Fixes #1070

## Verification

`make format` and `go test -shuffle=on -timeout=5m -count=1 ./tool/...` (what
`make test-unit/tool` runs) pass locally, including with shuffled test order — relevant
here since the fix itself is about ordering. The integration and e2e halves of `make test`
were verified through this PR's CI jobs rather than locally, since the testcontainers-based
suites had no usable Docker host here.

`.github/workflows/lint-golang.yaml` pins — 0 issues on the changed files. As noted on
#1045/#1046: `.tools/go.mod` pins a newer local linter version, and pristine `main` already
reports pre-existing findings under it in files this PR does not touch.

---

## Checklist

- [x] Tests pass: `make test`


tool/internal/instrument/instrument.go
```diff
@@ -74,7 +84,9 @@ func (ip *InstrumentPhase) instrument(ctx context.Context, rset *rule.InstRuleSe
 			return ex.Wrapf(err, "applying file rule %s to package %s", rule.Name, rset.PackageName)
 		}
 	}
-	for file, rules := range groupRules(ip.workDir, rset) {
+	file2rules := groupRules(ip.workDir, rset)
```
y1yang0: Sort within groupRules and add test in instrument_test.go

y1yang0 (changes requested):

tool/internal/instrument/instrument.go
darkraider01: Done — moved the sort into groupRules itself (now returns (map[string][]rule.InstRule, []string), with the second value pre-sorted) and dropped the separate sortedFiles helper, so there's no unordered map left to accidentally range over at any call site. Removed the standalone test file and added the coverage to TestGroupRules instead, including a new case with file names inserted out of order (zzz.go, aaa.go, mmm.go) to actually exercise the sort rather than just re-asserting stdlib behavior.

y1yang0 (approved):

PR #1074, opened 2026-08-12
feat(instrumentation): add gen_ai.client.operation.duration metric to OpenAI and error.type to Anthropic

### LINKED ISSUE #1025: bug(instrumentation): GenAI client duration metric omissions (OpenAI and Anthropic)
### Describe the bug

Hey folks, I noticed a couple of gaps in how we're handling the `gen_ai.client.operation.duration` metric across the GenAI middlewares. 

According to the GenAI Semantic Conventions, `gen_ai.client.operation.duration` is a Recommended (SHOULD) metric, and `error.type` is a Conditionally Required attribute on it when the operation ends in an error. Right now, neither middleware gets this quite right:

1. **OpenAI (`github.com/openai/openai-go`)**: The duration metric is completely missing. The middleware currently only records spans, and there is no meter or metric recording implemented at all.
2. **Anthropic (`github.com/anthropics/anthropic-sdk-go`)**: The duration metric exists, but it never gets the `error.type` attribute when the operation fails. The code defers adding this with a `// TODO(#679)` comment (which was added back in #775), so that gap is still open.
Related but minor issue:
When `error.type` *is* emitted (on spans for HTTP errors), the format doesn't match between the two middlewares. OpenAI uses the numeric status code (e.g. `"404"`), while Anthropic uses the full status string (e.g. `"404 Not Found"`). Figured it was worth mentioning here since it's a related parity issue!

### Environment

 - OS: macOS (but applies everywhere)
 - Go Version: 1.23
 - Version: `main` (cde2de2)

### To Reproduce

1. Instrument an app with `github.com/openai/openai-go`.
2. Make a request to the API and check the exported metrics,`gen_ai.client.operation.duration` is missing.
3. Instrument an app with `github.com/anthropics/anthropic-sdk-go`.
4. Trigger a failure (like an invalid API key).
5. Check the exported `gen_ai.client.operation.duration` metric,it's there, but `error.type` is missing.

### Expected behavior

- The `github.com/openai/openai-go` instrumentation should emit the `gen_ai.client.operation.duration` metric.
- The `github.com/anthropics/anthropic-sdk-go` instrumentation should include the `error.type` attribute on the duration metric when requests fail.
- Ideally, the `error.type` value format should be consistent across both middlewares.

### Additional context

The deferred Anthropic `error.type` on the metric is commented here:
https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/cde2de202279c67ad3a9da3db86e3f42b31495c6/instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go#L130

Labreo:
## Description

OpenAI (v1/v2/v3) emitted no `gen_ai.client.operation.duration` metric at all —
the middleware was spans-only with no meter or histogram anywhere. Anthropic emitted
the metric but never attached `error.type`, deferring it via a `// TODO(#679)`
comment added in #775.

This PR closes both gaps:

- **OpenAI v1/v2/v3** (`hook.go`): initialise a `Float64Histogram` for
  `gen_ai.client.operation.duration` in `initInstrumentation`, mirroring the
  existing Anthropic setup.
- **OpenAI v1/v2/v3** (`middleware.go`): register a single `defer` that calls
  `Record` after span start. An `errorAttrs` var is populated with `error.type`
  on each error path (transport error: `ErrorType(err)`; HTTP >=400: numeric
  status code string) and is empty on success.
- **Anthropic** (`middleware.go`): remove the `TODO(#679)` deferral and apply the
  same `errorAttrs` pattern. `error.type` on HTTP errors uses the numeric status
  code (consistent with OpenAI) rather than the full `resp.Status` string.
- **All four packages** (`middleware_test.go`): add unit tests asserting the metric
  is recorded exactly once per call and that `error.type` is present with the
  correct value on both transport and HTTP error paths.

## Motivation

`gen_ai.client.operation.duration` is a Recommended GenAI semantic-convention
metric that observability backends expect for latency dashboards and SLOs. OpenAI
users received no duration signal at all, while Anthropic users could not split
the histogram by error type to distinguish slow failures from fast ones.

Fixes #1025

---

## Checklist

- [x] Tests pass: `make test`


Labreo:
Hey folks, just wanted to clear up the recent force pushes on this PR.

I rebased the branch on the latest main to pull in recent setup and config changes. All unit tests, integration tests, and E2E suites are passing cleanly locally on my end.

The Verify Bundle check is currently showing red because of the instrumentation updates in the PR, so maintainers can trigger /regenerate-bundle whenever you get a chance to review. Ready for review when you are!

Labreo:
Added and rebased PR to resolve merge conflicts after latest changes.

vignesh2027:
Nice to see the OpenAI side finally get the histogram, and the transport error test is a good addition.

Two things from reading the Anthropic path.

The metric and the span now disagree on `error.type` for an HTTP error. The metric gets `strconv.Itoa(resp.StatusCode)` so `"429"`, while the span on line 152 still sets `attribute.String("error.type", resp.Status)` so `"429 Too Many Requests"`. Your own test pins `"429"` for the metric, which reads like the right call, so the span looks like the one that should move. Worth doing here rather than leaving the same attribute meaning two things in one operation.

The other is smaller, more of a note than a problem. `attrs := append(baseAttrs, errorAttrs...)` shares a base with the existing `spanAttrs = append(baseAttrs, spanAttrs...)` on line 118. It is fine as things stand, because `baseAttrs` is a composite literal so `len == cap` and both appends reallocate. It only bites if someone later gives `baseAttrs` spare capacity, at which point the deferred append would write over what `spanAttrs` is holding. A `slices.Concat`, or an explicit `make` with the right capacity, would close that off for nothing.

Labreo:
Thanks for catching both of those!

Updated the Anthropic HTTP error path so the span sets error.type to the numeric status code like "429", matching both the metric and how OpenAI handles it, and updated the test assertion as well. Also switched the deferred appends over to slices.Concat across Anthropic and OpenAI to avoid any capacity sharing down the line. Regenerated and verified the bundle too.

vignesh2027:
Had a look at the two red checks on `d313b6c`, since they are unrelated to each other.

**Verify Bundle** is the archive rather than anything you wrote. The job prints `Binary files a/tool/data/otelc-bundle.tgz and b/tool/data/otelc-bundle.tgz differ`, so `make package` and committing the updated archive should clear it. Worth knowing that `./.github/scripts/verify-bundle.sh` is exactly what CI runs, so it answers this locally before a push.

**Unit Tests** is a nil pointer panic rather than a failed assertion:

```
--- FAIL: TestOtelMiddleware_RecordsDurationOnHTTPError
[signal SIGSEGV: segmentation violation]
openai-go/v2.OtelMiddleware.func2
  instrumentation/github.com/openai/openai-go/v2/middleware.go:167
```

Line 167 is `tracer.Start(...)`, and `tracer` is nil there. `setupTestMeter` wires `operationDuration` to a manual reader but nothing sets the tracer, and the tests call `OtelMiddleware()` directly rather than going through `BeforeNewClient`, which is the only path that calls `initInstrumentation()`.

Two things that make it look smaller than it is. All three `TestOtelMiddleware_*` tests in `openai-go/v2` take the same path with a valid model, so all three reach line 167. The suite runs with `-test.shuffle`, and a panic takes the binary down, so whichever runs first is the only one reported. That also means the failing test name will move between runs.

The same gap is in v1 and v3, neither of which has any tracer setup in its test file. The anthropic side of this PR already does the right thing, `setupTestTracer` at `middleware_test.go:95` ending with `tracer = tp.Tracer("test")`, so mirroring that into the three OpenAI test files looks like the whole fix.

Labreo:
Done the required changes.Thanks @vignesh2027 :).Appreciate the help.

vignesh2027:
Verify Bundle is still red on `5d75b1a`, and it does look like the archive rather than anything else. I checked out that commit and ran `make package`, and the committed archive comes back different, having first confirmed on clean `main` that it reproduces there byte for byte, so the comparison means something.

The usual cause is ordering rather than the command. If the archive gets regenerated and then any file under `pkg/` or `instrumentation/` changes afterwards, the committed copy goes stale again, and the `slices.Concat` edits would have done exactly that. So it wants to be the last thing before the commit:

```
# after every code change is final
make package
git add tool/data/otelc-bundle.tgz
./.github/scripts/verify-bundle.sh   # same script CI runs, exits 0 when it matches
```

Your branch is level with main, so there is nothing to rebase first despite what the job message suggests.

vignesh2027:
Found why this keeps coming back, and it is not something you could have guessed from the error. Your archive has a `.DS_Store` baked into it.

Checked out `57b5d78`, regenerated, and compared the contents against your committed archive:

```
entries in yours: 299
entries in fresh: 298

only in yours:
  instrumentation_temp/.DS_Store
```

The bundler walks the filesystem rather than the git index, and its only filter is `.log`:

```go
func shouldExclude(name string) bool {
	return strings.HasSuffix(name, ".log")
}
```

`.DS_Store` is in `.gitignore`, so git never shows it and it never appears in a diff, but the packager picks it up all the same. Finder creates one as soon as you browse a directory. CI runs on Linux and never has one, so the archive it builds will never match yours, and re-running `make package` cannot fix it while the file is still on disk.

So the loop breaks with:

```
find instrumentation pkg -name .DS_Store -delete
make package
git add tool/data/otelc-bundle.tgz
./.github/scripts/verify-bundle.sh
```

Reproduced it from clean main to be sure: 298 entries normally, 299 after dropping a single `.DS_Store` into an instrumentation directory.

Sorry for pointing you at ordering earlier, that was the usual cause but not yours. Raising the packaging side separately, since any macOS contributor will hit this.

Labreo:
I am well aware of this @vignesh2027 .I am pretty sure this is well known issue and has been reported and discussed in #1027 .It should be fixed when the bundles are manaully regenerated.

Labreo:
Rebased the branch on latest main to clear the bundle conflict from the recent merges, cleaned out any local .DS_Store files, and regenerated the archive. Verified that ./.github/scripts/verify-bundle.sh passes locally and pushed the update.

Labreo:
Rebased and fixed CI/CD issues as well as merge conflicts.

Labreo:
Thanks for digging into that root cause, [@vignesh2027](https://github.com/vignesh2027).That .DS_Store bundler behavior explains the issues that were coming up with my PR's.It should be fine now.

Labreo:
Fixed and rebased this PR to reflect latest changes to the repo.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -169,12 +171,25 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
...
+		// Record the operation duration on every exit path below.
+		// errorAttrs carries error.type on error paths and is empty on success.
+		var errorAttrs []attribute.KeyValue
+		defer func() {
+			if operationDuration != nil {
+				attrs := slices.Concat(baseAttrs, errorAttrs)
+				operationDuration.Record(ctx, time.Since(start).Seconds(),
+					metric.WithAttributes(attrs...))
+			}
+		}()
```
darccio, line 189: For a streaming response, this defer fires when OtelMiddleware returns —
right after headers arrive — while the span stays open until
newStreamingReader finishes draining the body (see streaming_bridge.go).
So for the same call, gen_ai.client.operation.duration measures
time-to-first-byte while the span duration measures the full stream
lifetime; the two diverge for every streaming request.

Anthropic sidesteps this by not instrumenting streaming calls at all, so
it has no equivalent metric for them. Worth deciding which behavior is
intended here: record the duration metric when the stream actually
finishes (mirroring the span), or document that this histogram
intentionally captures time-to-headers for streaming operations.

This applies too for `v2` and `v3`.

instrumentation/github.com/openai/openai-go/middleware_test.go
```diff
@@ -4,12 +4,153 @@
...
+
+	_, err := middleware(req, next)
+	require.ErrorIs(t, err, wantErr)
+
+	dps := durationDataPoints(t, reader)
+	require.Len(t, dps, 1, "duration should be recorded once on transport error")
+	assert.Equal(t, uint64(1), dps[0].Count)
+	_, ok := dps[0].Attributes.Value(attribute.Key("error.type"))
+	require.True(t, ok, "error.type must be present on transport error")
+}
```
darccio, line 152: None of the new duration tests cover the streaming path. Given the
time-to-headers vs full-stream-duration divergence between the metric
and the span for streaming responses, a test asserting what
gen_ai.client.operation.duration actually records for a streaming call
(and when) would pin down the intended behavior and catch regressions
either way it gets resolved.

Applies to v2 and v3 too.

darccio (changes requested):

Labreo:
Thanks for the review, @darccio!

I updated the streaming flow so the duration metric waits until the stream finishes reading. I also added streaming tests for all three versions, rebased on `main`, and verified the bundle.

darccio (approved):
LGTM

PR #1076, opened 2026-08-12
feat(instrumentation/kafka-go): instrument (*kafka.Reader).FetchMessage

### LINKED ISSUE #1075: New Instrumentation Request for github.com/segmentio/kafka-go (*kafka.Reader).FetchMessage
**Package**

`github.com/segmentio/kafka-go`

**Version**

v0.4.x and later

**Environment**

Any Go application using compile-time instrumentation with `otelc`, deployed as a service consuming Kafka messages with manual offset management (i.e., calling `FetchMessage` + `CommitMessages` instead of `ReadMessage`).

**Use-case**

Applications that need fine-grained control over offset commits call `(*kafka.Reader).FetchMessage(ctx)` to receive a message without automatically committing the offset, then process it and call `CommitMessages` only on success. This is a common pattern for at-least-once processing guarantees.

Currently, the instrumentation in `instrumentation/github.com/segmentio/kafka-go/consumer/otelc.yaml` only hooks `ReadMessage`. When an application uses `FetchMessage` instead:

- No consumer `receive` spans are emitted.
- Trace context injected by the producer into message headers is never extracted, breaking distributed traces across the producer and consumer.

**Telemetry**

`FetchMessage` has the same signature as `ReadMessage` — `func (r *Reader) FetchMessage(ctx context.Context) (Message, error)` — so the same `BeforeReadMessage` / `AfterReadMessage` hooks can be reused. The expected telemetry is:

- A `<topic> receive` span with `SpanKindConsumer` on every `FetchMessage` call.
- Trace context extracted from message headers and linked to the producer span.
- Attributes matching the semantic conventions already applied for `ReadMessage`: `messaging.system`, `messaging.destination.name`, `messaging.operation`, `messaging.kafka.consumer.group`, `messaging.kafka.message.key`, `messaging.kafka.message.offset`, `messaging.kafka.destination.partition`.

**Anything else?**

The fix only requires adding a second rule entry in `otelc.yaml`:

```yaml
kafka_reader_fetchmessage:
  target: github.com/segmentio/kafka-go
  where:
    func: FetchMessage
    recv: "*Reader"
  do:
    - inject_hooks:
        before: BeforeReadMessage
        after: AfterReadMessage
        path: "go.opentelemetry.io/otelc/instrumentation/github.com/segmentio/kafka-go/consumer"
```

No changes to `consumer_hook.go` are needed since the function signatures are identical. A unit test case in `consumer_hook_test.go` and an integration test should accompany the change.

bhuvan-somisetty:
## Description

Add auto-instrumentation for `(*kafka.Reader).FetchMessage`, the manual-commit API of `github.com/segmentio/kafka-go`.

- **`otelc.yaml`**: Add `kafka_reader_fetchmessage` rule targeting `(*Reader).FetchMessage` with `BeforeFetchMessage`/`AfterFetchMessage` hooks.
- **`consumer_hook.go`**: Add `BeforeFetchMessage` and `AfterFetchMessage`, which delegate to the existing `BeforeReadMessage`/`AfterReadMessage` implementations since both methods share identical parameter and return types. Update `ExtractContext` doc to mention `FetchMessage`.
- **`consumer_hook_test.go`**: Add three tests — trace context propagation, error recording, and disabled-instrumentation — mirroring the `ReadMessage` test suite.

## Motivation

`FetchMessage` is widely used by applications that need fine-grained offset control (at-least-once processing): call `FetchMessage`, process the message, then call `CommitMessages` only on success. Before this change, those code paths emitted no consumer spans and trace context propagated by producers into Kafka message headers was silently dropped, breaking distributed traces across producer and consumer processes.

Fixes #1075

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


instrumentation/github.com/segmentio/kafka-go/consumer/consumer_hook.go
```diff
@@ -139,13 +139,31 @@ func AfterReadMessage(ictx hook.HookContext, msg kafka.Message, err error) {
...
+}
+
+// AfterFetchMessage creates a consumer span for a message received via
+// FetchMessage. It delegates to AfterReadMessage because the two methods share
+// the same parameter layout and span semantics.
+//
+// The Enable() check is intentionally omitted — see AfterReadMessage.
+func AfterFetchMessage(ictx hook.HookContext, msg kafka.Message, err error) {
+	AfterReadMessage(ictx, msg, err)
+}
```
amazingakai, line 185: Not much knowledgable about kafka but it seems we'll double instrument `ReadMessage` (since it internally calls `FetchMessage`) now? https://github.com/segmentio/kafka-go/blob/main/reader.go#L793

bhuvan-somisetty: Good catch, thanks! You're right, ReadMessage calls FetchMessage internally so this was double-instrumenting every ReadMessage call.

Fixed it by having BeforeReadMessage mark its ctx and write it back via SetParam, so that mark flows into ReadMessage's own internal r.FetchMessage(ctx) call. BeforeFetchMessage checks for that marker and skips creating span data when it's present, so the nested call no longer produces its own span — only the outer ReadMessage span is created. Direct FetchMessage calls (not going through ReadMessage) are unaffected since their ctx never carries the marker.

Added regression tests covering both the nested case and a direct FetchMessage call to make sure they behave correctly. Pushed the fix.

amazingakai (approved):
LGTM

amazingakai:
The bundle needs to be regenerated though.

bhuvan-somisetty:
Good catch, regenerated it in b1a6acf. Thanks for the review!

PR #1078, opened 2026-08-12
fix(tool/internal/ast): reject empty keys in scanArgs directive argument parsing

### LINKED ISSUE #1022: bug(tool/internal/ast): scanArgs splits directive args on quote-unaware colons and accepts empty keys
### Describe the bug

`scanArgs` in `tool/internal/ast/directive.go` splits each token with
`strings.Cut(tok, ":")`, but `tokenize` deliberately keeps the quote characters in
the token. The split is therefore quote-unaware, and the resulting key is never
validated, so malformed directives parse successfully into malformed args.

Verified against `ParseDirectiveArgs` on main:

| input after `//otelc:span ` | parsed | error |
|---|---|---|
| `"key:value"` | `{Key: "key, Value: value"}` | none |
| `:v` | `{Key: "", Value: v}` | none |
| `"k":v` | `{Key: "k", Value: v}` | none |

For contrast `k:"a:b"` parses correctly to `{Key: k, Value: a:b}`, because there
the first colon falls outside the quotes. That is why this is easy to miss.

### Why it matters

Quoting the whole argument (`//otelc:span "span.name:my op"`) is a natural
mistake, and it currently produces an attribute key of `"span.name` rather than
an error. The empty-key case is worse: `:v` yields an arg with no key at all.
Both flow downstream instead of being rejected at parse time, and
`scanArgs` already rejects other malformed input, so the missing colon case
errors while these do not.

### Suggested fix

Reject a token whose colon falls inside a quoted region, and reject an empty key,
in the same place the existing "missing colon separator" and "single-quoted
values are not supported" checks live.

### Notes

I checked for existing reports and found none; the code came in with #375 and
the validation was not discussed there. Happy to fix this, but per the intake
discussion in #973 I will wait to be assigned before opening a PR.

SatyamPandey-07:
Fixes #1022

### Description
`scanArgs` used `strings.Cut(tok, ":")` to split `key:value` tokens without
validating that the key segment is non-empty. A token like `:value` silently
produced `{Key:"", Value:"value"}` instead of returning an error.

This PR adds an explicit check after the cut: if `key == ""`, return an error.

### Tests Added
- `empty key rejected` — `:value` now errors
- `quoted value containing colon` — `url:"https://example.com/path"` parses correctly  
- `bare value with multiple colons` — `key:a:b:c` produces `value="a:b:c"` (first colon only)
- `multiple args with quoted colon value` — combination case works correctly

### Checklist
- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/ast/directive.go
+++ b/tool/internal/ast/directive.go
@@ -81,9 +81,16 @@ func scanArgs(input string) ([]DirectiveArg, error) {
 	}
 	args := make([]DirectiveArg, 0, len(tokens))
 	for _, tok := range tokens {
-		key, value, found := strings.Cut(tok, ":")
+		key, value, found := cutUnquoted(tok, ':')
 		if !found {
-			return nil, ex.Newf("argument %q missing colon separator", tok)
+			return nil, ex.Newf("argument %q has no unquoted colon separator; "+
+				"quote only the value, as in key:\"a:b\"", tok)
+		}
+		if key == "" {
+			return nil, ex.Newf("argument %q has an empty key", tok)
+		}
+		if strings.ContainsRune(key, '"') {
+			return nil, ex.Newf("argument %q has a quoted or malformed key", tok)
 		}
 		if strings.HasPrefix(value, "'") {
 			return nil, ex.Newf("single-quoted values are not supported in argument %q", tok)
@@ -213,4 +220,29 @@ func tokenize(input string) ([]string, error) {
 		tokens = append(tokens, current.String())
 	}
 	return tokens, nil
+}
+
+// cutUnquoted splits tok at the first occurrence of sep that falls outside
+// a double-quoted region, mirroring tokenize's quote/escape tracking so a
+// separator inside a quoted key or value is never mistaken for the
+// key:value boundary. found is false if no such separator exists.
+//
+//nolint:revive // confusing-results conflicts with nonamedreturns; see doc comment above for result order
+func cutUnquoted(tok string, sep rune) (string, string, bool) {
+	inQuote := false
+	escaped := false
+
+	for i, ch := range tok {
+		switch {
+		case escaped:
+			escaped = false
+		case ch == '\\' && inQuote:
+			escaped = true
+		case ch == '"':
+			inQuote = !inQuote
+		case ch == sep && !inQuote:
+			return tok[:i], tok[i+utf8.RuneLen(sep):], true
+		}
+	}
+	return tok, "", false
 }
--- a/tool/internal/ast/directive.go
+++ b/tool/internal/ast/directive.go
@@ -90,6 +90,9 @@ func scanArgs(input string) ([]DirectiveArg, error) {
 		if !found {
 			return nil, ex.Newf("argument %q missing colon separator", tok)
 		}
+		if key == "" {
+			return nil, ex.Newf("argument %q has an empty key", tok)
+		}
 		if strings.HasPrefix(value, "'") {
 			return nil, ex.Newf("single-quoted values are not supported in argument %q", tok)
 		}
--- a/tool/internal/ast/directive.go
+++ b/tool/internal/ast/directive.go
@@ -86,12 +86,15 @@ func scanArgs(input string) ([]DirectiveArg, error) {
 	}
 	args := make([]DirectiveArg, 0, len(tokens))
 	for _, tok := range tokens {
-		key, value, found := strings.Cut(tok, ":")
+		key, value, found := cutUnquoted(tok, ':')
 		if !found {
 			return nil, ex.Newf("argument %q missing colon separator", tok)
 		}
 		if key == "" {
 			return nil, ex.Newf("argument %q has an empty key", tok)
+		}
+		if strings.ContainsRune(key, '"') {
+			return nil, ex.Newf("argument %q has a quoted or malformed key", tok)
 		}
 		if strings.HasPrefix(value, "'") {
 			return nil, ex.Newf("single-quoted values are not supported in argument %q", tok)
@@ -207,4 +210,29 @@ func tokenize(input string) ([]string, error) {
 		tokens = append(tokens, current.String())
 	}
 	return tokens, nil
+}
+
+// cutUnquoted splits tok at the first occurrence of sep that falls outside
+// a double-quoted region, mirroring tokenize's quote/escape tracking so a
+// separator inside a quoted key or value is never mistaken for the
+// key:value boundary. found is false if no such separator exists.
+//
+//nolint:revive // confusing-results conflicts with nonamedreturns; see doc comment above for result order
+func cutUnquoted(tok string, sep rune) (string, string, bool) {
+	inQuote := false
+	escaped := false
+
+	for i, ch := range tok {
+		switch {
+		case escaped:
+			escaped = false
+		case ch == '\\' && inQuote:
+			escaped = true
+		case ch == '"':
+			inQuote = !inQuote
+		case ch == sep && !inQuote:
+			return tok[:i], tok[i+utf8.RuneLen(sep):], true
+		}
+	}
+	return tok, "", false
 }
--- a/tool/internal/ast/directive_test.go
+++ b/tool/internal/ast/directive_test.go
@@ -90,10 +90,10 @@ func TestMatchDirective(t *testing.T) {
 
 func TestScanArgs(t *testing.T) {
 	tests := []struct {
-		name     string
-		input    string
-		expected []DirectiveArg
-		hasError bool
+		name        string
+		input       string
+		expected    []DirectiveArg
+		errContains string // non-empty pins the error to the expected branch
 	}{
 		{
 			name:     "simple key:value",
@@ -116,14 +116,14 @@ func TestScanArgs(t *testing.T) {
 			},
 		},
 		{
-			name:     "single quotes rejected",
-			input:    "key:'single'",
-			hasError: true,
+			name:        "single quotes rejected",
+			input:       "key:'single'",
+			errContains: "single-quoted values are not supported",
 		},
 		{
-			name:     "unclosed quote",
-			input:    `key:"unclosed`,
-			hasError: true,
+			name:        "unclosed quote",
+			input:       `key:"unclosed`,
+			errContains: "unclosed double quote",
 		},
 		{
 			name:     "empty input",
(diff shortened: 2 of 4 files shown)


SatyamPandey-07:
@darccio could you take a look when you have a chance?

Fixes #1022 — scanArgs was using strings.Cut(tok, ":") to split key:value tokens without checking that the key segment is non-empty, so a token like :value silently produced {Key:"", Value:"value"} instead of erroring. Added an explicit key == "" check plus tests for the empty-key case, quoted values containing colons, bare values with multiple colons, and a combined case.

Pushed a follow-up commit (53b7d0d6) fixing a gofmt alignment mistake in the test file that was failing the Lint Go Code check — that job should be green now. Ready for review/merge whenever convenient.

SatyamPandey-07:
@darccio Rebased cleanly onto latest main (df034c1b) so the branch is no longer out-of-date — no conflicts, both commits carried through as-is. Re-verified locally: go build, go vet, gofmt, and the full tool/... test suite all pass. Should just be waiting on your review now — thanks for your time!

dobbydobap:
Thanks for picking this up. The empty-key check is correct and the :value test
gates it properly.

One thing before this closes #1022: the issue listed three cases and this covers
one. I ran the other two against this branch:

    "key:value"  ->  {Key: "key, Value: value}   no error
    "k":v        ->  {Key: "k", Value: v}        no error

Both still produce a malformed key. The colon split is still quote-unaware,
which was the main part of the report, so merging with "Fixes #1022" would close
the issue with that still live.

Also worth knowing: I ran the three new quoted tests against main without your
change and all three already pass there, so they document existing behaviour
rather than gate the fix. Only "empty key rejected" fails on main.

Simplest options are to widen this to reject a token whose colon falls inside
quotes, or change Fixes to Refs so #1022 stays open for the rest. Either is fine
by me, I just did not want the quote case to get lost.

SatyamPandey-07:
@dobbydobap @darccio Good catch, thanks for actually testing those — you're right that the empty-key check only handled one symptom and the split itself was still quote-unaware.

Pushed 2ef86c48 with the real fix: colon-splitting now uses a cutUnquoted helper that mirrors tokenize's existing quote/escape tracking, so a colon inside a quoted region is never mistaken for the key:value separator. "key:value" now correctly errors as "missing colon separator" instead of splitting mid-quote. For "k":v, the split now lands in the right place (after the quoted region), but since keys were never meant to be quoted per this function's doc comment, I added an explicit check rejecting a key that still contains " — so that one errors too now.

Added both your cases as regression tests (confirmed they fail without the fix). Fixes #1022 should now be accurate — happy to switch it to Refs instead if you'd rather keep the issue open for anything else you want to verify first.

dobbydobap:
Verified, this covers it. All three now error:

    :v            -> has an empty key
    "key:value"   -> missing colon separator
    "k":v         -> has a quoted or malformed key

I also ran ten cases that were already correct to check for regressions and they
parse identically: k:"a:b", url:"https://example.com/path", key:a:b:c,
op:"http:post" tag:foo, the escaped k:"a\"b", empty k:"" and k:, and
k:"unclosed still errors as unclosed double quote.

cutUnquoted reading as a mirror of tokenize's own quote and escape tracking is
the right call, and RuneLen keeps the offset honest.

Fixes #1022 is accurate now, no need to switch it to Refs from my side.

tool/internal/ast/directive.go
```diff
@@ -86,10 +86,16 @@ func scanArgs(input string) ([]DirectiveArg, error) {
 	}
 	args := make([]DirectiveArg, 0, len(tokens))
 	for _, tok := range tokens {
-		key, value, found := strings.Cut(tok, ":")
+		key, value, found := cutUnquoted(tok, ':')
 		if !found {
 			return nil, ex.Newf("argument %q missing colon separator", tok)
 		}
```
darccio, line 88: The quote-aware split is the right fix, but `"key:value"` now fails with `argument
"\"key:value\"" missing colon separator` — and that token visibly *has* a colon. It's
accurate from the parser's point of view (there is no *unquoted* colon), yet it points
the author away from the actual mistake.

This is the exact case #1022 flagged as "a natural mistake" — quoting the whole
argument instead of just the value — so it's worth spending a few words to name the
real cause and the fix:

```suggestion
		if !found {
			return nil, ex.Newf("argument %q has no unquoted colon separator; "+
				"quote only the value, as in key:\"a:b\"", tok)
		}
```

tool/internal/ast/directive_test.go
```diff
@@ -150,6 +150,48 @@ func TestScanArgs(t *testing.T) {
...
+		{
+			name:     "empty key rejected",
+			input:    ":value",
+			hasError: true,
+		},
+		{
+			name:     "colon inside a fully quoted token is not a separator",
+			input:    `"key:value"`,
+			hasError: true,
+		},
```
darccio, line 167: This case passes through the `!found` / "missing colon separator" branch, not through
any quote-aware rejection — so the name describes a mechanism the assertion never
confirms. Renaming it to match what actually fires:

```suggestion
		{
			name:     "fully quoted token has no unquoted colon separator",
			input:    `"key:value"`,
			hasError: true,
		},
```

More generally: there are now three distinct rejection paths (`missing colon
separator`, `has an empty key`, `has a quoted or malformed key`) and every negative
case asserts only `require.Error(t, err)`. A case can pass for the wrong reason, and a
future refactor could reroute one of these without any test noticing.

Consider adding an `errContains string` field to the table struct and switching the
runner to `require.ErrorContains(t, err, tt.errContains)` when it's set, so each
negative case is pinned to the branch it means to exercise.

tool/internal/ast/directive_test.go
```diff
@@ -150,6 +150,48 @@ func TestScanArgs(t *testing.T) {
 				{Key: "key", Value: ""},
 			},
 		},
+		{
+			name:     "empty key rejected",
+			input:    ":value",
+			hasError: true,
+		},
```
darccio, line 162: Two corners of the new split are unpinned. Neither is a regression — both behave the
same as the old `strings.Cut` — but both run through `cutUnquoted` now, so they're
worth locking down while the table is being extended:

- a bare `:` (degenerate empty-key input, distinct from `:value`)
- `k\:v`, which yields `{Key: "k\\", Value: "v"}` — a backslash outside a quoted region
  isn't an escape in either `tokenize` or `cutUnquoted`, so the key keeps a literal
  backslash

```suggestion
		{
			name:     "empty key rejected",
			input:    ":value",
			hasError: true,
		},
		{
			name:     "bare colon rejected",
			input:    ":",
			hasError: true,
		},
		{
			name:     "backslash in a bare key is not an escape",
			input:    `k\:v`,
			expected: []DirectiveArg{
				{Key: `k\`, Value: "v"},
			},
		},
```

The backslash case is a characterization test — it records today's behavior rather than
endorsing it. If a literal backslash in an attribute key is considered wrong, that's a
separate issue, and having it pinned here makes the change visible when someone fixes it.

tool/internal/ast/directive.go
```diff
@@ -205,3 +211,28 @@ func tokenize(input string) ([]string, error) {
...
 	return tokens, nil
 }
+
+// cutUnquoted splits tok at the first occurrence of sep that falls outside
+// a double-quoted region, mirroring tokenize's quote/escape tracking so a
+// separator inside a quoted key or value is never mistaken for the
+// key:value boundary. found is false if no such separator exists.
+//
+//nolint:revive // confusing-results conflicts with nonamedreturns; see doc comment above for result order
+func cutUnquoted(tok string, sep rune) (string, string, bool) {
```
darccio, line 231: `cutUnquoted` adds a second quote/escape state machine alongside `tokenize`'s, and the
two have to agree for the parse to be coherent. `docs/testing.md` recommends a fuzz
target exactly here — "components that parse untrusted or free-form input, where the
surface is too large to enumerate by hand" — with `FuzzParseDSN` as the precedent.

A `FuzzScanArgs` asserting the never-panics invariant, seeded with the table inputs
from `TestScanArgs`, would cover the interaction between the two state machines more
cheaply than enumerating cases. It could also assert the cross-component invariant
that matters: for every token `tokenize` emits, if `scanArgs` accepts it then the key
is non-empty and quote-free.

darccio (changes requested):

dobbydobap:
On the FuzzScanArgs suggestion, I built one while investigating #1022, so here
is what it did in case it saves you the setup.

Seeded with the TestScanArgs table and asserting two invariants, never panics
and a parsed key never contains a quote, the quote invariant fails on
"key:value" from the seed corpus alone, before fuzzing explores anything new. So
on current main it reproduces the bug immediately, and on this branch it becomes
a regression gate for the same case darccio is asking you to name in the error
message.

The invariant worth the most is the cross-component one: every token tokenize
produces should either split into a non-empty unquoted key or be rejected. That
is what pins cutUnquoted and tokenize to the same view of quoting, which is the
part that can drift later since they are two separate state machines.

Happy to hand the harness over if it is useful, or leave it with you.

SatyamPandey-07:
@darccio Done in a7884ff3 — addressed all four points from your review:

Error message (directive.go): !found now says "has no unquoted colon separator; quote only the value, as in key:\"a:b\"" so the author knows exactly what the problem is and how to fix it.

Test naming + errContains (directive_test.go): Replaced hasError bool with errContains string across the whole table so every negative case is pinned to the branch it exercises via require.ErrorContains. The fully-quoted token case is now named "fully quoted token has no unquoted colon separator" to match what actually fires.

Unpinned corner cases: Added "bare colon rejected" (":" → "has an empty key") and "backslash in a bare key is not an escape" (k\:v → {Key: "k\", Value: "v"}) as a characterization test with a comment explaining its purpose.

FuzzScanArgs (directive_fuzz_test.go): Added the fuzz target seeded from the full TestScanArgs table. Asserts two invariants — never panics, and every successfully parsed key is non-empty and quote-free — which pins cutUnquoted and tokenize to the same view of quoting. All 23 seeds pass in the ordinary go test run.

SatyamPandey-07:
Thanks for the detailed verification and for sharing the harness — it's exactly what surfaced the gap. The fuzz target is now in directive_fuzz_test.go with the cross-component invariant you described: every token tokenize emits must either split into a non-empty, quote-free key or be rejected by scanArgs. Pushed in a7884ff3. @dobbydobap

darccio (approved):
LGTM

PR #1081, opened 2026-08-12
fix(genai): set server.address and server.port on GenAI spans

### LINKED ISSUE #1024: GenAI spans miss server.address and server.port when HTTP instrumentation is suppressed

Anshumancanrock:
## Summary

Fixes #1024.

OpenAI and Anthropic instrumentation creates a GenAI span and suppresses the nested HTTP client span to avoid duplicates. Endpoint attrs (`server.address` / `server.port`) usually come from that HTTP span, so GenAI spans were missing them.

This PR keeps HTTP suppression and sets those attrs on the GenAI span instead:
- `server.address` from the request URL hostname
- `server.port` from the URL, including default `80`/`443` when omitted (required by GenAI SemConv when address is set)

Covers OpenAI v1/v2/v3 and Anthropic. URL parsing is shared in `pkg/netutil`.


pkg/netutil/http_test.go
```diff
@@ -0,0 +1,82 @@
...
+		},
+		{
+			name:        "unsupported scheme",
+			rawURL:      "custom://example.com/v1/messages",
+			wantAddress: "example.com",
+		},
+		{
+			name:   "relative URL",
+			rawURL: "/v1/messages",
+		},
```
darccio, line 56: Codecov flags 1 missing + 1 partial line in this file's patch coverage — the `err != nil || port == 0` branch in `HTTPServerEndpoint` (http.go:32) is never exercised: no case here has a zero or out-of-range port. Add two cases to close it.

```suggestion
		{
			name:   "relative URL",
			rawURL: "/v1/messages",
		},
		{
			name:        "explicit zero port",
			rawURL:      "https://api.openai.com:0/v1/chat/completions",
			wantAddress: "api.openai.com",
		},
		{
			name:        "port out of range",
			rawURL:      "https://api.openai.com:99999/v1/chat/completions",
			wantAddress: "api.openai.com",
		},
```

instrumentation/github.com/openai/openai-go/middleware_integration_test.go
darccio, line 449: Every case here uses `http://`, so `deepseek`/`azure`/`groq` only ever exercise the default-80 path, not the https/443 path that's how these providers are actually reached. `TestOtelMiddleware_ChatCompletion` above already covers https/443 for `api.openai.com`, but switching this table to https closes the gap for the provider-detection cases specifically, since a scheme bug here would otherwise slip through unnoticed for every provider except OpenAI itself.

Identical file content in `v2/` and `v3/` needs the same change.

```suggestion
		{"deepseek", "api.deepseek.com", "deepseek", "api.deepseek.com", 443},
		{"azure", "myendpoint.azure.com", "azure", "myendpoint.azure.com", 443},
		{"local", "localhost:11434", "local", "localhost", 11434},
		{"groq", "api.groq.com", "groq", "api.groq.com", 443},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			sr := setupTestTracer(t)
			middleware := OtelMiddleware()

			reqBody := `{"model":"test-model"}`
			req, _ := http.NewRequest(
				"POST",
				"https://"+tt.host+"/v1/chat/completions",
```

darccio (changes requested):

PR #1083, opened 2026-08-12
fix(instrumentation): map openai max_completion_tokens to gen_ai.request.max_tokens

### LINKED ISSUE #1007: bug(instrumentation/openai): max_completion_tokens not mapped to gen_ai.request.max_tokens

shivv23:
## Description

The OpenAI `chat/completions` request parameter `max_completion_tokens` was not mapped to the GenAI semantic convention attribute `gen_ai.request.max_tokens`, so spans for requests using the newer parameter (`gpt-4.1`, `gpt-5`, `o-series`, etc.) were missing the attribute entirely. The instrumentation only read the legacy `max_tokens` field.

Each version of the OpenAI middleware (v1, v2, v3) now reads both fields and maps whichever is set:

```go
if req.MaxCompletionTokens != nil {
    attrs = append(attrs, semconv.GenAIRequestMaxTokens(*req.MaxCompletionTokens))
} else if req.MaxTokens != nil {
    attrs = append(attrs, semconv.GenAIRequestMaxTokens(*req.MaxTokens))
}
```

`max_completion_tokens` takes precedence over `max_tokens` because OpenAI marks it as the preferred parameter for the newer models; `max_tokens` remains as a fallback for models that only support the legacy field.

## Motivation

`gen_ai.request.max_tokens` is part of the GenAI semantic conventions. Requests sent with `max_completion_tokens` (the parameter required by OpenAI's newer models) previously produced spans with no token-limit attribute, making it impossible to correlate the configured completion budget with telemetry.

Adds `TestParseChatRequest_MaxCompletionTokens` (unit) and `TestOtelMiddleware_ChatCompletion_MaxCompletionTokens` (integration) in each module asserting the mapping, including precedence when both fields are present.

Fixes #1007

---

## Checklist

- [x] Tests pass: `make test`


Labreo (commented):
This is a duplicate issue as @Anshumancanrock has an open PR https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1008 for this issue.

darccio (approved):

darccio:
@shivv23 Please regenerate the bundle.

PR #1088, opened 2026-08-12
feat(anthropic): instrument messages count_tokens endpoint

krsatyamthakur-droid:
Instruments the Anthropic `POST /v1/messages/count_tokens` endpoint, which was previously passed through uninstrumented. It now emits a GenAI client span carrying the request model and `gen_ai.usage.input_tokens` from the response.

One open question for maintainers: the gen_ai semantic conventions have no standard `gen_ai.operation.name` value for token counting, so this uses `"count_tokens"` as a placeholder. Happy to change it to whatever value (or omit the attribute entirely) you'd prefer.


krsatyamthakur-droid:
Closing — this was opened against the wrong repo; the work continues in my personal fork only.

Issue #728, opened 2026-07-21
[DRAFT] Generate Go semconv instrument registration from the Weaver registry


txabman42:
> [!NOTE]
> This issue is a proposal still in draft

## Problem

schemas/otelc/ declares the telemetry each instrumentation emits, but nothing verifies the declaration matches the Go code. The three CI guarantees today are:

1. .semconv-version, registry_manifest.yaml, and Go import paths agree on the version string.
2. weaver registry check proves the YAML resolves against upstream.
3. Manual co-maintenance, via source-of-truth comments in the YAML.

None of these read the Go code.

The same contract is written out by hand in three places:

1. **Emission.** `instrumentation/net/http/semconv/client.go` registers six instruments with hardcoded names, units, descriptions, and bucket boundaries.
2. **Declaration.** `schemas/otelc/groups/http.yaml` restates them.
3. **Verification.** `test/testutil/semconv.go` hardcodes attribute names as bare strings: `"gen_ai.usage.total_tokens"`, `"k8s.object.kind"`, `"k8s.object.api_version"`.

The three are joined only by a comment. Concretely: remove `http.client.connection.duration` from the YAML, or add a new metric in Go without declaring it, and CI stays green. With 8 instrumentation files in the registry, drift is a matter of time.

## Proposal

Make the registry the thing the code is derived from or checked against, never a parallel description of it. Three tiers, because the three signal types are at different stages of upstream support.

### Tier 1: Metrics. Delete ours, adopt upstream's generated types.

Upstream already generates typed Go metric packages from this same registry, and `httpconv` contains exactly the 10 types `http.yaml` declares. The generated API makes drift structural rather than a matter of care: required attributes are positional and typed, so omitting one does not compile, and optional attributes exist only as `Attr*` methods on that specific metric type, so `AttrClientAddress` is not reachable from `ClientRequestDuration` at all. Names, units, descriptions and attribute sets all regenerate on every semconv release, which means a semconv upgrade surfaces as a compile failure at the exact call sites that need attention.

gRPC already works this way. `net/http` is the only instrumentation that hand-writes registration, and it is the one that drifted.

[...]

PR #1090, opened 2026-08-13
fix(instrumentation/openai): record cached prompt tokens

### LINKED ISSUE #1089: bug(instrumentation/openai): cached prompt tokens are never recorded
The openai-go instrumentation never records cached prompt tokens, while the anthropic-sdk-go instrumentation does. Credit to @dobbydobap, who spotted this in the #709 discussion.

## The gap

OpenAI has automatic prompt caching and reports it on every chat/completion response as `usage.prompt_tokens_details.cached_tokens` (exposed by openai-go v1.12.0 as `CompletionUsagePromptTokensDetails.CachedTokens`). There are no cache references anywhere under `instrumentation/github.com/openai/`, and no cache key in any of the three OpenAI semconv packages, so that portion of prompt spend is invisible for OpenAI users but visible for Anthropic ones.

## It is also in the registry, not just the code

`schemas/otelc/groups/anthropic.yaml` declares `gen_ai.usage.cache_read.input_tokens` with `id:` inside `registry.otelc.gen_ai.anthropic`, the provider-specific group, and the header comment states:

> The two prompt-cache attributes are Anthropic-specific and are declared locally with `id:` in the attribute group below.

Cache reads are not Anthropic-specific. Because the attribute is defined in a provider-scoped group, `openai.yaml` cannot declare it without referencing another provider's group, so the wrong assumption is encoded in the contract as well as in the middleware comment at `anthropic-sdk-go/middleware.go:264`.

`gen_ai.usage.cache_creation.input_tokens` is genuinely Anthropic-specific: OpenAI's cache is automatic with no separate cache-write step to report.

## The semantics differ, and that is the part worth getting right

The two providers count cached tokens differently:

- **Anthropic** excludes cache reads from `input_tokens`, so the instrumentation folds them back in (`middleware.go:248-251`) to report the full prompt per semconv.
- **OpenAI** already includes cached tokens in `prompt_tokens`.

So OpenAI must record the value as a breakdown only. Folding it in the way the Anthropic path does would double-count the cached portion of every cache-hit request.

## Suggested fix

1. Record `gen_ai.usage.cache_read.input_tokens` from `usage.prompt_tokens_details.cached_tokens`, only when greater than zero, matching the Anthropic instrumentation's conditional behaviour. Do not add it to `gen_ai.usage.input_tokens`.
2. Move the `gen_ai.usage.cache_read.input_tokens` declaration from `registry.otelc.gen_ai.anthropic` to the shared `registry.otelc.gen_ai` group, which is where `openai.yaml` already documents shared otelc-invented attributes as living. Leave `cache_creation` where it is.

Point 2 touches the group organisation that #728 lists as an open decision ("Registry files are per-library but attribute groups are per-domain"), so it is worth a maintainer opinion. Following the convention the two files already document seemed like the smallest step that does not pre-empt that decision.

Happy to send a PR.

pradhyum6144:
OpenAI has automatic prompt caching and reports it as `usage.prompt_tokens_details.cached_tokens`, but the openai-go instrumentation never read that field, so cached prompt spend was invisible for OpenAI while the anthropic-sdk-go instrumentation has recorded it since #689. Credit to @dobbydobap, who spotted the asymmetry in the #709 discussion.

The two providers count cached tokens differently, and that difference is the reason this is not a copy of the Anthropic code. Anthropic excludes cache reads from `input_tokens`, so that instrumentation folds them back in to report the full prompt. OpenAI already includes them in `prompt_tokens`. So here the value is recorded as a breakdown of the input count and is never added to it, otherwise every cache-hit request would double-count the cached portion. There is a test pinning exactly that, since it is the easy thing to get wrong later.

Recorded only when greater than zero, so a request that did not hit the cache omits the attribute rather than reporting a zero, matching the Anthropic behaviour.

No `cache_creation` counterpart: OpenAI's cache is automatic and has no separate cache-write step to report.

On the registry side, `gen_ai.usage.cache_read.input_tokens` was declared with `id:` inside `registry.otelc.gen_ai.anthropic`, the provider-specific group, with a header comment calling both prompt-cache attributes Anthropic-specific. Cache reads are not, and while the attribute lives in a provider-scoped group `openai.yaml` cannot declare it without referencing another provider's group. This moves that one declaration to the shared `registry.otelc.gen_ai` group, which is where both files already document shared otelc-invented attributes as living, and leaves `cache_creation` in the Anthropic group. `weaver registry check` passes.

That move touches group organisation that #728 lists as an open decision, so if you would rather resolve it there first I am happy to split the schema change out and land only the code fix.

Tested with unit tests covering a cache hit (asserting `input_tokens` is unchanged), a response with no `prompt_tokens_details`, and one with `cached_tokens: 0`. `make lint-schema`, `golangci-lint` and the module test suite all pass, and the bundle is regenerated.

Fixes #1089


Aryanbhargava18:
The double-counting handling is correct — nice attention to the Anthropic divergence.

One gap: `internal/streaming/streaming.go` also parses `Usage` in both `processChatChunk` and `processCompletionChunk`, and the reader already sets `gen_ai.usage.input_tokens` from it. But those inline structs don't include `PromptTokensDetails`, so when a client uses `stream_options: {"include_usage": true}`, cached prompt tokens in the final streaming chunk are silently dropped. Should this PR also cover the streaming path, or is that a follow-up?

dobbydobap:
Thanks for picking this up, and for the credit.

Building on @Aryanbhargava18's streaming point rather than repeating it: since
#990 merged on 11 Aug the streaming reader is one shared module at
internal/streaming, with 34-line streaming_bridge.go files per version. So that
fix only needs writing once and covers v1, v2 and v3 together.

The non-streaming path is the one still split three ways. middleware.go exists
separately under openai-go/, v2/ and v3/, each 386 lines with the same
three-field Usage struct and no cache handling. This PR updates the v1 copy, so
once it merges v2 and v3 users still get no cache attribute on non-streaming
calls.

There should be no blocker. #990 established that none of the three copies
depend on the versioned SDK, and the middleware unmarshals the wire JSON into
its own local struct, so the same PromptTokensDetails addition drops into v2 and
v3 unchanged.

Worth mentioning that @darccio raised the same shape of concern on #990, that
every consumer app exercised only the bare v1 SDK, which is what led to
test/apps/openaiclientv2 being added there.

Are the other two majors meant as follow-ups, or worth folding in here? Either
is fine, I just did not want the asymmetry to outlive the fix for it.

The registry find is a good one. anthropic.yaml lines 19 to 20 state it
outright, so the assumption was in the contract as well as the middleware
comment.

pradhyum6144:
> The double-counting handling is correct — nice attention to the Anthropic divergence.
> 
> One gap: `internal/streaming/streaming.go` also parses `Usage` in both `processChatChunk` and `processCompletionChunk`, and the reader already sets `gen_ai.usage.input_tokens` from it. But those inline structs don't include `PromptTokensDetails`, so when a client uses `stream_options: {"include_usage": true}`, cached prompt tokens in the final streaming chunk are silently dropped. Should this PR also cover the streaming path, or is that a follow-up?

Confirmed. The chunk's usage is the same CompletionUsage type the non-streaming path uses (ChatCompletionChunk at chatcompletion.go:558, usage field at :603), so prompt_tokens_details is part of the payload being unmarshalled when include_usage is set. The parser's local struct just has nowhere to put it. Both copies drop it, streaming.go:217 and :256, each with the same three field Usage struct.

I'll fold it in. It lands a bit differently from the non-streaming fix though: internal/streaming keeps its own attribute keys in attrs.go rather than importing the per-version semconv packages, so it needs a key there, a field on StreamingReader, and the emit in finalize. Upside is that since #990 made it one shared module, that single change covers v1, v2 and v3 together.

pradhyum6144:
> Thanks for picking this up, and for the credit.
> 
> Building on @Aryanbhargava18's streaming point rather than repeating it: since #990 merged on 11 Aug the streaming reader is one shared module at internal/streaming, with 34-line streaming_bridge.go files per version. So that fix only needs writing once and covers v1, v2 and v3 together.
> 
> The non-streaming path is the one still split three ways. middleware.go exists separately under openai-go/, v2/ and v3/, each 386 lines with the same three-field Usage struct and no cache handling. This PR updates the v1 copy, so once it merges v2 and v3 users still get no cache attribute on non-streaming calls.
> 
> There should be no blocker. #990 established that none of the three copies depend on the versioned SDK, and the middleware unmarshals the wire JSON into its own local struct, so the same PromptTokensDetails addition drops into v2 and v3 unchanged.
> 
> Worth mentioning that @darccio raised the same shape of concern on #990, that every consumer app exercised only the bare v1 SDK, which is what led to test/apps/openaiclientv2 being added there.
> 
> Are the other two majors meant as follow-ups, or worth folding in here? Either is fine, I just did not want the asymmetry to outlive the fix for it.
> 
> The registry find is a good one. anthropic.yaml lines 19 to 20 state it outright, so the assumption was in the contract as well as the middleware comment.

You're right. I checked and all three middleware.go copies are 386 lines with zero cache references in v2 and v3, so as it stands this only fixes a third of the non-streaming surface.

One thing I noticed while looking that makes it worse than uneven coverage: openai.yaml names v3/middleware.go as its source of truth (lines 5-8). So this PR declares gen_ai.usage.cache_read.input_tokens in a registry whose stated source of truth is the one copy that doesn't emit it. That's the declaration describing something that isn't true, not just an asymmetry, and it settles the question for me. I'll fold v2 and v3 in here rather than leaving them as follow ups.

Also while I'm in that file, the same header still points at v3/streaming.go, which #990 replaced with streaming_bridge.go and internal/streaming, so I'll refresh those source of truth lines since I'm touching it anyway.

Thanks for the #990 context on openaiclientv2. Once the code change is in I'll look at what integration coverage actually makes sense for the versioned copies.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming_test.go
```diff
@@ -381,3 +381,62 @@ func assertSliceAttribute(t *testing.T, attrs []attribute.KeyValue, key string,
...
+// are recorded, and that they are not added to gen_ai.usage.input_tokens:
+// OpenAI already counts them inside prompt_tokens.
+func TestStreamingReader_CacheReadTokens(t *testing.T) {
+	sr := tracetest.NewSpanRecorder()
+	tp := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))
+	tr := tp.Tracer("test")
+	_, span := tr.Start(t.Context(), "test-stream")
+
+	streamData := "data: {\"id\":\"chatcmpl-cache\",\"model\":\"gpt-4o-mini\",\"choices\":" +
+		"[{\"delta\":{\"content\":\"hi\"},\"finish_reason\":\"stop\"}]," +
```
Aryanbhargava18, line 396: Your mock payload crams both choices and usage into a single chunk:
```json
"choices":[{"delta":{"content":"hi"},"finish_reason":"stop"}],
"usage":{"prompt_tokens_details":{"cached_tokens":1024}}
```
OpenAI's streaming API never emits this. Per the [stream_options contract](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options), the API guarantees that when `include_usage: true` is set, `usage` is delivered in an isolated, terminal chunk where the `choices` array is strictly empty (`[]`).

By asserting against an impossible JSON shape, the test fails to protect the production code. If a future contributor optimizes `processChatChunk` to bypass usage extraction when `len(choices) > 0` (a completely valid optimization based on the API contract), your test will silently mask the regression. Please split this into two separate chunks in the test string to accurately mirror the wire protocol.

Aryanbhargava18 (changes requested):
The streaming coverage and the v2/v3 fixes look good. The core parsing logic is correct, but the new streaming test implementation fundamentally violates the API contract.

instrumentation/github.com/openai/openai-go/internal/streaming/streaming_test.go
```diff
@@ -381,3 +381,62 @@ func assertSliceAttribute(t *testing.T, attrs []attribute.KeyValue, key string,
...
+	require.NoError(t, reader.Close())
+
+	spans := sr.Ended()
+	require.Len(t, spans, 1)
+
+	attrs := spans[0].Attributes()
+	assertInt64Attribute(t, attrs, "gen_ai.usage.cache_read.input_tokens", 1024)
+	assertInt64Attribute(t, attrs, "gen_ai.usage.input_tokens", 2048)
+	assertInt64Attribute(t, attrs, "gen_ai.usage.total_tokens", 2112)
+}
```
darccio, line 414: `TestStreamingReader_CacheReadTokens` only exercises `OpChat`. `processChatChunk`
and `processCompletionChunk` are separate, non-shared blocks in streaming.go, so
a bug specific to the completion path's cache handling would not be caught here.

```suggestion
}

// TestStreamingReader_CompletionCacheReadTokens verifies the completion-chunk
// path records cached prompt tokens the same way the chat-chunk path does.
func TestStreamingReader_CompletionCacheReadTokens(t *testing.T) {
	sr := tracetest.NewSpanRecorder()
	tp := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))
	tr := tp.Tracer("test")
	_, span := tr.Start(t.Context(), "test-stream")

	streamData := "data: {\"id\":\"cmpl-cache\",\"model\":\"gpt-4o-mini\",\"choices\":" +
		"[{\"text\":\"hi\",\"finish_reason\":\"stop\"}]," +
		"\"usage\":{\"prompt_tokens\":2048,\"completion_tokens\":64,\"total_tokens\":2112," +
		"\"prompt_tokens_details\":{\"cached_tokens\":1024}}}\n\ndata: [DONE]\n\n"

	body := io.NopCloser(bytes.NewReader([]byte(streamData)))
	reader := NewStreamingReader(body, span, time.Now(), OpCompletion)

	_, err := io.ReadAll(reader)
	require.NoError(t, err)
	require.NoError(t, reader.Close())

	spans := sr.Ended()
	require.Len(t, spans, 1)

	attrs := spans[0].Attributes()
	assertInt64Attribute(t, attrs, "gen_ai.usage.cache_read.input_tokens", 1024)
	assertInt64Attribute(t, attrs, "gen_ai.usage.input_tokens", 2048)
	assertInt64Attribute(t, attrs, "gen_ai.usage.total_tokens", 2112)
}
```

instrumentation/github.com/openai/openai-go/cache_tokens_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			attrs := chatSpanAttrsForUsage(t, tt.usage)
+
+			assertNoAttribute(t, attrs, cacheReadTokensKey)
+			assertInt64Attribute(t, attrs, "gen_ai.usage.input_tokens", 10)
+		})
+	}
+}
```
darccio, line 88: `chatSpanAttrsForUsage` always POSTs to `chat/completions`, which `classifyOperation`
routes to `parseChatResponse`. `parseCompletionResponse` (middleware.go:351) has its
own, separately-gated cache_read.input_tokens check at middleware.go:394, but it has
no test coverage of its own — a regression specific to that branch would not be caught.
This mirrors the streaming-side gap already flagged on `processCompletionChunk`.

```suggestion
}

// completionSpanAttrsForUsage runs one non-streaming legacy completion call
// through the middleware with the given usage JSON and returns the recorded
// span attributes.
func completionSpanAttrsForUsage(t *testing.T, usage string) []attribute.KeyValue {
	t.Helper()
	sr := setupTestTracer(t)

	req, _ := http.NewRequest(
		http.MethodPost,
		"http://api.openai.com/v1/completions",
		io.NopCloser(bytes.NewReader([]byte(`{"model":"gpt-3.5-turbo-instruct"}`))),
	)
	respBody := `{"id":"cmpl-1","model":"gpt-3.5-turbo-instruct","choices":[{"finish_reason":"stop"}],"usage":` + usage + `}`
	next := func(_ *http.Request) (*http.Response, error) {
		return &http.Response{
			StatusCode: http.StatusOK,
			Header:     http.Header{"Content-Type": []string{"application/json"}},
			Body:       io.NopCloser(bytes.NewReader([]byte(respBody))),
		}, nil
	}

	_, err := OtelMiddleware()(req, next)
	require.NoError(t, err)

	spans := sr.Ended()
	require.Len(t, spans, 1)
	return spans[0].Attributes()
}

// TestCompletionRecordsCacheReadTokens verifies the legacy completions
// endpoint records cache_read.input_tokens the same way the chat endpoint
// does, and that it is not folded into gen_ai.usage.input_tokens.
func TestCompletionRecordsCacheReadTokens(t *testing.T) {
	attrs := completionSpanAttrsForUsage(t,
		`{"prompt_tokens":2048,"completion_tokens":64,"total_tokens":2112,`+
			`"prompt_tokens_details":{"cached_tokens":1024}}`)

	assertInt64Attribute(t, attrs, cacheReadTokensKey, 1024)
	assertInt64Attribute(t, attrs, "gen_ai.usage.input_tokens", 2048)
	assertInt64Attribute(t, attrs, "gen_ai.usage.total_tokens", 2112)
}

// TestCompletionOmitsCacheReadTokensWhenUnused verifies the attribute is
// absent rather than recorded as zero when the request did not hit the cache.
func TestCompletionOmitsCacheReadTokensWhenUnused(t *testing.T) {
	attrs := completionSpanAttrsForUsage(t,
		`{"prompt_tokens":10,"completion_tokens":20,"total_tokens":30,`+
			`"prompt_tokens_details":{"cached_tokens":0}}`)

	assertNoAttribute(t, attrs, cacheReadTokensKey)
	assertInt64Attribute(t, attrs, "gen_ai.usage.input_tokens", 10)
}
```

darccio (changes requested):

PR #1094, opened 2026-08-13
fix(net/http): end spans after instrumentation is disabled

SSOHEB:
## Description

Fix the HTTP server instrumentation lifecycle so spans created before a
runtime instrumentation state change are always ended.

`AfterServeHTTP` previously re-checked the instrumentation enablement state before retrieving the span stored by `BeforeServeHTTP`. If instrumentation was disabled after the span was created but before the request completed, the after-hook returned early and skipped `span.End()`.

This change removes that redundant enablement check and relies on the presence of the span in the hook context to determine whether there is a span to end.

A regression test covers the enabled to disabled transition.

## Motivation

Instrumentation enablement can be evaluated dynamically between the
`BeforeServeHTTP` and `AfterServeHTTP` hooks.

Previously, this sequence could occur:

1. Instrumentation is enabled and `BeforeServeHTTP` creates a span.
2. The instrumentation is disabled while the request is in flight.
3. `AfterServeHTTP` observes the disabled state and returns early.
4. The span created by `BeforeServeHTTP` is never ended.

The regression test reproduces this behavior against the previous
implementation and verifies that the span is ended after the fix.

The existing test expectation was also updated to reflect the corrected
span lifecycle when a span is already present in the hook context.

## Testing

- `go test ./instrumentation/net/http/server/...`
- `go test ./...`

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


Labreo (commented):
Nice catch, this is a neat find. Since runtime.Instrumented() reads the environment variables on every call, the flag can definitely change while a request is in flight. Re-checking Enable() in AfterServeHTTP meant any span created in BeforeServeHTTP would get skipped if the flag turned off mid-request, leaving the span hanging without calling span.End().

Relying on the span being present in the hook context makes a lot of sense, and it matches what kafka-go consumer hook is doing. The new test covers the enabled to disabled transition really nicely too.

Also noticed that net/http client hook has a similar setup in AfterRoundTrip where it checks Enable() before ending the span. Might be worth taking a look at that one as well for a quick follow-up fix.

SSOHEB:
@Labreo Thanks for the review! Glad the lifecycle reasoning and regression test make sense.

For the client hook, I’ll first verify whether the same lifecycle issue can occur when instrumentation is enabled when the span is created but becomes disabled before AfterRoundTrip runs. If I can reproduce it, I’ll add a regression test and then make the minimal fix.

If it has the same issue, would you prefer that I fix it in this PR or raise a separate PR?

Labreo:
I would wait for feedback from a maintainer personally and defer to their opinion regarding the follow up.Also it would be helpful if you could link the issue if there is one.

SSOHEB:
> I would wait for feedback from a maintainer personally and defer to their opinion regarding the follow up.Also it would be helpful if you could link the issue if there is one.

That makes sense. I’ll wait for maintainer feedback before deciding how to handle the client hook follow up. If I reproduce the same issue, I’ll link the relevant issue if one already exists, or open one if there isn’t one.  😊

DavidGrath:
Good Day,  

Please there's a clarification I'd like to ask for
From my understanding, environment variables are inherited by a process at startup and aren't dynamically updated when the environment outside the process changes. So, ordinarily, reading an environment variable again at runtime would return the same value unless the process itself modifies its environment.
This question isn't just for this PR, I'm mainly curious as to where the gap in my understanding comes from concerning the behavior of `runtime.Instrumented`

SSOHEB:
@DavidGrath Good point. You're correct that changes to environment variables made outside the running process aren't propagated to it. The part I was referring to is that runtime.Instrumented() calls os.Getenv() each time it is invoked, so if the running process itself changes the environment using os.Setenv(), a subsequent call can observe a different value.
I verified that Enable() in the HTTP server hook calls runtime.Instrumented() directly, so there is no cached enablement state between BeforeServeHTTP and AfterServeHTTP. The regression test uses t.Setenv() to reproduce that process local state change between the two hooks. I agree that this is different from an external environment change after process startup.

DavidGrath:
@SSOHEB , thanks

y1yang0:
> Also noticed that net/http client hook has a similar setup in AfterRoundTrip where it checks Enable() before ending the span. Might be worth taking a look at that one as well for a quick follow-up fix.

Good catch. Thanks for taking a look at this. I suggest fixing AfterRoundTrip in this PR as well

SSOHEB:
> > Also noticed that net/http client hook has a similar setup in AfterRoundTrip where it checks Enable() before ending the span. Might be worth taking a look at that one as well for a quick follow-up fix.
> 
> Good catch. Thanks for taking a look at this. I suggest fixing AfterRoundTrip in this PR as well

Thanks for the suggestion. I fixed the same issue in AfterRoundTrip as well and pushed the changes to this PR.

y1yang0 (approved):

PR #1097, opened 2026-08-13
fix(tool): parse +incompatible and build metadata in module cache paths

### LINKED ISSUE #1096: bug(setup): findModVersion cannot parse +incompatible versions in module cache paths
### Describe the bug

Version-gated instrumentation rules are silently skipped when a dependency is resolved from the Go module cache with a `+incompatible` version suffix.

`findModVersion` in `tool/internal/setup/find.go` derives a dependency's version by regex-parsing the compile source path for a module cache segment (`...@vX.Y.Z/...`):

```go
var versionRegexp = regexp.MustCompile(`@v\d+\.\d+\.\d+(-.*?)?/`)

func findModVersion(path string) string {
    version := versionRegexp.FindString(filepath.ToSlash(path))
    if version == "" {
        return ""
    }
    return version[1 : len(version)-1]
}
```

For legacy pre-module repositories (v2+ tagged without a `go.mod`), the Go module cache stores the `+incompatible` suffix in the path directory name. Because `versionRegexp` only matches hyphenated suffixes (`-`), `findModVersion` fails:

| Source Path | `findModVersion()` Result | Expected |
|---|---|---|
| `/go/pkg/mod/github.com/foo/bar@v1.2.3/pkg/foo.go` | `v1.2.3` | `v1.2.3` |
| `/go/pkg/mod/github.com/foo/bar@v1.2.3-rc.1/pkg/foo.go` | `v1.2.3-rc.1` | `v1.2.3-rc.1` |
| `/go/pkg/mod/github.com/foo/bar@v1.2.3+incompatible/pkg/foo.go` | `""` | `v1.2.3+incompatible` |
| `/go/pkg/mod/github.com/foo/bar@v0.0.0-20210101120000-abcdef/pkg/foo.go` | `v0.0.0-20210101120000-abcdef` | `v0.0.0-20210101120000-abcdef` |

An empty `Dependency.Version` causes `VersionInRange("", range)` to evaluate to `false`, causing the rule to be dropped during both the setup phase (`tool/internal/setup/match.go`) and the pin phase (`tool/internal/setup/pin.go`).

### Environment

 - OS: Linux / macOS / Windows
 - Go Version: 1.23+
 - Version: `main` branch

### To Reproduce

1. Target a dependency resolved from the Go module cache with a `+incompatible` tag (e.g. `...@v1.2.3+incompatible/...`).
2. Define a version-gated instrumentation rule targeting this dependency.
3. Run setup / pin with `otelc`.
4. See error/warning: The rule does not match because the dependency version is extracted as `""`, emitting a warning that the version is unresolved even though it is present in the path.

### Expected behavior

`findModVersion` should extract `v1.2.3+incompatible` from `/go/pkg/mod/github.com/foo/bar@v1.2.3+incompatible/pkg/foo.go` so version-gated rules function properly for `+incompatible` dependencies.

### Additional context

Fixing `versionRegexp` to accept `[-+]` (e.g. `@v\d+\.\d+\.\d+([-+][^/]*)?/`) covers both `+incompatible` and build metadata (`+build`) alongside existing pre-release tags (`-rc.1`).

Labreo:
## Description

This PR updates `versionRegexp` in `tool/internal/setup/find.go` to support `+incompatible` suffixes and `+buildmetadata` tags when parsing module versions from compile source paths in Go's module cache.

Unit test cases have been added to `TestFindModVersion` in `tool/internal/setup/find_test.go` to verify parsing of both `+incompatible` and build metadata path formats.

## Motivation

In Go's module cache (`$GOPATH/pkg/mod`), pre-module legacy libraries tagged at v2+ are stored with a `+incompatible` suffix (e.g., `/go/pkg/mod/github.com/foo/bar@v2.1.0+incompatible/pkg/foo.go`).

Previously, `findModVersion` relied on a regex (`@v\d+\.\d+\.\d+(-.*?)?/`) that only matched hyphenated pre-release tags (`-rc.1`). For `+incompatible` dependencies, `findModVersion` returned `""`, causing `VersionInRange("", range)` to evaluate to `false` and silently drop or skip version-gated instrumentation rules during setup and pinning.

Fixes #1096  

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


vignesh2027:
Checked this against a real module cache and the fix holds up. Four modules in mine land on the `+` form:

```
gotest.tools@v2.2.0+incompatible
github.com/evanphx/json-patch@v5.9.11+incompatible
github.com/pierrec/lz4@v2.6.1+incompatible
github.com/peterbourgon/diskv@v2.0.1+incompatible
```

Every one of those returned an empty version under the old pattern, so this is a real gap rather than a theoretical one.

One thought on the second test case. `+incompatible` is the only `+` form that reaches a cache path, since Go strips ordinary semver build metadata, so `v1.2.3+20220101` is covering a shape that does not occur. Not harmful, and the regex handles it either way, but naming a real module in the case might document the intent better than a synthetic version.

Also worth noting that swapping `.*?` for `[^/]*` makes the boundary explicit rather than relying on laziness stopping at the separator. Same result, easier to reason about.

Labreo:
Thanks for checking this against real module cache paths and confirming those examples!

That point on build metadata makes sense. I updated the test cases in find_test.go to use realistic module paths (github.com/evanphx/json-patch and gotest.tools) instead of the synthetic version so the test suite documents the actual cache shape better. Also rebased the branch on the latest main.

y1yang0 (approved):

PR #1098, opened 2026-08-13
fix(redis): instrument ring shards created before the hook runs

vignesh2027:
## Description

`NewRing` builds a shard for every entry in `RingOptions.Addrs` during construction. `afterNewRingClientV9` runs after that and calls `OnNewNode`, which only invokes its callback for shards created after it is registered. The hook therefore never reached any shard the caller configured up front.

The fix keeps the `OnNewNode` registration, which still matters for shards added later through `SetAddrs`, and adds a pass over the shards that already exist.

`client_hook.go` had no coverage on any of its six hook functions, so this also adds tests for all of them. Each builds a client against a port nothing listens on and asserts the command produced a client span carrying the expected endpoint. The hook opens its span before handing off to the next hook in the chain, so a failed dial still exercises the full path and the tests need no Redis server.

`TestAfterNewRingClientV9_AttachesHookToEachNode` fails on main and passes with the change, so it pins the behaviour rather than describing it. Package coverage goes from 89.2% to 96.9%.

## Motivation

Configuring shards up front is how a ring is normally set up, so in practice Redis Ring clients were producing no telemetry at all. The failure is silent: no error, no warning, just an absence of spans. Nothing tested those six functions, which is why it went unnoticed.

## Compatibility / behavioral change

This changes the telemetry contract for existing `redis.Ring` users, so it is worth calling out in the release notes:

- **Pre-configured ring shards are now instrumented.** Shards configured up front via `RingOptions.Addrs` (the common case) produced zero spans before this change and now emit per-command client spans. To a telemetry pipeline this can look like a traffic spike: new time series, higher storage cost, and potentially false alerts on first rollout.
- **Manual workarounds now double-count.** Anyone who worked around the bug by attaching `otelRedisHook` to ring shards themselves (their own `ForEachShard` + `AddHook`) will get two hook entries in the chain and two identical spans per command. Those workarounds should be removed once this ships.

The code fix itself is unaffected by the above; these are operational notes for upgraders.

---

## Checklist

- [x] Tests pass: `make test`


instrumentation/github.com/redis/go-redis/v9/client_hook.go
```diff
@@ -36,6 +38,21 @@ func afterNewRingClientV9(call hook.HookContext, client *redis.Ring) {
 	client.OnNewNode(func(rdb *redis.Client) {
 		rdb.AddHook(newOtelRedisHook(rdb.Options().Addr))
 	})
+	// NewRing builds a shard for every entry in RingOptions.Addrs before this
+	// hook runs, and OnNewNode only fires for shards created after it is
+	// registered. Without this pass the shards a caller configured up front
+	// are never instrumented, which is the common case.
+	attachHookToExistingShards(client)
+}
```
NameHaibinZhang, line 46: **[Major]** This change silently changes the telemetry contract for existing users of `redis.Ring`.

Two follow-on effects that need release-note coverage:

1. **Newly emitted spans.** Users who configure shards up front via `RingOptions.Addrs` (the common case) get zero spans today. After this PR, the same code path suddenly emits per-command client spans. From a telemetry pipeline's perspective this looks like a traffic spike and can create new time series, higher storage cost, and false alerts.
2. **Duplicate spans for manual workarounds.** Anyone who patched around the bug by iterating shards themselves (e.g. their own `ForEachShard` + `AddHook`) will now have two `otelRedisHook` entries in the chain, producing two identical client spans per command. That doubles QPS/latency counts derived from spans.

Neither is a code fix — the fix itself is correct — but the CHANGELOG / release notes should call out:
- Pre-configured ring shards are now instrumented (previously they were silently dropped).
- Any manual workaround that adds `otelRedisHook` to ring shards should be removed to avoid duplicate spans.

Optional but valuable: consider a small idempotency guard in `newOtelRedisHook`/`AddHook` (e.g. skip if a hook of this type is already attached), if go-redis exposes enough to check that.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	afterNewFailOverRedisClientV9(hooktest.NewMockHookContext(), client)
+
+	spans := pingAndCollect(t, func(ctx context.Context) error {
+		return client.Ping(ctx).Err()
+	}, sr)
+	assertRedisSpan(t, spans, unreachableHost)
+}
+
+func TestAfterNewRingClientV9_AttachesHookToEachNode(t *testing.T) {
```
NameHaibinZhang: **[Major]** This test proves the fix on a single shard, but the fix is specifically about instrumenting **every** pre-configured shard, and also about keeping the `OnNewNode` path working for shards added after construction. Neither is verified today, so a regression on either half would slip through.

Please strengthen coverage with two additions:

**(a) Multiple pre-configured shards** — prove the iteration touches all of them:

```go
ring := redis.NewRing(&redis.RingOptions{
    Addrs: map[string]string{
        "shard0": "127.0.0.1:1",
        "shard1": "127.0.0.2:1",
    },
    DialTimeout: 50 * time.Millisecond,
    MaxRetries:  -1,
})
t.Cleanup(func() { _ = ring.Close() })

afterNewRingClientV9(hooktest.NewMockHookContext(), ring)

// Fire a command against each shard directly so each hook has to run.
_ = ring.ForEachShard(ctx, func(c context.Context, rdb *redis.Client) error {
    _ = rdb.Ping(c).Err()
    return nil
})

// Assert both server.address values appear.
```

**(b) A regression test for the retained `OnNewNode` path** — add a shard via `SetAddrs` **after** `afterNewRingClientV9` runs, and assert its spans still carry the new endpoint. Without this, a future refactor that drops `OnNewNode` (thinking `ForEachShard` alone is enough) would silently regress dynamic-shard support.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
```diff
@@ -0,0 +1,203 @@
...
+	// the case OnNewNode alone does not cover.
+	afterNewRingClientV9(hooktest.NewMockHookContext(), ring)
+
+	spans := pingAndCollect(t, func(ctx context.Context) error {
+		return ring.Ping(ctx).Err()
+	}, sr)
+	assertRedisSpan(t, spans, unreachableHost)
+}
+
+func TestAfterNewClusterClientV9_RegistersNodeCallback(t *testing.T) {
```
NameHaibinZhang: **[Minor]** After this PR, `afterNewRingClientV9` uses `OnNewNode` **plus** `ForEachShard`, while `afterNewClusterClientV9` still uses `OnNewNode` alone. If cluster nodes are ever materialized before this hook is registered (analogous to the ring bug being fixed here), the same silent-drop issue would exist for clusters.

It looks like `NewClusterClient` defers node discovery to first use, so the current cluster code is likely fine — but that guarantee is not obvious from the code. Two low-cost options to future-proof:

- Add a short comment on `afterNewClusterClientV9` in `client_hook.go` explaining why `OnNewNode` alone is sufficient there (i.e. nodes are created lazily after registration).
- Or add a companion test similar to (b) above: construct a `ClusterClient`, then attach the hook, then simulate node discovery and assert spans carry the new endpoint.

Without one of these, the ring/cluster asymmetry is a latent trap for the next contributor.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	afterNewRedisClientV9(hooktest.NewMockHookContext(), client)
+
+	spans := pingAndCollect(t, func(ctx context.Context) error {
+		return client.Ping(ctx).Err()
+	}, sr)
+	assertRedisSpan(t, spans, unreachableHost)
+}
+
+func TestAfterNewFailOverRedisClientV9_AttachesHook(t *testing.T) {
```
NameHaibinZhang: **[Suggestion]** The comment acknowledges this, but using `redis.NewClient` here means the test exercises `afterNewFailOverRedisClientV9` against a client that was **never** built through the failover construction path. Functionally the hook does the same thing, but the test does not distinguish a regression in the failover-specific hook from a regression in the plain-client hook.

A more faithful version:

```go
client := redis.NewFailoverClient(&redis.FailoverOptions{
    MasterName:    "mymaster",
    SentinelAddrs: []string{unreachableAddr},
    DialTimeout:   50 * time.Millisecond,
    MaxRetries:    -1,
})
```

If sentinel discovery makes the test flaky or slow, keeping `NewClient` is defensible — but please extend the comment to explain that limitation explicitly, so the next reader does not assume the failover-specific path is covered.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
```diff
@@ -0,0 +1,203 @@
...
+	return &redis.Options{
+		Addr:        unreachableAddr,
+		DialTimeout: 50 * time.Millisecond,
+		MaxRetries:  -1,
+	}
+}
+
+// newRecordingClient prepares a span recorder and resets the package-level
+// init guard, which otherwise leaks state between tests in this package.
+func newRecordingClient(t *testing.T) *tracetest.SpanRecorder {
```
NameHaibinZhang: **[Suggestion]** `newRecordingClient` resets the package-level `initOnce` and installs a new global `TracerProvider`. This is safe today because no test in this package uses `t.Parallel()`, but the new tests here extend that constraint to more entry points, making the assumption more load-bearing.

Two options:

1. Add a one-line comment on this helper stating that tests using it must remain serial — so a future author does not add `t.Parallel()` and get flaky span races.
2. Longer term, consider refactoring `initInstrumentation` to accept an injected tracer (or read from a testable indirection), which would eliminate the `initOnce` reset trick entirely and remove the serial-execution constraint.

NameHaibinZhang (commented):
### Summary

Nice, focused fix — the `NewRing` + `OnNewNode` timing gap is a real bug and `ForEachShard` is the right hook to use for it. Correctness review found no blockers: `Ring.ForEachShard` iterates a shadow copy of the current shard list, and `OnNewNode` only fires for shards created *after* registration, so there is no confirmed double-hook path. `rdb.Options()` and the ignored `ForEachShard` error are both safe under the current go-redis contract. Test coverage on all six hook entrypoints is very welcome.

A few items are worth addressing before merge:

**Should fix**
- The ring-shard regression test only uses a single pre-configured shard and does not verify that the `OnNewNode` path still instruments shards added later via `SetAddrs`. Both are inline.
- This change silently produces spans that never existed on `main` for the common case (pre-configured `Addrs`). Users who worked around the bug with their own `AddHook`/`ForEachShard` loop will now emit duplicate spans. Please flag this in the release notes / CHANGELOG.

**Nice to have**
- `afterNewClusterClientV9` still relies on `OnNewNode` alone. If pre-configured cluster nodes ever suffer the same asymmetry, this fix is incomplete for clusters. A short comment (or a follow-up) confirming the cluster case is fine would prevent the same regression re-appearing there.
- The failover test uses `redis.NewClient`; using `redis.NewFailoverClient` would exercise the actual failover construction path.
- The test file adds more usages of the `initOnce = *new(sync.Once)` reset pattern. It works today because these tests run serially, but that assumption is now load-bearing across even more tests; a note near `newRecordingClient` would document the constraint.

Full details in the inline comments.

vignesh2027:
Thanks for the thorough review, this made the change materially better. All items addressed in the latest push.

**Test coverage (Major).** Both halves are now pinned:
- `TestAfterNewRingClientV9_AttachesHookToEachNode` now configures two shards on distinct hosts and drives each directly through `ForEachShard`, asserting a span for both `server.address` values. It fails on main (zero pre-configured shards instrumented) and passes with the fix, so it pins the ForEachShard pass.
- `TestAfterNewRingClientV9_InstrumentsShardAddedViaSetAddrs` adds a shard via `SetAddrs` after the hook is registered and asserts the new endpoint is instrumented. That guards the retained `OnNewNode` path, so a future refactor dropping it would be caught.

I used `ForEachShard` to drive both rather than `ring.Ping`, because `Ping` routes by key and would only touch one shard, which made an earlier version of this test nondeterministic.

**One thing I hit while doing this, worth flagging.** Adding the multi-shard and SetAddrs tests surfaced a data race under `go test -race`: `NewRing` starts a background heartbeat goroutine that pings every shard on a 500ms ticker, and those background pings run through the hook and kept touching the global tracer state the tests reset between runs. CI does not run `-race`, so it was invisible, but it is real. Fixed by setting `HeartbeatFrequency` far into the future in the test ring options, which stops the background pings entirely and leaves each test to drive its shards explicitly. The suite now passes `-race -shuffle=on -count=20`.

**Release notes (Major).** I could not find a CHANGELOG or `.chloggen` in the repo, so I have put the two operational notes (pre-configured shards now instrumented; manual workarounds should be removed to avoid duplicate spans) in a Compatibility section in the PR description. Happy to move that wherever release notes are actually collected if you point me at it.

On the optional idempotency guard: I looked, and `hooksMixin` is unexported with no public getter for installed hooks, so there is no clean way to check for an existing `otelRedisHook` before adding one without reaching into go-redis internals. The workaround-removal note seemed the safer route than a reflective guard.

**Cluster comment (Minor).** Added a comment on `afterNewClusterClientV9` explaining why `OnNewNode` alone is sufficient there: `NewClusterClient` creates no nodes at construction (its node map starts empty) and materialises them lazily on first use, after the callback is registered, so unlike `NewRing` there are no pre-existing nodes to miss. I verified this against go-redis v9.21.0 (`newClusterNodes` initialises an empty map; nodes are created via `GetOrCreate` which fires `onNewNode`). The comment also notes it would need the same treatment if that ever changes.

**Failover test (Suggestion).** Switched to `redis.NewFailoverClient` so it exercises the real failover construction path. Sentinel discovery against the unreachable address fails fast (well under a second, measured across runs) so it stays deterministic and needs no server. The expected host is derived from the client rather than hard-coded, since a failover client with no resolved master reports a placeholder addr.

**Serial-execution note (Suggestion).** Expanded the comment on `newRecordingClient` to state that tests using it must stay serial and must not add `t.Parallel()`, with the reason (shared recorder plus the init-guard reset), and a note that an injectable tracer would remove the constraint.

Package coverage is now 97.9% with all six hook entrypoints at 100%.

vignesh2027:
Thanks again for such a careful review, Haibin. I really appreciate the depth of it, the SetAddrs regression point and the cluster asymmetry in particular made the change noticeably better, and the note about the serial-execution constraint was a good catch. All six items are addressed in the latest push, with the specifics in my reply above.

One heads-up so the red checks do not mislead you when you get back to it. The failures are the Install protoc step reporting "unable to get latest version", which is a GitHub infrastructure flake rather than anything in this change. Main's own Integration Tests hit the identical failure at 13:34 today with none of this code in them, while the 11:38, 12:22 and 12:50 runs passed, so it is intermittent on their side. Everything passes locally, including go test -race -shuffle=on, the golden-files check, and verify-bundle. A re-run should clear it whenever it is convenient for you, since fork PRs cannot re-run Actions themselves.

No rush at all, and thanks again for the time you put into this.

vignesh2027:
Thanks again for such a careful review, Haibin. Every item is addressed in the latest push, with the specifics in my earlier reply.

Two follow-ups since then. First, a correction on the release-note point: I had said there was no CHANGELOG, but the repo has one now, so I moved the note into a proper entry under Fixed, with the two compatibility callouts you asked for (pre-configured shards now emit spans, and manual AddHook/ForEachShard workarounds should be removed to avoid duplicates).

Second, in case yesterday's red checks looked concerning: those were the Install protoc setup step failing with "unable to get latest version", a GitHub infrastructure flake that also hit main's own Integration Tests at 13:34 yesterday, not anything in this change. It has since cleared. I have rebased onto latest main and regenerated the bundle, so the PR is mergeable and CI is clean again. No rush at all on the re-review.

vignesh2027:
Thanks for the thorough review. All five points are addressed.

**1. Release-note coverage (Major).** Added a CHANGELOG entry under Fixed. It states that rings built from `RingOptions.Addrs` now emit client spans where they emitted none before, notes that this will look like new traffic to a telemetry pipeline, and tells anyone who attached the hook to ring shards themselves to remove that workaround so commands are not recorded twice.

On the idempotency guard: go-redis does not expose the installed hook chain, so a hook cannot check whether one of its own is already attached. I left it out rather than guess, and covered the duplicate-span case in the release note instead.

**2. Test coverage (Major).** Both halves are covered now.
- `TestAfterNewRingClientV9_AttachesHookToEachNode` configures two shards up front on distinct hosts and asserts a span for each, so the `ForEachShard` pass has to reach both.
- `TestAfterNewRingClientV9_InstrumentsShardAddedViaSetAddrs` installs the hook first, then adds a shard through `SetAddrs`, and asserts the new endpoint's span. That pins the retained `OnNewNode` path, so dropping it on the assumption that `ForEachShard` alone is enough would fail here.

**3. Ring and cluster asymmetry (Minor).** Added a comment on `afterNewClusterClientV9` explaining that cluster nodes are created lazily on first use, after the callback is registered, so every node passes through `OnNewNode`, and that the ring needs the extra pass only because `NewRing` builds its configured shards during construction. It also says what would have to change if go-redis ever materialises cluster nodes eagerly.

**4. Failover path (Suggestion).** Switched to `redis.NewFailoverClient` so the test exercises the failover construction path rather than a plain client standing in for it. Sentinel discovery against an unreachable address fails fast, so it stays deterministic without a server. Since an unresolved master reports a placeholder address, the expected host is read from the client rather than hardcoded.

**5. Serial execution constraint (Suggestion).** Documented on `newRecordingClient`: it mutates package-level `initOnce` and installs a process-global `TracerProvider`, so tests using it must stay serial and must not add `t.Parallel()`. The comment also records the longer-term fix of injecting a tracer into `initInstrumentation`, which would remove the reset trick and the constraint together.

instrumentation/github.com/redis/go-redis/v9/client_hook.go
vignesh2027: Added a CHANGELOG entry covering both effects you listed: that rings configured through `RingOptions.Addrs` now emit per-command client spans where they previously emitted none, which will read as new traffic to a pipeline, and that anyone who worked around the bug by attaching the hook to shards themselves should remove that code or each command will record two identical spans.

On the optional idempotency guard: go-redis keeps its hooks in the unexported `hooksMixin` and exposes only `AddHook`, with no way to read back what is already attached. So a skip-if-present check is not implementable from outside the library today. I left it out rather than approximate it.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
vignesh2027: Both added.

`TestAfterNewRingClientV9_AttachesHookToEachNode` builds a ring with two pre-configured shards, runs the hook, then drives a command through each shard and asserts both `server.address` values appear. I drove them with `ForEachShard` rather than `ring.Ping`, since Ping routes by key and would only ever touch one shard, which is what let the single-shard version look like it passed.

`TestAfterNewRingClientV9_InstrumentsShardAddedViaSetAddrs` covers the other half: it adds a shard with `SetAddrs` after the hook is registered, so that shard arrives through the retained `OnNewNode` path rather than the `ForEachShard` pass. That pins exactly the behaviour a refactor dropping `OnNewNode` would silently regress.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
vignesh2027: Went with the comment option. `afterNewClusterClientV9` now explains why `OnNewNode` alone is sufficient there: `NewClusterClient` creates no nodes during construction, its node map starts empty and nodes are created lazily on first use, after the callback is registered, so every node passes through `OnNewNode`. The ring needs the extra pass only because `NewRing` builds its configured shards during construction, before the hook runs.

The comment also records the condition that would invalidate it, that if go-redis ever materialises cluster nodes eagerly at construction the same `ForEachShard` treatment would be needed, so the asymmetry is no longer a silent trap.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
vignesh2027: Switched to the real constructor. The test now builds through `redis.NewFailoverClient` with an unreachable sentinel address and fail-fast options, so it exercises the failover-specific construction path rather than a plain client standing in for it. It stays fast because sentinel discovery never resolves and the options cut the retries.

instrumentation/github.com/redis/go-redis/v9/client_hook_attach_test.go
vignesh2027: Took option 1 for now. `newRecordingClient` carries a comment stating that tests using it must stay serial, that adding `t.Parallel()` would race the shared recorder and the init guard and make span capture nondeterministic, and that the longer-term fix is giving `initInstrumentation` an injectable tracer so the reset trick and the constraint both disappear. That felt like the right scope for this PR, with the refactor left as a follow-up rather than bundled into a bug fix.

PR #1100, opened 2026-08-14
test(util): cover the fatal paths in assert.go

vignesh2027:
## Description

Adds tests for the fatal paths in `assert.go`. `Assert`'s failing branch, `ShouldNotReachHere` and `Unimplemented` were at 50%, 0% and 0%.

Each test re-execs the test binary with only the target test selected and a guard variable set, so the fatal branch runs in a child process while the parent asserts on its exit status and output. `assert_test.go` already uses this shape for `AssertType`; the boilerplate is pulled into one helper rather than repeated three times.

All four functions in `assert.go` now report 100%. The package goes from 89.2% to 90.2%.

## Motivation

The gap was structural rather than neglect. All three call `ex.Fatalf`, which exits the process, so none can be driven from the test process without taking the run down with it.

Every case asserts the message reaches the output, not just the exit code. The entire value of `Unimplemented` and of a failed `Assert` is telling the reader which condition broke.

This finishes the `util` half of what #248 set out to do before it went stale. The `ex` half landed in #928.

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/util/assert_fatal_test.go
+++ b/tool/util/assert_fatal_test.go
@@ -1,0 +1,83 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package util
+
+import (
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// fatalCaseVar names the environment variable that tells a re-executed test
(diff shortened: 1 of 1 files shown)


darccio (approved):
LGTM

y1yang0:
Please consider https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/1128 test file naming convention, maybe we can add a CI check to enforce it

PR #1101, opened 2026-08-14
fix(instrumentation/redis): redact AUTH and HELLO credentials from db.query.text

### LINKED ISSUE #1091: Redis `db.query.text` includes AUTH/HELLO passwords
### Describe the bug

Redis spans dump the raw command args into `db.query.text`, including the connection handshake.

If the client has `Password` set, first command triggers `HELLO … AUTH <password>` (or `AUTH` on fallback) through `ProcessHook`. I got:

```
db.query.text = "hello 3 auth default s3cret"
```

`getRedisV9Statement` just joins `cmd.Args()` with no masking.

### To Reproduce

1. Use a Redis client with a password:

```go
rdb := redis.NewClient(&redis.Options{
    Addr:     "localhost:6379",
    Password: "s3cret",
})
rdb.Ping(ctx)
```

2. Look at the client spans.
3. The `hello` (or `auth`) span has the password in `db.query.text`.

### Expected behavior

Keep the span, redact the secret. e.g. `HELLO 3 AUTH default s3cret` → `HELLO 3 AUTH ? ?`

ipsitapp8:
## Description

Redis spans were putting the connection password into `db.query.text`.

`getRedisV9Statement` joins every command argument, and go-redis sends the handshake before any of your own commands run, so with `Options.Password` set you get:

```
db.query.text = "hello 3 auth default s3cret"
```

This redacts the credential arguments to `?` and keeps the span, so the handshake is still visible and timed:

```
hello 3 auth default s3cret   ->   hello 3 auth ? ?
```

It covers `AUTH password`, the ACL form `AUTH username password`, and the optional `AUTH` section inside `HELLO`. The AUTH section of `HELLO` is optional and comes after the protocol version, so I look for the `auth` token rather than assuming an index. A trailing `SETNAME clientname` stays visible since only the two arguments after `auth` are hidden, and only `args[0]` is matched so a key that happens to be named `auth` is untouched.

## Motivation

It affects any app that sets a Redis password, not only ones that call `AUTH` themselves, and the value gets exported to whatever tracing backend is configured.

Out of scope here: `CONFIG SET requirepass` also carries a secret. Happy to do that separately if you want it.

Fixes #1091

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### Notes on the checklist

`make lint` and `make format` both flag a lot outside this change: 40 lint issues in `tool/`, and `make format` reflows ~95 unrelated YAML and Go files. Neither touches the two files here, so I left all of that alone. Looks like a local tool version differing from CI's pinned one.

`make test-unit/instrumentation` is green for 21 of 22 modules. `mongo-driver/mongo` is killed by the 6m timeout locally, which is unrelated to this change.


ipsitapp8:
@amazingakai @NameHaibinZhang please check the pr whenever you get a chance!

NameHaibinZhang (approved):

PR #1102, opened 2026-08-14
fix(tool/internal/instrument): prevent nil pointer panic on bodyless functions in directive and raw rules

### LINKED ISSUE #1093: fix(tool/internal/instrument): prevent nil pointer panic on bodyless functions in directive and raw rules
### Describe the bug

In `applyDirectiveRule` (`tool/internal/instrument/apply_directive.go`) and `insertRaw` (`tool/internal/instrument/apply_raw.go`), statements are prepended to function bodies without checking whether the function has a body:

```go
funcDecl.Body.List = append(stmts, funcDecl.Body.List...)
```

In Go AST representations, external assembly routines and forward declarations (for example, `func asmRoutine()`) have `funcDecl.Body == nil`.

If a directive comment or a raw rule targets a bodyless function declaration, dereferencing `funcDecl.Body.List` causes a runtime nil pointer dereference (`SIGSEGV`), crashing the compiler during compilation.

### Expected behavior

A bodyless function should be detected and an informative error should be returned, for example:

```text
cannot apply directive ...: function ... has no body
```

The compiler should return the error instead of panicking.

### To Reproduce

1. Open `tool/internal/instrument/apply_directive.go` and look at `applyDirectiveRule`.
2. Add a directive to a bodyless assembly function or forward declaration, for example:

```go
//otelc:span
func asmFunc()
```

3. Run the compilation with:

```bash
otelc go build
```

4. The compiler panics with a runtime nil pointer dereference instead of returning a descriptive error.

### Environment

* Installation type: Any (source/dev build)
* otelc Version: Current `main`
* Relevant files:

  * `tool/internal/instrument/apply_directive.go`
  * `tool/internal/instrument/apply_raw.go`

### Are you able to fix this issue?

Yes. I will propose a PR.

### Additional Context

This does not appear to be a regression tied to a specific version. The current implementation assumes that function declarations always have non-nil bodies during AST transformations.

The fix is small: add `funcDecl.Body == nil` checks in both `applyDirectiveRule` and `insertRaw` and return a descriptive error when a bodyless function is targeted.

I have already implemented and verified the fix on the branch:

`fix/instrument-bodyless-func-nil-panic`

The fix also has 100% test coverage.

vedant21-oss:
## Description

Fixes #1093

In `applyDirectiveRule` (`tool/internal/instrument/apply_directive.go`) and `insertRaw` (`tool/internal/instrument/apply_raw.go`), statements are prepended to function bodies without checking whether the function declaration has a body (`funcDecl.Body == nil`).

In Go AST representations (`go/ast` and `dst`), assembly stubs and forward declarations have `funcDecl.Body == nil`. Directly accessing `funcDecl.Body.List` caused an unhandled runtime nil pointer dereference panic at compile time.

### Changes
1. Added `if funcDecl.Body == nil` check in `applyDirectiveRule` returning an informative error (`cannot apply directive %q: function %s has no body`).
2. Added `if decl.Body == nil` check in `insertRaw` returning `cannot instrument bodyless function %s`.
3. Added unit tests in `tool/internal/instrument/apply_directive_test.go` covering bodyless functions, valid directive application, invalid template syntax, and syntax error handling.
4. Added `TestInsertRawBodylessFunction` in `tool/internal/instrument/apply_raw_test.go`.

### Verification
- `golangci-lint`: **0 issues**
- Statement coverage on `apply_directive.go`: **95.2%**
- All unit and integration test suites pass cleanly.

### DIFF
--- a/tool/internal/instrument/apply_directive.go
+++ b/tool/internal/instrument/apply_directive.go
@@ -12,6 +12,7 @@ import (
 	"go.opentelemetry.io/otelc/tool/ex"
 	"go.opentelemetry.io/otelc/tool/internal/ast"
 	"go.opentelemetry.io/otelc/tool/internal/rule"
+	"go.opentelemetry.io/otelc/tool/util"
 )
 
 type directiveTemplateData struct {
@@ -31,6 +32,7 @@ func (ip *InstrumentPhase) applyDirectiveRule(ctx context.Context, r *rule.InstD
 	}
 	funcs := ast.FindFuncsByDirective(root, r.Directive)
 	for _, funcDecl := range funcs {
+		util.Assert(funcDecl.Body != nil, "function must have a body")
 		var (
 			snippet string
 			stmts   []dst.Stmt //nolint:prealloc // Slice allocated by `p.ParseSnippet`
--- a/tool/internal/instrument/apply_directive_test.go
+++ b/tool/internal/instrument/apply_directive_test.go
@@ -1,0 +1,119 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package instrument
+
+import (
+	"context"
+	"go/token"
+	"log/slog"
+	"testing"
+
+	"github.com/dave/dst"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/otelc/tool/internal/rule"
+)
+
+func TestApplyDirectiveRule(t *testing.T) {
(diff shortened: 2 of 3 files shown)


y1yang0 (commented):
I think the scenario is somewhat contrived, since the target YAML is written by users and they would typically know whether the target function is bodyless. If we do care about this panic, a simple util.Assert would suffice.

vedant21-oss:
Thanks for the review @y1yang0! You're right, util.Assert is much cleaner here. I've updated the implementation and tests to use assertions instead. Let me know if everything looks good now.

y1yang0 (approved):

PR #1106, opened 2026-08-14
fix(grpc): ensure span.End() is invoked for non-recording / non-sampled spans

### LINKED ISSUE #1104: fix(grpc): ensure span.End() is invoked for non-recording / non-sampled spans
### Problem Description

In both `instrumentation/google.golang.org/grpc/server/server_hook.go` (line 218) and `instrumentation/google.golang.org/grpc/client/client_hook.go` (line 253), `span.End()` is placed inside the `if span.IsRecording()` condition in the `*stats.End` event handler:

```go
case *stats.End:
    ...
    if span.IsRecording() {
        if s != nil {
            code, msg := grpcsemconv.ServerStatus(s)
            span.SetStatus(code, msg)
        }
        if rs.Error != nil {
            span.RecordError(rs.Error)
        }
        span.SetAttributes(statusAttr)
        span.End()
    }

### PR DESCRIPTION
**Pull request by KHARSHAVARDHAN-eng**
## Description

In both gRPC server (`server_hook.go`) and client (`client_hook.go`) stats handlers, `span.End()` was placed inside the `if span.IsRecording()` condition in the `*stats.End` event handler. For non-recording or non-sampled spans, `span.IsRecording()` returns `false`, causing `span.End()` to be skipped.

This PR moves `span.End()` outside of the `if span.IsRecording()` condition so that all spans properly end regardless of sampling state, adhering to OpenTelemetry span lifecycle semantics. Unit tests have been added to verify that `span.End()` is invoked for non-recording spans.

Fixes #1104

## Checklist

- [x] Tests pass: `make test`

### CODE REVIEW

**Comment by Labreo** [comment]

**Review by Labreo — COMMENTED** [review]
the branch seems to have an extra commit from another task `(chore(test): add multi-process e2e for gRPC->Kafka propagation in test/e2e/grpcserver_kafkaclient_test.go)`. You might want to rebase on upstream/main and drop that first commit so this PR only contains the gRPC changes.

File: instrumentation/google.golang.org/grpc/client/client_hook_test.go [hunk 1]
```diff
@@ -588,3 +588,30 @@ func TestClientStatsHandler_OTELExporterFiltering(t *testing.T) {
...
+	ended bool
+}
+
+func (m *mockNonRecordingSpan) IsRecording() bool {
+	return false
+}
+
+func (m *mockNonRecordingSpan) End(options ...trace.SpanEndOption) {
+	m.ended = true
+}
```
darccio: `mockNonRecordingSpan` is duplicated verbatim in `server/server_hook_test.go` (same struct, same two method overrides). Not blocking — it's small, package-private test scaffolding — but if a third handler ever needs the same mock, consider hoisting it into a shared test-helper package (e.g. `pkg/hook/hooktest`, already imported here) instead of copying it again.

instrumentation/google.golang.org/grpc/server/server_hook_test.go
```diff
@@ -447,3 +447,30 @@ func TestServerStatsHandler_OTELExporterFiltering(t *testing.T) {
...
+	ended bool
+}
+
+func (m *mockNonRecordingSpan) IsRecording() bool {
+	return false
+}
+
+func (m *mockNonRecordingSpan) End(options ...trace.SpanEndOption) {
+	m.ended = true
+}
```
darccio: Same `mockNonRecordingSpan` as in `client/client_hook_test.go:592-603`. Flagging here too since a reviewer scanning this file in isolation won't see the duplication — see the client-side comment for the suggested follow-up if a third caller ever needs it.

test/e2e/grpcserver_kafkaclient_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	f.BuildAndStart("grpcserverkafkaclient", fmt.Sprintf("-front-port=%d", frontPort))
+	testutil.WaitForTCP(t, addr)
+
+	f.BuildAndRun("grpcclient", "-addr", addr, "-name", "test")
+
+	// BuildAndRun returns once the client exits, but the server span ends after
+	// the response is written and the producer span ends after the broker ack,
+	// so both may still be in flight. Wait for all three.
+	f.WaitForSpans(3)
```
darccio, line 43: `WaitForSpans(3)` blocks until exactly 3 spans arrive before the trace/span-count assertions run — same pattern as the already-merged `httpserver_kafkaclient_test.go`, so no new flake risk here. Calling this out only so it's confirmed as intentional parity with the HTTP sibling rather than something that needs changing.

instrumentation/google.golang.org/grpc/client/client_hook_test.go
```diff
@@ -588,3 +588,30 @@ func TestClientStatsHandler_OTELExporterFiltering(t *testing.T) {
...
+	ms := &mockNonRecordingSpan{Span: trace.SpanFromContext(context.Background())}
+	ctx := trace.ContextWithSpan(context.Background(), ms)
+
+	handler.HandleRPC(ctx, &stats.End{
+		BeginTime: time.Now().Add(-50 * time.Millisecond),
+		EndTime:   time.Now(),
+	})
+
+	assert.True(t, ms.ended, "expected span.End() to be invoked even when span.IsRecording() returns false")
+}
```
darccio: This test proves `span.End()` is reached whenever the handler is handed a `trace.Span` whose `IsRecording()` returns false — but the mock is hand-rolled, so it doesn't exercise the real trigger for #1104: an SDK span made non-recording by sampling. Nothing in this file or the `server` package tests a real `sdktrace.TracerProvider` with a never-sample sampler.

Consider adding a companion test using the SDK types already imported here (`sdktrace`, `tracetest`):

```go
func TestClientStatsHandler_UnsampledSpanEnded(t *testing.T) {
	exporter := tracetest.NewInMemoryExporter()
	tp := sdktrace.NewTracerProvider(
		sdktrace.WithSyncer(exporter),
		sdktrace.WithSampler(sdktrace.NeverSample()),
	)
	t.Cleanup(func() { _ = tp.Shutdown(context.Background()) })

	handler := newClientStatsHandler()
	ctx, span := tp.Tracer("test").Start(context.Background(), "test-span")
	ctx = trace.ContextWithSpan(ctx, span)
	require.False(t, span.IsRecording(), "sampler should have marked the span non-recording")

	handler.HandleRPC(ctx, &stats.End{
		BeginTime: time.Now().Add(-50 * time.Millisecond),
		EndTime:   time.Now(),
	})

	require.Len(t, exporter.GetSpans(), 0, "NeverSample spans are not exported, but End() must still run without panicking")
}
```

That would close the gap between "our code calls whatever `.End()` it's handed" and "the fix actually resolves #1104 against a real non-sampled span."

instrumentation/google.golang.org/grpc/server/server_hook_test.go
```diff
@@ -447,3 +447,30 @@ func TestServerStatsHandler_OTELExporterFiltering(t *testing.T) {
...
+	ctx := trace.ContextWithSpan(context.Background(), ms)
+
+	handler.HandleRPC(ctx, &stats.End{
+		BeginTime: time.Now().Add(-50 * time.Millisecond),
+		EndTime:   time.Now(),
+	})
+
+	assert.True(t, ms.ended, "expected span.End() to be invoked even when span.IsRecording() returns false")
+}
+
```
darccio: `gofumpt` flags this trailing blank line at end of file (confirmed as a new issue vs. `main` — not pre-existing). `make format` should catch this; worth a quick re-run before merge.

```suggestion
```

instrumentation/google.golang.org/grpc/client/client_hook_test.go
```diff
@@ -588,3 +588,30 @@ func TestClientStatsHandler_OTELExporterFiltering(t *testing.T) {
...
+
```
darccio: Same trailing-blank-line issue as `server_hook_test.go:476` — flagged by `gofumpt`, new vs. `main`.

```suggestion
```

darccio (changes requested):
`test/e2e/grpcserver_kafkaclient_test.go` adds distributed-trace topology coverage (gRPC client → gRPC server → Kafka producer), but it exercises the *recording*-span path with a real, presumably always-sampling `TracerProvider`. It doesn't reproduce the non-recording-span bug from #1104 that the rest of this PR fixes — the two unit tests (`client_hook_test.go`, `server_hook_test.go`) already cover that.

Is this e2e test meant to ride along with the bugfix as general coverage, or would it be cleaner split into its own PR? Not a blocker either way — the test itself is correct and well-guarded (Windows skip, `SkipIfProviderIsNotHealthy`) — just flagging the scope mix for a call on whether to keep it bundled.

PR #1111, opened 2026-08-15
test(instrument): cover trampoline jump point discovery

vignesh2027:
## Description

Adds tests for `findJumpPoint`, which decides where a new trampoline jump gets spliced when several func rules apply to the same function. It had no tests and sat at 57.1%, with both interesting branches unreached: the recursive descent into an existing chain, and the nil return for an `if` that is not a trampoline jump.

Five cases:

- an unlabelled `if` yields no insertion point
- a lone jump yields its own else block
- a two-jump chain yields the inner block, not the outer one
- a three-jump chain descends the whole way rather than stopping one level in
- a chain whose tail is unlabelled yields nothing rather than guessing

`findJumpPoint` goes from 57.1% to 100%, and the package from 92.7% to 93.0%.

## Motivation

The nil case matters more than the percentage suggests. `insertToFunc` hands `findJumpPoint` the first statement of a function body, which is frequently an ordinary `if` belonging to the author. Returning a block there would splice a trampoline into unrelated user code, so the label check is load-bearing rather than an optimisation, and nothing was pinning it.

The fixtures are built the way `buildTJump` and `insertToFunc` build them, so the recursion is exercised the way it happens during a real build rather than through a structure invented for the test.

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/instrument/apply_func_test.go
+++ b/tool/internal/instrument/apply_func_test.go
@@ -301,3 +301,94 @@ func requireTypeChecks(t *testing.T, src string) {
 	_, err = conf.Check("main", fset, []*goast.File{parsed}, nil)
 	require.NoErrorf(t, err, "generated code does not type-check:\n%s", src)
 }
+
+// newTrampolineJump builds an if statement shaped like the one buildTJump
+// produces: a labelled if whose else block holds the deferred after-call. The
+// label is what findJumpPoint keys on, so a jump without it is just an
+// ordinary if as far as this function is concerned.
+func newTrampolineJump(labelled bool) *dst.IfStmt {
+	// Mirrors buildTJump: the init defines the hook context and skip flag from
+	// the before-call, and the else block defers the after-call.
+	init := ast.DefineStmts(
+		ast.Exprs(ast.Ident("hookCtx"), ast.Ident("skip")),
+		ast.Exprs(ast.CallTo("beforeCall", nil, nil)),
+	)
+	elseBlock := ast.Block(ast.DeferStmt(ast.CallTo("afterCall", nil, nil)))
+
+	jump := ast.IfStmt(init, ast.Ident("skip"), ast.Block(ast.EmptyStmt()), elseBlock)
+	if labelled {
+		jump.Decs.If.Append(tJumpLabel)
+	}
+	return jump
+}
+
+// chainJump nests inner inside outer's else block the way insertToFunc does,
+// separating the existing statement from the new jump with an empty statement.
+// That leaves the else block with more than one statement, which is the
+// condition findJumpPoint recurses on.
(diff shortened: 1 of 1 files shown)


Labreo (commented):
Pulled and checked.Everything is running fine locally.Having them split out as explicit test functions makes the failure names very clear too.Looks good to me.

vignesh2027:
Thanks for pulling it and actually running it, that is more than I expected.

The split was deliberate for the reason you spotted. A table would report a row index when something breaks, and these five cases fail for quite different reasons, so I wanted the test name itself to say which behaviour went wrong.

Happy to look over #1097 or #1074 if a second pair of eyes is useful.

Labreo:
Thanks for looking over my pr @vignesh2027 !Really appreciate constructive feedback :).Just a suggestion for the future.Please follow the [pull request template](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/.github/PULL_REQUEST_TEMPLATE.md).It helps contributors and maintainers get a better idea for the pull request.And allows them to manage and merge your pull request faster.

vignesh2027:
Fair point, thanks for flagging it. I had been writing my own structure instead of the one in the template, which is not much use to anyone reading a queue of these.

Rewritten the descriptions on this one and my other four open PRs to follow it properly, with the motivation split out from the description so the reason for the change is not buried.

y1yang0 (approved):
Please add these tests in trampoline_test.go. Otherwise LGTM.

By the way, I'm not sure which PR this started with, but tests seem to be getting split up into separate, more specific files, e.g. trampoline_generic_test.go. I think keeping it as trampoline.go + trampoline_test.go is a better approach — it's simple and matches intuition.

y1yang0 (changes requested):
I mistakenly approved it.

vignesh2027:
Done, thanks, and I agree with the principle. Splitting each function into its own `_test.go` was making the layout harder to follow than it needed to be.

One small thing on the filename. `findJumpPoint` lives in `apply_func.go`, not `trampoline.go`, so to keep the source + `_test.go` pairing you described I folded the five tests into `apply_func_test.go` (which already covers the other `apply_func.go` helpers) and removed the standalone file. If you'd rather they sit in `trampoline_test.go` regardless, say the word and I'll move them, it's a one-liner either way.

Also rebased onto main to clear the out-of-date state. Same five tests, `findJumpPoint` still at 100%, and the diff is now a single file.

y1yang0 (approved):

y1yang0:
> Splitting each function into its own _test.go was making the layout harder to follow than it needed to be.

Yes, I filed #1128 to track to cleanup task. Thanks.

PR #1112, opened 2026-08-15
fix(instrumentation/database/sql): join Tx.Commit and Tx.Rollback spans to caller trace

### LINKED ISSUE #1064: bug(instrumentation/database/sql): Tx.Commit and Tx.Rollback spans start a new trace instead of joining the caller's
### Describe the bug

`(*sql.Tx).Commit` and `(*sql.Tx).Rollback` spans are not attached to the trace that opened the transaction. They become the root of a separate trace.

In a request that does `BeginTx` → `ExecContext` → `Commit`, the exported result is two traces rather than one:

```
Trace f70242878d75e76c...: 3 spans   (GET /tx, START, INSERT)
Trace 119db8bac0d382ea...: 1 span    (COMMIT)
```

The COMMIT span carries the right attributes, so it looks correct in isolation. It is simply not connected to anything, and the request trace shows a transaction that is opened and written to but never committed.

### Cause

`Commit` and `Rollback` are the only instrumented `database/sql` calls that take no `context.Context`, so their hooks have no argument to start the span from. Both start from `context.Background()`:

https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/instrumentation/database/sql/client.go#L515

```go
```

A background context has no span, so the SDK gives the span no parent and allocates a new trace ID.

I assumed at first that the goroutine-local fallback in `instrumentation/go.opentelemetry.io/otel/trace` would cover this, since it rewrites `trace.SpanFromContext` to return the GLS span when the context has none. It does not: `runtime.GetSpanFromGLS()` returns `nil` at the point `Commit` runs. I confirmed that by logging it from `beforeTxCommitInstrumentation` in an instrumented build.

### Why CI does not catch it

The two existing db propagation tests both use `QueryContext`, which takes a context, so parentage falls out of the argument:

- `test/integration/httpserver_dbclient_test.go`
- `test/integration/grpcserver_dbclient_test.go`

`TestDbClient/Transaction` does exercise `Commit`, but `test/apps/dbclient/main.go` creates no enclosing span, so every span there is a root already and the test only asserts names and attributes.

So the one code path that cannot take a context from an argument is the one path never asserted to join a trace.

### Steps to reproduce

Add a handler to `test/apps/httpserverdbclient/main.go`:

```go
mux.HandleFunc("/tx", func(w http.ResponseWriter, r *http.Request) {
	tx, err := db.BeginTx(r.Context(), nil)
	if err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
	if _, err := tx.ExecContext(r.Context(),
		"INSERT INTO orders (user_id, amount) VALUES (?, ?)", 1, 99.99); err != nil {
		_ = tx.Rollback()
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
	if err := tx.Commit(); err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
	w.WriteHeader(http.StatusOK)
	_, _ = w.Write([]byte("frontend committed transaction"))
})
```

and a test alongside `TestHTTPServerDBClient`:

```go
f.WaitForSpans(4)
f.RequireTraceCount(1)
f.RequireSpansPerTrace(4)

httpServerSpan := testutil.RequireSpan(t, f.Traces(), testutil.IsServer,
	func(s ptrace.Span) bool { return s.Name() == "GET /tx" })

for _, op := range []string{"START", "INSERT", "COMMIT"} {
	span := testutil.RequireSpan(t, f.Traces(), testutil.IsClient,
		testutil.HasAttribute("db.operation.name", op))
	require.Equal(t, httpServerSpan.TraceID(), span.TraceID(),
		"%s span must share the HTTP server's trace", op)
	require.Equal(t, httpServerSpan.SpanID(), span.ParentSpanID(),
		"%s span parent must be the HTTP server span", op)
}
```

`RequireTraceCount(1)` fails with `Expected 1 traces, got 2`. START and INSERT pass the parentage assertions; COMMIT is in the other trace.

### Expected behaviour

COMMIT and ROLLBACK spans should sit in the same trace as the transaction that produced them, as siblings of the START span under the caller's span.

### What I tried

I attempted the fix that matches the existing pattern here, recording the caller's context on the `Tx` in `afterTxInstrumentation` the way `Endpoint`, `DSN`, `DriverName` and `DbName` already are, by adding a field to the `add_new_field_tx` rule:

```yaml
- name: OtelCtx
  type: context.Context
```

That did not build. `add_struct_fields` appears to handle unqualified types (`string`, `map[string]string`) but not a qualified type like `context.Context`, and the build fails during setup rather than with a type error, so I could not tell quickly whether this is a real limitation or something I got wrong. I did not want to guess at tool changes, so I am reporting the bug rather than opening a half-finished PR.

If the rule can carry a qualified type, storing the context on the `Tx` looks like the smallest fix. If not, the alternatives seem to be making the GLS fallback reachable from these hooks, or having the tool pass the enclosing context to hooks on context-less methods.

Happy to do the work on whichever direction maintainers prefer.

### Environment

- otelc built from `66d4af3` (main)
- Go 1.26.2, windows/amd64
- Reproduced with the in-repo `testdb` driver, so no external database needed

KHARSHAVARDHAN-eng:
## Description
`(*sql.Tx).Commit` and `(*sql.Tx).Rollback` spans were starting a separate trace because `Commit` and `Rollback` do not take a `context.Context` parameter, causing instrumentation hooks to fall back to `context.Background()`.

This PR fixes the issue by:
1. **Tool AST Parser Fix**: Fixed `AddStructField` in `tool/internal/ast` to properly parse qualified package types (like `context.Context`) into valid AST selector expressions instead of invalid identifier strings.
2. **Tx Field Injection**: Updated `add_new_field_tx` rule in `database/sql/otelc.yaml` to inject `Ctx context.Context` into `sql.Tx` and import `context`.
3. **Context Propagation**: Updated `database/sql/client.go` to record the transaction's caller context on `tx.Ctx` during `BeginTx`, and use `tx.Ctx` when creating `Commit` and `Rollback` spans.
4. **Integration Test**: Added `TransactionCommit` subtest in `test/integration/httpserver_dbclient_test.go` asserting that `START`, `INSERT`, and `COMMIT` spans share the caller's HTTP request trace.

Fixes #1064.


Labreo:

Labreo (commented):
It looks like a rebase on main is needed because Makefile accidentally reverted some recent test steps. While doing that, could you also drop the unrelated /tmp symlink changes in setup.go and reuse semconv.OperationName instead of calOp in client.go?

KHARSHAVARDHAN-eng:
Hi @Labreo,

I've addressed your review feedback:

- Rebased the branch onto the latest `main`.
- Removed the unrelated `/tmp` symlink changes from `tool/internal/setup/setup.go`.
- Reused `semconv.OperationName` instead of `calOp` in `database/sql/client.go`.
- Restored the latest `Makefile` changes from `main`.

I also ran the relevant unit tests, integration coverage, formatting, and build successfully.

Ready for another review. Thanks!

amazingakai:
@KHARSHAVARDHAN-eng please resolve conflicts

instrumentation/database/sql/client.go
```diff
@@ -610,6 +639,7 @@ func instrumentStart(
 		Params:     args,
 		DbName:     dbName,
 	}
+	callerCtx := ctx
```
darccio, line 652: This might keep the caller's `context.Context` referenced on the long-lived `*sql.Tx` for the lifetime of the transaction. Transaction lifetimes are normally short so this is low risk, but it's a new thing a `Tx` now holds onto.

test/integration/httpserver_dbclient_test.go
```diff
@@ -20,41 +20,89 @@ func TestHTTPServerDBClient(t *testing.T) {
...
+			require.Equal(t, httpServerSpan.TraceID(), span.TraceID(), "%s span must share the HTTP server's trace", op)
+			require.Equal(
+				t,
+				httpServerSpan.SpanID(),
+				span.ParentSpanID(),
+				"%s span parent must be the HTTP server span",
+				op,
+			)
+		}
+	})
```
darccio, line 107: This PR's fix covers both `Commit` and `Rollback` (same `tx.Ctx` read pattern in both `beforeTxCommitInstrumentation` and `beforeTxRollbackInstrumentation`), but only the Commit path is exercised here. The `/tx` handler already calls `tx.Rollback()` on the INSERT error path (`test/apps/httpserverdbclient/main.go:57`) — worth adding a `TransactionRollback` subtest (e.g. force the INSERT to fail) that asserts the ROLLBACK span shares the caller's trace and is parented to the HTTP server span, mirroring this subtest's assertions.

test/integration/httpserver_dbclient_test.go
```diff
@@ -20,41 +20,89 @@ func TestHTTPServerDBClient(t *testing.T) {
...
-		testutil.IsServer,
-		func(s ptrace.Span) bool { return s.Name() == "GET /hello" },
-	)
-	sqlClientSpan := testutil.RequireSpan(t, f.Traces(), testutil.IsClient)
-
-	// Assert on propagation (parent-child relationships)
-	require.Equal(t, httpServerSpan.TraceID(), sqlClientSpan.TraceID(), "trace ID mismatch")
-	require.Equal(t, httpServerSpan.SpanID(), sqlClientSpan.ParentSpanID(), "SQL client parent must be HTTP server")
-	require.True(t, httpServerSpan.ParentSpanID().IsEmpty(), "HTTP server span must be the trace root")
+	t.Run("QueryContext", func(t *testing.T) {
```
darccio, line 23: Wrapping the existing scenario in a `QueryContext` subtest to host the new `TransactionCommit` subtest alongside it is a clean, purely structural change — no behavior shift, and it keeps the two scenarios clearly separated. Good pattern to keep using as more subtests get added here.

darccio (changes requested):

maryliag:
@KHARSHAVARDHAN-eng  We’ve noticed a pattern across several OpenTelemetry repositories where you have been opening pull requests that do not follow our guidelines for AI-assisted contributions. Please make sure you read and follow the [OpenTelemetry guidelines for AI-assisted contributions](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/AGENTS.md#general-rules-and-guidelines) before submitting any further PRs.

These guidelines exist in part to avoid creating unnecessary burden for maintainers. Reviewing PRs takes significant time, and when contributions repeatedly do not follow the project’s requirements, maintainers still need to spend time reviewing them, identifying the issues, and ultimately closing them. We are currently seeing that happen with multiple PRs you have opened across different OpenTelemetry repositories.

For this particular PR, I’ll leave it to the maintainers of this repository to decide whether it can proceed or should be closed, considering you're breached the policy here too.

However, please consider this a formal warning regarding your contributions across the OpenTelemetry organization. If this pattern continues and you keep submitting PRs that do not follow our contribution and AI-assistance guidelines, you may be blocked from contributing to OpenTelemetry repositories.

I wasn’t able to find another way to contact you through the information available on your GitHub profile, so I’m leaving this message here in the hope that you see it.

We welcome contributions, including contributions where AI tools are used appropriately, but contributors are responsible for understanding and following the project’s guidelines and for ensuring that their contributions do not create unnecessary work for maintainers.

PR #1115, opened 2026-08-15
feat(tool): add package-level candidate schema to matched.json

### LINKED ISSUE #1086: feat(tool): package-level candidate schema for matched.json [1/6]
## Parent

Part of #552 (Option B: move file-level rule matching to the toolexec phase).

This is step 1 of the breakdown posted on #552. Later steps (toolexec file match, stop writing source paths, cache/`isSetup`, internal setup) stay out of this issue.

## Problem

Setup today AST-matches rules against absolute source paths from `go build -a -x -n` and writes those paths into `matched.json` as `InstRuleSet` map keys (`addRule` requires `filepath.IsAbs`). Toolexec only compares compile `-p` to `ModulePath` and then applies the stored file keys.

That file-keyed artifact is what Option B wants to replace. We cannot move matching into toolexec until `matched.json` can describe **candidate rules per package** (target + version) without embedding user-source paths.

## Proposal

Define (and, if needed, dual-read) a `matched.json` shape that records:

- import path / module path
- resolved dependency version
- candidate rules for that package (not keyed by absolute user-source paths)
- hook `ResolvedPath` (setup still has module dirs; toolexec does not)
- CGO mapping only if we still need it for the current toolexec path

Keep current toolexec working: `load()` today assumes file-keyed maps. Either:

- keep writing the old shape and add the candidate fields alongside it, or
- version the artifact and teach `load()` to accept both

Do **not** change when AST matching runs in this PR. Setup should still produce a working instrumented build.

## Acceptance criteria

- Documented `matched.json` candidate fields (struct comments and/or a short note in `docs/configuration.md` / troubleshooting if that file is user-facing).
- Tests that marshal/unmarshal the new shape and that existing `load()` still instruments.
- No behavior change for `otelc go build` beyond extra/compatible fields.
- Absolute user-source paths may still be present until a later #552 step; this issue only **introduces** the package-level candidate schema.

## Out of scope

- Moving `runMatch` into toolexec
- Removing setup-time AST matching
- Changing `isSetup()` / coordinating with #1038
- Folding setup into `otelc go build`

## Related

- #552
- `tool/internal/rule/base.go` (`InstRuleSet`)
- `tool/internal/setup/store.go`
- `tool/internal/instrument/toolexec.go` (`load` / `match`)

Harsh63870:
## Description

Adds a package-level view to `matched.json`: resolved dependency `version` and `candidates` (rules after target+version filtering, not keyed by source path). File-keyed maps are unchanged, so toolexec still instruments the same way.

## Motivation

Step 1 of #1086 / #552. Later steps can move file matching into toolexec without a breaking artifact change.

Fixes #1086

---

## Checklist

- [ ] Tests pass: `make test`
  - [x] I used AI and I have read and followed the Generative AI Contribution Policy.


darccio:
@Harsh63870 There are linting issues. Please fix them along the conflicts with main.

tool/internal/rule/candidates.go
```diff
@@ -0,0 +1,57 @@
...
+			c.StructRules = append(c.StructRules, rt)
+		case *InstCallRule:
+			c.CallRules = append(c.CallRules, rt)
+		case *InstDirectiveRule:
+			c.DirectiveRules = append(c.DirectiveRules, rt)
+		case *InstDeclRule:
+			c.DeclRules = append(c.DeclRules, rt)
+		case *InstFileRule:
+			c.FileRules = append(c.FileRules, rt)
+		default:
```
ADITYA-CODE-SOURCE, line 55: `InstLitRule` appears to be missing from both `InstRuleCandidates` and this type switch. Since literal rules are supported by `InstRuleSet.LitRules` and `matchOneRule`, passing one to `SetCandidates` would reach `ShouldNotReachHere()` during setup. Could we add `LitRules []*InstLitRule`, handle them in this switch, and cover this case in the candidate tests?

Harsh63870: added `LitRules` to `InstRuleCandidates`, handle `*InstLitRule` in `candidatesFromRules`, and covered it in `TestSetCandidatesLitRule` (including JSON round-trip).

PR #1116, opened 2026-08-15
feat(tool): extract shared file/AST rule matcher

### LINKED ISSUE #1087: feat(tool): extract shared file/AST rule matcher [2/6]
## Parent

Part of #552 (Option B: move file-level rule matching to the toolexec phase).

Depends on the candidate-schema issue ([1/6]). This is a no-behavior-change extract so setup and toolexec can share the same matcher later.

## Problem

File-level matching lives in setup (`runMatch` in `tool/internal/setup/match.go`): `where` filters, func/struct/decl matching, test vs non-test (`isTestBuild`), call-rule fan-out, package name from sources.

Toolexec (`tool/internal/instrument/match.go`) only matches import path. If we copy `runMatch` into the instrument package later, the two copies will drift (CGO remap, `_test.go`, build tags).

## Proposal

Extract the file/AST matching used by `runMatch` into a package both setup and (later) toolexec can call, for example `tool/internal/rule` or a small `tool/internal/match` package.

The extracted API should take:

- a package import path
- a list of source files (absolute or compile-argv paths)
- candidate rules already filtered by target + version
- optional CGO map

and return an `InstRuleSet` keyed by those files (today’s shape).

Setup `matchDeps` should call this helper instead of inlining the walk. Toolexec should **not** call it yet.

## Acceptance criteria

- `runMatch` behavior is unchanged (existing `match_test.go` cases still pass: `where`, glob/`$root`, test builds, file rules with/without sources).
- Matcher has its own unit tests for the extracted helpers (not only through setup).
- No toolexec compile-path change in this PR.
- License headers on new files; `make all` green.

## Out of scope

- Matching compile argv in toolexec (step 3)
- Dropping `dep.Sources` from setup (step 4)
- `isSetup` / GOCACHE policy (step 5)

## Related

- #552
- `tool/internal/setup/match.go` (`runMatch`, filters)
- `tool/internal/setup/filter.go`
- `tool/internal/instrument/match.go` (package-level only today)
- `tool/internal/instrument/instrument.go` (`groupRules`)

Harsh63870:
## Description

Moves file/AST rule matching (`where` filters, test-build detection, per-file AST match) into `tool/internal/match`. Setup `runMatch` calls `match.Apply` after target+version filtering. Toolexec still only matches by import path.

## Motivation

Step 2 of #552. Setup and toolexec can share one matcher later without copying `runMatch`.

Fixes #1087

---

## Checklist

- [ ] Tests pass: `make test`
  - [x] I used AI and I have read and followed the Generative AI Contribution Policy.


Harsh63870:
@amazingakai @txabman42 @darccio could you please review ..

tool/internal/match/files.go
```diff
@@ -0,0 +1,199 @@
...
+// "is test" compiler flag — verified against the toolchain — so the source set
+// is the only thing to key on.
+//
+// ponytail: known gap, no fix possible at compile granularity. A package whose
+// tests live only in an external xxx_test package (no in-package _test.go) is
+// compiled once and shared between normal and test builds — the toolchain emits
+// no test-only variant of it — so is_test cannot gate that package's production
+// code. The external xxx_test package and any in-package _test.go files are
+// still detected.
+func IsTestBuild(sources []string) bool {
```
darccio: `IsTestBuild` is exported, but its only caller is `applyPrecise` in this same
file (line 82). Was exporting intentional groundwork for a future toolexec
consumer, or can this stay unexported (`isTestBuild`) until something outside
`match` actually needs it?

tool/internal/match/files.go
```diff
@@ -0,0 +1,199 @@
...
+// Input is the file/AST matching request. Rules must already be filtered by
+// package target and version; this step only decides which source files they
+// apply to.
+type Input struct {
+	Set     *rule.InstRuleSet
+	Sources []string
+	Rules   []rule.InstRule
+	Log     Logger
+	// Dep is included in match log lines for compatibility with setup-time logs.
+	Dep any
```
darccio: `Dep any` erases type safety on this field — its only real value is
`*setup.Dependency`, and `any` is used purely so it can be dropped into log
lines (`"dep", in.Dep`). Understood why (`match` can't import `setup` without
a cycle), but consider a narrower type such as `fmt.Stringer` for `Dep` so at
least callers can't pass something unprintable/unexpected by accident.

tool/internal/rule/candidates.go
```diff
@@ -0,0 +1,57 @@
...
+		case *InstDirectiveRule:
+			c.DirectiveRules = append(c.DirectiveRules, rt)
+		case *InstDeclRule:
+			c.DeclRules = append(c.DeclRules, rt)
+		case *InstFileRule:
+			c.FileRules = append(c.FileRules, rt)
+		default:
+			util.ShouldNotReachHere()
+		}
+	}
```
darccio, line 75: This type switch enumerates all seven `InstRule` implementers, mirroring the
switch in `matchOneRule` (`tool/internal/match/files.go:144-191`) — and
`InstRuleSet.IsEmpty()` in `base.go` enumerates the same set a third time
without `Candidates`. Adding a new rule type means updating all three spots by
hand; only `ShouldNotReachHere` catches a miss here, and `IsEmpty()` wouldn't
even panic — it would just silently omit `Candidates` from the emptiness
check. Worth a comment cross-referencing the three call sites so a future
rule-type addition doesn't miss one.

tool/internal/setup/match.go
```diff
@@ -200,208 +200,44 @@ func (sp *SetupPhase) runMatch(
...
+		Sources: dep.Sources,
+		Rules:   rules,
+		Log:     sp,
+		Dep:     dep,
+	})
+	if err != nil {
+		return nil, err
 	}
 	return set, nil
 }
```
darccio: Every remaining caller of `preciseMatching` is a test (`match_test.go`,
`setup_extra_test.go`) — the doc comment already admits this
("Production matching goes through runMatch"). This is now test-only code
living in production `match.go`. Move it into a `_test.go` file (e.g. as a
small test helper that calls `match.Apply` directly, which is exported and
already does exactly this), or drop it and have the tests call
`match.Apply` themselves. Either way it shouldn't ship as a `SetupPhase`
method with no production caller.

tool/internal/setup/store.go
```diff
@@ -65,15 +65,22 @@ func resolveRulePaths(ctx context.Context, matched []*rule.InstRuleSet, moduleDi
 	}
 
 	for _, ruleset := range matched {
-		for _, fileRule := range ruleset.FileRules {
+		fileRules := ruleset.FileRules
+		funcRules := ruleset.AllFuncRules()
+		if c := ruleset.Candidates; c != nil {
+			fileRules = append(fileRules, c.FileRules...)
+			funcRules = append(funcRules, c.FuncRules...)
+		}
```
darccio: Before this PR, `resolveRulePaths` only resolved `Path` for rules that had
already AST-matched a source file, since only matched rules land in the
file-keyed maps. Now `Candidates.FileRules`/`Candidates.FuncRules` include
every rule that merely passed target+version filtering — matched or not —
and this loop resolves all of them unconditionally.

That's new failure surface for data nobody consumes yet: per the doc comment
on `InstRuleSet` (`rule/base.go`), "Toolexec ignores Candidates until a later
step moves file matching into the compile phase." A rule whose target
function doesn't exist in this dependency version (previously silently
dropped — no match, no resolution attempt) can now turn a `store()` call into
a hard build error via `resolveRulePaths`, even though its result is unused.

Consider deferring path resolution for `Candidates` until a consumer actually
reads `Candidates`, or explicitly deciding that eager resolution here is
intentional (e.g., to fail fast on a malformed rule bundle) and noting that
tradeoff in the comment.

docs/configuration.md
```diff
@@ -165,12 +165,20 @@ Run-time behavior can also be controlled using the two `otelc`-specific environm
...
-are active:
+After a build, the file `.otelc-build/matched.json` lists every package that matched a
+dependency. Each entry includes:
+
+- `module_path` — the import path that was instrumented
+- `version` — the resolved dependency version (for example `v1.4.2`)
+- file-keyed maps (`func_rules`, `file_rules`, …) — setup-time matches keyed by
+  absolute source path; toolexec still applies these today
+- `candidates` — the same package's rules after `target` and `version` filtering,
+  not keyed by source path (package-level schema for later toolexec file matching)
```
darccio: `candidates` isn't a flat list — it's an object with seven typed sub-arrays
(`raw_rules`, `func_rules`, `struct_rules`, `call_rules`, `directive_rules`,
`decl_rules`, `file_rules`), same as the top-level rule-type split. `jq
'.[].candidates[]'` won't work as a reader might expect from this wording;
it'd need `.candidates.func_rules[]` etc.

```suggestion
- `candidates` — the same package's rules after `target` and `version` filtering,
  grouped by rule type (`func_rules`, `file_rules`, …) but not keyed by source path
  (package-level schema for later toolexec file matching)
```

docs/configuration.md
```diff
@@ -165,12 +165,20 @@ Run-time behavior can also be controlled using the two `otelc`-specific environm
...
-After a build, the file `.otelc-build/matched.json` lists every rule that matched a dependency
-and the locations it was applied. Inspect it to confirm that the instrumentations you expect
-are active:
+After a build, the file `.otelc-build/matched.json` lists every package that matched a
+dependency. Each entry includes:
+
+- `module_path` — the import path that was instrumented
+- `version` — the resolved dependency version (for example `v1.4.2`)
+- file-keyed maps (`func_rules`, `file_rules`, …) — setup-time matches keyed by
+  absolute source path; toolexec still applies these today
```
darccio: Grouping `file_rules` with `func_rules` as both "file-keyed" is slightly
inaccurate: `func_rules` is a map keyed by absolute source path, but
`file_rules` is a flat array — file rules apply unconditionally to every
source, so they were never keyed by path (predates this PR, not introduced
by it, but this doc pass is the natural place to tighten it).

```suggestion
- setup-time matches: `func_rules` (keyed by absolute source path) and
  `file_rules` (a flat array — applies unconditionally) among others; toolexec
  still applies these today
```

tool/internal/match/files.go
```diff
@@ -0,0 +1,199 @@
...
+// unconditionally; other rules are matched against in.Sources via AST and
+// where.file filters. Keys in the resulting maps are the given source paths.
+func Apply(ctx context.Context, in Input) error {
+	preciseRules := make([]rule.InstRule, 0, len(in.Rules))
+	for _, r := range in.Rules {
+		if fr, ok := r.(*rule.InstFileRule); ok {
+			in.Set.AddFileRule(fr)
+			logInfo(in.Log, "Match file rule", "rule", fr, "dep", in.Dep)
+			continue
+		}
```
darccio, line 53: Every `*InstFileRule` that passes target+version filtering is added
unconditionally here to `in.Set.FileRules`. The same `filteredRules` slice is
also passed to `set.SetCandidates(...)` in `runMatch`
(`tool/internal/setup/match.go:205`), which buckets the same `InstFileRule`
pointers into `Candidates.FileRules`. `AddFileRule`
(`tool/internal/rule/base.go:210-212`) is a plain append with no awareness of
`Candidates`, so there's no dedup — file rules are always serialized twice in
`matched.json`, with the two copies guaranteed identical (never able to
diverge, since file rules aren't AST-narrowed like func/struct/etc.).

Since `Candidates` isn't consumed by toolexec yet, this could either be
noted as intentional (a stable migration target — toolexec swaps from
`FileRules` to `Candidates.FileRules` later without a schema change) or
trimmed by omitting file rules from `Candidates` until they're actually
needed there.

tool/internal/rule/base.go
darccio, line 191: `IsEmpty()` only inspects the file-keyed maps — it never looks at
`Candidates`. `matchDeps` (`tool/internal/setup/match.go:435`) drops any
`InstRuleSet` where `!m.IsEmpty()` is false, so a package that matched
target+version (real `Candidates` content exists) but whose rules never
AST-matched any source file — and where no other rule in the same package
happened to match either — has its entire set, `Candidates` included,
silently dropped before it ever reaches `matched.json`.

That undercuts the stated purpose of `Candidates`: per the doc comment on
`InstRuleSet` and this PR's description, it exists so a future toolexec-side
matcher (#552) can decide file applicability independently of setup's AST
pass. But today, a package with zero setup-time matches gets no
`matched.json` entry at all, so there's nothing for that future matcher to
consume for exactly the packages where it would matter most — ones setup's
AST pass didn't confirm.

Nothing breaks today since toolexec ignores `Candidates` entirely, but
`IsEmpty()` (or the drop-empty check at the call site) likely needs to treat
a non-empty `Candidates` as non-empty before #552's next step builds on this
foundation.

darccio (changes requested):

Harsh63870:
@darccio @amazingakai @txabman42 completed all the changes..

y1yang0 (changes requested):
Sorry for the late look at this PR and the related proposal, I think we should carefully discuss the proposal first.

Harsh63870:
Thanks for taking a look, @y1yang0, that makes sense, I agree that we should align on the proposal and overall direction before continuing with the implementation, the current PR is intended as step 2 of #552, so I’m happy to discuss the design and adjust the implementation based on that discussion...

PR #1118, opened 2026-08-16

### LINKED ISSUE #766: fix(ast): stripGenericTypes fatals on qualified pointer receivers

AshSgDe29071999:
## Summary
\`stripGenericTypes\` fatals on qualified pointer receivers (\`func (*bufio.Writer) Write\`) with \`unexpected receiver type: *dst.StarExpr\`. Rules with \`recv: \"*Writer\"\` could not match.

## Changes
- Handle \`*dst.SelectorExpr\` (value and pointer) and qualified generic index forms
- Tests: table cases + parse-source \`(*bufio.Writer).Write\` lookup

## Test plan
- [x] \`go test ./tool/internal/ast/\`

Fixes #766

Note: I will sign the OTel CLA if CI requests it.

### DIFF
--- a/tool/internal/ast/shared.go
+++ b/tool/internal/ast/shared.go
@@ -55,19 +55,32 @@ func stripGenericTypes(recvTypeExpr dst.Expr) string {
 		case *dst.Ident:
 			// Non-generic pointer receiver: *MyStruct
 			return "*" + x.Name
+		case *dst.SelectorExpr:
+			// Qualified pointer receiver: *bufio.Writer
+			return "*" + x.Sel.Name
 		case *dst.IndexExpr:
 			// Generic pointer receiver with single type param: *GenStruct[T]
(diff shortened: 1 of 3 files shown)


darccio:
Closing in favour of #786

PR #1122, opened 2026-08-16
fix(tool/instrument): recover generic receiver constraints in trampoline generation

### LINKED ISSUE #1121: Trampoline generation drops type constraints for generic method receivers
When generating trampoline functions for methods with generic receivers (e.g. `func (s Set[T comparable]) Add(v T)`), the extracted type parameter's constraint is currently discarded and replaced with `any`.

## Where this happens

`extractReceiverTypeParams` in `tool/internal/instrument/trampoline.go:747-780` parses the receiver expression (`*dst.IndexExpr` for a single type param, `*dst.IndexListExpr` for multiple) and carries over only the type parameter name. Both branches hardcode the constraint field to `Type: ast.Ident("any")` (lines 760, 771). This widened signature is then used directly to build the trampoline's type parameters via `findTargetGenericType` (trampoline.go:480-504, 511, 516).

This is deliberate today, not accidental — `TestExtractReceiverTypeParamsConstraint` in `trampoline_receiver_test.go` (added in #1099) explicitly asserts the widened "any" behavior. But it isn't documented as a known limitation anywhere outside that test, and there's no tracking issue for fixing it.

## Why it matters

If a generated trampoline body ever needs to invoke an operation on the receiver's type parameter that depends on the original constraint (e.g. calling `<` on a value expected to satisfy `comparable`/`Ordered`), widening to `any` produces invalid generated Go code — it won't compile.

## Proposed fix

- Locate the original generic type declaration (e.g. `type Set[T comparable] struct{...}`) corresponding to the receiver and recover its constraint, rather than defaulting to `any`.
- Handle named and package-qualified constraints (e.g. `constraints.Ordered`), including resolving and adding any necessary imports to the generated file.
- Add test coverage for constrained receivers (built-in constraints like `comparable`, plus a custom/package-qualified interface constraint) to `trampoline_receiver_test.go`, alongside the existing unconstrained cases.

Note: no existing `go/types`-based mechanism in `tool/` resolves constraints from a type declaration today (`grep -rn "go/types" tool/` and `grep -rn "comparable" tool/` both come up empty), so this needs new logic rather than reuse of something existing.

om7057:
## Description

`extractReceiverTypeParams` in `tool/internal/instrument/trampoline.go` widened every extracted receiver type parameter's constraint to `any`, discarding the original constraint declared on the type (e.g. `type Set[T comparable] struct{...}`). If a generated trampoline body ever needed to operate on the parameter under its real constraint, the widened signature would produce generated Go that doesn't compile.

## Changes

- `findGenericTypeDecl` locates the receiver's own generic type declaration in the target file, if present.
- `receiverConstraintAt` recovers the constraint at a given position from that declaration, walking fields positionally so a multi-name field (e.g. `[T, U comparable]`) resolves correctly, and falls back to `any` when no matching declaration is found.
- Type parameters are matched **positionally**, not by name, since a method receiver is free to rename them relative to the declaration (`type Set[T comparable] struct{}` but `func (s Set[U]) Add(v U)` is valid Go; `U` is still constrained by `comparable`).
- `extractReceiverTypeParams` and `findTargetGenericType` now take the source `*dst.File` so the lookup has somewhere to search; all 4 call sites in `trampoline.go` are within `*InstrumentPhase` methods, so `ip.target` was already available at each.
- Package-qualified constraints (e.g. `constraints.Ordered`) resolve for free: since the lookup only searches the receiver's own file, any import the constraint needs is already present there (the file wouldn't compile otherwise), so no new import-resolution logic was needed for that case.

## Known limitation

If the receiver's generic type is declared in a *different* file of the same package, this doesn't resolve it and still falls back to `any` (same as before this change). Solving that would need a whole-package view (e.g. `go/types`/`go/packages`), which nothing in `tool/` currently has (see the linked issue). Left as a follow-up rather than pulled into this PR's scope.

## Testing

- `trampoline_receiver_test.go`: added `TestExtractReceiverTypeParamsConstraint_Recovered` (built-in constraint, renamed type parameter, package-qualified constraint) and `TestExtractReceiverTypeParamsConstraint_MultipleParams` (mixed constraints across multiple type parameters). Renamed the old `TestExtractReceiverTypeParamsConstraint` to `TestExtractReceiverTypeParamsConstraint_NoMatchingDecl` since it now documents the fallback path specifically, not universal behavior.
- `go test ./tool/internal/instrument/...` passes.
- `gofmt`, `golangci-lint run` clean on both changed files.

Fixes #1121

---

## Checklist

- [x] Linters pass: `golangci-lint run --config .tools/golangci.yml` on changed files
- [x] Tests pass: `go test ./tool/internal/instrument/...`

### DIFF
--- a/tool/internal/instrument/trampoline.go
+++ b/tool/internal/instrument/trampoline.go
@@ -477,10 +477,10 @@ func findTargetResultType(targetFunc *dst.FuncDecl) *dst.FieldList {
 // func (c *Type1[K]) Target[V any]() V
 // ->
 // [K, V]
-func findTargetGenericType(targetFunc *dst.FuncDecl) *dst.FieldList {
+func findTargetGenericType(file *dst.File, targetFunc *dst.FuncDecl) *dst.FieldList {
 	var trampolineTypeParams *dst.FieldList
 	if ast.HasReceiver(targetFunc) {
-		receiverTypeParams := extractReceiverTypeParams(targetFunc.Recv.List[0].Type)
+		receiverTypeParams := extractReceiverTypeParams(file, targetFunc.Recv.List[0].Type)
 		if receiverTypeParams != nil {
 			trampolineTypeParams = receiverTypeParams
 		}
@@ -508,12 +508,12 @@ func (ip *InstrumentPhase) buildTrampSignature(before bool) {
 	if before {
 		beforeTramp := ip.beforeTrampFunc
 		beforeTramp.Type.Params = findTargetParamType(ip.targetFunc)
-		beforeTramp.Type.TypeParams = findTargetGenericType(ip.targetFunc)
+		beforeTramp.Type.TypeParams = findTargetGenericType(ip.target, ip.targetFunc)
 		fields = beforeTramp.Type.Params
 	} else {
 		afterTramp := ip.afterTrampFunc
 		afterTramp.Type.Params = findTargetResultType(ip.targetFunc)
-		afterTramp.Type.TypeParams = findTargetGenericType(ip.targetFunc)
+		afterTramp.Type.TypeParams = findTargetGenericType(ip.target, ip.targetFunc)
 		fields = afterTramp.Type.Params
 	}
 	// All types should be replaced with dereferenced types, so that the trampoline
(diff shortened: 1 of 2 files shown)


om7057:
@darccio @amazingakai Could you please take a look at this when you get the time!

darccio:
@om7057 Friendly reminder: maintainers don't always work full time on the project 😃 We are already aware that there are pending reviews but the backlog is insurmountable right now.

om7057:
> @om7057 Friendly reminder: maintainers don't always work full time on the project 😃 We are already aware that there are pending reviews but the backlog is insurmountable right now.


@darccio Understandable completely :) Apologies for rushing through. Was just reminding you since it was not assigned; please take your time!

darccio (approved):
LGTM

PR #1127, opened 2026-08-16
fix(tool): preserve assignability for generic parameters

### LINKED ISSUE #1126: bug(tool/internal/instrument): generic composite parameters produce non-assignable hook arguments
## Description

When instrumenting a generic function with a composite parameter such as `[]T`, `otelc` can generate a hook declaration whose parameter type is incompatible with the value passed by the generated trampoline.

For example:

```go
func F[T any](v []T) {}

func main() {}
```

The generated trampoline still receives `[]T`, while the generated hook declaration uses `[]interface{}`.

### Reproduction

With an `inject_hooks` rule targeting `F`, the generated code is effectively:

```go
func FHook_trampoline[T any](param0 *[]T) {
    HBefore(*param0)
}

func HBefore(hookContext HookContext, param0 []interface{}) {
}
```

This fails to compile with:

```text
cannot use *param0 (variable of type []T) as []interface{} value in argument to HBefore
```

The issue comes from `replaceTypeParamsWithAny` in `tool/internal/instrument/trampoline.go`, which currently rewrites composite types containing type parameters. For example:

```text
[]T -> []interface{}
```

The existing unit test also expects this transformation.

### Expected behavior

The generated hook declaration should remain type compatible with the value passed by the generated trampoline, so that valid generic source code does not result in uncompilable generated code.

This also applies to other composite types containing type parameters, such as pointers, maps, channels, and function types.

This is separate from the constrained generic receiver issue in #1121.

SSOHEB:
## Description

Fixes an issue where generic function parameters containing type parameters inside composite types, such as `[]T`, can result in incompatible generated hook signatures.

The generated trampoline may continue to pass `[]T` while the generated hook declaration expects `[]interface{}`, causing the generated code to fail compilation.

This change detects type parameters within composite type expressions and widens the complete parameter to `interface{}` at the generated hook boundary. It also adjusts variadic handling for generic parameters.

## Motivation

Generic composite parameter types must remain assignable across the generated trampoline and hook boundary. Replacing only the inner type parameter is not sufficient because Go does not treat composite types such as `[]T` and `[]interface{}` as assignable.

This change ensures the generated hook signature remains compatible with the value passed by the trampoline.

Fixes #1126


darkraider01:
The core fix looks right — widening the entire composite type to interface{} (instead of rebuilding it with []T → []interface{} etc.) correctly addresses the assignability issue, and the variadic handling change in callBeforeHook (passing the whole slice as one boxed interface{} arg instead of spreading) is a sensible, deliberate tradeoff for that.

One thing worth double-checking before merge:

Silently-swallowed default case (trampoline.go:148)
The rewritten containsTypeParameter's default case now returns false for any dst.Expr node type the switch doesn't handle. The old replaceTypeParamsWithAny called util.Unimplemented(...) in that spot, failing loudly on unrecognized type-expression forms. Now, if a target parameter's type uses some AST node shape not enumerated here, containsTypeParameter will quietly assume "no type parameter present" — which means the generic param won't get widened to interface{}, and the exact []T vs []interface{} assignability bug this PR fixes could resurface silently rather than being caught with a clear error. Worth considering whether the default case should still fail loudly (or log) rather than fail open.

amazingakai:
@SSOHEB please resolve merge conflicts

SSOHEB:
> @SSOHEB please resolve merge conflicts

done

vignesh2027 (commented):
I read through this and it looks right to me.

The core change is the important one: replacing the whole type with `interface{}` when it contains a type parameter, instead of structurally rewriting it. `[]T` becoming `interface{}` rather than `[]interface{}` is what restores assignability, since the trampoline passes a `[]T` value and any concrete type is assignable to `interface{}`, while `[]T` is not assignable to `[]interface{}`. The map, chan, struct and func cases follow the same reasoning.

One thing worth calling out: the old `IndexExpr` and `IndexListExpr` branches returned `interface{}` unconditionally, so this also fixes a case beyond the reported `[]T`. A concrete instantiation like `Container[int]`, which contains no type parameter, was being collapsed to `interface{}` too, even on a non-generic function, since `replaceTypeParamsWithAny` runs on every parameter. The new `containsTypeParameter(tType.X) || containsTypeParameter(tType.Index)` preserves it, because neither `Container` nor `int` matches a type-param name.

That distinction is really the heart of the fix, so it might be worth locking into `TestContainsTypeParameter`, which currently exercises Star, Array, Map, Struct, Interface, Paren and Func but not the Index forms:

- `containsTypeParameter(Container[T])` -> true
- `containsTypeParameter(Container[int])` -> false
- the `IndexListExpr` equivalents

Non-blocking. The fix itself reads correct to me, and turning the structural rewrite into a predicate is a nice cleanup.

SSOHEB:
> I read through this and it looks right to me.
> 
> The core change is the important one: replacing the whole type with `interface{}` when it contains a type parameter, instead of structurally rewriting it. `[]T` becoming `interface{}` rather than `[]interface{}` is what restores assignability, since the trampoline passes a `[]T` value and any concrete type is assignable to `interface{}`, while `[]T` is not assignable to `[]interface{}`. The map, chan, struct and func cases follow the same reasoning.
> 
> One thing worth calling out: the old `IndexExpr` and `IndexListExpr` branches returned `interface{}` unconditionally, so this also fixes a case beyond the reported `[]T`. A concrete instantiation like `Container[int]`, which contains no type parameter, was being collapsed to `interface{}` too, even on a non-generic function, since `replaceTypeParamsWithAny` runs on every parameter. The new `containsTypeParameter(tType.X) || containsTypeParameter(tType.Index)` preserves it, because neither `Container` nor `int` matches a type-param name.
> 
> That distinction is really the heart of the fix, so it might be worth locking into `TestContainsTypeParameter`, which currently exercises Star, Array, Map, Struct, Interface, Paren and Func but not the Index forms:
> 
> * `containsTypeParameter(Container[T])` -> true
> * `containsTypeParameter(Container[int])` -> false
> * the `IndexListExpr` equivalents
> 
> Non-blocking. The fix itself reads correct to me, and turning the structural rewrite into a predicate is a nice cleanup.

Alright,  I added the requested IndexExpr and IndexListExpr coverage and pushed the changes. The local instrument tests, go vet, and git diff --check are passing.

SSOHEB (commented):
LGTM

tool/internal/instrument/trampoline.go
```diff
@@ -1109,103 +1114,97 @@ func makeMethodPanic(method *dst.FuncDecl, message string) {
...
 		}
-		newField.Type = replaceTypeParamsWithAny(field.Type, typeParams)
-		result[i] = newField
 	}
-	return result
+	return false
+}
+
+// replaceTypeParamsWithAny replaces type parameters with interface{} for use in
+// non-generic contexts like HookContextImpl methods
```
darccio: The function doc doesn't currently explain *why* widening is all-or-nothing rather than structural, which is the whole point of this fix (issue #1126: `[]T` is not assignable to `[]interface{}`, but is assignable to `interface{}`). Worth spelling out so a future reader doesn't "simplify" this back into a structural rewrite.

```suggestion
// replaceTypeParamsWithAny replaces type parameters with interface{} for use in
// non-generic contexts like HookContextImpl methods. Widening is all-or-nothing:
// composite types that merely mention a type parameter (e.g. []T, map[K]V) become
// a bare interface{}, not a structurally rewritten equivalent (e.g. []interface{}),
// since only the former is assignable from the concrete value the trampoline holds.
// Hook authors must type-assert accordingly.
```

tool/internal/instrument/trampoline.go
```diff
@@ -1109,103 +1114,97 @@ func makeMethodPanic(method *dst.FuncDecl, message string) {
...
+		return containsTypeParameter(tType.X, typeParams) || containsTypeParameter(tType.Index, typeParams)
 	case *dst.IndexListExpr:
-		// GenStruct[T, U] -> interface{} (for generic receiver methods with multiple type params)
-		return ast.InterfaceType()
-	case *dst.Ellipsis:
-		// ...T -> []T
-		// Preserve variadic syntax. This maintains variadic semantics in the
-		// generated hook signatures
-		return ast.Ellipsis(replaceTypeParamsWithAny(tType.Elt, typeParams))
+		return containsTypeParameter(tType.X, typeParams) || containsTypeParameterInExprs(tType.Indices, typeParams)
```
darccio, line 1163: Nice incidental fix here beyond the reported `[]T` case: the old `IndexExpr`/`IndexListExpr` branches in `replaceTypeParamsWithAny` returned `interface{}` unconditionally, so a concrete instantiation like `Container[int]` (no type parameter involved) was being collapsed to `interface{}` too. Gating on `containsTypeParameter(tType.X) || containsTypeParameter(tType.Index)` preserves it correctly, and `TestContainsTypeParameter` locks in both the positive and negative case.

tool/internal/instrument/trampoline.go
```diff
@@ -465,7 +465,11 @@ func (ip *instrumentPhase) checkHookDecl(hookFunc *dst.FuncDecl, before bool) er
 func (ip *instrumentPhase) callBeforeHook(t *rule.InstFuncRule) {
 	// Query whether the parameter is a variadic parameter in the target function
 	targetParams := findTargetParamType(ip.targetFunc)
-	isEllipsis := func(i int) bool { return ast.IsEllipsis(targetParams.List[i].Type) }
+	genericTypes := findTargetGenericType(ip.target, ip.targetFunc)
+	isEllipsis := func(i int) bool {
+		return ast.IsEllipsis(targetParams.List[i].Type) &&
+			!containsTypeParameter(targetParams.List[i].Type, genericTypes)
+	}
```
darccio, line 472: This PR's actual fix has no end-to-end coverage. `TestContainsTypeParameter` and `TestReplaceTypeParamsWithAny` verify the type-widening helpers in isolation, but nothing exercises this `isEllipsis` change or the generated trampoline+hook pair for the cases #1126 reports: a composite generic parameter (`[]T`, `map[K]V`) or a generic variadic (`...T`) on a real target function.

`testdata/golden/generic-functions` only covers a bare `T` parameter (`GenericFunc[T any](p1 T, p2 int)`), which never hit the bug — the bug is specific to `T` nested inside a composite type. Consider extending that golden fixture (or adding a new one) with a case like `func GenericFunc[T any](p1 []T, p2 ...T)`, so the suite proves the generated code actually compiles for the reported scenario rather than only proving the predicate is correct in isolation.

darccio (changes requested):

tool/internal/instrument/trampoline.go
SSOHEB: Thanks @darccio for reviewing this, so i extended the generic functions golden fixture to cover  the composite generic parameter and the generic variadic parameter, as you requested. Let me know if this looks right to you.

tool/internal/instrument/trampoline.go
SSOHEB: thanks 😁

tool/internal/instrument/trampoline.go
SSOHEB: thanks for the review of this , you can check it now, if i updated the documentation approriately , if you want anymore chances let me know
!

darccio (approved):
LGTM

PR #1132, opened 2026-08-17
fix(database/sql): parse non-parenthesized MySQL Unix-socket DSNs

### LINKED ISSUE #1131: fix(database/sql): parse non-parenthesized MySQL Unix-socket DSNs
## Problem

The MySQL DSN parser incorrectly handles non-parenthesized Unix-socket DSNs.

For example:

```text
unix:/tmp/mysql.sock/db
```

This should produce:

```text
server.address = /tmp/mysql.sock
server.port = ""
db.namespace = db
```

The current parser instead splits at the first `/`, which is part of the absolute socket path, and returns incorrect address and database values.

This gap was documented as a follow-up in #785 and was confirmed with @NameHaibinZhang.

## Root Cause

In the non-parenthesized branch of `parseMySQLDSN`, the parser uses the first `/` as the separator between the address and database name.

For a Unix-socket DSN, the address itself begins with `/`, so the first slash cannot be used as that separator.

## Proposed Fix

For a recognized non-parenthesized `unix:` DSN whose remaining address starts with `/`, split at the final `/` instead.

Preserve the current behavior for TCP, parenthesized Unix-socket DSNs, unknown protocols, and other DSN forms.

## Tests

Add focused coverage for:

- `unix:/tmp/mysql.sock/db`
- `user:pass@unix:/tmp/mysql.sock/mydb`
- `unix:/tmp/mysql.sock/`
- Existing parenthesized form: `unix(/tmp/mysql.sock)/db`
- Regression coverage for TCP and other existing address forms

The no-database form `unix:/tmp/mysql.sock` is structurally ambiguous when the final slash is also used as the database separator. Its expected behavior should be confirmed before encoding it in the parser.

## References

- #785
@NameHaibinZhang

ADITYA-CODE-SOURCE:
## Summary

Fix parsing of non-parenthesized MySQL Unix-socket DSNs such as:

```text
unix:/tmp/mysql.sock/db
```

The parser now uses the final `/` as the database separator when handling the non-parenthesized `unix:/` form. Other MySQL DSN forms retain their existing behavior.

## Changes

- Parse the Unix socket address as `/tmp/mysql.sock`
- Parse the database namespace as `db`
- Preserve an empty port for Unix sockets
- Add unit coverage for DSNs with credentials and empty database names
- Add integration coverage for the emitted database span attributes
- Regenerate the embedded `otelc` bundle

## Testing

- `go test ./dsnparse -race -shuffle=on -count=1`
- `go -C test test -v -timeout=10m -count=1 -tags=integration -run 'TestDBClient/DSN_Parsing' ./integration/...`
- `make test-unit`
- `make package`
- `git diff --check`

`make all` currently reaches lint but fails on pre-existing lint findings in files unchanged by this PR.

Fixes #1131


darccio:
@ADITYA-CODE-SOURCE I reviewed a sibling PR that happens to be a bit more complete. I'm closing your PR in favour of #1157. I requested their contributor to cherry-pick some parts of yours (integration test, mostly).

PR #1136, opened 2026-08-17
fix(tool/internal/setup): skip no-op tool file rewrite and go mod tid…

### LINKED ISSUE #1134: refactor(tool/internal/setup): every build reruns go mod tidy even when the pin changed nothing

For a project with a committed `otel.instrumentation.go`, every `otelc go build`
goes through `updateToolFile()` in `tool/internal/setup/pin.go`, which
unconditionally rewrites the tool file — even when the rendered result is
byte-identical to what's already on disk — and then runs `go mod tidy`.

`go mod tidy` loads the whole module graph and can take seconds, and may touch
the network on a cold cache. In the steady state, where setup was done once and
the files are committed, that work repeats on every build and cannot produce a
different result. It's the difference between a build that shells out to the Go
toolchain and one that doesn't.

Skip both when this run provably changed nothing. The signal already exists:
`ensureOtelcRequire()` returns a `modified` bool reporting whether it touched
go.mod, and `updateToolFile` has been discarding it since the pin command landed
in #655. Rendering the updated AST to bytes first and comparing against the file
on disk gives the same signal for the tool file. Tidy then runs only when the
tool file changed, go.mod changed, or go.sum is missing.

- The broader `isSetup()` TODO in #161. This is adjacent in spirit but much
  narrower: it doesn't try to detect "already set up" globally, only whether
  this particular run changed anything tidy could react to. The narrow version
  seemed safer to land first.
- A timestamp or mtime check on the tool file. Fragile across checkouts and
  clean builds; a byte comparison of the rendered output is exact.
- Caching a content hash of the inputs in a stamp file. Closes one more edge
  case (a stale-but-present go.sum from a tidy that died mid-run) but adds state
  to manage. I left it out and can add it if reviewers want that case fully
  covered.

**Additional context**

Rough size of the effect: in the test suite, `TestUpdateToolFile` — which
reaches the real tidy — takes ~16s on a cold module cache and ~1s warm  while the steady-state call skips the Go toolchain entirely, and cold caches or network-touching tidies widen the gap.
cache, while a second `updateToolFile` call in the steady state completes
without invoking the Go toolchain at all. Cold caches or network-touching tidies
widen that, not narrow it.

I have this implemented with tests covering both the skip and the deleted-go.sum
re-tidy, and can open a PR.

itssaharsh:
## Description

`updateToolFile` now:

- renders the updated AST to bytes (new `ast.PrintFile` helper, which
  `ast.WriteFileAtomic` also reuses) and wrthe
  rendered content differs from what's on disk, and
- uses the `modified` bool from `ensureOtelly
  discarded, running `go mod tidy` only when the tool file changed, go.mod
  changed, or go.sum is missing.

Every path that mutates module state still rough
the generate path, which has its own tidy. Prunes and generate-directive changes
make the rendered bytes differ. Adding the ping the
version flips the go.mod bool. A hand-deleted or never-created go.sum fails the
existence check. The skip only fires when tng tidy
could react to.

One gap I know of: a stale-but-present go.sum left behind by a tidy that died
mid-run. In the default `otelc go build` fle
manager reverts go.mod after the failed build, so the next run re-adds the
require and tidies. In the standalone pin frts
exactly what to do ("missing go.sum entry … to add it: go mod tidy"). I can add
a content-hash stamp to close that case ful

### Effect

Measured on go1.25.0 linux/amd64, i7-13620H
reaches the real tidy, takes ~16s on a cold module cache (it resolves modules
over the network) and ~1.2s warm. The stead
`updateToolFile` call completes without invoking the Go toolchain at all. That
gap is per module, per build, for anyone oncold
cache or a network-touching tidy widens it rather than narrowing it.

No microbenchmark here — the saving is a skipped subprocess, not a hot path, so
the test timings seemed like the more hones

### Behavior

New tests cover both sides: `TestUpdateTool
(nothing changed → no rewrite, no tidy) and
`TestUpdateToolFile_MissingGoSumRunsTidy` (d
restores it). All existing paths — first-time pins, prunes, directive changes,
version bumps — tidy exactly as before.

## Motivation

`go mod tidy` is the most expensive fixed c
committed-setup flow — seconds of pure waste per build in the steady state.

Fixes #1134

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format                          - [x] Code formatted: `make format`
- [ ] Linters pass: `make lint` (reports 41 findings that pristine `main` also reports with the repo-pinned linter; none in files this PR touches)
- [ ] Tests pass: `make test` (tool unit suites pass locally, including the setup and ast packages; full e2e not run)
- [x] Tests follow [testing guidelines](https://github.com/open-telemetnstrumentation/blob/main/docs/testing.md)
- [ ] Documentation updated (not applicable)
- [ ] OpenTelemetry Registry updated (not a
  - [] I used AI and I have read and follobutionPolicy](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md).
- [x] I have the experience and knowledge nw, and validate all content in this PR.


Labreo (commented):
Nice fix. The steady-state skip makes sense and the tests cover the important cases well.

One thing I was curious about was if a user manually edits go.mod outside of otelc, does goModChanged still catch that? I'm guessing the Go toolchain just complains about missing sums at build time, but wanted to double check since that path isn't covered in the tests.

itssaharsh:
goModChanged doesn't detect manual edits, and that's intentional. It is recomputed from go.mod each run and only reports whether otelc had to change it. So if otelc's entries are broken, they get fixed and tidied again, while the user's dependencies are left to the normal Go toolchain. This keeps the invariant that otelc only tidies when it writes to go.mod, avoiding a module graph load on every build. Happy to add a small test showing that a manual require is left untouched when the tidy is skipped.

amazingakai:
This is intentional, see: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/655#discussion_r3535254025

You can maybe add a comment about it there.

Labreo:
Ok thanks for clarity @amazingakai and @itssaharsh .

tool/internal/ast/parser.go
```diff
@@ -119,16 +119,26 @@ func writeFile(w writeCloser, filePath string, root *dst.File) error {
...
+	}
+
+	return buf.Bytes(), nil
+}
+
+// WriteFileAtomic writes the AST to a file atomically.
+func WriteFileAtomic(filePath string, root *dst.File) error {
+	data, err := PrintFile(root)
+	if err != nil {
 		return ex.Wrapf(err, "failed to restore AST for file %s", filePath)
```
darccio: PrintFile already attaches the message "failed to restore AST". This wrap
repeats the operation and adds only the file path, so the final error reads
"failed to restore AST for file otel.instrumentation.go: failed to restore
AST". Keep the operation name in PrintFile. Add only the path here.

```suggestion
		return ex.Wrapf(err, "for file %s", filePath)
```

tool/internal/setup/pin.go
```diff
@@ -422,15 +428,48 @@ func updateToolFile(ctx context.Context, toolFile string, prunedImports map[stri
...
+	// The one gap left open is a stale-but-present go.sum after a tidy that
+	// errored out mid-run: the default build flow self-heals because the
+	// state manager reverts go.mod (so the next run re-adds the require and
+	// tidies), and in the standalone pin flow the next go build names the
+	// fix itself ("missing go.sum entry ... to add it: go mod tidy").
+	//
+	// Manual user edits to go.mod or the tool file are deliberately out of
+	// scope here: tidying those up is the user's responsibility, the same as
+	// in an uninstrumented project. That split was settled during the #655
+	// review, where this conditional shape was first proposed.
```
darccio: Lines 462-465 state the scope for manual edits, then cite the review that
approved the scope. The reader of this file needs the scope statement, not the
review history. The citation gives the reader no rule to follow, and it breaks
when the repository moves the discussion. Keep the scope sentence. Delete the
provenance.

```suggestion
	// in an uninstrumented project.
```

tool/internal/setup/pin.go
```diff
@@ -422,15 +428,48 @@ func updateToolFile(ctx context.Context, toolFile string, prunedImports map[stri
...
+
+	toolFileChanged := !bytes.Equal(updated, original)
+	if toolFileChanged {
+		if writeErr := util.WriteFileAtomic(toolFile, updated); writeErr != nil {
+			return ex.Wrapf(writeErr, "writing tool file %s", toolFile)
+		}
 	}
 
-	_, ensureErr := ensureOtelcRequire(filepath.Dir(toolFile), util.Version)
+	goModChanged, ensureErr := ensureOtelcRequire(filepath.Dir(toolFile), util.Version)
```
darccio, line 474: A dev build of otelc reports util.Version as "v0.0.0", "(devel)", or a
pseudo-version. ensureOtelcRequireVersion returns (false, nil) for all three
(pin.go:243), so goModChanged never flips for a dev build.

When the tool directive exists, the otelc require line is absent, and go.sum
exists, the skip then fires on every run. Only `go tool otelc pin` breaks: the
hook code uses replaced instrumentation and pkg modules, so otelc go build
still works. Go prints the repair command in the error.

A standalone otelc pin that fails at the tidy leaves this state. The build flow
cannot produce it, because the state manager reverts go.mod and go.sum together
(state.go:100).

Release builds do not reach this state: ensureOtelcRequireVersion calls
AddRequire when the require line is absent, so goModChanged flips and the tidy
runs.

Fix: in ensureOtelcRequire, report modified=true when the require line is
absent, also for invalid versions. The tidy then adds the require at the
version Go resolves, and the skip keeps its invariant.

tool/internal/setup/pin.go
darccio: This is a bit verbose. Please make it shorter.

tool/internal/setup/pin_test.go
```diff
@@ -814,6 +815,90 @@ go 1.25
...
+	require.NoError(t, os.Remove(goSumPath))
+
+	var logs bytes.Buffer
+	debugLogger := slog.New(slog.NewTextHandler(&logs, &slog.HandlerOptions{Level: slog.LevelDebug}))
+	ctx := util.ContextWithLogger(t.Context(), debugLogger)
+
+	require.NoError(t, updateToolFile(ctx, toolFile, nil, opts))
+	require.NotContains(t, logs.String(), "skipping go mod tidy")
+	require.FileExists(t, goSumPath, "go mod tidy should have restored go.sum")
+}
```
darccio, line 997: You offered this test in the review thread: a manual require stays untouched
while the tidy is skipped. The thread settled the contract that otelc tidies
only the state it writes. This test adds the contract to the suite, so a later
change cannot break it silently.

```suggestion
}

func TestUpdateToolFile_SkippedTidyKeepsManualRequire(t *testing.T) {
	trueValue := true

	dir := t.TempDir()

	require.NoError(t, os.WriteFile(
		filepath.Join(dir, "go.mod"),
		[]byte(`module example.com/test

go 1.25
`),
		0o644,
	))

	toolFile := filepath.Join(dir, toolFileCanonical)
	writeToolFile(t, toolFile, "fmt")

	opts := PinOptions{
		Prune:    true,
		Generate: &trueValue,
	}

	// First run reaches the steady state.
	require.NoError(t, updateToolFile(t.Context(), toolFile, nil, opts))

	goModPath := filepath.Join(dir, "go.mod")
	goModAfterFirst, err := os.ReadFile(goModPath)
	require.NoError(t, err)

	// The user adds a manual require. otelc owns only the lines it
	// writes, so the manual require must survive the skipped tidy.
	goModManual := string(goModAfterFirst) + "\nrequire example.com/manual v1.2.3\n"
	require.NoError(t, os.WriteFile(goModPath, []byte(goModManual), 0o644))

	var logs bytes.Buffer
	debugLogger := slog.New(slog.NewTextHandler(&logs, &slog.HandlerOptions{Level: slog.LevelDebug}))
	ctx := util.ContextWithLogger(t.Context(), debugLogger)

	require.NoError(t, updateToolFile(ctx, toolFile, nil, opts))
	require.Contains(t, logs.String(), "skipping go mod tidy")

	goModAfterSecond, err := os.ReadFile(goModPath)
	require.NoError(t, err)
	require.Equal(t, goModManual, string(goModAfterSecond))
}
```

tool/internal/setup/pin_test.go
```diff
@@ -814,6 +815,90 @@ go 1.25
...
+	goModAfterFirst, err := os.ReadFile(filepath.Join(dir, "go.mod"))
+	require.NoError(t, err)
+
+	// Second run changes nothing, so go mod tidy must be skipped.
+	var logs bytes.Buffer
+	debugLogger := slog.New(slog.NewTextHandler(&logs, &slog.HandlerOptions{Level: slog.LevelDebug}))
+	ctx := util.ContextWithLogger(t.Context(), debugLogger)
+
+	require.NoError(t, updateToolFile(ctx, toolFile, nil, opts))
+	require.Contains(t, logs.String(), "skipping go mod tidy")
```
darccio: Both new tests assert on the literal debug message "skipping go mod tidy"
(here and in TestUpdateToolFile_MissingGoSumRunsTidy). A wording change in
pin.go breaks both tests at once.

The log text is also the only direct evidence for the skip: a tidy on a clean
module writes nothing, so file equality cannot prove that the skip fired.

Declare the message as a constant in pin.go, and reference the constant from
both tests. The tests keep the direct signal and lose the copy of the text.

tool/internal/setup/pin_test.go
```diff
@@ -814,6 +815,90 @@ go 1.25
...
+	var logs bytes.Buffer
+	debugLogger := slog.New(slog.NewTextHandler(&logs, &slog.HandlerOptions{Level: slog.LevelDebug}))
+	ctx := util.ContextWithLogger(t.Context(), debugLogger)
+
+	require.NoError(t, updateToolFile(ctx, toolFile, nil, opts))
+	require.NotContains(t, logs.String(), "skipping go mod tidy")
+	require.FileExists(t, goSumPath, "go mod tidy should have restored go.sum")
+}
+
 func TestUpdateToolFile_ParseError(t *testing.T) {
```
darccio, line 1109: The comment above the skip (pin.go:448) claims that pruning an import flips
toolFileChanged and forces the tidy. No test exercises that claim after the
steady state. The gate now carries the correctness of the whole skip, so the
claim needs a direct test.

```suggestion
func TestUpdateToolFile_PruneAfterSteadyStateRunsTidy(t *testing.T) {
	trueValue := true

	dir := t.TempDir()

	require.NoError(t, os.WriteFile(
		filepath.Join(dir, "go.mod"),
		[]byte(`module example.com/test

go 1.25
`),
		0o644,
	))

	toolFile := filepath.Join(dir, toolFileCanonical)
	writeToolFile(t, toolFile, "fmt")

	opts := PinOptions{
		Prune:    true,
		Generate: &trueValue,
	}

	// First run reaches the steady state.
	require.NoError(t, updateToolFile(t.Context(), toolFile, nil, opts))

	var logs bytes.Buffer
	debugLogger := slog.New(slog.NewTextHandler(&logs, &slog.HandlerOptions{Level: slog.LevelDebug}))
	ctx := util.ContextWithLogger(t.Context(), debugLogger)

	// A prune must flip toolFileChanged and force the tidy.
	require.NoError(t, updateToolFile(ctx, toolFile, map[string]bool{"fmt": true}, opts))
	require.NotContains(t, logs.String(), "skipping go mod tidy")

	toolFileAfter, err := os.ReadFile(toolFile)
	require.NoError(t, err)
	require.NotContains(t, string(toolFileAfter), `"fmt"`)
}

func TestUpdateToolFile_ParseError(t *testing.T) {
```

darccio (changes requested):
I addressed the conflicts to ensure that `ex.Wrapf` aren't re-added again.

darccio:
@itssaharsh Check the E2E in this PR: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/34162642856/job/101867460933?pr=1136

itssaharsh:
@darccio I can't understand why the tests are failing :)

PR #1137, opened 2026-08-17
fix(instrumentation/net/http): cut per-request allocations in client 

### LINKED ISSUE #1133: refactor(instrumentation/net/http): hooks allocate a map and build debug log arguments on every request

The net/http client and server hooks do two things on every request that cost
allocations without doing useful work in the default configuration.

First, the Before hooks pass state to the After hooks with
`ictx.SetData(map[string]interface{}{...})`. That allocates a map and hashes
string keys per request just to carry a span and a start time. Two of the stored
entries are never read at all: `"ctx"` on both sides, and `"req"` on the client
side.

Second, the hooks call `logger.Debug(...)` with computed arguments such as
`r.URL.String()`. slog evaluates arguments before it checks the level, so at the
default INFO level every request serializes the full URL and boxes attributes
that are then discarded. The disabled path logs per request too.

Benchmarking the Before+After pair (mock hook context, SDK tracer, no exporter)
on go1.25.0 linux/amd64 shows 27 allocs/op on the server side and 29 on the
client side. Six and nine of those respectively come from the two items above.
For a project whose pitch is low runtime overhead, that seemed worth raising.

Replace the map with a small typed struct holding just the span and start time,
and gate the argument-taking `Debug` calls behind a level check so the URL
serialization and attribute boxing only happen when debug logging is on. Also
stop logging on the disabled path, which is the principle the gin hook already
documents ("Keep the disabled path free of logging", `context_hook.go`) but
net/http never picked up.

- Keeping the map but dropping only the unread `"ctx"` / `"req"` entries. Saves
  the least and leaves the map allocation and key hashing in place.
- `sync.Pool` for the carried state. More machinery than a two-field struct
  needs, and the struct is a single small allocation anyway  a pool is more machinery than two fields justify.
- Converting the Debug calls to `slog` `LogAttrs` instead of gating them. That
  removes the boxing but not the `r.URL.String()` call, which is the larger
  cost.
- Leaving the disabled path logging as-is. Reasonable if it's deliberate for
  troubleshooting, which is why I'd rather ask than assume.

**Additional context**

I have this implemented and benchmarked with benchstat (10 runs per side,
`-benchtime=2s`, all comparisons p=0.000) and can open a PR. Happy to hear first
if either of these is intentional, or if you'd prefer the two changes split into
separate PRs.

itssaharsh:
## Description

Two code changes, both in the hook files.

**Typed struct instead of a map for Before→After state.** The
`map[string]interface{}` becomes a small `hookData` struct holding the span and
the start time. One small allocation instead of a map, no key hashing, and the
never-read `"ctx"` / `"req"` entries are gone. This data is written and read by
the same hook pair within these files, so it isn't part of any exported
contract.

**Level check before building debug arguments.** The `Debug` calls that take
computed arguments are gated behind a level check, so `r.URL.String()` and
attribute boxing only run when debug logging is actually enabled. The disabled
path no longer logs per request, matching the principle the gin hook documents.

The diff also updates the hook unit tests and `tool/data/otelc-bundle.tgz`. The
bundle embeds these hook sources and the Verify Bundle check requires it to be
in sync; since the regenerate bot can't push to fork PRs, it's regenerated here
with `make package`.

### Benchmarks

Added in this PR. Environment: go1.25.0 linux/amd64, 13th Gen Intel Core
i7-13620H (16 threads), WSL2. Each side:

    OTEL_LOG_LEVEL=warn go test -bench=. -benchmem -run '^$' -count=10 -benchtime=2s

compared with benchstat.

    goos: linux
    goarch: amd64
    cpu: 13th Gen Intel(R) Core(TM) i7-13620H
                      │    main     │             this PR                 │
                      │   sec/op    │   sec/op     vs base                │
    ServeHTTPHooks-16   9.609µ ± 7%   7.816µ ± 7%  -18.66% (p=0.000 n=10)
    RoundTripHooks-16   8.088µ ± 3%   6.441µ ± 3%  -20.36% (p=0.000 n=10)

                      │     B/op     │     B/op      vs base               │
    ServeHTTPHooks-16   4.312Ki ± 0%   3.930Ki ± 0%   -8.88% (p=0.000 n=10)
    RoundTripHooks-16   3.312Ki ± 0%   2.867Ki ± 0%  -13.44% (p=0.000 n=10)

                      │  allocs/op  │ allocs/op   vs base                │
    ServeHTTPHooks-16    27.00 ± 0%   21.00 ± 0%  -22.22% (p=0.000 n=10)
    RoundTripHooks-16    29.00 ± 0%   20.00 ± 0%  -31.03% (p=0.000 n=10)

All six comparisons are p=0.000 at n=10; allocation and byte counts are exact.

Two notes for anyone reproducing this. `OTEL_LOG_LEVEL=warn` is only there
because the one-time init INFO log interleaves with the benchmark output and
breaks benchstat's parsing; it doesn't tilt the comparison, since the
Debug-argument cost this PR removes is paid on main at every log level. And
`-benchtime=2s` matters — at 1s the client sec/op result wasn't claimable
(p=0.052, ±22% spread) and only stabilized at 2s. The allocation numbers were
stable either way.

The benchmark includes the mock context and SDK span start/end, so the hook-only
relative improvement is larger than the headline percentages.

### Behavior

No telemetry change: same spans, same attributes, same status handling.

Debug logging changes slightly at `OTEL_LOG_LEVEL=debug`. The informative lines
(`BeforeServeHTTP called`, `AfterRoundTrip called`, and so on) are unchanged.
The redundant trailing `... completed` lines and the per-request
`instrumentation disabled` lines are gone.

## Motivation

These hooks run on every request of every instrumented HTTP app; fixed
per-request overhead is the project's whole cost story.

Fixes #1133

---

## Checklist

- [ ] Linters pass: `make lint` (reports 41 findings that pristine `main` also reports with the repo-pinned linter; none in files this PR touches)
- [ ] Tests pass: `make test` (unit suites and the net/http-scoped integration tests pass locally; full e2e not run)
- [ ] Documentation updated (not applicable)
- [ ] OpenTelemetry Registry updated (not applicable)
- [] This PR has content that I did not fully write myself.
  - [] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md).

---


vignesh2027:
Nice change, and the allocation win is real. I ran the before and after locally, driving BeforeRoundTrip and AfterRoundTrip in a loop:

```
main:    2147 ns/op   4320 B/op   34 allocs/op
this PR: 1820 ns/op   3864 B/op   25 allocs/op
```

So roughly 9 fewer allocations and about 15% faster per request, and the typed hookData read is nil-safe: the `!ok || data == nil || data.span == nil` guard covers both the wrong-type and nil cases. Gating the Debug calls behind debugEnabled() is right too, since slog evaluates its arguments before the level check.

One thing I think is a regression, on both hooks. AfterRoundTrip and AfterServeHTTP now start with an Enable() re-check that returns before the span is retrieved and ended:

```go
if !clientEnabler.Enable() {
    return
}
```

On main, AfterRoundTrip has no such check: it goes straight to reading the span and `defer span.End()`. With the check added, if instrumentation is enabled when BeforeRoundTrip creates the span but disabled at runtime before AfterRoundTrip runs, the After hook returns early and the span is never ended.

I reproduced it on this branch: enable, run BeforeRoundTrip, set OTEL_GO_DISABLED_INSTRUMENTATIONS=nethttp, run AfterRoundTrip, and the recorder shows 0 ended spans where there should be 1.

This is the same lifecycle issue #1094 is fixing on the server side (its description is literally that AfterServeHTTP re-checked enablement before retrieving the span, so the span never gets ended). This PR keeps that shape on the server side and adds it to the client, so the two would collide.

The allocation refactor does not actually need the Enable() check to keep its win. Stored hookData only exists if Before ran while enabled, so the existing `!ok || data == nil` guard is already enough to key off, and it ends the span regardless of a later toggle. Dropping the two added Enable() returns keeps the allocation improvement and avoids the leak.

itssaharsh:
The scenario is real, but this PR didn't introduce it. The Enable() re-check already exists on main in both After hooks (client_hook.go, server_hook.go). It only appears in my diff because I added a comment and removed the Debug line inside it.

Your repro matches main's current tested behavior. TestAfterRoundTrip already covers the case of a disabled env with a span in hook data and expects 0 ended spans, with the comment "Span should not be ended because instrumentation is disabled."

On #1094 , it's not server-only. It changes both client and server hooks, along with tests and the bundle, and implements exactly this fix: remove the After-side re-check, rely on the stored span, and regression-test the mid-flight disable case.

So I'd rather not duplicate that fix in this PR. The allocation refactor doesn't depend on the check, and my hookData is already the state #1094's approach wants to use. I'll keep this PR behavior-neutral and rebase on #1094 when it lands. The overlap is just one comment and one deleted log line, so rebasing either way should be trivial.

vignesh2027:
You're right, and thanks for walking through it. I misread the diff: the `if !clientEnabler.Enable()` block in AfterRoundTrip is context on your base (6b37f50a), not something this PR adds. Your change there is just the dropped Debug line and the comment, so that early return is main's behavior, not yours.

I also had #1094 backwards. It merged on the 18th (3cb602e6) and it changes both the client and server hooks plus their tests, so it is not server-only, it is exactly this fix. Current main already carries it: AfterRoundTrip no longer re-checks Enable(), and TestAfterRoundTrip_DisabledAfterStart_Regression asserts the span still ends after a mid-flight disable.

That block is already gone on main, so once you rebase, this PR won't carry it and comes out as purely the allocation refactor with no behavior change. Keeping it behavior-neutral is the right call, and the allocation numbers stand on their own.

instrumentation/net/http/client/client_hook.go
```diff
@@ -99,30 +120,28 @@ func BeforeRoundTrip(ictx hook.HookContext, transport *http.Transport, req *http
...
+		start: time.Now(),
 	})
 }
 
 func AfterRoundTrip(ictx hook.HookContext, res *http.Response, err error) {
+	// This runs once per outbound request; keep the disabled path free of logging.
 	if !clientEnabler.Enable() {
-		logger.Debug("HTTP client instrumentation disabled")
 		return
 	}
```
darccio: This PR's base (`6b37f50a`) predates #1094 (`3cb602e6`, "end spans after instrumentation is disabled"), which is now on `main`. On `main`, `AfterRoundTrip` has no `Enable()` re-check — it goes straight to reading the stored span and ending it. This PR still carries the re-check:

```suggestion
func AfterRoundTrip(ictx hook.HookContext, res *http.Response, err error) {
	data, ok := ictx.GetData().(*hookData)
```

That means if instrumentation is enabled when `BeforeRoundTrip` creates the span but gets disabled (e.g. via remote config) before `AfterRoundTrip` runs, this returns before `defer span.End()` and the span leaks — the exact regression #1094 fixed on `main`, reintroduced here.

The same collision exists in three more places and all four need to move together:
- `instrumentation/net/http/server/server_hook.go` — `AfterServeHTTP` has the identical `if !serverEnabler.Enable() { return }` re-check that main also removed.
- `instrumentation/net/http/client/client_hook_test.go` — the `"instrumentation disabled"` case in `TestAfterRoundTrip` asserts `0` spans ended; on `main` it asserts `1`, and `TestAfterRoundTrip_DisabledAfterStart_Regression` is missing from this PR entirely.
- `instrumentation/net/http/server/server_hook_test.go` — same pattern: `"instrumentation disabled"` in `TestAfterServeHTTP` asserts `0` spans ended instead of `1`, and `TestAfterServeHTTP_DisabledAfterStart_Regression` is missing.

Please rebase onto current `main`, drop the `Enable()` re-check from both After hooks, and carry forward both `*_DisabledAfterStart_Regression` tests along with the corrected assertions. `tool/data/otelc-bundle.tgz` will need to be regenerated again afterward since it embeds this source.

darccio (changes requested):

itssaharsh:
Reran the benchmarks against the new base, since removing that re check made main itself faster: server 25 to 19 allocs/op (sec/op −10.32%), client 27 to18 (−19.83%), all p=0.000 at n=10. Same 6 and 9 allocations removed as before  just measured against a faster baseline now.

itssaharsh:
@darccio completed the changes
new benchmarks
ServeHTTPHooks  sec/op -11.70%  B/op  -8.96%  allocs/op 24 -> 18
RoundTripHooks  sec/op  -6.18%  B/op -13.60%  allocs/op 26 -> 17
It seems the #1016 did something

darccio:
@itssaharsh The trick is based on the fact that #1016 turned `Instrumented()` almost free. The pre-#1016 implementation re-parsed the env var on every call: `os.Getenv` → `parseInstrumentationList` (`strings.SplitSeq` + `TrimSpace`/`ToLower` + `append` into a fresh []string). With `nethttp` configured that's a slice allocation 
 plus the split/trim CPU work, paid once per request, forever.

Once you rebased, both sides of the benchstat comparison inherited the cheaper `Instrumented()`.

#1094 introduced another improvement too, removing the After-sided `Enabled()` re-check entirely, thus killing one whole `Instrumented()` call per request.

darccio (approved):
LGTM

itssaharsh:
@darccio thanks for the explanation of `1016` and how it the performance increases could you explain why the E2E tests didn't pass on this and if merging was a safe choice (since the E2E test didn't pass).

darccio:
@itssaharsh It isn't possible to merge with failing checks. I'm not sure which E2E didn't pass. They only had some warnings: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/34130381559

What I did see now is that unit tests seem to contain failures but they don't fail: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/actions/runs/34130381628

PR #1139, opened 2026-08-17
test(k8s.io/client-go): replace context.TODO() with t.Context()

kehach07:
Refactors tests in instrumentation/k8s.io/client-go to use t.Context() instead of context.TODO() for improved context cancellation management during test execution.

## Description

This PR replaces `context.TODO()` with `t.Context()` in the `k8s.io/client-go` instrumentation tests (`instrumentation/k8s.io/client-go/k8s_otel_handler_test.go`). It also cleans up the now-unused `"context"` import in that file.

## Motivation

Using `t.Context()` is a best practice introduced in Go 1.24 that automatically handles context cancellation when the test completes or times out. This avoids leaving hanging resources or routines in case a test is interrupted or fails prematurely. `context.TODO()` should only be used as a temporary placeholder when it's unclear which Context to use, whereas `t.Context()` provides the precise lifecycle scoping needed for testing.

Fixes # 

---

## Checklist

- [x] Tests pass: `make test`
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](https://github.com/


y1yang0 (approved):

kehach07:
@y1yang0  Thank you for the approval! 

I had to force-push to rebase the branch because it fell out of date with `main`, which automatically disabled the auto-merge. 

Also, the `Verify Bundle` check is currently failing because I'm developing on Windows. When I package the `otelc-bundle.tgz` locally, it creates a slightly different binary signature (due to OS-level file permissions/timestamps) than what the Ubuntu CI expects, causing the diff to fail.

Could you please comment `/regenerate-bundle` to let the GitHub Actions bot fix the archive for my branch, and then re-enable auto-merge? Thank you!

Issue #1143, opened 2026-08-18
Reduce exported API surface in tool packages


y1yang0:
A lot of identifiers in the tool packages are exported but never used
outside their own package. This mostly affects `tool/internal/*`, where
exporting serves no purpose at all since `internal` packages can't be
imported from outside the module.

A rough count using a go/packages-based usage scan: ~100 such symbols.
The biggest offenders:

- `tool/internal/rule`: all of the `Selector*`/`Modifier*`/`Sel*`/`Key*`
  constants and the schema validation helpers in `schema.go`
- `tool/internal/setup`: the filter types (`Filter`, `FuncFilter`,
  `AllOf`, `Not`, ...), `StateManager` and its helpers,
  `InstrumentationConfig`, `BuildWithToolexec`, `AutoPin`, ...
- `tool/internal/ast`, `manifest`, `profile`, `imports`, `pkgload`:
  assorted helpers with no external users

They should be unexported

darkraider01:
I'd like to take this. Spot-checked a handful of the cited symbols before committing to make sure I understood the actual boundary, not just the ~100 count:

setup.BuildWithToolexec, setup.AutoPin, rule.Selector*/Modifier*/Key* — confirmed zero references outside their own package. Clean.
One conflict, though: tool/internal/setup/filter_test.go is package setup_test (external, black-box), and it references setup.Filter, setup.FuncFilter, setup.AllOf, setup.Not, setup.OneOf, setup.StructFilter, setup.PackageNameFilter, setup.DirectiveFilter, setup.MatchContext, setup.Build — exactly the filter types you listed for unexporting. Unexporting them breaks that test file as-is. Straightforward fix (fold it into package setup as an internal test, alongside the rest), but wanted to flag it rather than silently convert it, since it changes that file from black-box to white-box testing.
tool/internal/rule's two external test files (normalize_test.go, target_glob_test.go, both package rule_test) only touch Normalize, ValidateTarget, MatchGlobTarget, IsGlobTarget, IsRootTarget, TargetRoot — none of those are on your unexport list, so no conflict there.

Planning to run the real go/packages-based usage scan (same kind you used for the count) as the actual first step, rather than trust the cited list by hand, in case anything else has a similar external-test dependency I haven't spotted.

y1yang0:
@darkraider01 Assigned, thanks for taking this 👍 

> tool/internal/setup/filter_test.go is package setup_test (....

[...]

PR #1145, opened 2026-08-18
chore(tool): unexport internal identifiers with no cross-package use

darkraider01:
## Description

First slice of #1143, covering the packages with only a handful of affected symbols so the
approach can be reviewed cheaply before the larger ones (`ast`, `setup`, `rule`) follow.

Unexported:

- `util.IsUnix` → `isUnix`
- `pkgload.ResolveModuleDir` → `resolveModuleDir`
- `manifest.Load` → `load`
- `profile.CPU` / `Heap` / `Trace` → `typeCPU` / `typeHeap` / `typeTrace`

The profile constants get a `type` prefix rather than plain `cpu`/`heap`/`trace` because
`profile.go` imports `runtime/trace`; a bare `trace` constant would shadow the package
within that file.

### How the list was produced

Rather than work from a hand-written list, I wrote a `go/packages` analyzer (full type
info, `Tests: true`) that collapses a package with its internal test variant
(`p [p.test]`) and its external test package (`p_test`) into one logical package — an
external `_test` package is not a real API consumer, since it can be converted to an
internal test instead. It then reports exported package-scope identifiers with zero
references from any other logical package.

I sanity-checked it against ground truth before trusting it: `cmd/otelc` references
`setup.{Cleanup,GoBuild,Pin,PinOptions,Setup}` and `instrument.{Toolexec,EnableNestedToolexec}`,
and none of those appear in the analyzer's output, while `setup.Build`/`AutoPin`/
`BuildWithToolexec` (distinct from `GoBuild`) correctly do.

### One correction to the issue's count

Six of the ~100 identifiers **cannot** be unexported, because they are reachable from the
signature of an exported function that external code calls. Callers never name them —
they get them via `:=` inference — so a naive reference scan reports them as unused:

| Type | Reached through |
|---|---|
| `profile.Type` | `ParseTypes`, `Start`, `Merge` |
| `setup.PinResult` | `Pin`, `AutoPin` |
| `setup.Dependency` | exported setup signatures |
| `imports.Resolution` | exported imports signature |
| `manifest.Entry` | exported manifest signature |
| `ast.FuncDirectiveMatch` | exported ast signature |

Unexporting these would leave an exported function returning a type callers cannot name,
which revive's `unexported-return` flags — so the PR would fail lint. They are left
exported here and called out so the same trap does not catch the later PRs.

Note `setup.Filter` is *not* in that group: `Build`, which returns it, is itself
unexportable, so the two move together in a later PR.

### Deferred to its own PR

`instrument.InstrumentPhase` and `instrument.TJump` are only two symbols but 94 occurrences
across 23 files. Bundling that rename here would bury this diff, so it gets its own PR.

## Motivation

Part of #1143

## Verification

This is a pure rename with no behaviour change, so the check that matters is that the whole
suite still passes and nothing outside the four packages moved.

`go build ./tool/...`, `go vet ./tool/...` and `go test -shuffle=on -timeout=15m -count=1
./tool/...` all pass locally. The integration and e2e halves of `make test` were verified
through this PR's CI jobs rather than locally, since the testcontainers-based suites had no
usable Docker host here.

`.github/workflows/lint-golang.yaml` pins, run across the whole repo the way CI invokes it
— 0 issues.

Two over-eager substitutions were caught by the compiler and test suite during the rename
and fixed: a `manifest_test.go` call site, and three assertions comparing against runtime
error strings that legitimately contain the word "CPU". Re-running the analyzer afterwards
confirms `util`, `pkgload` and `manifest.Load` no longer appear, and `manifest`/`profile`
now list only their signature-blocked types.

---

## Checklist

- [x] Tests pass: `make test`
- [ ] Tests added for new functionality (N/A — rename only, no behaviour change)

[^I-know-my-stuff]:

### DIFF
--- a/tool/internal/manifest/manifest.go
+++ b/tool/internal/manifest/manifest.go
@@ -73,7 +73,7 @@ func Generate(instrumentationRoot string) (Manifest, error) {
 	return manifest, nil
 }
 
-func Load() (Manifest, error) {
+func load() (Manifest, error) {
 	var manifest Manifest
 	if err := json.Unmarshal(data.GetManifestJSON(), &manifest); err != nil {
 		return nil, ex.Wrapf(err, "loading embedded instrumentation manifest")
--- a/tool/internal/manifest/manifest_test.go
+++ b/tool/internal/manifest/manifest_test.go
@@ -147,7 +147,7 @@ func TestIsRuleFile(t *testing.T) {
 }
 
 func TestLoad(t *testing.T) {
-	got, err := Load()
+	got, err := load()
 	require.NoError(t, err)
 	require.NotEmpty(t, got)
 
--- a/tool/internal/pkgload/pkgload.go
+++ b/tool/internal/pkgload/pkgload.go
@@ -207,8 +207,8 @@ func ResolveModule(ctx context.Context, pkgDir string) (*packages.Module, error)
 	return pkg.Module, nil
 }
 
-// ResolveModuleDir returns the module directory for a given package directory.
-func ResolveModuleDir(ctx context.Context, pkgDir string) (string, error) {
+// resolveModuleDir returns the module directory for a given package directory.
+func resolveModuleDir(ctx context.Context, pkgDir string) (string, error) {
 	mod, err := ResolveModule(ctx, pkgDir)
 	if err != nil {
 		return "", err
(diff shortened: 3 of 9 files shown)


darkraider01:
Superseded by #1146 — closing this triggered the fork's auto-delete-branch setting, and GitHub won't let a PR be reopened once its head branch is deleted, even with an identically-named branch pushed back. #1146 has the same content plus the `instrument.InstrumentPhase`/`TJump` rename this one was missing.

PR #1152, opened 2026-08-18
fix(grpc): HandleRPC must not touch caller span on skipped RPCs

### LINKED ISSUE #1147: grpc: HandleRPC ends and overwrites caller span on skipped OTLP export calls
### Description

When `TagRPC` opts out of instrumenting OTLP export RPCs via `grpcsemconv.IsOTELExporterPath(info.FullMethodName)`, it returns the caller's `context.Context` unmodified without attaching `gRPCContext` or creating a gRPC span.

However, gRPC's stats handler lifecycle still executes `HandleRPC(ctx, rs)`. In `HandleRPC`:
1. `span := trace.SpanFromContext(ctx)` retrieves the caller's active span.
2. `*stats.OutHeader` and `*stats.End` inspect `if span.IsRecording()` without checking `gctx != nil`.
3. As a result, the caller's span is stamped with `rpc.grpc.status_code` / server address attributes and ended prematurely via `span.End()`.

### Common Reproduction Scenario

This occurs during standard graceful shutdown or flush flows:
```go
ctx, span := tracer.Start(ctx, "shutdown-task")
defer span.End()

tp.ForceFlush(ctx) // OTLP exporter gRPC call executes HandleRPC with ctx and ends 'shutdown-task'

### PR DESCRIPTION
**Pull request by gyanranjanpanda**
## Summary

Fixes #1147 `HandleRPC` in the gRPC client and server stats handlers was silently
corrupting and prematurely ending the caller's active span whenever
`TagRPC` opted out of instrumenting an OTLP export RPC.

## Root cause

`TagRPC` signals "I skipped this RPC" by returning the caller's
`context.Context` unchanged — no `gRPCContext` is attached. The metric
branches in `HandleRPC` (`OutPayload`, `InPayload`, and the `End`
metrics block) already respected this contract with `if gctx != nil`
guards. The span branches (`OutHeader` and the span portion of `End`)
did not: they only checked `span.IsRecording()`.

Because `trace.SpanFromContext(ctx)` was called before the `gctx`
check, on the skipped path it resolved to **the caller's active span**.
Any caller span that happened to be recording would:

- Have `rpc.grpc.status_code` and peer address attributes written onto it.
- Be ended via `span.End()` during the OTLP export RPC.

This is silently destructive: no error, no log — the span simply
reports the wrong duration, wrong attributes, and is exported at the
wrong time.

## Reproduction

The standard graceful-shutdown pattern triggers this:

```go
ctx, span := tracer.Start(ctx, "graceful-shutdown")
defer span.End()
tp.ForceFlush(ctx)  // → OTLP export gRPC call → HandleRPC ends "graceful-shutdown" here
```

`tp.ForceFlush(ctx)` → `bsp.exportSpans(ctx)` → `UploadTraces(ctx)` →
gRPC Export. The context carries the caller span through because
`exportContext` derives from the caller's `ctx`. The OTLP exporter uses
the instrumented gRPC connection, so our `HandleRPC` fires with a
context that still has the caller's recording span on it.

## Fix

Add an early return at the top of `HandleRPC` when `gctx == nil`:

```go
gctx, _ := ctx.Value(gRPCContextKey{}).(*gRPCContext)
if gctx == nil {
    return
}
span := trace.SpanFromContext(ctx)
```

This makes all branches consistent with the existing metric guards and
ensures the caller's span is never reachable on the skipped path.
The `span` lookup is hoisted below the guard so it cannot be resolved
against a caller-owned span.

The redundant per-branch `if gctx != nil` guards on `OutPayload` and
`InPayload` are removed since `gctx` is guaranteed non-nil past the
early return.

## Note on #1104

Issue #1104 proposes decoupling `span.End()` from the
`span.IsRecording()` guard. If that change is made without the `gctx`
guard established here, the skipped path would unconditionally call
`span.End()` on the caller's span. The early return makes that safe.

## Tests

Two regression tests are added — one per package:

- `TestClientStatsHandler_SkippedRPC_DoesNotTouchCallerSpan`
- `TestServerStatsHandler_SkippedRPC_DoesNotTouchCallerSpan`

Each test:
1. Starts a real recording caller span (`"graceful-shutdown"`).
2. Calls `TagRPC` with an OTLP export method name (skip path).
3. Drives all five `HandleRPC` event types through the skipped context.
4. Asserts the caller span is **still recording** (not ended).
5. Ends the span explicitly and asserts only 1 span is exported with
   the correct name and no `rpc.grpc.status_code` attribute.

These tests would have failed against the code before this fix.

## Affected files

| File | Change |
|------|--------|
| `instrumentation/google.golang.org/grpc/client/client_hook.go` | Early-return guard + restructure |
| `instrumentation/google.golang.org/grpc/server/server_hook.go` | Same |
| `instrumentation/google.golang.org/grpc/client/client_hook_test.go` | Regression test |
| `instrumentation/google.golang.org/grpc/server/server_hook_test.go` | Regression test |


Labreo:

Labreo:
Looks good to me.Nice fix.

darccio (approved):
LGTM

darccio:
@gyanranjanpanda Regenerate the bundle, please.

gyanranjanpanda:
i did it could @darccio check it now

PR #1153, opened 2026-08-18
feat(instrumentation/genai): add compile-time instrumentation for google.golang.org/genai

prabindersinghh:
## Description

Adds compile-time instrumentation for the Google Gen AI Go SDK (`google.golang.org/genai`), so applications using Gemini get GenAI client spans without any code changes.

The SDK is the single Go client for both of Google's generative AI surfaces, the Gemini Developer API and Vertex AI. Both route `GenerateContent` through the same `generateContent` HTTP method, so one instrumentation covers both. `client.Chats` sessions are covered too, since they issue the same requests underneath.

What is included:

- Rule (`otelc.yaml`) matching `genai.NewClient`, with an after hook.
- Hook and tracing `http.RoundTripper` under `instrumentation/google.golang.org/genai/`.
- Unit tests for the hook and the transport.
- Integration test in `test/integration/` that builds an instrumented binary against a mock endpoint and asserts the emitted span.
- Telemetry contract in `schemas/otelc/groups/gemini.yaml`.
- Docs and supported-library table updates.

### Why an after hook rather than a before hook

`NewClient` fills in a nil `ClientConfig.HTTPClient` itself, and for Vertex AI without explicit credentials the client it builds carries application default credentials. A before hook that substituted its own client would break that authentication. Hooking the return value means the instrumentation decorates whatever client the SDK settled on, so authentication is left alone.

`Client.ClientConfig()` returns a copy of the config, but `HTTPClient` is a pointer that the copy shares with the internal API client, so replacing its transport takes effect for every subsequent call. Clients sharing one `*http.Client` are wrapped once, not once per client.

### Provider detection

`gen_ai.provider.name` and `gen_ai.system` are read from the backend the client was configured for rather than guessed from the request host, so the Gemini Developer API reports `gcp.gemini` / `gemini` and Vertex AI reports `gcp.vertex_ai` / `vertex_ai`. This avoids the host-matching problem that #824 fixed for the other providers.

### Token accounting

Two Gemini specific details are normalised so the span means the same thing it does for the other GenAI instrumentations:

- `thoughtsTokenCount` is generated and billed as output but reported separately from `candidatesTokenCount`, so the two are summed into `gen_ai.usage.output_tokens`.
- `promptTokenCount` already covers cached content, so it maps onto `gen_ai.usage.input_tokens` unchanged. This is unlike Anthropic's `input_tokens`, which excludes cache reads and has to be folded back together.

`gen_ai.usage.total_tokens` uses the API's `totalTokenCount` when present and is otherwise derived.

### Not included

- Streaming (`streamGenerateContent`) passes through without a span. Usage totals only arrive in the final server sent event and this does not accumulate SSE yet. Same gap as #679 left for Anthropic, and #970 tracks the wider SSE work.
- `countTokens`, `embedContent`, image and video generation, and the remaining model RPCs are not instrumented.
- Spans only, no metrics.
- Prompt and response content is never captured. The semantic conventions gate that behind `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`, which is not implemented here, so metadata only is the only behaviour.

`error.type` uses the numeric status code, following #976 rather than the older full status string that the Anthropic middleware still uses.

## Motivation

Gemini is one of the three major model providers, and otelc already instruments the other two (`openai-go` and `anthropic-sdk-go`). Go services calling Gemini currently get either nothing or a bare `net/http` client span with no model, token usage, or finish reason on it, which leaves a visible hole when a service calls more than one provider and you want to compare cost and latency across them.

This fills that hole using the same shape as the existing GenAI instrumentations, so the three providers emit comparable spans.

---

## Checklist

- [ ] Tests pass: `make test`

[^I-know-my-stuff]:


darccio:
@prabindersinghh Is this related to any issue?

prabindersinghh:
No, not to a dedicated one. That's an omission on my part: I left the `Fixes #` line in the template empty instead of filling it in.

The closest precedent I found is #679, which tracks the same work for `anthropic-sdk-go`, so the convention here looks like one issue per provider. I'm happy to open the equivalent for `google.golang.org/genai` and link it here. If you'd rather it hang off #709, which scopes GenAI provider coverage more broadly, that works too.

Just say which you prefer and I'll sort it out.

vedant21-oss (commented):
Hi @prabindersinghh, I went through this one properly since it overlaps with the GenAI
work I'm doing in #1370. Overall this is a really clean PR — the body preservation, the
nil-Transport fallback and the double-wrap guard are all handled, and the tests are
better than most. A few things below, mostly questions.

**The branch won't merge right now**

It's 87 commits behind `main` and conflicts in three files. Two of them regenerate, so it's
less painful than it sounds:

1. **`CHANGELOG.md`** — this one's a real delete/modify conflict. The file was removed from
   the repo root in #1240, so the entry this PR adds has nowhere to live. Just drop that hunk.
2. **`tool/data/instrumentation-manifest.json`** — a pile of new instrumentations landed
   since you branched (amqp #1349, elastic #1340, zap #1339, gocql #809, redis #1369). Don't
   hand-merge it, `make manifest` regenerates it.
3. **`tool/data/otelc-bundle.tgz`** — the usual binary one. `make package` after everything
   else is merged.

If you haven't already, `make setup-git` registers the merge driver that lets the bundle
conflict resolve so the merge can complete; then regenerate it properly afterwards.

**Streaming is the big gap**

`streamGenerateContent` is passed straight through with no span. The comment explains why
(usage only arrives in the last SSE event), and I agree that's better than emitting a span
with no response attributes.

But the openai-go module already solved exactly this — `newStreamingReader` in
`instrumentation/github.com/openai/openai-go/v3/middleware.go` wraps the response body and
finishes the span when the stream ends. Is there a reason that can't be reused here, or is
it just scope? I'm not asking you to grow this PR, but `GenerateContentStream` is what most
chat apps actually call, so if it's a follow-up it'd be good to have an issue tracking it so
it doesn't get lost.

**The SSE branch in the non-streaming path looks unreachable**

In `RoundTrip`, after `classifyOperation` has already sent `streamGenerateContent` down the
pass-through path, there's still a check for `Content-Type: text/event-stream` that sets
`gen_ai.request.is_stream`. As far as I can tell a plain `generateContent` never answers
with SSE, so that branch can't be hit. Is it defensive, or does Vertex do something here I
don't know about? If it's defensive a one-line comment saying so would help, because right
now `TestRoundTrip_SSEResponseMarkedAsStream` makes it look like a real path.

**Another copy of semconv/genai.go**

This isn't your fault and I don't think it should block the PR, but flagging it because
it's literally what I'm writing an RFC about.

`instrumentation/google.golang.org/genai/semconv/genai.go` is now the fifth hand-written
copy of this file (three openai modules, anthropic, now this). I checked all 15 keys
against `go.opentelemetry.io/otel/semconv/v1.37.0` — the version `transport.go` already
imports as `otelsemconv` for `ErrorType` — and **12 of the 15 already exist there**:

- provider.name, operation.name, request.model, request.max_tokens, request.temperature,
  request.top_p, request.top_k, response.id, response.model, response.finish_reasons,
  usage.input_tokens, usage.output_tokens → all in upstream v1.37.0
- only `gen_ai.system`, `gen_ai.request.is_stream` and `gen_ai.usage.total_tokens` are
  genuinely local

Your registry file gets this right, by the way — it `ref:`s everything and reuses the two
`id:` declarations from `groups/openai.yaml` instead of re-declaring them. It's just the Go
side that duplicates. Matching openai is the correct call for this PR; I'm working on
generating these from the registry so this stops happening, and I'll make sure this module
is covered.

**Small thing: the Transport swap isn't synchronised**

`AfterNewClient` reads `httpClient.Transport`, checks the type, then writes it back. If an
app calls `genai.NewClient` twice from different goroutines with the same
`ClientConfig.HTTPClient`, that's a data race on the caller's client. Probably rare enough
not to care about, but worth knowing about. Related: when the caller *does* supply their own
`*http.Client` that they also use for non-Gemini traffic, this puts `otelTransport` in front
of all of it. `classifyOperation` makes that a cheap no-op so I don't think it's a problem,
just noting the mutation lands on an object the caller owns.

**Stuff I checked that's fine, so you don't have to re-explain it**

- `roundTripper()` falling back to `http.DefaultTransport` is necessary, not just defensive —
  the SDK really does leave `Transport` nil (`client.go:382` in genai v1.66.0 does
  `cc.HTTPClient = &http.Client{}`).
- The nil-`HTTPClient` early return can't actually fire, since `NewClient` always populates
  it, but it's harmless.
- `gcp.gemini` and `gcp.vertex_ai` both match the upstream `gen_ai.provider.name` enum.
- The `GetBody` / `TeeReader` + `MultiReader` handling matches openai's byte for byte, and
  the tests assert both bodies survive.
- Folding `thoughtsTokenCount` into output tokens is right and it's tested both ways.
- `SuppressHTTPClientInstrumentation` is there and the integration test's
  `RequireSingleSpan` should catch it if it ever goes missing.

One test suggestion: there's nothing covering a response body larger than
`maxResponseBodySize`. That's the path where the tee/MultiReader reassembly could drop bytes,
and it'd be a nasty one to find in production. A 4MB+ body test that just asserts the caller
still reads the whole thing would pin it.

Nice work on this — the comments explaining *why* are exactly what this repo asks for.

PR #1155, opened 2026-08-18
fix(tool/internal/setup): ensure deterministic rule load order

### LINKED ISSUE #1144: bug(tool/internal/setup): rule load order is nondeterministic, producing different instrumented output between identical builds
### Describe the bug

`parseRuleFromYaml` in `tool/internal/setup/match.go` iterates the unmarshalled YAML rule map directly and appends to the returned slice in that order:

```go
func parseRuleFromYaml(content []byte) ([]rule.InstRule, error) {
	var h map[string]map[string]any
	err := yaml.Unmarshal(content, &h)
	...
	rules := make([]rule.InstRule, 0)
	for name, fields := range h {   // <- randomized map iteration
		...
		rules = append(rules, r)
	}
	return rules, nil
}
```

Go randomizes map iteration order, so the returned rule order differs between runs of the same input. Nothing sorts it afterwards: both call sites (`loadRulesFromToolFiles`, and the `-rules` path) append it straight into the rule set, and `instrument()` applies rules to a file in slice order.

The `-rules` path adds a second layer of the same problem at `match.go:480`:

```go
return slices.Concat(slices.Collect(maps.Values(ruleSet))...), nil
```

Because rule application order is semantically meaningful — multiple func rules on one function nest their trampoline-jump-ifs in application order — **the instrumented output is not reproducible**. Two identical builds of the same project with the same rule files can produce different generated code.

### Environment

 - OS: Linux (WSL2 / Ubuntu); not platform-specific — this is map iteration order
 - Go Version: 1.25.7
 - Version: `d95997a` (`main`)

### To Reproduce

**1. Show the parse order itself varies.** Add this to `tool/internal/setup/`:

```go
func TestParseRuleFromYamlOrder(t *testing.T) {
	names := []string{"aaa", "bbb", "ccc", "ddd", "eee", "fff", "ggg", "hhh"}
	yamlSrc := ""
	for _, n := range names {
		yamlSrc += fmt.Sprintf(
			"rule_%s:\n  target: main\n  where:\n    func: Func1\n  do:\n    - inject_hooks:\n        before: %sBefore\n        path: p\n",
			n, n)
	}

	seen := map[string]int{}
	for range 40 {
		rules, err := parseRuleFromYaml([]byte(yamlSrc))
		if err != nil {
			t.Fatalf("parse failed: %v", err)
		}
		order := ""
		for _, r := range rules {
			order += r.GetName() + " "
		}
		seen[order]++
	}
	if len(seen) > 1 {
		t.Errorf("identical input produced %d different rule orders", len(seen))
	}
}
```

Result — 40 parses of byte-identical YAML:

```text
distinct rule orders over 40 parses of the SAME yaml input: 8
   order: rule_fff rule_ggg rule_hhh rule_aaa rule_bbb rule_ccc rule_ddd rule_eee
   order: rule_ddd rule_eee rule_fff rule_ggg rule_hhh rule_aaa rule_bbb rule_ccc
   order: rule_aaa rule_bbb rule_ccc rule_ddd rule_eee rule_fff rule_ggg rule_hhh
```

**2. Show it reaches generated code.** The golden harness hides this, because `instrument_test.go:186` sorts rule names before building the rule set — with the comment *"Sort rule names to ensure deterministic order in tests"*. Production has no such sort.

Remove that one line so the harness matches production, then run the existing `multiple-func-rules` fixture (two rules, `hook_func_1` and `hook_func_2`, both targeting `Func1`) repeatedly:

```
go test -count=1 -run 'TestInstrumentation_Integration/multiple-func-rules' ./tool/internal/instrument/
```

Over 12 runs: **11 passed, 1 failed**. The failure is a swap of the two trampolines:

```text
-	if hookContext3592294264, _ := OtelBeforeTrampoline_Func13592294264(&p1, &p2); false {
+	if hookContext1830170046, _ := OtelBeforeTrampoline_Func11830170046(&p1, &p2); false {
-		if hookContext1830170046, _ := OtelBeforeTrampoline_Func11830170046(&p1, &p2); false {
+		if hookContext3592294264, _ := OtelBeforeTrampoline_Func13592294264(&p1, &p2); false {
```

The nesting order of the trampoline-jump-ifs inverts, and the declaration order in the generated globals file inverts with it. Same source, same rules, different output.

### Expected behavior

Rules load in a stable order, so a given set of rule files always produces byte-identical instrumented output.

### Additional context

**Why this has gone unnoticed.** The golden suite sorts rule names in its own harness, so CI is deterministic by construction and can never surface this. The intermittency in step 2 (1 failure in 12) is also characteristic: for a two-key map Go's randomization does not flip the order on every run, so a naive repeat-run check usually passes.

**Severity.** This is the same class the project has already fixed twice — #986 (sorting `moduleDirs` in `generatePinnedProjects`/`getBackupFiles`) and #971 (sorting `moduleDirs` in `resolveRulePaths`) — and more recently #1070 for `instrument()`'s file iteration. Unlike #1070, which only changed which error message surfaced, this one changes the generated code, so it affects build reproducibility and any downstream artifact hashing or caching that assumes identical inputs give identical output.

**Suggested fix.** Sort the rule names in `parseRuleFromYaml` before building the slice (`slices.Sorted(maps.Keys(h))`), and give `match.go:480` a deterministic order rather than `maps.Values`. Once production is deterministic, the sort in `instrument_test.go` can stay as a belt-and-braces measure or be dropped — but it should no longer be the only thing providing the guarantee.

**Related, not filed separately.** `StateManager.Revert()` (`tool/internal/setup/state.go`) iterates `s.files` unsorted while `Commit()` in the same file explicitly sorts "for deterministic behavior". That one only affects the ordering of joined error messages on a partially-failed revert, so it is far less serious, but it is the same inconsistency and could be swept up by the same PR.

shivv23:
## Description

`parseRuleFromYaml` iterates over a Go map (`for name, fields := range h`) which has nondeterministic iteration order. This causes rule application order to vary across runs, leading to non-reproducible instrumentation behavior. Similarly, `loadCustomRules` collects rule values via `slices.Concat(slices.Collect(maps.Values(ruleSet))...)` which also depends on map iteration order.

Both functions now iterate over map keys in sorted (lexicographic) order using `slices.Sorted(maps.Keys(...))`.

## Motivation

Reproducible builds and deterministic instrumentation output are important for debugging and correctness. With nondeterministic map iteration, the same YAML rule file can produce different instrumentation results across runs.

Fixes #1144

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### DIFF
--- a/tool/internal/setup/match.go
+++ b/tool/internal/setup/match.go
@@ -62,7 +62,8 @@ func parseRuleFromYaml(content []byte) ([]rule.InstRule, error) {
 		return nil, ex.Wrap(err)
 	}
 	rules := make([]rule.InstRule, 0)
-	for name, fields := range h {
+	for _, name := range slices.Sorted(maps.Keys(h)) {
+		fields := h[name]
 		flatRules, normErr := rule.Normalize(fields)
 		if normErr != nil {
 			return nil, normErr
@@ -477,7 +478,15 @@ func loadCustomRules(ruleConfig string) ([]rule.InstRule, error) {
 		}
 	}
 
-	return slices.Concat(slices.Collect(maps.Values(ruleSet))...), nil
+	total := 0
+	for _, rules := range ruleSet {
+		total += len(rules)
+	}
+	result := make([]rule.InstRule, 0, total)
+	for _, name := range slices.Sorted(maps.Keys(ruleSet)) {
+		result = append(result, ruleSet[name]...)
+	}
+	return result, nil
 }
 
 func loadRulesFromToolFiles(ctx context.Context, toolFiles []string) ([]rule.InstRule, error) {
--- a/tool/internal/setup/match_test.go
+++ b/tool/internal/setup/match_test.go
@@ -1168,6 +1168,60 @@ func Target(value string) error { return nil }
 	assert.Equal(t, "matching", matchedFuncRules[0].Name)
 }
 
+func TestParseRuleFromYamlDeterministicOrder(t *testing.T) {
+	yamlContent := []byte(`
+zebra:
+  target: main
+  func: Example
+  raw: "_ = 1"
+alpha:
+  target: main
+  func: Example
+  raw: "_ = 1"
+mangle:
+  target: main
+  func: Example
+  raw: "_ = 1"
+`)
+
+	rules, err := parseRuleFromYaml(yamlContent)
+	require.NoError(t, err)
+	require.Len(t, rules, 3)
+
+	names := make([]string, len(rules))
+	for i, r := range rules {
+		names[i] = r.GetName()
+	}
+	require.Equal(t, []string{"alpha", "mangle", "zebra"}, names)
+}
+
+func TestLoadCustomRulesDeterministicOrder(t *testing.T) {
+	content := `zebra:
+  target: main
+  func: Example
+  raw: "_ = 1"
+alpha:
+  target: main
+  func: Example
+  raw: "_ = 1"
+mangle:
+  target: main
+  func: Example
+  raw: "_ = 1"`
+
+	p := writeCustomRules(t, "order.yaml", content)
+
+	rules, err := loadCustomRules(p)
+	require.NoError(t, err)
+	require.Len(t, rules, 3)
+
+	names := make([]string, len(rules))
+	for i, r := range rules {
+		names[i] = r.GetName()
+	}
+	require.Equal(t, []string{"alpha", "mangle", "zebra"}, names)
+}
+
 func TestRunMatch_EmptyRules(t *testing.T) {
 	dep := &Dependency{
 		ImportPath: "example.com/noop",


darkraider01:
Hey, heads up — I opened #1144 and commented there wanting to take it, but it's not actually assigned yet. Per the intake discussion in #973 (maintainers only want work to start once an issue's been marked ready/assigned), might be worth pinging a maintainer to sort out who's picking this up before sinking more time into the PR.

Labreo:
@darkraider01 I think it is fine as #973 is still in discussion as of now.And the change itself seems to be trivial.

Labreo:
@shivv23 .Just a suggestion for the future.Please follow the [pull request template](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/.github/PULL_REQUEST_TEMPLATE.md).It helps contributors and maintainers get a better idea for the pull request.And allows them to manage and merge your pull request faster.

Labreo (commented):
Looks fine.Just that the commit title is not according to the repository's conventional commits convention.It should be scoped accordingly.For this it could probably be` fix(tool/internal/setup): ensure deterministic rule load order in parseRuleFromYaml and loadCustomRules.`

shivv23:
@Labreo All done! Please have a look.

darccio:
@darkraider01 You decide. Usually we close the PR in favour of the issue's assignee but this PR seems complete and there isn't another PR to defer.

darkraider01:
All good, no worries , I never got around to opening a PR for [#1144](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/1144), so nothing to defer to on my end. Go ahead with this one. @darccio

tool/internal/setup/match.go
```diff
@@ -477,7 +478,11 @@ func loadCustomRules(ruleConfig string) ([]rule.InstRule, error) {
 		}
 	}
 
-	return slices.Concat(slices.Collect(maps.Values(ruleSet))...), nil
+	result := make([]rule.InstRule, 0, len(ruleSet))
+	for _, name := range slices.Sorted(maps.Keys(ruleSet)) {
+		result = append(result, ruleSet[name]...)
+	}
```
darccio: `len(ruleSet)` undercounts capacity here — `ruleSet` maps each name to a *slice* of rules (a single YAML entry can expand into several), so the real element count is the sum of each slice's length, not the map's key count. Not a correctness bug (the slice still grows), just a capacity hint that doesn't do what it looks like it does.

```suggestion
	total := 0
	for _, rules := range ruleSet {
		total += len(rules)
	}
	result := make([]rule.InstRule, 0, total)
	for _, name := range slices.Sorted(maps.Keys(ruleSet)) {
		result = append(result, ruleSet[name]...)
	}
```

tool/internal/setup/match_test.go
```diff
@@ -1168,6 +1168,33 @@ func Target(value string) error { return nil }
...
+	rules, err := parseRuleFromYaml(yamlContent)
+	require.NoError(t, err)
+	require.Len(t, rules, 3)
+
+	names := make([]string, len(rules))
+	for i, r := range rules {
+		names[i] = r.GetName()
+	}
+	require.Equal(t, []string{"alpha", "mangle", "zebra"}, names)
+}
```
darccio, line 1196: This locks down `parseRuleFromYaml`'s ordering, but the bug in #1144 was demonstrated to reach further than that function — the issue's step-2 repro shows the trampoline nesting in *generated code* flips depending on rule order, and that only surfaces because `instrument_test.go` sorts rule names before the golden comparison (masking production's behavior in tests). Consider also covering `loadCustomRules`'s ordering, or temporarily dropping that harness-side sort in an integration test to assert the golden output stays byte-identical across repeated runs — closer to what actually broke.

darccio (changes requested):

tool/internal/setup/match.go
shivv23: Fixed

tool/internal/setup/match_test.go
shivv23: Fixed

shivv23:
@darccio All done. Please have a look.

darccio (approved):

PR #1157, opened 2026-08-18
fix(database/sql): parse non-parenthesized MySQL Unix-socket DSNs

### LINKED ISSUE #1131: fix(database/sql): parse non-parenthesized MySQL Unix-socket DSNs

ipsitapp8:
## Description

`parseMySQLDSN` splits the address from the database name at the first `/`. That works for tcp/host addresses since they never contain a `/`, but not for a unix socket, where the address itself is an absolute path.

For `unix:/tmp/mysql.sock/db`, the first `/` is the one right after `unix:`, so the parser split it into address `unix:` and db `/tmp/mysql.sock/db`. `mysqlStripProtocol` then saw an empty address after the protocol and returned an empty host, and the db name came out as `tmp/mysql.sock/db` instead of `db`.

## Fix

Added `isMySQLUnixAddr`, which checks the non-parenthesized address for an exact `unix:` prefix. When it matches, the address/database separator is the last `/` in the DSN instead of the first, since the socket path itself owns every `/` before that one. Every other non-parenthesized form (tcp, bare host, bare port) has no `/` in the address at all, so first and last agree there and nothing changes for them.

The parenthesized form (`unix(/tmp/mysql.sock)/db`) was already correct and is untouched.

## Tests

Added `TestParseDSN_MySQLUnixSocketNonParenthesized`, covering:
- `unix:/tmp/mysql.sock/db`
- `user:pass@unix:/tmp/mysql.sock/mydb`
- `unix:/tmp/mysql.sock/` (trailing slash, empty db)
- a socket path nested under multiple directories
- the parenthesized form, as a regression check
- a protocol name that starts with "unix" but isn't it (`unixish`), so the match stays exact
- tcp and bare host forms, to confirm they still use the first `/`

Also ran the existing `FuzzParseDSN` for 600k+ executions after the change, no crashes.

One thing I left alone on purpose: `unix:/tmp/mysql.sock` with no trailing slash at all is ambiguous, the last `/` would land inside the socket path itself instead of at a real separator. The issue flagged this as needing confirmation before encoding, so I didn't add a test or a rule for it. Behavior there is unchanged from before this fix.

Fixes #1131

---


instrumentation/database/sql/dsnparse/parse.go
```diff
@@ -310,7 +310,16 @@ func parseMySQLDSN(dsn string) DSNInfo {
...
+		// A unix-socket address is itself an absolute filesystem path and so
+		// contains '/' before the database name does, meaning the first '/'
+		// belongs to the address, not the separator. Every other form here
+		// (tcp, bare host, bare port) has no '/' in the address at all, so
+		// using the last '/' there gives the same split as the first.
+		sep := strings.IndexByte
+		if isMySQLUnixAddr(rest) {
+			sep = strings.LastIndexByte
+		}
+		if sl := sep(rest, '/'); sl >= 0 {
```
darccio: For a non-parenthesized unix-socket DSN, the separator search runs against
`rest` before the query string is stripped. If a param value contains `/`,
`LastIndexByte` lands inside the query instead of at the real separator:

`unix:/tmp/mysql.sock/db?tls=a/b` → Host=`/tmp/mysql.sock/db?tls=a`, DB=`b`
(both wrong; should be Host=`/tmp/mysql.sock`, DB=`db`).

Limiting the search to the portion before `?` fixes this without touching
the parenthesized branch or any non-unix form.

```suggestion
		sep := strings.IndexByte
		searchIn := rest
		if isMySQLUnixAddr(rest) {
			sep = strings.LastIndexByte
			if q := strings.IndexByte(rest, '?'); q >= 0 {
				searchIn = rest[:q]
			}
		}
		if sl := sep(searchIn, '/'); sl >= 0 {
```

instrumentation/database/sql/dsnparse/parse.go
```diff
@@ -324,6 +333,14 @@ func parseMySQLDSN(dsn string) DSNInfo {
 	return DSNInfo{Host: host, Port: port, DBName: dbName}
 }
 
+// isMySQLUnixAddr reports whether a non-parenthesized MySQL address begins
+// with the "unix:" protocol keyword, meaning what follows the colon is a
+// filesystem path rather than a host[:port].
+func isMySQLUnixAddr(addr string) bool {
+	colon := strings.IndexByte(addr, ':')
+	return colon >= 0 && addr[:colon] == "unix"
+}
```
darccio, line 346: `isMySQLUnixAddr` re-derives the protocol name from the first colon, which
duplicates the classification `mysqlStripProtocol` already does a few lines
below against `mysqlKnownProtocols`. Not a bug, but "what counts as the
protocol" is now encoded in two places that could drift apart. Worth folding
into a single lookup — e.g. have `mysqlStripProtocol` return the matched
protocol name so both call sites share it, or have `isMySQLUnixAddr` check
`mysqlKnownProtocols[addr[:colon]] == "unix"` explicitly.

darccio (changes requested):

darccio:
There's a sibling PR, #1132, fixing the same issue (#1131) with an equivalent approach (last-`/` separator for non-parenthesized `unix:` addresses). One thing it has that this PR doesn't: an integration test in `test/integration/db_client_test.go` asserting the emitted span attributes end-to-end for a unix-socket DSN:

```go
{
    name:       "MySQL non-parenthesized unix socket",
    driverName: "mysql",
    dsn:        "user:pass@unix:/tmp/mysql.sock/inventory",
    wantAddr:   "/tmp/mysql.sock",
    wantPort:   0,
    wantDb:     "inventory",
},
```

Worth cherry-picking that case into this PR's integration suite alongside the unit tests already here — it's the one thing 1132 verifies that this PR currently doesn't (that the parsed `DSNInfo` actually flows through to the correct span attributes, not just the parser output in isolation). I'd suggest closing 1132 in favor of this PR once that's added, since this one otherwise has broader unit coverage and already addresses the query-string separator edge case.

ipsitapp8:
Fixed the separator bug, thanks for the clear repro. The LastIndexByte search now stops at the query string so a / inside a param value like tls=a/b doesn't get picked up as the host/db split anymore. Added a test with your exact example (unix:/tmp/mysql.sock/db?tls=a/b) to cover it. Left the isMySQLUnixAddr/mysqlStripProtocol duplication alone for now since it's not causing a bug, happy to fold those together in a follow up if you'd rather do it here.

ipsitapp8:
Added the integration case you asked for, in the `DSN Parsing` table in `test/integration/db_client_test.go`:

```go
{
    name:       "MySQL non-parenthesized unix socket",
    driverName: "mysql",
    dsn:        "user:pass@unix:/tmp/mysql.sock/inventory",
    wantAddr:   "/tmp/mysql.sock",
    wantPort:   0,
    wantDb:     "inventory",
},
```

Confirmed the parser produces exactly that before wiring up the assertion: `Host="/tmp/mysql.sock"`, `Port=""`, `DBName="inventory"`. So the end-to-end path from `DSNInfo` through to `server.address` / `db.namespace` is covered now, not just the parser in isolation.

Also rebased onto main and regenerated the bundle, so the conflict is cleared. I see #1132 is already closed, so this should be everything that was outstanding.

darccio (approved):

PR #1158, opened 2026-08-18
fix(instrumentation/mongo): stop injected CommandMonitor from overriding a user's Monitor

### LINKED ISSUE #1148: bug(instrumentation/mongo): injected OTel CommandMonitor can silently override a user-supplied Monitor
### Describe the bug

`BeforeConnect`/`BeforeNewClient` inject the OTel `CommandMonitor` into `mongo.Connect`/`mongo.NewClient` calls by looping over the variadic `opts ...*options.ClientOptions` and calling `opt.SetMonitor(monitor)` on any struct whose own `Monitor` field is `nil`:

```go
// instrumentation/go.mongodb.org/mongo-driver/mongo/client_hook.go
// (identical logic in BeforeConnect and BeforeNewClient)
for _, opt := range opts {
    if opt == nil {
        continue
    }
    if opt.Monitor == nil {
        opt.SetMonitor(monitor)
    }
}
```

This checks each `*ClientOptions` struct in isolation, but `mongo.NewClient`/`mongo.Connect` don't use the structs in isolation — they call `options.MergeClientOptions(opts...)`, which folds the whole slice into one struct by taking the **last non-nil** value of each field, in slice order:

```go
// go.mongodb.org/mongo-driver/mongo/options/clientoptions.go (MergeClientOptions)
if opt.Monitor != nil {
    c.Monitor = opt.Monitor
}
```

So if the caller passes multiple `*ClientOptions` and their own `CommandMonitor` is set on one that is **not the last struct in the slice**, the hook injects the OTel monitor into a later, still-`nil` struct — and since that struct comes later, the merge picks the OTel monitor and silently discards the user's own `CommandMonitor`. This is a realistic pattern: the driver's own doc comment on `NewClient` says "later Set* methods will overwrite the values from previous Set* method invocations", which explicitly encourages composing options across multiple structs (e.g. a shared base-options builder plus a per-call URI/override struct).

The bug is present in both driver versions instrumented by this repo (identical loop in both files):
- `instrumentation/go.mongodb.org/mongo-driver/mongo/client_hook.go` (`BeforeConnect`, `BeforeNewClient`)
- `instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go` (same pattern)

The existing test suite doesn't cover this: `client_hook_test.go` only asserts the single-struct case (monitor on the *only* struct is left untouched) and the both-nil case (both structs get injected). It never exercises "user monitor set on an earlier struct, later struct has none" — exactly the ordering `MergeClientOptions` treats specially.

### Environment

- OS: any (bug is in Go logic, not OS-specific)
- Go Version: 1.23+
- mongo-driver Version: v1.17.7 (also applies to the v2 driver instrumentation)

### To Reproduce

Standalone repro against the real `go.mongodb.org/mongo-driver v1.17.7` package (not a mock), replicating the hook's exact injection loop and then running the driver's real `options.MergeClientOptions`:

```go
func injectMonitorLikeHook(opts []*options.ClientOptions, otelMonitor *event.CommandMonitor) []*options.ClientOptions {
	for _, opt := range opts {
		if opt == nil {
			continue
		}
		if opt.Monitor == nil {
			opt.SetMonitor(otelMonitor)
		}
	}
	return opts
}

func TestUserMonitorInEarlierOptionsIsSilentlyDropped(t *testing.T) {
	userMonitor := &event.CommandMonitor{}
	otelMonitor := &event.CommandMonitor{}

	base := options.Client().SetMonitor(userMonitor)
	uriOpts := options.Client().ApplyURI("mongodb://localhost:27017")

	opts := []*options.ClientOptions{base, uriOpts}
	opts = injectMonitorLikeHook(opts, otelMonitor)

	merged := options.MergeClientOptions(opts...)

	if merged.Monitor != userMonitor {
		t.Fatalf("expected user's CommandMonitor to win, got otel monitor overriding it (merged.Monitor == otelMonitor: %v)", merged.Monitor == otelMonitor)
	}
}
```

Result:

```
=== RUN   TestUserMonitorInEarlierOptionsIsSilentlyDropped
    repro_test.go:40: expected user's CommandMonitor to win, got otel monitor overriding it (merged.Monitor == otelMonitor: true)
--- FAIL: TestUserMonitorInEarlierOptionsIsSilentlyDropped (0.00s)
```

Steps against the actual instrumentation:
1. Call `mongo.Connect(ctx, options.Client().SetMonitor(yourMonitor), options.Client().ApplyURI(uri))` in a project built with `otelc`.
2. `BeforeConnect` injects the OTel monitor into the second (`ApplyURI`) struct because its `Monitor` field is `nil`.
3. `mongo.NewClient` merges `opts` and keeps the *last* non-nil `Monitor` — the injected OTel one — discarding `yourMonitor`.

### Expected behavior

Injecting the OTel `CommandMonitor` should never cause an explicitly user-configured `CommandMonitor` to be dropped from the effective (post-merge) options, regardless of how many `*ClientOptions` structs are passed or which one carries the user's monitor. A compile-time instrumentation tool should be a safe, additive drop-in — it should not silently delete a user's pre-existing observability hook with no error or log.

### Additional context

Root cause: the hook decides whether to inject per-struct (`opt.Monitor == nil`), but the driver decides the *effective* value across the whole slice with last-non-nil-wins semantics. These two decision rules disagree whenever the user's `Monitor` isn't on the last struct in `opts`.

Suggested fix direction: don't mutate individual structs based on their own nil-check. Instead, determine the effective merged `Monitor` first (`options.MergeClientOptions(opts...).Monitor`). If it's `nil` (no user monitor anywhere in the slice), append a new trailing `options.Client().SetMonitor(monitor)` to `opts` rather than mutating an existing struct — because it's last, `MergeClientOptions`'s last-non-nil-wins rule guarantees it always takes effect precisely when needed, and it can never be overridden by (or override) any struct already in the slice, since a fresh `options.Client()` leaves every other field `nil`. Apply the same fix to both `mongo-driver/mongo` and `mongo-driver/v2/mongo` instrumentation, and add regression tests for the ordering case above alongside the existing single-struct/both-nil cases.

Happy to send a PR for this if assigned.

bhuvan-somisetty:
## Description

`BeforeConnect`/`BeforeNewClient` in both the v1 and v2 mongo-driver instrumentation injected the OTel `CommandMonitor` by checking each `*options.ClientOptions` struct's own `Monitor` field in isolation:

```go
for _, opt := range opts {
    if opt == nil {
        continue
    }
    if opt.Monitor == nil {
        opt.SetMonitor(monitor)
    }
}
```

`mongo.NewClient`/`mongo.Connect` don't resolve `opts` in isolation, though — they call `options.MergeClientOptions(opts...)`, which folds the whole slice into one struct by keeping the **last non-nil** value of each field, in slice order. If a caller's own `CommandMonitor` was set on a struct that wasn't last in `opts`, this loop would inject the OTel monitor into a later, still-nil struct, and the merge would pick the injected monitor over the user's — silently dropping it.

## Root cause

The hook decided whether to inject per-struct (`opt.Monitor == nil`), but the driver decides the *effective* value across the whole slice with last-non-nil-wins semantics. These two rules disagree whenever the user's `Monitor` isn't on the last struct in `opts`.

I verified this against the real `mongo-driver` `MergeClientOptions` source and with a standalone reproduction before writing the fix — details in #1148.

## Fix

Instead of mutating individual structs, `injectMonitor` now checks the *effective* merged `Monitor` once. If it's already non-nil (a user monitor is present anywhere in `opts`), nothing changes. Otherwise it appends a new trailing `*options.ClientOptions` carrying the OTel monitor. Being last, `MergeClientOptions`'s last-non-nil-wins rule guarantees it always applies when needed, and — since a fresh `options.Client()` leaves every other field `nil` — it can never override, or be overridden by, anything the caller already passed in.

Applied identically to both `instrumentation/go.mongodb.org/mongo-driver/mongo` and `instrumentation/go.mongodb.org/mongo-driver/v2/mongo`, which had the same logic.

## Testing

- Updated existing unit tests for the new "append a trailing struct" behavior (previously asserted injection into every struct, which is exactly what caused the bug).
- Added a regression test for the reported scenario: a user monitor on an earlier `ClientOptions` struct, with a later struct left at its zero value — asserts the *merged* result still keeps the user's monitor.
- `go test ./...` passes for both `instrumentation/go.mongodb.org/mongo-driver/mongo` and `instrumentation/go.mongodb.org/mongo-driver/v2/mongo`.
- `golangci-lint run` on both packages reports only pre-existing issues (globals, local `replace` directives, existing function signatures) that are unchanged by this diff — verified by comparing against the pre-fix code.

## Impact

Without this fix, a user with their own `CommandMonitor` (e.g. existing APM/logging) silently loses it the moment `otelc` instruments their `mongo.Connect`/`mongo.NewClient` call, with no error or log — a correctness regression for a tool whose whole value proposition is being a safe, additive drop-in.

Fixes #1148

---

## Checklist

- [x] Code formatted: `golangci-lint fmt` (no changes needed)
- [x] Linters pass: `golangci-lint run` on both affected packages (only pre-existing, unrelated issues remain)
- [x] Tests pass: `go test ./...` on both affected packages


Labreo:
Nice fix.Just curious though,did you think about chaining both monitors together instead of bailing early when a user one is present? Like, could both the user's monitor and the OTel one fire simultaneously? Not sure if that's practical with the current CommandMonitor interface but would be cool if it is.

bhuvan-somisetty:
Good call, thanks @Labreo! Turns out chaining is straightforward since `event.CommandMonitor` is just a plain struct of three optional callback fields, not an interface. Pushed a change where the injected monitor now wraps and calls both the caller's own `Started`/`Succeeded`/`Failed` and the OTel ones, instead of bailing out when a user monitor is present. Added tests covering the chaining itself plus the two existing scenarios.

bhuvan-somisetty:
Hi @darccio @y1yang0 @NameHaibinZhang — this PR is ready for review whenever you get a chance. It fixes a bug where the injected OTel CommandMonitor could silently drop a caller's own CommandMonitor (or, per @Labreo's suggestion, now chains with it instead). CI is green and the branch has been rebased on main. Thanks!

instrumentation/go.mongodb.org/mongo-driver/mongo/client_hook.go
```diff
@@ -24,33 +25,70 @@ func (g mongoEnabler) Enable() bool {
...
+func injectMonitor(opts []*options.ClientOptions) []*options.ClientOptions {
+	otelMonitor := otelmongo.NewMonitor()
+	monitor := otelMonitor
+	if existing := options.MergeClientOptions(opts...).Monitor; existing != nil {
+		monitor = chainMonitors(existing, otelMonitor)
 	}
+	// Full slice expression forces a new backing array so this never mutates a
+	// caller-owned slice passed in via `opts...`.
+	return append(opts[:len(opts):len(opts)], options.Client().SetMonitor(monitor))
+}
```
darccio, line 92: The appended element is `options.Client()`, which is not field-empty: at the pinned
v1.17.9 it returns `&ClientOptions{HTTPClient: httputil.DefaultHTTPClient}`. Since
`MergeClientOptions` merges `HTTPClient` under the same last-non-nil-wins rule, and
this element is deliberately last, its default `HTTPClient` wins over a custom
`http.Client` the caller set on an earlier element (proxy, custom TLS, timeouts,
OIDC/AWS auth transport).

That's the same silent-override class this PR fixes, displaced from `Monitor` onto
`HTTPClient` — and it's newly introduced here: the old per-struct mutation never
constructed a fresh `options.Client()`, so `HTTPClient` was never touched. It also
makes the doc comment above ("can never be overridden by, or silently override,
anything the caller already passed in") inaccurate as written.

A bare `&options.ClientOptions{Monitor: monitor}` alone isn't sufficient either.
v2's `MergeClientOptions` short-circuits on `len(opts) == 1` and returns `opts[0]`
verbatim, so with empty `opts` the injected element becomes the whole merged result
and `HTTPClient` stays nil — and `newClient` does `client.httpClient =
clientOpts.HTTPClient` with no nil fallback. Carrying the effective `HTTPClient`
forward covers that, and preserves pointer identity so the driver's
`c.httpClient == httputil.DefaultHTTPClient` ownership check still behaves.

```suggestion
func injectMonitor(opts []*options.ClientOptions) []*options.ClientOptions {
	merged := options.MergeClientOptions(opts...)
	otelMonitor := otelmongo.NewMonitor()
	monitor := otelMonitor
	if merged.Monitor != nil {
		monitor = chainMonitors(merged.Monitor, otelMonitor)
	}
	// Set only Monitor, and carry the caller's effective HTTPClient forward, so the
	// appended element is a no-op for every field except Monitor. Using
	// options.Client() here would inject its default HTTPClient which, being last in
	// the merge, silently replaces a caller's custom one.
	injected := &options.ClientOptions{Monitor: monitor, HTTPClient: merged.HTTPClient}
	// Full slice expression forces a new backing array so this never mutates a
	// caller-owned slice passed in via `opts...`.
	return append(opts[:len(opts):len(opts)], injected)
}
```

Worth a regression test alongside the existing ones: caller sets a custom
`HTTPClient` on an earlier element, assert `MergeClientOptions(newOpts...).HTTPClient`
is still that client. No current test asserts any non-`Monitor` field survives
injection, which is why this slipped through.

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go
```diff
@@ -22,31 +25,68 @@ func (g mongoEnabler) Enable() bool {
...
+func injectMonitor(opts []*options.ClientOptions) []*options.ClientOptions {
+	otelMonitor := otelmongo.NewMonitor()
+	monitor := otelMonitor
+	if existing := options.MergeClientOptions(opts...).Monitor; existing != nil {
+		monitor = chainMonitors(existing, otelMonitor)
 	}
+	// Full slice expression forces a new backing array so this never mutates a
+	// caller-owned slice passed in via `opts...`.
+	return append(opts[:len(opts):len(opts)], options.Client().SetMonitor(monitor))
+}
```
darccio, line 92: Same issue as the v1 hook, and v2 is the version where it bites hardest.

v2.8.0's `options.Client()` does `opts.SetHTTPClient(httputil.NewHTTPClient())`, so
each call allocates a *fresh* `http.Client`. Being last in the merge, this element's
client wins over anything the caller configured on an earlier element. v2's
`MergeClientOptions` uses reflection over exported fields with `!field.IsZero()`, so
the non-zero `HTTPClient` is copied over just like any other field.

Applying the same fix as v1 keeps the two hooks byte-identical, which is the pattern
here, and additionally avoids the `len(opts) == 1` short-circuit hazard noted on the
v1 comment (that shortcut only exists in v2).

```suggestion
func injectMonitor(opts []*options.ClientOptions) []*options.ClientOptions {
	merged := options.MergeClientOptions(opts...)
	otelMonitor := otelmongo.NewMonitor()
	monitor := otelMonitor
	if merged.Monitor != nil {
		monitor = chainMonitors(merged.Monitor, otelMonitor)
	}
	// Set only Monitor, and carry the caller's effective HTTPClient forward, so the
	// appended element is a no-op for every field except Monitor. Using
	// options.Client() here would inject its default HTTPClient which, being last in
	// the merge, silently replaces a caller's custom one.
	injected := &options.ClientOptions{Monitor: monitor, HTTPClient: merged.HTTPClient}
	// Full slice expression forces a new backing array so this never mutates a
	// caller-owned slice passed in via `opts...`.
	return append(opts[:len(opts):len(opts)], injected)
}
```

instrumentation/go.mongodb.org/mongo-driver/mongo/client_hook_test.go
```diff
@@ -98,26 +143,75 @@ func TestBeforeConnect(t *testing.T) {
...
+
+			newOpts, ok := mockCtx.GetParam(1).([]*options.ClientOptions)
+			require.True(t, ok)
+			merged := options.MergeClientOptions(newOpts...)
+			require.NotNil(t, merged.Monitor)
+
+			merged.Monitor.Started(context.Background(), &event.CommandStartedEvent{})
+			assert.True(t, userStarted, "user's monitor set on an earlier option must still fire")
+		},
+	)
```
darccio, line 214: No test currently asserts that a caller-set field other than `Monitor` survives
injection — every non-empty `opts` fixture in this file (`options.Client()`,
`options.Client().SetMonitor(...)`, `options.Client().ApplyURI(...)`) leaves every
other field zero. That gap is exactly why the `HTTPClient` override (see the Must
Fix comment on `client_hook.go:73-82`) wasn't caught here.

This test asserts the fix from that comment — it will fail against the current
`injectMonitor` (which starts from `options.Client()`) and should pass once that
suggestion is applied. Needs `"net/http"` added to the import block above.

```suggestion
	)

	t.Run("carries a caller's custom HTTPClient through injection instead of losing it to options.Client()'s default", func(t *testing.T) {
		t.Setenv("OTEL_GO_ENABLED_INSTRUMENTATIONS", "MONGODB")

		customClient := &http.Client{}
		base := options.Client().SetHTTPClient(customClient)
		mockCtx := hooktest.NewMockHookContext(t.Context())

		BeforeConnect(mockCtx, t.Context(), base)

		newOpts, ok := mockCtx.GetParam(1).([]*options.ClientOptions)
		require.True(t, ok)
		merged := options.MergeClientOptions(newOpts...)
		assert.Same(t, customClient, merged.HTTPClient,
			"caller's HTTPClient must survive injection, not be replaced by the appended element's default")
	})
```

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook_test.go
```diff
@@ -98,26 +143,75 @@ func TestBeforeConnect(t *testing.T) {
...
+
+			newOpts, ok := mockCtx.GetParam(0).([]*options.ClientOptions)
+			require.True(t, ok)
+			merged := options.MergeClientOptions(newOpts...)
+			require.NotNil(t, merged.Monitor)
+
+			merged.Monitor.Started(context.Background(), &event.CommandStartedEvent{})
+			assert.True(t, userStarted, "user's monitor set on an earlier option must still fire")
+		},
+	)
```
darccio, line 214: Same gap as the v1 test file, for the same reason: `injectMonitor` is identical
here, so the `HTTPClient`-override bug and this missing regression test both carry
over unchanged. Needs `"net/http"` added to the import block above.

```suggestion
	)

	t.Run("carries a caller's custom HTTPClient through injection instead of losing it to options.Client()'s default", func(t *testing.T) {
		t.Setenv("OTEL_GO_ENABLED_INSTRUMENTATIONS", "MONGODB_V2")

		customClient := &http.Client{}
		base := options.Client().SetHTTPClient(customClient)
		mockCtx := hooktest.NewMockHookContext()

		BeforeConnect(mockCtx, base)

		newOpts, ok := mockCtx.GetParam(0).([]*options.ClientOptions)
		require.True(t, ok)
		merged := options.MergeClientOptions(newOpts...)
		assert.Same(t, customClient, merged.HTTPClient,
			"caller's HTTPClient must survive injection, not be replaced by the appended element's default")
	})
```

instrumentation/go.mongodb.org/mongo-driver/mongo/client_hook.go
```diff
@@ -24,33 +25,70 @@ func (g mongoEnabler) Enable() bool {
 
 var enabler = mongoEnabler{}
 
-// BeforeConnect intercepts mongo.Connect and injects the OTel command monitor
-func BeforeConnect(ictx hook.HookContext, ctx context.Context, opts ...*options.ClientOptions) {
-	if !enabler.Enable() {
-		return
+// chainMonitors returns a CommandMonitor that invokes each non-nil callback of
+// both user and otel in turn (user first), so a caller's own CommandMonitor
+// keeps firing alongside the OTel one instead of being replaced by it.
```
darccio, line 30: Verified against the pinned v1.17.9 source that `event.CommandMonitor` has exactly
these three callback fields today, so this is correct as written — but the function
hardcodes that shape with no note of the assumption. If a future mongo-driver
release adds a fourth callback, it would silently not be chained, the same
"silently dropped" failure mode this PR exists to fix for `Monitor` itself. A short
note makes that assumption visible to the next person touching this function.

```suggestion
// chainMonitors returns a CommandMonitor that invokes each non-nil callback of
// both user and otel in turn (user first), so a caller's own CommandMonitor
// keeps firing alongside the OTel one instead of being replaced by it.
//
// Hardcodes the three CommandMonitor callbacks (Started/Succeeded/Failed). If a
// future mongo-driver release adds another callback field, it must be added here
// too, or it will be silently dropped from the chain.
```

instrumentation/go.mongodb.org/mongo-driver/v2/mongo/client_hook.go
```diff
@@ -22,31 +25,68 @@ func (g mongoEnabler) Enable() bool {
 
 var enabler = mongoEnabler{}
 
-// BeforeConnect intercepts mongo.Connect and injects the OTel command monitor
-func BeforeConnect(ictx hook.HookContext, opts ...*options.ClientOptions) {
-	if !enabler.Enable() {
-		return
+// chainMonitors returns a CommandMonitor that invokes each non-nil callback of
+// both user and otel in turn (user first), so a caller's own CommandMonitor
+// keeps firing alongside the OTel one instead of being replaced by it.
```
darccio, line 30: Same note as the v1 hook — `event.CommandMonitor` has the identical three fields at
the pinned v2.8.0, so today's behavior is correct, but the assumption is worth
naming explicitly since both hook copies share this exact function body.

```suggestion
// chainMonitors returns a CommandMonitor that invokes each non-nil callback of
// both user and otel in turn (user first), so a caller's own CommandMonitor
// keeps firing alongside the OTel one instead of being replaced by it.
//
// Hardcodes the three CommandMonitor callbacks (Started/Succeeded/Failed). If a
// future mongo-driver release adds another callback field, it must be added here
// too, or it will be silently dropped from the chain.
```

darccio (changes requested):

bhuvan-somisetty:
Good catch, thanks — the appended element was carrying its own default `HTTPClient` and silently winning the merge, same failure mode as the `Monitor` bug just moved to a different field. Fixed by only setting `Monitor` on the injected struct and carrying the merged `HTTPClient` forward, added the doc note about `chainMonitors` hardcoding the three callbacks, added the regression tests for both v1 and v2, and rebased onto main to clear the conflicts. Ready for another look.

darccio (approved):

Labreo:
Pulled and checked locally.Looks good to me.

PR #1165, opened 2026-08-19
fix(tool): exclude OS junk files from otelc bundle archive

### LINKED ISSUE #1123: bug(tool): make package archives untracked files, so a stray .DS_Store makes Verify Bundle unfixable
### Describe the bug

`make package` walks the filesystem rather than the git index, and the bundler's only exclusion is `.log`:

```go
// .tools/bundle/main.go
func shouldExclude(name string) bool {
	return strings.HasSuffix(name, ".log")
}
```

So any untracked junk sitting under `pkg/` or `instrumentation/` is archived, including files listed in `.gitignore`. `.DS_Store` is the one that bites, because macOS Finder creates it as soon as a directory is browsed, and `.gitignore` line 11 already hides it from every diff and every `git status`.

The result on a macOS machine is a bundle containing an entry that CI, running on Linux, will never produce. Verify Bundle then fails permanently, and the advice it prints does not help:

```
Binary files a/tool/data/otelc-bundle.tgz and b/tool/data/otelc-bundle.tgz differ
archive verification failed.
To fix: rebase your branch on main, run "make package", and commit the updated archive.
```

Rebasing does nothing, and re-running `make package` re-adds the file every time, because it is still on disk. Nothing in git, the diff, or the job output points at the cause.

### Steps to reproduce

On clean `main`:

```
make package && tar tzf tool/data/otelc-bundle.tgz | wc -l      # 298
touch instrumentation/github.com/redis/go-redis/v9/.DS_Store
make package && tar tzf tool/data/otelc-bundle.tgz | wc -l      # 299
tar tzf tool/data/otelc-bundle.tgz | grep DS_Store              # instrumentation/.../.DS_Store
```

### Seen in the wild

#1074 is currently stuck on exactly this. Its committed archive carries `instrumentation_temp/.DS_Store`, 299 entries against 298 from a fresh build, and the author has pushed twice trying to clear it.

It may also explain #737 and the archive discussions around #585, where the suspicion was that some bundles had been "generated wrongly" without a cause being identified.

### Expected behavior

The archive should depend only on tracked source, so that the same commit produces the same bundle on any machine.

### Suggested fix

Widening `shouldExclude` to cover `.DS_Store`, `Thumbs.db` and similar would fix the immediate case. Excluding anything git ignores, or packaging from the git index rather than the filesystem, would close the class rather than the instance, since the next stray file will behave the same way.

Happy to send a patch if that is useful, though I appreciate the review queue is long at the moment, so equally happy for someone else to pick it up.

ipsitapp8:
Fixes #1123

`make package` walks the filesystem, so any untracked file sitting under `pkg/` or `instrumentation/` gets archived even though it's not part of the source. `.DS_Store` is the common case since macOS Finder writes it as soon as a directory is opened, and it's already in `.gitignore` so nothing about it shows up in git status or diffs. The result is a bundle that depends on what happened to be on the machine that ran `make package`, and Verify Bundle fails with no clue why.

This widens `shouldExclude` to also drop `.DS_Store`, `Thumbs.db`, and `desktop.ini` by exact filename, alongside the existing `.log` suffix check. Added a test for `shouldExclude` and one that reproduces the bug report's repro steps directly (archive a dir with a `.DS_Store` in it, check it's not in the output).

## Test plan
- [x] `go test ./...` in `.tools/bundle`
- [x] `make package` still produces a valid archive

### DIFF
--- a/.tools/bundle/main_test.go
+++ b/.tools/bundle/main_test.go
@@ -8,6 +8,7 @@ import (
 	"compress/gzip"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 )
 
@@ -59,6 +60,35 @@ func TestArchive_ExcludesOSJunk(t *testing.T) {
 	}
 	if !containsSuffix(names, "client.go") {
 		t.Errorf("archive is missing expected source file, got entries: %v", names)
+	}
+}
+
+// TestArchive_PrunesJunkOnlyDir verifies a directory containing only OS junk
+// is treated as effectively empty and omitted from the archive, matching
+// git's behavior for directories with no tracked content.
+func TestArchive_PrunesJunkOnlyDir(t *testing.T) {
+	srcDir := t.TempDir()
+	if err := os.WriteFile(filepath.Join(srcDir, "client.go"), []byte("package redis\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	junkOnlyDir := filepath.Join(srcDir, "empty")
+	if err := os.Mkdir(junkOnlyDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(junkOnlyDir, ".DS_Store"), []byte("junk"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	outPath := filepath.Join(t.TempDir(), "out.tgz")
+	if err := archive(outPath, []string{srcDir}); err != nil {
+		t.Fatalf("archive: %v", err)
+	}
+
+	names := readTarNames(t, outPath)
+	for _, name := range names {
+		if strings.Contains(name, "empty") {
+			t.Errorf("archive contains junk-only directory %q, want it pruned", name)
+		}
 	}
 }
 
--- a/.tools/bundle/main_test.go
+++ b/.tools/bundle/main_test.go
@@ -1,0 +1,99 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestShouldExclude(t *testing.T) {
+	tests := []struct {
+		name string
+		want bool
+	}{
+		{"instrumentation/redis/.DS_Store", true},
+		{".DS_Store", true},
+		{"instrumentation/redis/Thumbs.db", true},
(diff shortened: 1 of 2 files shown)


Labreo:
Nice fix, been a pain for macOS contributors. filepath.Base(name) is the right call so it catches junk files anywhere in the tree without false positives. Tests cover the exact bug scenario. LGTM.

.tools/bundle/main_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func containsSuffix(names []string, suffix string) bool {
+	for _, name := range names {
+		if filepath.Base(name) == suffix {
+			return true
+		}
+	}
+	return false
+}
```
darccio, line 129: No test exercises `isEffectivelyEmpty` when a directory contains only junk
files (e.g. a subdirectory holding nothing but `.DS_Store`), which should be
pruned as empty per the same logic this PR relies on. Worth a regression
guard so a future change to `isEffectivelyEmpty` can't silently regress this.

```suggestion
}

// TestArchive_PrunesJunkOnlyDir verifies a directory containing only OS junk
// is treated as effectively empty and omitted from the archive, matching
// git's behavior for directories with no tracked content.
func TestArchive_PrunesJunkOnlyDir(t *testing.T) {
	srcDir := t.TempDir()
	if err := os.WriteFile(filepath.Join(srcDir, "client.go"), []byte("package redis\n"), 0o644); err != nil {
		t.Fatal(err)
	}
	junkOnlyDir := filepath.Join(srcDir, "empty")
	if err := os.Mkdir(junkOnlyDir, 0o755); err != nil {
		t.Fatal(err)
	}
	if err := os.WriteFile(filepath.Join(junkOnlyDir, ".DS_Store"), []byte("junk"), 0o644); err != nil {
		t.Fatal(err)
	}

	outPath := filepath.Join(t.TempDir(), "out.tgz")
	if err := archive(outPath, []string{srcDir}); err != nil {
		t.Fatalf("archive: %v", err)
	}

	names := readTarNames(t, outPath)
	for _, name := range names {
		if strings.Contains(name, "empty") {
			t.Errorf("archive contains junk-only directory %q, want it pruned", name)
		}
	}
}
```

darccio (changes requested):

.tools/bundle/main_test.go
ipsitapp8: Added, thanks. Pushed as a separate commit with the exact case you flagged (client.go alongside a dir holding nothing but .DS_Store), rebased on main to pick up the bundle changes from the meantime too.

darccio (approved):

PR #1167, opened 2026-08-19
fix(instrumentation/logrus): lazily init hook maps to survive go:linkname init-order

### LINKED ISSUE #1028: bug(instrumentation/logrus): AfterLogrusNew panics with "assignment to entry in nil map" during logrus's own package init
### Describe the bug

Building and running an otelc-instrumented app that has `github.com/sirupsen/logrus` anywhere in its dependency graph, even if the application never calls a single logrus function, currently prints this before any application code runs:

```
failed to exec After hook AfterLogrusNew
assignment to entry in nil map
```

The panic fires before `main()`, from logrus's own package-level `var std = New()` initializer, and is silently recovered by the generated hook trampoline (`tool/internal/instrument/impl.tmpl`), so the process keeps running but every user sees this panic-looking message on every startup.

### Cause

`AfterLogrusNew` (the compile-time hook on `logrus.New()`, wired via `//go:linkname`) fires while logrus's own `var std = New()` package-level initializer runs. `go:linkname` doesn't create a normal Go import edge, so Go's init-order guarantees don't apply here: the hook implementation package's own state (`hookInitMap` in `instrumentation/github.com/sirupsen/logrus/logrus_hook.go`) is not guaranteed to be initialized yet when the hook fires. The resulting nil-map write panics.

### Environment

- OS: any (not platform-specific)
- Go Version: 1.25.0
- Version: main

### To Reproduce

This is not specific to any particular usage pattern. It reproduces unconditionally, even when the application code never references `github.com/sirupsen/logrus` at all:

```go
package main

import (
	"fmt"

	_ "github.com/sirupsen/logrus"
)

func main() {
	fmt.Println("app never touches logrus directly")
}
```

Built and run with `otelc go build -a`:

```
$ /tmp/logrusblankimport-repro
failed to exec After hook AfterLogrusNew
assignment to entry in nil map
{"time":"...","level":"INFO","msg":"trace provider initialized with auto-export"}
...
app never touches logrus directly
```

A skipped regression test documenting this panic was added in #752 (`test/integration/logs_test.go`, subtest `"AfterLogrusNew hook does not panic during logrus init"`).

### Expected behavior

`AfterLogrusNew` should not panic regardless of when it fires relative to the hook implementation package's own init.

ipsitapp8:
Fixes #1028

Any app with `github.com/sirupsen/logrus` in its dependency graph, even without calling into logrus directly, prints this on every startup once instrumented:

```
failed to exec After hook AfterLogrusNew
assignment to entry in nil map
```

`AfterLogrusNew` is wired to `logrus.New()` via `//go:linkname`, which doesn't create a normal Go import edge. So when logrus's own `var std = New()` fires during logrus's package init, there's no guarantee this hook package's own `var hookInitMap = make(...)` has run yet. The hook can fire against a still-nil map and panic on the write.

Fix: stop initializing `hookInitMap` and `fieldInitMap` at var-declaration time, and lazily create them under `hookInitMu` inside `AfterLogrusNew` / `AfterLogrusWithField` instead. That's safe no matter which order the two packages' inits run in. `formatterInit` is a plain bool so its zero value was never a problem.

Added two tests that set both maps to nil right before calling the hooks (simulating the pre-init state) and assert no panic, alongside the existing coverage.

## Test plan
- [x] `go test ./...` in `instrumentation/github.com/sirupsen/logrus`
- [x] `go vet ./...`

---


instrumentation/github.com/sirupsen/logrus/logrus_hook_test.go
```diff
@@ -126,6 +126,63 @@ func TestTraceHook_Fire_NoTraceContext(t *testing.T) {
...
+	hasHook := false
+	for _, hooks := range logger.Hooks {
+		for _, h := range hooks {
+			if _, ok := h.(*traceHook); ok {
+				hasHook = true
+			}
+		}
+	}
+	assert.True(t, hasHook)
+}
```
darccio, line 160: This test leaves `hookInitMap`/`fieldInitMap`/`formatterInit` nil'd out after
it runs, relying on every other test's `resetHookState()` call at its own
start rather than restoring state itself. That's safe under the current
sequential (non-parallel) test execution in this file, but it breaks the
file's existing convention and would silently start leaking state if a
future test added `t.Parallel()`. Consider calling `resetHookState()` at the
end (or via `t.Cleanup`) to match the rest of the suite.
```suggestion
func TestAfterLogrusNew_NilMapAtCallTime(t *testing.T) {
	hookInitMu.Lock()
	hookInitMap = nil
	fieldInitMap = nil
	formatterInit = false
	hookInitMu.Unlock()
	t.Cleanup(resetHookState)

	ictx := hooktest.NewMockHookContext()
	logger := logrus.New()
	assert.NotPanics(t, func() {
		AfterLogrusNew(ictx, logger)
	})

	assert.True(t, hasTraceHook(logger))
}
```

instrumentation/github.com/sirupsen/logrus/logrus_hook_test.go
```diff
@@ -126,6 +126,63 @@ func TestTraceHook_Fire_NoTraceContext(t *testing.T) {
...
+	})
+
+	hasHook := false
+	for _, hooks := range logger.Hooks {
+		for _, h := range hooks {
+			if _, ok := h.(*traceHook); ok {
+				hasHook = true
+			}
+		}
+	}
```
darccio: This hook-scan loop is duplicated verbatim in both new tests
(TestAfterLogrusNew_NilMapAtCallTime and
TestAfterLogrusWithField_NilMapAtCallTime). Worth factoring into a small
helper, e.g.:

```go
func hasTraceHook(logger *logrus.Logger) bool {
	for _, hooks := range logger.Hooks {
		for _, h := range hooks {
			if _, ok := h.(*traceHook); ok {
				return true
			}
		}
	}
	return false
}
```

Then each test becomes `assert.True(t, hasTraceHook(logger))`.
```suggestion
	assert.True(t, hasTraceHook(logger))
```

darccio (changes requested):

ipsitapp8:
Fixed both. The two nil-map tests now call t.Cleanup(resetHookState) so they don't leave hookInitMap/fieldInitMap nil for the next test to depend on, and pulled the hook-scan loop out into a hasTraceHook helper so it's not copy pasted between the two tests. Ran the package tests locally, all green.

ipsitapp8:
Rebased onto main and regenerated the bundle, so the conflict is cleared. Both review points from earlier are in: the two nil-map tests now use `t.Cleanup(resetHookState)` so they restore state themselves rather than relying on the next test, and the duplicated hook-scan loop is factored into a `hasTraceHook` helper. Ready for another look whenever you have time.

darccio (approved):

PR #1170, opened 2026-08-19
fix(tool): re-scope recovered generic constraints to receiver names

### LINKED ISSUE #1169: bug(tool/internal/instrument): recovered generic constraint keeps declaration's param names, breaks on renamed receiver with inter-parameter constraint
### Describe the bug

#1122 recovers a generic receiver's real type-parameter constraints instead of widening them to `any`, which is a genuine improvement. But it clones each constraint verbatim from the type declaration without remapping the type-parameter names, so when a method receiver renames the type parameters and a constraint references another parameter, the generated trampoline references a name that is not in scope and fails to compile.

### Where

`extractReceiverTypeParams` / `receiverConstraintAt` in `tool/internal/instrument/trampoline.go`. `receiverConstraintAt` returns `dst.Clone(field.Type)` from the type declaration, keeping the declaration's parameter names, while the surrounding field uses the receiver's (possibly renamed) parameter names.

### Reproduction

Valid Go, where the receiver renames the type parameters (`K -> A`, `V -> B`) and `V`'s constraint references `K`:

```go
type M[K any, V ~[]K] struct {
	k K
	v V
}

func (m M[A, B]) Method() {}
```

Running the receiver through the current code produces:

```
param[0] name=A constraint=any
param[1] name=B constraint=~[]K     <-- references K, but the params are A and B
```

The generated trampoline is therefore equivalent to:

```go
func otel_trampoline[A any, B ~[]K](recv *M[A, B]) { /* ... */ }
```

which does not compile:

```
undefined: K
B does not satisfy ~[]A
```

The correct output re-scopes the constraint to the receiver's names, `B ~[]A`, which compiles.

### Steps to reproduce

Parse the source above, call `extractReceiverTypeParams(file, receiverType)`, and inspect the second field: its name is `B` but its constraint is `~[]K`. Building `func f[A any, B ~[]K]() {}` directly gives `undefined: K` on Go 1.24.

### Scope

This needs both a receiver that renames the type parameters and a constraint that references another parameter (an inter-parameter constraint such as `~[]K` or `map[K]V`). Both are valid but uncommon, so the common case #1122 targets is unaffected and its tests pass. It is the same class of failure #1122 set out to remove (uncompilable generated code), for an input its tests do not cover.

### Suggested fix

When cloning a constraint in `receiverConstraintAt`, substitute the type declaration's parameter names with the receiver's names before use. A small `dst` walk that rewrites `*dst.Ident` nodes matching a declaration parameter name to the corresponding receiver name would cover `~[]K`, `map[K]V`, and any other inter-parameter constraint. A regression test with a renamed receiver over an inter-parameter constraint would lock it down, since the current tests use matching names only.

I verified the recovered constraint and the compile failure directly; happy to send the fix if it is useful.

vignesh2027:
## Description

`extractReceiverTypeParams` recovers a generic type parameter constraint from the receiver's own type declaration and attaches it to the receiver's parameter name. When the receiver renames a parameter, an inter-parameter constraint kept the declaration's name instead of the receiver's.

Concretely, for:

```go
type M[K any, V ~[]K] struct{ /* ... */ }

func (m M[A, B]) InstrumentedMethod() { /* ... */ }
```

the recovered type parameter list came out as `[A any, B ~[]K]`. The `~[]K` refers to `K`, which the generated trampoline never declares (it declares `A` and `B`), so the emitted code fails to compile with `undefined: K`.

The fix maps the declaration's parameter names to the receiver's names positionally (`K -> A`, `V -> B`), then rewrites the identifiers inside each recovered constraint, so `~[]K` becomes `~[]A` and every reference stays in scope. Constraints with no rename are returned untouched, so the common case allocates nothing extra.

## Motivation

A receiver is free to rename a generic type's parameters, and constraints that reference a sibling parameter (`~[]K`, `map[K]V`, and similar) are common on container types. Without this, instrumenting a method on such a type emits code that does not build, which fails the whole compile-time rewrite for that package.

New unit test `TestExtractReceiverTypeParamsConstraint_RenamedInterParam` covers the renamed receiver with an inter-parameter constraint and asserts the recovered constraint reads `~[]A`.

Fixes #1169

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### DIFF
--- a/tool/internal/instrument/trampoline.go
+++ b/tool/internal/instrument/trampoline.go
@@ -923,22 +923,24 @@ func extractReceiverTypeParams(file *dst.File, recvType dst.Expr) *dst.FieldList
 		// GenStruct[T] - single type parameter
 		if ident, ok := t.Index.(*dst.Ident); ok {
 			original := findGenericTypeDecl(file, receiverBaseTypeName(t.X))
+			nameMap := receiverNameMap(original, []dst.Expr{t.Index})
 			return &dst.FieldList{
 				List: []*dst.Field{{
 					Names: []*dst.Ident{ident},
-					Type:  receiverConstraintAt(original, 0),
+					Type:  remapConstraintNames(receiverConstraintAt(original, 0), nameMap),
 				}},
 			}
 		}
 	case *dst.IndexListExpr:
 		// GenStruct[T, U, ...] - multiple type parameters
 		original := findGenericTypeDecl(file, receiverBaseTypeName(t.X))
+		nameMap := receiverNameMap(original, t.Indices)
 		fields := make([]*dst.Field, 0, len(t.Indices))
 		for i, idx := range t.Indices {
 			if ident, ok := idx.(*dst.Ident); ok {
 				fields = append(fields, &dst.Field{
 					Names: []*dst.Ident{ident},
-					Type:  receiverConstraintAt(original, i),
+					Type:  remapConstraintNames(receiverConstraintAt(original, i), nameMap),
 				})
 			}
 		}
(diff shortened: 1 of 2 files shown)


vignesh2027:
Good catch, this was a real bug and I have fixed it.

The walk renamed every identifier whose name matched a declaration parameter, including positions that name something rather than reference a type parameter. Your example reproduced exactly: with `type M[Stringer any, V fmt.Stringer]` and receiver `M[A, B]`, the selector in `fmt.Stringer` became the nonexistent `fmt.A`. Interface method names and struct field names inside a constraint had the same problem.

The remap now collects those naming positions first, the selected symbol and its package qualifier in a `SelectorExpr`, and the names a `Field` declares, then skips them and renames only genuine references.

Added two regression tests: `..._NameCollision` for your qualified-constraint case and `..._MethodNameCollision` for an interface method whose name matches a type parameter. I checked both fail against the previous implementation, with the selector coming back as `A` instead of `Stringer` and the method name as `A` instead of `T`, and pass with the fix.

tool/internal/instrument/trampoline.go
```diff
@@ -949,6 +951,69 @@ func extractReceiverTypeParams(file *dst.File, recvType dst.Expr) *dst.FieldList
...
+// renames the constraint is returned untouched.
+func remapConstraintNames(constraint dst.Expr, nameMap map[string]string) dst.Expr {
+	if len(nameMap) == 0 {
+		return constraint
+	}
+	dst.Inspect(constraint, func(node dst.Node) bool {
+		if ident, ok := node.(*dst.Ident); ok {
+			if renamed, found := nameMap[ident.Name]; found {
+				ident.Name = renamed
+			}
```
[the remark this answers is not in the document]
vignesh2027: Good catch, fixed.

`remapConstraintNames` now walks the constraint once to collect the identifiers that name something rather than reference a type parameter, and skips them on the renaming pass. That covers a selector's package and symbol (the `fmt` and `Stringer` in `fmt.Stringer`) and the names a field or interface method declares. Only what is left is renamed.

Two regression tests added. `TestExtractReceiverTypeParamsConstraint_NameCollision` uses your exact example, `type GenStruct[Stringer any, V fmt.Stringer]` with receiver `GenStruct[A, B]`, and asserts the constraint stays `fmt.Stringer` rather than becoming the nonexistent `fmt.A`. `TestExtractReceiverTypeParamsConstraint_MethodNameCollision` covers the same hazard for a name the constraint declares, an interface method whose name matches a type parameter.

tool/internal/instrument/trampoline.go
```diff
@@ -923,22 +923,24 @@ func extractReceiverTypeParams(file *dst.File, recvType dst.Expr) *dst.FieldList
 		// GenStruct[T] - single type parameter
 		if ident, ok := t.Index.(*dst.Ident); ok {
 			original := findGenericTypeDecl(file, receiverBaseTypeName(t.X))
+			nameMap := receiverNameMap(original, []dst.Expr{t.Index})
 			return &dst.FieldList{
 				List: []*dst.Field{{
 					Names: []*dst.Ident{ident},
-					Type:  receiverConstraintAt(original, 0),
+					Type:  remapConstraintNames(receiverConstraintAt(original, 0), nameMap),
```
darccio, line 930: The rename/collision fix is exercised only through the `IndexListExpr`
branch (multi-parameter receivers): `TestExtractReceiverTypeParamsConstraint_RenamedInterParam`,
`_NameCollision`, and `_MethodNameCollision` all use two-parameter receivers
like `GenStruct[A, B]`.

This `IndexExpr` branch (single-parameter receivers, e.g. `GenStruct[T]`) calls
the same `receiverNameMap`/`remapConstraintNames` functions but has no test
covering a rename here. A single-parameter receiver can't have an
inter-parameter constraint (there's no sibling to reference), but it can still
hit the name-collision hazard on its own constraint, e.g.:

```go
type GenStruct[Stringer fmt.Stringer] struct{}
func (g GenStruct[A]) M() {}
```

Worth a small test asserting the recovered constraint stays `fmt.Stringer`
rather than becoming `fmt.A`, to cover this branch combination explicitly
rather than relying on it sharing code with the tested `IndexListExpr` branch.

darccio (changes requested):

tool/internal/instrument/trampoline.go
vignesh2027: Added. `TestExtractReceiverTypeParamsConstraint_SingleParamNameCollision` uses your example, `type GenStruct[Stringer fmt.Stringer]` with receiver `GenStruct[A]`, and asserts the recovered constraint stays `fmt.Stringer`.

You were right that the existing ones all went through `IndexListExpr`. To confirm the new test actually covers the `IndexExpr` branch rather than duplicating that, I reverted `remapConstraintNames` to the blanket rename and it fails there with `actual: "A"` against `expected: "Stringer"`.

PR #1178, opened 2026-08-20
fix(instrumentation/kafka-go): mark async producer spans and log otherwise-invisible async failures

### LINKED ISSUE #1177: bug(instrumentation/kafka-go): producer spans always end as Ok with meaningless duration when Writer.Async is true
### Describe the bug

`AfterWriteMessages` in the kafka-go producer instrumentation ends every span created by `BeforeWriteMessages` using the error returned directly from `(*kafka.Writer).WriteMessages`. This assumes `WriteMessages` only returns after the write has actually completed (or failed) against the broker. That assumption is false whenever the writer is configured with `Async: true`.

In `github.com/segmentio/kafka-go` (`writer.go`), `WriteMessages` batches the messages and then:

```go
batches := w.batchMessages(msgs, assignments)
if w.Async {
    return nil
}
```

With `Async: true`, the call returns `nil` immediately after handing the messages to per-partition batch queues — the actual produce-to-broker happens later on background goroutines (`partitionWriter.writeBatch`). Any resulting error is only ever reported through the user-supplied `Writer.Completion` callback:

```go
// writer.go
Completion func(messages []Message, err error)
...
if ptw.w.Completion != nil {
    ptw.w.Completion(batch.msgs, err)
}
```

The instrumentation's `AfterWriteMessages` never hooks `Completion`, so for an async writer it always sees `err == nil` and ends every span as `codes.Ok`, essentially instantly.

### Environment

- Go Version: 1.23+
- kafka-go Version: v0.4.51 (instrumentation/github.com/segmentio/kafka-go/go.mod)

### To Reproduce

1. Configure a `kafka.Writer` with `Async: true` and a `Completion` callback.
2. Call `w.WriteMessages(ctx, msg)` in a project built with `otelc`.
3. `BeforeWriteMessages` starts a producer span for `msg`.
4. `WriteMessages` returns `nil` immediately (per `writer.go:667-669`, before the message is actually sent).
5. `AfterWriteMessages` is invoked with `err == nil` and ends the span as `codes.Ok` — before the message has even been dispatched to the broker, let alone acknowledged.
6. If the write subsequently fails (broker unreachable, leader not available, etc.), that failure is reported only via `Completion`, which the instrumentation never observes — the already-closed, already-Ok span has no way to reflect it.

This can be shown directly from the kafka-go source without needing a live broker: any test that instruments `WriteMessages` on an `Async: true` writer and asserts `AfterWriteMessages` fires before `Completion` (or before any injected transport delay/error resolves) demonstrates the gap, since `WriteMessages` unconditionally returns at line 668 without waiting on `batch.done`.

### Expected behavior

For an async writer, spans should not be finalized with a success status based on `WriteMessages`'s immediate (and essentially meaningless) return value. Either:
- the span shouldn't be ended until the actual produce result is known (via `Completion`), or
- at minimum, spans for async writes should be marked in a way that distinguishes "enqueued, delivery unknown" from "confirmed delivered", so consumers of the trace data don't read span duration/status as delivery latency/success.

### Additional context

Impact: for any application using kafka-go's `Async: true` mode — the standard configuration recommended by kafka-go itself for high-throughput producers — every producer span otelc emits is silently wrong: duration measures only local enqueue time (microseconds) rather than the actual Kafka round-trip, and status is always OK regardless of whether the broker write actually succeeds. This makes latency dashboards and error-rate alerting built on these spans meaningless for async producers, and is a correctness bug in a fairly common configuration, not an edge case.

Affected component: `instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go` (`BeforeWriteMessages`/`AfterWriteMessages`).

Suggested direction: when `w.Async` is true, wrap `w.Completion` (preserving any user-supplied callback) to end the pending spans there instead of in `AfterWriteMessages`, matching messages back to spans (e.g. by tracking span-per-message alongside a correlation key, since `Completion` receives the batch of `[]kafka.Message`). If matching messages back to spans reliably isn't feasible for all cases, an acceptable fallback is to at least avoid asserting `codes.Ok` for async writes and to annotate the span (e.g. an event or attribute) noting delivery confirmation is unavailable.

bhuvan-somisetty:
## Description

`kafka.Writer.WriteMessages` returns as soon as messages are enqueued when `Async` is true — before the broker write happens. `AfterWriteMessages` (in `instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go`) ends the producer spans created by `BeforeWriteMessages` using that immediate return value, so:

- Span duration only ever reflects local enqueue time (microseconds), never the actual Kafka round-trip.
- Any real delivery failure is reported later, solely through `Writer.Completion` — which this instrumentation never hooked — so such failures were invisible everywhere: not on the span, not in any log.

Verified against the real `kafka-go` v0.4.51 source: `WriteMessages` unconditionally returns `nil` at the `if w.Async { return nil }` early-return in `writer.go`, before any batch is awaited, and `Completion func(messages []Message, err error)` is the only place a delivery result for an async write surfaces.

## Root cause

The instrumentation assumed `WriteMessages`'s return value always reflects the outcome of the broker write. That's true for the default synchronous writer, but not for `Async: true`, which is kafka-go's documented high-throughput mode.

## Solution

- Added a `messaging.kafka.async` span attribute (`instrumentation/github.com/segmentio/kafka-go/semconv`), set on producer-send spans when the writer is async, so consumers of the trace data know duration and status here reflect hand-off, not confirmed delivery — never emitted for the sync path or for consumer spans.
- Wrapped `Writer.Completion` (`ensureAsyncFailureLogging`), installed once per `*kafka.Writer` and chaining any callback the caller already configured, so an async write failure is at least logged instead of silently disappearing.

I deliberately did **not** attempt full per-message span correlation through `Completion`. I considered it, but both realistic approaches have real problems:
- Correlating via the propagated trace context in message headers silently finds nothing when no `propagation.TextMapPropagator` is configured — which is the OTel SDK default.
- Adding a dedicated correlation header would work, but that header gets sent to the broker as part of the message and is then visible to every real Kafka consumer reading the topic, for the lifetime of that message — a bigger footprint than is justified here.

Given that, logging the failure (with the caller's own `Completion` still firing normally) is the honest, dependency-free improvement over the previous behavior, where these failures were entirely invisible.

## Testing

- `instrumentation/github.com/segmentio/kafka-go/producer`: added `TestBeforeWriteMessages_AsyncWriterMarksSpans`, `TestBeforeWriteMessages_SyncWriterOmitsAsyncAttr`, `TestEnsureAsyncFailureLogging_LogsFailureAndChainsOriginalCompletion` (asserts the failure is logged *and* the caller's own `Completion` still receives the real error), and `TestEnsureAsyncFailureLogging_WrapsOnce` (asserts a writer reused across multiple `WriteMessages` calls only gets one logging wrapper, not a new layer per call).
- `instrumentation/github.com/segmentio/kafka-go/semconv`: added `TestKafkaRequestTraceAttrs_ProducerAsync`, `TestKafkaRequestTraceAttrs_ProducerSyncOmitsAsync`, `TestKafkaRequestTraceAttrs_ConsumerIgnoresAsync`.
- `go test ./...` passes for both packages.
- `golangci-lint run` on both packages: no new issues beyond this repo's existing accepted baseline for these files (verified against a clean run of the unmodified files); the two genuinely new findings my diff introduced (`gochecknoglobals` on the wrap-tracking map, `sloglint` on the new log call) are addressed with targeted `//nolint` comments, matching the existing pattern used elsewhere in this repo (e.g. `openai-go/middleware.go`).

## Impact

For any application using `kafka.Writer{Async: true}` — kafka-go's own recommended configuration for high-throughput producers — every producer span was silently misleading (duration too short, status never reflecting real failures). Latency dashboards and error-rate alerting built on these spans were meaningless for async producers. This PR doesn't fully close that gap (a fully reliable fix would require changes upstream in kafka-go itself, e.g. a stable per-message completion handle), but it stops the span from lying and gives operators a real signal (a log line) for failures that were previously invisible.

Fixes #1177

---

## Checklist

- [x] Code formatted: `golangci-lint fmt` (no changes needed)
- [x] Linters pass: `golangci-lint run` on both affected packages (no new issues vs. baseline)
- [x] Tests pass: `go test ./...` on both affected packages


instrumentation/github.com/segmentio/kafka-go/semconv/client.go
```diff
@@ -68,6 +79,9 @@ func KafkaRequestTraceAttrs(req KafkaRequest) []attribute.KeyValue {
 	switch req.Operation {
 	case KafkaOperationSend:
 		attrs = append(attrs, semconv.MessagingOperationTypeSend)
+		if req.Async {
+			attrs = append(attrs, messagingKafkaAsyncKey.Bool(true))
+		}
```
darccio, line 89: This only ever emits `messaging.kafka.async=true`; a sync producer span
carries no attribute at all rather than `messaging.kafka.async=false`.

That keeps cardinality down, but it means a query/dashboard can't
distinguish "definitely sync" from "predates this attribute" /
"instrumentation disabled" — both read as absence. Given `req.Async` is
already known and cheap to encode, was the omit-on-false behavior a
deliberate low-cardinality choice, or would you want it explicit?

instrumentation/github.com/segmentio/kafka-go/semconv/client.go
```diff
@@ -47,8 +47,19 @@ type KafkaRequest struct {
...
+	// the client library, not confirmed delivery, so its duration and status
+	// must not be read as broker round-trip latency or delivery success.
+	Async bool
 }
 
+// messagingKafkaAsyncKey has no OpenTelemetry semantic convention yet; it marks
+// producer spans for kafka.Writer.Async writes so consumers of the trace data
+// don't mistake enqueue-time duration and Unset/Ok status for confirmed
+// delivery (see KafkaRequest.Async).
+const messagingKafkaAsyncKey = attribute.Key("messaging.kafka.async")
```
darccio, line 66: This adds a custom attribute but skips this repo's own registration
workflow for non-upstream attributes. Per docs/semantic-conventions.md
("Local Registry"), library-specific attributes not covered upstream must
be declared with `id:` in `schemas/otelc/groups/<library>.yaml`.

schemas/otelc/groups/kafka.yaml isn't touched by this PR, and its header
comment currently reads:

> Every attribute is standard upstream OpenTelemetry messaging telemetry,
> referenced with `ref:`.

That statement is now false for `messaging.kafka.async`, and the
producer span group (`span.otelc.messaging.kafka.producer`) doesn't list
it either. `make lint-schema` / CI's registry-diff check won't catch this
gap — they validate the registry's internal consistency against the
pinned upstream version, not that the registry covers everything the Go
code actually emits — so schemas/otelc/ silently stops being the
accurate contract it's meant to be.

Please add an `id:` declaration for `messaging.kafka.async` to
kafka.yaml, reference it (conditionally) from the producer span group,
and update the file's header comment to note the one non-upstream
attribute.

instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go
```diff
@@ -106,12 +126,51 @@ func BeforeWriteMessages(
...
 // kafka.WriteMessages may return kafka.WriteErrors — a []error aligned with
 // the message slice — to indicate partial success. When that happens, only the
 // spans for messages whose entry is non-nil are marked as Error; the rest stay
 // Ok. For any other error type, the error is applied to every span.
+//
+// For an Async writer, a nil err here only means the messages were handed off
+// to the client library, not that the broker write succeeded: WriteMessages
+// returns before that happens. These spans are marked via the
+// messaging.kafka.async attribute (set in BeforeWriteMessages) so their
+// duration and Unset/Ok status aren't misread as confirmed delivery.
```
darccio: Worth being explicit in this comment (and in the PR description) that
this is a labeling fix, not a correctness fix for the span itself: after
this change, an async producer span still ends immediately via
`AfterWriteMessages`, still gets `codes.Ok`/Unset, and still shows
enqueue-time duration — none of that changes here. The only new
signal is that `messaging.kafka.async=true` tells consumers of the trace
not to trust it.

That's a reasonable, deliberately scoped choice (matches the fallback
described in #1177), but the current wording — "these spans are marked
... so duration and status aren't misread" — could read as if the
duration/status problem itself is fixed rather than just flagged.
Suggest tightening it so a future reader doesn't assume more was fixed
here than actually was.

instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go
```diff
@@ -106,12 +126,51 @@ func BeforeWriteMessages(
...
+		if err != nil {
+			//nolint:sloglint // matches the existing package-level logger usage in this file
+			logger.Error("kafka async write failed after WriteMessages returned; "+
+				"the producer span(s) for the affected message(s) were already ended without this outcome",
+				"error", err, "messageCount", len(msgs))
+		}
+		if original != nil {
+			original(msgs, err)
+		}
+	}
```
darccio: Data race, confirmed against kafka-go v0.4.51 source. `LoadOrStore`
only guards against installing the wrapper twice — it does not
synchronize this write with kafka-go's own reads of `w.Completion`.

Sequence: goroutine A calls `w.WriteMessages(...)` on a writer that's
never been wrapped yet; its batch is handed off to a background
goroutine that will later read `ptw.w.Completion` unsynchronized
(kafka-go `writer.go:1180`: `if ptw.w.Completion != nil {
ptw.w.Completion(batch.msgs, err) }`). Concurrently, goroutine B calls
`w.WriteMessages(...)` on the *same* writer — writer sharing across
goroutines is documented-safe kafka-go usage — and wins the
`LoadOrStore`, executing `w.Completion = func(...)` here. If A's
background goroutine's read races with B's write, that's an
unsynchronized concurrent read/write of the same field: a real data
race under the Go memory model, not just a theoretical one. None of
the new tests exercise concurrent `WriteMessages` calls against a
shared writer with real background completion, so `go test -race`
won't surface it.

Wrapping `Completion` lazily on first use can't be made safe without
also serializing against kafka-go's own internal reads, which this
package doesn't control. Consider wrapping at writer-construction time
instead (before any `WriteMessages` call can occur), or explicitly
documenting/enforcing that `ensureAsyncFailureLogging` must run before
the writer is shared across goroutines.

instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go
```diff
@@ -40,6 +40,15 @@ var (
...
 	initOnce   sync.Once
+
+	// asyncCompletionWrapped tracks which *kafka.Writer instances already have
+	// a logging wrapper installed on Completion (see ensureAsyncFailureLogging),
+	// so a writer reused across many WriteMessages calls is only wrapped once.
+	// Entries are keyed by the writer's own pointer and are never removed, but
+	// kafka.Writer values are conventionally constructed once per process and
+	// reused for its lifetime, not created per call, so this stays bounded by
+	// the number of distinct writers an application creates, not by traffic.
+	asyncCompletionWrapped sync.Map //nolint:gochecknoglobals // per-writer wrap tracking, see comment above
```
darccio: `asyncCompletionWrapped` keys by `*kafka.Writer` and never removes
entries, so every wrapped writer is pinned for the process's lifetime.
The comment's justification — writers are conventionally constructed
once and reused — holds for the common case, but nothing in kafka-go
enforces it; an application that creates short-lived `Writer` values
(uncommon, but not forbidden) leaks one map entry per writer, forever.

Not blocking, but worth a `sync.Map`-based `Writer` finalizer or a
one-line acknowledgment that this trades a theoretical leak for
simplicity, if that's the intended tradeoff.

instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go
```diff
@@ -106,12 +126,51 @@ func BeforeWriteMessages(
...
+// configured propagator (silently finding nothing when one isn't set, which is
+// the OTel default) or attaching a tracking header to every outgoing message
+// that would then be visible to real Kafka consumers — both worse than not
+// correlating. Logging the failure is the honest, dependency-free improvement
+// over the previous behavior, where such failures were entirely invisible.
+func ensureAsyncFailureLogging(w *kafka.Writer) {
+	if _, alreadyWrapped := asyncCompletionWrapped.LoadOrStore(w, struct{}{}); alreadyWrapped {
+		return
+	}
+	original := w.Completion
```
darccio: If application code reassigns `w.Completion` after the first
`WriteMessages` call on this writer (e.g. swapping in a new callback
at runtime), that assignment silently overwrites this wrapper's
closure — plain field assignment, not append. Because
`asyncCompletionWrapped` already has an entry for `w`,
`ensureAsyncFailureLogging` will never re-wrap on a later call, so the
failure-logging this PR adds just stops working with no error, no log
line, no signal that it happened.

This may not be common (most callers set `Completion` once, at
construction), but there's nothing here that detects or guards against
it. At minimum, worth a comment noting the assumption that `Completion`
is set once and not mutated post-construction — or, if that assumption
can't be guaranteed, detect drift (e.g. by comparing the currently
installed `Completion` against what this package last set) and
re-wrap.

instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go
darccio: Noting for the record, not asking for a change: kafka-go's own docs
say `Writer.Close` blocks on `Completion` callback calls, so this
wrapper's extra work (the conditional `logger.Error` call) runs inside
that blocking window. The overhead is negligible, but it's worth
knowing that any future change to what this wrapper does on failure
(e.g. richer structured logging, metrics) will extend how long `Close`
blocks.

instrumentation/github.com/segmentio/kafka-go/producer/producer_hook.go
```diff
@@ -106,12 +126,51 @@ func BeforeWriteMessages(
...
+	w.Completion = func(msgs []kafka.Message, err error) {
+		if err != nil {
+			//nolint:sloglint // matches the existing package-level logger usage in this file
+			logger.Error("kafka async write failed after WriteMessages returned; "+
+				"the producer span(s) for the affected message(s) were already ended without this outcome",
+				"error", err, "messageCount", len(msgs))
```
darccio: The `//nolint:sloglint` comment says this "matches the existing
package-level logger usage in this file," but it doesn't quite — the
other two calls in this file (`logger.Info(...)` at line 61,
`logger.Debug(...)` at line 79) are single string literals with no
concatenation. This is the first call built via `"a" + "b"`, which is
likely what trips sloglint's static-message check even though the
result is a compile-time constant. Merging the message into one
literal avoids both the mismatch in the comment and the need for the
nolint:

```suggestion
			logger.Error("kafka async write failed after WriteMessages returned; the producer span(s) for the affected message(s) were already ended without this outcome",
				"error", err, "messageCount", len(msgs))
```

instrumentation/github.com/segmentio/kafka-go/producer/producer_hook_test.go
```diff
@@ -203,3 +205,112 @@ func TestAfterWriteMessages_AlwaysEndsSpans(t *testing.T) {
...
+// writer's real result surfaces, well after the (already-ended) spans stop
+// listening — is at least logged, and that any Completion callback the caller
+// already configured still runs.
+func TestEnsureAsyncFailureLogging_LogsFailureAndChainsOriginalCompletion(t *testing.T) {
+	setupTest(t)
+
+	var buf bytes.Buffer
+	oldLogger := logger
+	logger = slog.New(slog.NewTextHandler(&buf, nil))
+	t.Cleanup(func() { logger = oldLogger })
```
darccio: This is the first place in the repo that swaps a package-level var
(`logger`) to capture output in a test — no other test under
`instrumentation/` does this. It's safe today because nothing in this
file calls `t.Parallel()`, but the pattern is fragile: if a future edit
parallelizes any test/subtest in this package, this global mutation
(and the identical one in TestEnsureAsyncFailureLogging_WrapsOnce)
would race across subtests, and nothing here would signal that at
review time.

Worth considering a testable-logger seam instead — e.g. having
`runtime.Logger()` return an interface the test can inject a
capturing implementation into, rather than reassigning the package
var directly — so this doesn't quietly become the template for future
tests in other instrumentation packages.

darccio (changes requested):

bhuvan-somisetty:
Thanks for the deep review, especially catching the data race — you're right that LoadOrStore alone didn't synchronize with kafka-go's own read of Completion. Switched to a sync.Once so every concurrent caller blocks until the wrap is actually installed, which closes that. Also registered messaging.kafka.async in kafka.yaml (id + producer span ref + header comment), tightened the AfterWriteMessages comment so it doesn't overclaim, folded the log message into one literal to drop the nolint, added inline notes on the map-eviction and Completion-reassignment tradeoffs, and dropped the logger-swapping test pattern in favor of asserting on the chaining/wrap-once behavior directly. Rebased onto main too. Should be good for another pass.

darccio (approved):

PR #1179, opened 2026-08-20
fix(tool): centralize rule file name matching

ADITYA-CODE-SOURCE:
## Summary

Centralize otelc rule-file name matching so setup, pin discovery, manifest generation, and latest-library compatibility tests recognize the same files.

Previously, setup and manifest generation maintained duplicate filename checks, while `InstrumentedTargets` accepted every `.yaml` file and ignored valid `.yml` rule files.

## Changes

- Add a shared matcher for:
  - `otelc.yaml`
  - `otelc.yml`
  - `*.otelc.yaml`
  - `*.otelc.yml`
- Use the shared matcher in setup, pin discovery, manifest generation, and latest-library scanning
- Ignore unrelated YAML files during latest-library rule discovery
- Add focused coverage for valid and invalid filenames
- Run LatestLibBuild when instrumentation YAML or YML files change
- Remove duplicate matcher tests from setup and manifest packages

## Testing

- `go test ./tool/util ./tool/internal/setup ./tool/internal/manifest -count=1`
- `go -C test test ./testutil/... -count=1`
- `make test-unit/tool`
- `make test-unit/helper`
- `make lint/action`
- `make lint/yaml`
- `git diff --check`

### DIFF
--- a/tool/util/shared.go
+++ b/tool/util/shared.go
@@ -36,6 +36,14 @@ const (
 	OtelcToolExe           = "otelc"
 )
 
+// IsRuleFileName reports whether name identifies an otelc rule file.
+func IsRuleFileName(name string) bool {
+	return name == "otelc.yml" ||
+		name == "otelc.yaml" ||
+		strings.HasSuffix(name, ".otelc.yml") ||
+		strings.HasSuffix(name, ".otelc.yaml")
+}
+
 func GetMatchedRuleFile() string {
 	const matchedRuleFile = "matched.json"
 	return GetBuildTemp(matchedRuleFile)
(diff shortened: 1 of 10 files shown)


tool/util/shared.go
```diff
@@ -36,6 +36,14 @@ const (
 	OtelcToolExe           = "otelc"
 )
 
+// IsRuleFileName reports whether name identifies an otelc rule file.
+func IsRuleFileName(name string) bool {
```
y1yang0: Rename to IsRuleFile

y1yang0 (approved):
Otherwise LGTM

PR #1182, opened 2026-08-20
fix(ci): propagate change detection failures

### LINKED ISSUE #1181: fix(ci): required done jobs pass when change detection fails
### Describe

Several required workflow aggregate jobs treat skipped downstream jobs as success without first checking whether their `Detect Changes` job succeeded.

If `actions/checkout` or `dorny/paths-filter` fails in the `changes` job, the filtered downstream job is skipped because the filter output is missing. The final required `done` job then accepts that `skipped` result and exits successfully, so the required check can go green even though change detection failed.

Affected workflows:

- `.github/workflows/lint-golang.yaml`
- `.github/workflows/check-license-headers.yaml`
- `.github/workflows/verify-bundle.yaml`
- `.github/workflows/test-latestlibbuild.yaml`
- `.github/workflows/check-consistency.yaml`

### Expected behavior

The final required `done` job should:

- fail if `needs.changes.result != "success"`
- pass if change detection succeeds and the relevant filter output is explicitly `false`
- fail if the filter output is missing or unexpected
- require the downstream jobs to succeed when the filter output is `true`

This is already the pattern used by:

- `.github/workflows/test-unit.yaml`
- `.github/workflows/test-integration.yaml`
- `.github/workflows/test-e2e.yaml`

### Why this matters

These are required workflows. A change-detection infrastructure failure should not be hidden as an intentional path-filter skip.

### Reproduction

Using the current logic:

1. Make the `changes` job fail, for example by temporarily adding a failing step after checkout.
2. The filtered downstream job is skipped because the path-filter output is unavailable.
3. The final `done` job sees the downstream result as `skipped`.
4. The final job exits successfully.

This means the required check can pass even though change detection failed.

### Suggested fix

Update the five affected `done` jobs to follow the existing `test-unit`, `test-integration`, and `test-e2e` pattern:

- capture `needs.changes.result`
- capture the relevant `needs.changes.outputs.*` value
- fail when change detection failed
- pass only for an explicit `false` filter output
- fail on missing or unexpected filter output
- require all downstream jobs to succeed when the filter output is `true`

For `check-consistency.yaml`, require all four consistency jobs to succeed when `code == true`.

ADITYA-CODE-SOURCE:
## Description

Update required workflow aggregate jobs so a failed `Detect Changes` job cannot be mistaken for an intentional path-filter skip.

The affected jobs now:

- fail when change detection fails
- pass when change detection succeeds and reports no relevant changes
- reject missing or unexpected filter outputs
- require downstream jobs to succeed when relevant changes are present

This follows the existing pattern in the Unit, Integration, and E2E workflows.

Fixes #1181

## Verification

- `make format/yaml`
- `make lint/yaml`
- `make lint/action`
- `git diff --check`

### DIFF
--- a/.github/workflows/check-consistency.yaml
+++ b/.github/workflows/check-consistency.yaml
@@ -114,21 +114,36 @@ jobs:
     steps:
       - name: Check result
         run: |
+          changes_result="${{ needs.changes.result }}"
+          code_changed="${{ needs.changes.outputs.code }}"
           api_result="${{ needs.check-api-sync.result }}"
           golden_result="${{ needs.check-golden-files.result }}"
           mod_tidy_result="${{ needs.check-mod-tidy.result }}"
           manifest_result="${{ needs.verify-manifest.result }}"
+          echo "Changes result: $changes_result"
+          echo "Code changed: $code_changed"
           echo "api-sync result: $api_result"
           echo "golden-files result: $golden_result"
           echo "mod-tidy result: $mod_tidy_result"
           echo "manifest result: $manifest_result"
-          if [[ ("$api_result" == "success" || "$api_result" == "skipped") && \
-                ("$golden_result" == "success" || "$golden_result" == "skipped") && \
-                ("$mod_tidy_result" == "success" || "$mod_tidy_result" == "skipped") && \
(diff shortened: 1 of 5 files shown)


darccio (approved):
LGTM

PR #1185, opened 2026-08-20
fix(tool): reject invalid manifest version ranges

### LINKED ISSUE #1184: Manifest generator accepts invalid rule.version ranges
### Description

The manifest generator accepts malformed `rule.version` ranges that are rejected during normal rule loading.

For example:

```yaml
rule:
  target: example.com/target
  version: v1.0.0,
```

Running:

```bash
go run ./tool/cmd/gen-manifest
```

successfully generates the manifest and writes:

```json
"versionRange": "v1.0.0,"
```

However, normal rule loading validates version ranges and rejects this malformed value.

### Expected behavior

Manifest generation should validate `rule.version` using the same validation logic as normal rule loading and return an error for malformed version ranges.

### Actual behavior

Manifest generation accepts malformed version ranges and includes them in `instrumentation-manifest.json`.

### Proposed fix

Validate `rule.version` while loading manifest entries, preferably by reusing the existing version-range validation helper.

Add regression tests covering:

* Rejection of malformed version ranges
* Acceptance of valid version ranges
* Propagation of a clear validation error

### Scope

I plan to keep the PR focused on:

* Validating manifest-side `rule.version` values
* Returning an error for malformed ranges
* Adding regression tests for valid and invalid version ranges

ADITYA-CODE-SOURCE:
## Summary

Validate rule version ranges during manifest generation using the same validation applied during normal rule loading.

Malformed ranges now fail with rule-file context instead of being written to `instrumentation-manifest.json`.

Fixes #1184

## Testing

- `go test -count=1 ./tool/internal/manifest ./tool/util ./tool/internal/setup`
- `make verify-manifest`
- `make lint/license-header`
- `git diff --check`

### DIFF
--- a/tool/internal/manifest/manifest.go
+++ b/tool/internal/manifest/manifest.go
@@ -136,6 +136,9 @@ func loadModuleEntries(moduleDir, modulePath string) (Manifest, error) {
 			if rule.Target == "" {
 				continue
 			}
+			if validateErr := util.ValidateVersionRange(rule.VersionRange); validateErr != nil {
+				return ex.Wrapf(validateErr, "validating version in rule file %s", path)
+			}
 			entries = append(entries, Entry{
 				ModulePath:   modulePath,
 				Target:       rule.Target,
--- a/tool/internal/manifest/manifest.go
+++ b/tool/internal/manifest/manifest.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io/fs"
+	"maps"
 	"os"
 	"path/filepath"
 	"slices"
@@ -132,12 +133,13 @@ func loadModuleEntries(moduleDir, modulePath string) (Manifest, error) {
 		if unmarshalErr := yaml.Unmarshal(content, &rules); unmarshalErr != nil {
 			return ex.Wrapf(unmarshalErr, "parsing rule file %s", path)
 		}
-		for _, rule := range rules {
+		for _, name := range slices.Sorted(maps.Keys(rules)) {
+			rule := rules[name]
+			if validateErr := util.ValidateVersionRange(rule.VersionRange); validateErr != nil {
+				return ex.Wrapf(validateErr, "validating version for rule %q in file %s", name, path)
+			}
 			if rule.Target == "" {
 				continue
-			}
-			if validateErr := util.ValidateVersionRange(rule.VersionRange); validateErr != nil {
-				return ex.Wrapf(validateErr, "validating version in rule file %s", path)
 			}
 			entries = append(entries, Entry{
 				ModulePath:   modulePath,
--- a/tool/internal/manifest/manifest_test.go
+++ b/tool/internal/manifest/manifest_test.go
@@ -5,6 +5,7 @@ package manifest
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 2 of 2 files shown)


tool/internal/manifest/manifest.go
```diff
@@ -136,6 +136,9 @@ func loadModuleEntries(moduleDir, modulePath string) (Manifest, error) {
 			if rule.Target == "" {
 				continue
 			}
+			if validateErr := util.ValidateVersionRange(rule.VersionRange); validateErr != nil {
+				return ex.Wrapf(validateErr, "validating version in rule file %s", path)
+			}
```
darccio: An empty-target rule with a malformed version is silently skipped here — `continue` fires before `ValidateVersionRange` runs. Validating the version first means a bad range is caught even on a rule that will never produce a manifest entry, closer to "same validation as normal rule loading" (setup rejects target first, but for a different reason: target is required there, not skippable).

```suggestion
			if validateErr := util.ValidateVersionRange(rule.VersionRange); validateErr != nil {
				return ex.Wrapf(validateErr, "validating version in rule file %s", path)
			}
			if rule.Target == "" {
				continue
			}
```

darccio: The wrapped error only names the rule *file* (`path`), not which rule inside it failed. `rules` is a `map[string]yamlRule` and the loop currently discards the key. For a file with several rules, capturing the name makes the failure immediately actionable instead of requiring the reader to scan the whole file.

```suggestion
			for name, rule := range rules {
				if rule.Target == "" {
					continue
				}
				if validateErr := util.ValidateVersionRange(rule.VersionRange); validateErr != nil {
					return ex.Wrapf(validateErr, "validating version for rule %q in file %s", name, path)
				}
```

tool/internal/manifest/manifest.go
darccio: `rules` is a map, so Go's randomized iteration order means the specific error reported when a file has more than one invalid rule varies between runs — generation always fails, but the message isn't reproducible. Sorting the keys first fixes that and also gives finding #2 a `name` to report. (`slices` is already imported in this file for the manifest sort.)

```suggestion
			var rules map[string]yamlRule
			if unmarshalErr := yaml.Unmarshal(content, &rules); unmarshalErr != nil {
				return ex.Wrapf(unmarshalErr, "parsing rule file %s", path)
			}
			names := make([]string, 0, len(rules))
			for name := range rules {
				names = append(names, name)
			}
			slices.Sort(names)
			for _, name := range names {
				rule := rules[name]
```

tool/internal/manifest/manifest_test.go
```diff
@@ -51,11 +56,28 @@ server:
...
+  target: example.com/target
+  version: v1.0.0,
+`)
+
+	_, err := Generate(root)
+	require.ErrorContains(t, err, `version "v1.0.0," must use non-empty start and end bounds`)
+	require.ErrorContains(t, err, "validating version in rule file otelc.yaml")
+	require.ErrorContains(t, err, "loading rules for module example.com/test")
+	require.ErrorContains(t, err, "generating manifest from")
+}
```
darccio, line 145: As flagged previously that a rule with an empty target skips version validation entirely (the `continue` fires before `ValidateVersionRange` runs). There's no test pinning that behavior today, so it's easy to break silently — either by accident, or as a side effect of fixing the previously flagged issue without meaning to. Add a test that documents current behavior explicitly, so any future change to the ordering is a deliberate, visible diff here.

```suggestion
}

func TestGenerateSkipsVersionValidationForEmptyTarget(t *testing.T) {
	root := t.TempDir()
	writeModule(t, root, "module", "example.com/test")
	writeRuleFile(t, root, "module/otelc.yaml", `
notarget:
  version: v1.0.0,
`)

	got, err := Generate(root)
	require.NoError(t, err)
	require.Empty(t, got)
}
```

darccio, line 145: Only the empty-end-bound form (`"v1.0.0,"`) is exercised through `Generate`. `ValidateVersionRange` also rejects too-many-commas, invalid-semver, and inverted-bounds forms (covered directly in `tool/util/shared_test.go`), but never through the manifest path — so a future change that breaks how `Generate` wires up the helper (e.g. an argument swap) wouldn't be caught here. A small table test closes that gap without duplicating `shared_test.go`'s exhaustive coverage of the helper itself.

```suggestion
}

func TestGenerateRejectsOtherInvalidVersionRanges(t *testing.T) {
	tests := []struct {
		name    string
		version string
	}{
		{name: "too many commas", version: "v1.0.0,v2.0.0,v3.0.0"},
		{name: "invalid semver", version: "not-a-version"},
		{name: "inverted bounds", version: "v2.0.0,v1.0.0"},
	}

	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			root := t.TempDir()
			writeModule(t, root, "module", "example.com/test")
			writeRuleFile(t, root, "module/otelc.yaml", "invalid:\n  target: example.com/target\n  version: "+test.version+"\n")

			_, err := Generate(root)
			require.Error(t, err)
		})
	}
}
```

tool/internal/manifest/manifest.go
darccio: Setup's rule loading (`tool/internal/setup/match.go:87-98`) rejects an empty target with an error and validates non-empty targets with `rule.ValidateTarget`. The manifest loader only mirrors the version check added in this PR — a malformed glob target (the same class of "generator accepts what loading rejects" bug #1184 describes for versions) can still reach `instrumentation-manifest.json` unchanged. If target validity matters at generation time the way version validity does, it's worth mirroring here too; if the manifest is intentionally scoped to accept any non-empty target regardless of glob validity, it'd help to say so in a comment.

```suggestion
			if rule.Target == "" {
				continue
			}
			if validateErr := ruleutil.ValidateTarget(rule.Target); validateErr != nil {
				return ex.Wrapf(validateErr, "validating target in rule file %s", path)
			}
			if validateErr := util.ValidateVersionRange(rule.VersionRange); validateErr != nil {
				return ex.Wrapf(validateErr, "validating version in rule file %s", path)
			}
```

darccio (changes requested):
`setup/match.go` and `manifest.go` both now call `util.ValidateVersionRange`, but the surrounding checks (target validation, error wrapping, empty-value handling) are separate hand-maintained code paths. Nothing keeps them in sync going forward — issue #1184 happened because one loader's rules drifted from the other's, and the same class of gap (see the target-validation note in this review) can reappear silently. Not a blocker for this focused bugfix, but worth a follow-up issue to extract a single shared rule-validation function that both `setup` and `manifest` call, so "same validation as normal rule loading" is structurally guaranteed rather than something each PR has to remember to re-check.

ADITYA-CODE-SOURCE:
@darccio Thanks for the detailed review I updated the PR to address the version ordering, deterministic rule iteration, rule-name error context, and extra manifest tests. I kept target validation as a separate follow-up to keep #1184 focused.

darccio (approved):

PR #1186, opened 2026-08-20
fix(tool/internal/instrument): resolve cross-file generic receiver constraints 

### LINKED ISSUE #1171: bug(tool/internal/instrument): generic receiver constraint recovery doesn't search other files of the same package, falls back to `any`
> **Note:** this is a known, self-flagged gap, not a new discovery. PR #1122 (fixes #1121) explicitly documented it in a "Known limitation" section and referenced "the linked issue" for the follow-up — but no such issue was ever filed, as far as I can find (searched `findGenericTypeDecl`, "cross-file", "different file", "sibling file", "whole-package", "go/packages" as both issue and PR terms, and #1121's full cross-reference timeline, which only shows #1122 and #1126). Filing it now so it's tracked.

### Problem

When instrumenting a method with a generic receiver, the tool tries to recover the receiver's real type constraint (e.g. `comparable`) so the generated trampoline's own type-parameter declaration matches the source, rather than widening it to `any`. This recovery only searches the single already-parsed file containing the instrumented method. If the generic type is declared in a *different* file of the same package — an ordinary Go layout — the lookup misses it and the constraint silently falls back to `any`, with no error or warning.

### Why it happens

`findGenericTypeDecl` (`tool/internal/instrument/trampoline.go:809-826`) only receives the single `*dst.File` being instrumented (`ip.target`), which `parseFile` sets from a strictly one-file-at-a-time processing loop (`instrument.go`'s per-file loop calling `parseFile`, `apply_func.go:370-376`). No code path combines a package's files into one AST before this lookup runs, so a type declared in a sibling file is structurally invisible to it.

```go
// findGenericTypeDecl searches file's top-level declarations for a generic type
// declaration named typeName ... The declaration may live in a different file
// of the same package; that case is not resolved today, so it falls through
// to the caller's any fallback.
func findGenericTypeDecl(file *dst.File, typeName string) *dst.FieldList {
	if file == nil || typeName == "" {
		return nil
	}
	for _, decl := range file.Decls {
		...
	}
	return nil
}
```

When it returns `nil`, `receiverConstraintAt` (`trampoline.go:833-848`) skips its positional-match loop and returns `ast.Ident("any")` at line 847.

### Minimal reproducer

```go
// file_a.go
package pkg

type Box[T comparable] struct { value T }

// file_b.go — same package, different file
package pkg

func (b *Box[T]) Get() T { return b.value }
```

Instrument `Get` with a func rule. `findGenericTypeDecl` receives only `file_b.go`'s AST, finds no `type Box[...]` there, and returns `nil`.

### Current behavior

The generated trampoline's type-parameter declaration is emitted as `[T any]`, silently, purely because of file layout.

### Expected behavior

The generated trampoline's type-parameter declaration should be `[T comparable]`, matching what already happens correctly when the type declaration and the method are in the same file (covered by `trampoline_receiver_test.go`).

### Relevant code path / file locations

- `tool/internal/instrument/trampoline.go`
  - `findGenericTypeDecl` (809-826) — the single-file-scoped lookup
  - `extractReceiverTypeParams` (758-791) — calls it, only production call sites
  - `receiverConstraintAt` (833-848) — the `any` fallback, line 847
  - `findTargetGenericType` (480-504) — combines receiver + method-own type params
  - `buildTrampSignature` (506-531, calls at 511 and 516) — the only place the result is consumed for codegen; this is what actually changes the emitted trampoline signature
- `tool/internal/instrument/apply_func.go:370-376` (`parseFile`) and `tool/internal/instrument/instrument.go` (per-file loop) — confirm files are parsed and processed one at a time, never as a combined multi-file AST

### Why this matters

This is a real, deterministic gap triggered by ordinary Go file layout (splitting a generic type's declaration from its methods across files), not a hypothetical edge case. The degradation is silent — nothing indicates to the author that their instrumentation is running with a weaker constraint than the source declares. Whether this manifests as a generated-code compile failure (as #1125 showed for the same-file case before #1122 fixed it) depends on whether a given trampoline body path exercises the constraint; I haven't audited every `impl.tmpl` path for that, so I'm not asserting compile failure as a certainty — only that the constraint information is genuinely lost, which is the exact class of problem #1121/#1122 were filed to fix, just for the file layout #1122 explicitly left unhandled.

### Suggested test cases

- A two-file unit test: parse the type declaration and the method into two separate `*dst.File` values, run `extractReceiverTypeParams` against the method's file only, and assert today's actual (degraded) fallback as a baseline.
- Same setup, asserting the constraint is correctly recovered once fixed.
- A golden end-to-end test with the split-file layout, compared against the existing same-file golden case's precision.
- A regression test confirming the genuine "no declaration exists anywhere in the package" case still correctly falls back to `any` rather than erroring.

### Possible direction 

`findGenericTypeDecl` would need visibility into the rest of the package's files, not just the one being instrumented. Since the pipeline currently parses and instruments one file at a time, this likely means threading some form of package-level type-declaration index into the instrument phase — a structural change to a currently file-scoped step, not a one-line fix.

darkraider01:
## Description
When instrumenting a method on a generic struct whose `type` definition is located in a different file of the same package (rather than in the same file as the method), `otelc` previously failed to resolve the type parameter constraints. This caused trampoline generation to either fall back to `any` or fail compilation when constraints referenced imported packages that were not present in the target file.
This PR introduces the following changes:
1. **Package File Discovery & Caching (`tool/internal/instrument/package_files.go`):**
   - Added `packageGoFiles()` to retrieve package `.go` files from `ip.compileArgs`.
   - Added `loadSiblingASTs()` to lazily parse and cache sibling package ASTs with `ast.ParseFileFast`.
   - Ensured deterministic sibling traversal by sorting map keys.
2. **Cross-file Constraint Recovery & Import Injection (`tool/internal/instrument/trampoline.go`, `tool/internal/instrument/imports.go`):**
   - Added `resolveGenericTypeDecl()` to look up generic struct declarations across package sibling files when not found in the target file.
   - Added `injectConstraintImports()` to inspect constraint expressions, resolve their import paths via `imports.ResolveAlias()`, and inject missing imports into the target file AST and `importcfg` via `addRuleImports()`.
   - Refactored `findTargetGenericType()` to run once per rule at the top of `createTrampoline()` and propagate errors cleanly.
3. **Import Alias Helper (`tool/internal/imports/imports.go`):**
   - Added `ResolveAlias()` to look up the import path for a given alias or inferred package name within an AST file.
4. **Testing:**
   - Added unit tests in `package_files_test.go`, `trampoline_test.go`, and `imports_test.go`.
   - Added end-to-end golden test under `tool/internal/instrument/testdata/golden/generic-receiver-cross-file/`.
## Motivation
Fixes #1171
Generic receiver methods whose type declarations reside in sibling files could not recover constraints (e.g. `[T fmt.Stringer]`), leading to missing type parameters in trampoline functions or compilation failures when constraints required package imports in the target file.

## Checklist
- [x] Tests pass: `make test`
[^I-know-my-stuff]:

### DIFF
--- a/tool/internal/imports/imports.go
+++ b/tool/internal/imports/imports.go
@@ -106,6 +106,19 @@ func FindNew(ctx context.Context, root *dst.File, ruleImports map[string]string,
 	return result
 }
 
+// ResolveAlias reports the import path alias refers to within root's own
+// import declarations, using the same alias resolution parseFile already
+// performs (explicit aliases, or pkgload.ResolvePackageName for implicit
+// ones). This is used to look up what a package-qualifier in an expression
+// copied out of root (e.g. "constraints" in constraints.Ordered) refers to,
+// when that expression is about to be written into a different file that may
+// not share the same alias for that import path.
+func ResolveAlias(ctx context.Context, root *dst.File, alias string, buildFlags ...string) (string, bool) {
+	m := parseFile(ctx, root, buildFlags...)
+	path, ok := m.AliasToPath[alias]
+	return path, ok
+}
+
 // getExisting returns a map of alias -> path for all imports in the file.
 // For imports without explicit aliases, the package name is used as the key.
 func getExisting(ctx context.Context, root *dst.File, buildFlags ...string) map[string]string {
--- a/tool/internal/imports/imports_test.go
+++ b/tool/internal/imports/imports_test.go
@@ -116,6 +116,67 @@ func TestGetExisting(t *testing.T) {
 	}
 }
 
+func TestResolveAlias(t *testing.T) {
+	tests := []struct {
+		name     string
+		root     *dst.File
+		alias    string
+		wantPath string
+		wantOK   bool
+	}{
+		{
+			name: "explicit alias",
+			root: &dst.File{
+				Decls: []dst.Decl{
+					&dst.GenDecl{
+						Tok: token.IMPORT,
+						Specs: []dst.Spec{
+							&dst.ImportSpec{
+								Name: dst.NewIdent("ctx"),
+								Path: &dst.BasicLit{Value: `"context"`},
+							},
+						},
+					},
+				},
+			},
+			alias:    "ctx",
+			wantPath: "context",
+			wantOK:   true,
+		},
+		{
+			name: "implicit alias resolved from package name",
+			root: &dst.File{
+				Decls: []dst.Decl{
+					&dst.GenDecl{
+						Tok: token.IMPORT,
+						Specs: []dst.Spec{
+							&dst.ImportSpec{Path: &dst.BasicLit{Value: `"net/http"`}},
+						},
+					},
+				},
+			},
+			alias:    "http",
+			wantPath: "net/http",
+			wantOK:   true,
+		},
+		{
+			name:     "alias not found",
+			root:     &dst.File{},
+			alias:    "constraints",
+			wantPath: "",
+			wantOK:   false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			path, ok := ResolveAlias(t.Context(), tt.root, tt.alias)
+			assert.Equal(t, tt.wantOK, ok)
+			assert.Equal(t, tt.wantPath, path)
+		})
+	}
+}
+
 func TestCreateSpec(t *testing.T) {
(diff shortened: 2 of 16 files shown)


y1yang0 (changes requested):
Let's discuss proposal/bug first

PR #1188, opened 2026-08-20
fix(pkg): preserve trace context in database/sql transactions

### LINKED ISSUE #1187: [database/sql] Transaction Commit and Rollback spans are detached from distributed traces
### Describe the bug

In `instrumentation/database/sql/client.go`, the instrumentation hooks for `Tx.Commit()` and `Tx.Rollback()` create new spans using `context.Background()` instead of the original transaction's context. Because they use an empty background context, the `commit` and `rollback` spans will not be linked to the parent trace, resulting in orphaned spans and broken distributed traces for database transactions.

### Environment

 - OS: All
 - Go Version: All
 - Version: `main` branch

### To Reproduce

1. Instrument an application using `go.opentelemetry.io/otelc/instrumentation/database/sql`.
2. Begin a transaction using `db.BeginTx(ctx, nil)`.
3. Perform a query (`tx.ExecContext(ctx, ...)`).
4. Commit the transaction (`tx.Commit()`).
5. Observe the generated traces in your tracing backend (e.g. Jaeger).
6. **See error:** You will notice that the `COMMIT` span is not a child of the original trace context. It appears as an isolated, single-span trace.

**Code Pointers:**
In `instrumentation/database/sql/client.go`, the `beforeTxCommitInstrumentation` and `beforeTxRollbackInstrumentation` hooks hardcode `context.Background()`:
```go
// Line 515
func beforeTxCommitInstrumentation(ictx hook.HookContext, tx *sql.Tx) {
	// ...
}

// Line 532
func beforeTxRollbackInstrumentation(ictx hook.HookContext, tx *sql.Tx) {
	// ...
	instrumentStart(ictx, context.Background(), "rollback", "ROLLBACK", tx.Endpoint, tx.DriverName, tx.DSN, tx.DbName)
}
```

### Expected behavior

The `COMMIT` and `ROLLBACK` spans should be correctly linked as children of the trace context in which the transaction was initiated (`BeginTx`), providing end-to-end visibility into the transaction's lifecycle.

### Additional context

**Impact:**
Any application relying on compile-time instrumentation for distributed tracing will have broken, detached spans for transaction finalization. This pollutes trace backends with orphaned traces and deprives developers of complete transaction latency visibility.

**Proposed Solution:**
1. In `instrumentation/database/sql/otelc.yaml`, update the `add_new_field_tx` directive to inject a new context field (e.g., `OtelCtx context.Context`) into the `Tx` struct.
2. In `client.go`, modify `beforeTxInstrumentation` (which hooks `BeginTx` and receives the transaction's context) to store the original `ctx` in `ictx`.
3. In `afterTxInstrumentation`, retrieve the context from `ictx` and assign it to the newly injected `tx.OtelCtx`.
4. In `beforeTxCommitInstrumentation` and `beforeTxRollbackInstrumentation`, use `tx.OtelCtx` instead of `context.Background()` to start the span (falling back to `context.Background()` only if `tx.OtelCtx` is nil).

vishalmore90:
## Description
This PR fixes an issue where database spans generated during a transaction (such as `START TRANSACTION`, `INSERT`, `COMMIT`, or `ROLLBACK`) were orphaned and disconnected from their parent HTTP server trace.

The root cause was that `tx.Commit()` and `tx.Rollback()` in `database/sql` do not take a `context.Context` parameter, causing the `otelc` tracing hooks to fallback to `context.Background()`. 

To fix this, we've utilized `otelc`'s code injection capabilities to extend the `sql.Tx` struct:
- Added an `OtelCtx context.Context` field to `database/sql.Tx` via `instrumentation/database/sql/otelc.yaml`.
- Modified `beforeTxInstrumentation` to save the parent context inside the hook data map.
- Modified `afterTxInstrumentation` to inject the parent context into the newly initialized `*sql.Tx` instance.
- Modified `beforeTxCommitInstrumentation` and `beforeTxRollbackInstrumentation` to use `tx.OtelCtx` as the active context instead of falling back to `context.Background()`.

An E2E test `TestHTTPServerDBClient_Transaction` has been added to verify that trace IDs remain consistent across the transaction boundaries and span parenting correctly follows the HTTP server root span.

## Motivation

Database transaction spans were incorrectly creating disjointed traces instead of being grouped under the active parent HTTP trace. Preserving the context in the `sql.Tx` struct allows end-to-end trace correlation and correctly associates subsequent queries and commits to the original caller.

Fixes #1187

## Checklist

- [x] Tests pass: `make test`


instrumentation/database/sql/client.go
```diff
@@ -214,6 +217,10 @@ func afterTxInstrumentation(ictx hook.HookContext, tx *sql.Tx, err error) {
 	tx.DriverName = dbRequest.DriverName
 	tx.DSN = dbRequest.Dsn
 	tx.DbName = dbRequest.DbName
+	if parentCtx, ok := callData["parentCtx"].(context.Context); ok {
+		tx.OtelCtx = parentCtx
+	}
+	instrumentEnd(ictx, err)
```
darkraider01, line 223: wait, this fires `instrumentEnd` twice - there's already `defer instrumentEnd(ictx, err)` a few lines up (line 204). SDK guards `End()` so it's harmless, but I think this line's just dead weight, the defer already covers it

test/apps/httpserverdbclient/go.mod
```diff
@@ -4,4 +4,12 @@ go 1.25.0
...
-require go.opentelemetry.io/otelc/test/shared/testdb v0.0.0-00010101000000-000000000000
+require (
+	go.opentelemetry.io/otel/trace v1.45.0
+	go.opentelemetry.io/otelc/test/shared/testdb v0.0.0-00010101000000-000000000000
+)
+
+require (
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
+	go.opentelemetry.io/otel v1.45.0 // indirect
+)
```
darkraider01, line 15: heads up, `go mod tidy` reverts this whole block in CI - the "Check go.mod/go.sum Tidy" job is failing on exactly this. Looked at main.go and it never actually imports otel/trace, so I think this was leftover from an earlier version of the test. Running `go mod tidy` locally should clean it up

test/apps/httpserverdbclient/go.sum
```diff
@@ -0,0 +1,16 @@
+github.com/cespare/xxhash/v2 v2.3.0 h1:UL815xU9SqsFlibzuggzjXhog7bL6oX9BbNZnL2UFvs=
```
darkraider01, line 1: same as the go.mod comment - CI deletes this file back to empty right now

test/e2e/httpserver_dbclient_test.go
```diff
@@ -60,3 +61,52 @@ func TestHTTPServerDBClient(t *testing.T) {
...
+	f := testutil.NewTestFixture(t)
+
+	frontPort := testutil.FreePort(t)
+
+	f.BuildAndStart("httpserverdbclient", fmt.Sprintf("-front-port=%d", frontPort))
+	testutil.WaitForTCP(t, fmt.Sprintf("127.0.0.1:%d", frontPort))
+
+	// We use the regular instrumented httpclient but we modify the addr to include the path if we couldn't,
+	// wait, httpclient hardcodes /hello. Let's just use standard Go http client. It won't generate a client span,
+	// but the server span will become the root span.
```
darkraider01, line 75: think this still has some thinking-out-loud left in it ("wait, ... Let's just"). the actual point is solid, just needs tightening - something like "a plain http.Get won't produce a client span, so the server span becomes the trace root" covers it in one line

test/e2e/httpserver_dbclient_test.go
```diff
@@ -60,3 +61,52 @@ func TestHTTPServerDBClient(t *testing.T) {
...
+	resp, err := http.Get(fmt.Sprintf("http://127.0.0.1:%d/tx", frontPort))
+	require.NoError(t, err)
+	resp.Body.Close()
+
+	// Wait for the server span to be exported.
+	f.WaitForSpans(4) // HTTP server -> START -> INSERT -> COMMIT
```
darkraider01, line 81: this test timed out in CI btw - full 10m budget, panicked. my guess is spinning up a second `BuildAndStart` on top of the existing one is what's eating the time. might be worth seeing if this can reuse the fixture from `TestHTTPServerDBClient` instead of building a whole second app instance

instrumentation/database/sql/client.go
```diff
@@ -192,6 +192,9 @@ func beforeTxInstrumentation(ictx hook.HookContext, db *sql.DB, ctx context.Cont
 		return
 	}
 	instrumentStart(ictx, ctx, "begin", "START TRANSACTION", db.Endpoint, db.DriverName, db.DSN, db.DbName)
+	if data, ok := ictx.GetData().(map[string]interface{}); ok {
+		data["parentCtx"] = ctx
```
darkraider01, line 196: quick question - `instrumentStart` already stashes the span's own context under `"ctx"` a few lines above this. is `parentCtx` meant to be different from that on purpose? because using the pre-transaction context here means COMMIT/ROLLBACK end up as siblings of START under the HTTP span, not children of it. could be exactly what you want, just wasn't called out anywhere and it's not obvious from the diff alone

instrumentation/database/sql/client.go
```diff
@@ -528,7 +539,11 @@ func beforeTxRollbackInstrumentation(ictx hook.HookContext, tx *sql.Tx) {
 	if tx == nil {
 		return
 	}
-	instrumentStart(ictx, context.Background(), "rollback", "ROLLBACK", tx.Endpoint, tx.DriverName, tx.DSN, tx.DbName)
+	ctx := context.Background()
+	if tx.OtelCtx != nil {
+		ctx = tx.OtelCtx
+	}
+	instrumentStart(ictx, ctx, "rollback", "ROLLBACK", tx.Endpoint, tx.DriverName, tx.DSN, tx.DbName)
```
darkraider01, line 546: same OtelCtx wiring as commit but I don't see a rollback case in the new test - might be worth adding one, since that's the branch most likely to actually diverge in behavior

darkraider01 (commented):
Left a few line comments below

PR #1189, opened 2026-08-21
chore(tool): consolidate tool tests into per-source-file pairing

y1yang0:

Issue #1128, opened 2026-08-17
Consolidate tests into a single _test.go per source file


y1yang0:
Tests are currently tending to get split out into narrowly scoped files, for example trampoline_generic_test.go alongside trampoline_test.go, plus some odd ones like helpers_extra_test.go, shared_paths_test.go, and toolexec_dropin_test.go. I think we should only accept the {trampoline.go, trampoline_test.go} pairing, keeping all test code for a source file in that single _test.go. It's simple and intuitive: we can always find the relevant tests in <file>_test.go, and our attention isn't scattered across a sea of files.

Orphaned test files (no matching source file):
```
{ex_test.go, fatal_test.go}
{ast_test.go, funcmatch_test.go, parser_extra_test.go, primitives_builders_test.go, shared_extra_test.go, shared_helpers_test.go}
{api_sync_test.go, call_placeholder_test.go, call_template_extra_test.go, hookcontext_runtime_test.go, impl_template_test.go, logger_delegate_test.go, tool_version_test.go, toolexec_exec_test.go, toolexec_extra_test.go, trampoline_generic_test.go}
{profile_extra_test.go}
{extra_helpers_test.go, logger_delegate_test.go, match_one_rule_test.go, pin_helpers_test.go, setup_extra_test.go}
{helpers_extra_test.go, shared_paths_test.go, toolexec_dropin_test.go}
```
Segmented test files(many test files map to a single source):
```
{parser_extra_test.go, shared_extra_test.go, shared_helpers_test.go, call_template_extra_test.go, toolexec_extra_test.go, trampoline_generic_test.go, profile_extra_test.go, extra_helpers_test.go, setup_extra_test.go, pin_helpers_test.go, helpers_extra_test.go, shared_paths_test.go, toolexec_dropin_test.go}
```

darkraider01:
i'd like to take it !

y1yang0:
@darkraider01 Thanks for doing this! Assigned.

darkraider01:
Happy to take this. Surveyed the tree — 28 orphaned test files across tool/ex, tool/internal/{ast,instrument,profile,setup} and tool/util (a few more than the issue lists; trampoline_receiver_test.go and apply_directive_test.go landed since). Targets are mostly obvious, and I checked there are no helper-function name collisions, so the merges shouldn't hit redeclaration errors.

Two spots where the rule collides with build constraints, which I'd rather not guess at:

[...]

PR #1190, opened 2026-08-21
feat(tool): enforce test file naming conventions in CI

darkraider01:
## Summary

- Adds `tool/cmd/check-test-names`, a Go checker enforcing exact `foo_test.go` <-> `foo.go` pairing across `tool/`, `pkg/`, and `instrumentation/` unit packages (no prefix matching, per discussion on #1128 — that's what let segmented test files creep in).
- Exempts dedicated scenario suites (`test/integration`, `test/e2e`, `test/bench`, `test/latestlibbuild`, `test/latestlibrun`, `test/versionmatrix`), `testdata/` fixtures, `demo/`, and dot-directories.
- Registers legitimate 1:1 exceptions in an explicit allowlist with documented rationale (platform-specific builds, fuzz targets, shared test helpers, and end-to-end tests spanning more than one source file within a package).
- Wires the check into `make check-test-file-naming` and a new `check-test-file-naming` CI job in `check-consistency.yaml`, gated into the `done` status check.
- Renames `tool/internal/instrument/logger_delegate_test.go` to `instrument_delegate_test.go` to better associate it with `instrument.go`, as agreed with @y1yang0.

Note: `render_raw_code_test.go` and `assert_fatal_test.go` are allowlisted for now rather than folded into their paired files, since #1189 (open) already does that consolidation — allowlisting avoids duplicating/conflicting with that PR. The allowlist entries can be dropped once #1189 merges.

## Test plan

- [x] `go test ./tool/cmd/check-test-names/...` — new unit tests cover exact match, orphan detection, allowlisting, exempt directories, testdata, and demo/dot-dir skipping
- [x] `go run ./tool/cmd/check-test-names` passes clean against the current tree
- [x] Manually verified an injected orphan test file is caught and reported, then passes again once removed
- [x] `go test ./tool/internal/instrument/... -run TestInstrumentPhaseLogDelegators` passes after the rename
- [x] `gofmt` and the license-header check pass on the new files


tool/cmd/check-test-names/main.go
```diff
@@ -0,0 +1,128 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Command check-test-names verifies that every unit test file under this
+// repository's Go packages pairs 1:1 with a source file of the same name in
+// the same directory (foo_test.go <-> foo.go). Prefix-style pairing (e.g.
+// foo_extra_test.go treated as belonging to foo.go) is rejected on purpose:
+// it is how segmented, hard-to-navigate test files crept into the tree
+// before. Legitimate exceptions are registered in allowlist.go together with
+// the reason they can't follow the 1:1 rule.
```
y1yang0:
```suggestion
// Command check-test-names verifies that every unit test file under this
// repository's Go packages pairs 1:1 with a source file of the same name in
// the same directory (foo_test.go <-> foo.go).
```

tool/cmd/check-test-names/allowlist.go
```diff
@@ -0,0 +1,48 @@
...
+// renamed or split to satisfy the exact-match rule. Keeping the reason next
+// to the entry is what keeps this a registry of deliberate exceptions rather
+// than an escape hatch for the segmented test files the naming rule exists
+// to prevent (see the discussion on #1128). Paths are slash-separated,
+// relative to the repository root.
+// seeV1PackageRationale is shared by the v2/v3 openai-go allowlist entries,
+// which mirror their v1 counterpart's rationale exactly.
+const seeV1PackageRationale = "see the v1 package rationale."
+
+var allowlist = map[string]string{ //nolint:gochecknoglobals // private lookup table
```
y1yang0: I'd suggest a plain list is enough, no need to spell out a reason for every entry.

The {foo.go + foo_test.go} pairing is an easy rule to follow, so there's no real risk of someone dropping a file in as an "escape hatch" out of laziness. In other words, if a file shows up in allowlist.go, that already implies it was a deliberate choice. If the author feels like it, they're welcome to add a one-line comment explaining why, but it shouldn't be required.

tool/cmd/check-test-names/main.go
darkraider01: Applied in 18301b0  thanks, the trimmed version reads better.

tool/cmd/check-test-names/allowlist.go
darkraider01: Agreed, changed in 18301b0. `allowlist` is now a plain `[]string`; `checkTree` builds the lookup set internally. I kept short inline comments on the four entries where the reason isn't obvious from the filename (the two pending #1189 folds, the `!windows` exec path, and the fuzz target) and dropped the rest.

y1yang0 (approved):

PR #1191, opened 2026-08-21
docs(tool): document HookContext accessor panics for generic targets

### LINKED ISSUE #1174: docs(tool): HookContextImpl.GetParam/SetParam/GetReturnVal/SetReturnVal panic for any generic target, undocumented and undiscoverable until runtime
### Problem

For any generic function or generic-receiver method target, the generated `HookContextImpl{suffix}`'s `GetParam`, `SetParam`, `GetReturnVal`, and `SetReturnVal` methods are replaced with unconditional `panic(...)` bodies. A hook author who writes `ctx.GetParam(0)` (or any of the other three) inside their own Before/After hook function for a generic target gets a runtime panic — with no compile-time warning, no lint, and nothing in the tool's documentation or golden examples flagging that these specific methods are unsafe to call for generic targets. The same `hook.HookContext` interface is handed to hook authors regardless of whether the underlying function is generic, so there's no type-level signal that these four methods have silently become unsafe.

Related but independent finding from the same investigation: #1173 — same neighborhood of code (`rewriteHookContextMethods`), different symptom (dead code, not a runtime panic risk), no shared fix.

### Why it happens

`rewriteHookContextMethods` (`tool/internal/instrument/trampoline.go:895-938`) checks whether the target is generic:

```go
if findTargetGenericType(ip.target, ip.targetFunc) != nil {
	makeMethodPanic(methodGetParam, "GetParam is unsupported for generic functions")
	makeMethodPanic(methodGetRetVal, "GetReturnVal is unsupported for generic functions")
	makeMethodPanic(methodSetParam, "SetParam is unsupported for generic functions")
	makeMethodPanic(methodSetRetVal, "SetReturnVal is unsupported for generic functions")
	return
}
```

`makeMethodPanic` (`trampoline.go:940-951`) replaces the method body wholesale with a single `panic("...")` statement. This is gated purely on genericity — it applies identically whether or not an After hook is present, and regardless of how the hook author intends to use the accessor methods.

### Current behavior

Confirmed by tracing `populateHookContext`, `callBeforeHook`, and `callAfterHook`: the tool's own generated code never calls these four methods — it only ever writes directly to the `HookContextImpl`'s `Params`/`ReturnVals` fields, and passes the raw parameter values as separate positional arguments to the hook function (e.g. `GenericFuncBefore(ctx, p1, p2)`), which is presumably why the repo's own golden testdata hook examples never call `ctx.GetParam(...)` and so never trip this path. The panic is real, but only reachable through hand-written hook code that a user could plausibly write, especially since nothing distinguishes a generic-target `HookContext` from a non-generic one at the type level — both satisfy the same `hook.HookContext` interface.

### Minimal reproducer

```go
// target
func GenericFunc[T any](p1 T, p2 int) (T, error) { return p1, nil }
```

```go
// hook.go — a plausible, not-unreasonable hook author mistake
func GenericFuncBefore(ctx hook.HookContext, p1 interface{}, p2 int) {
	v := ctx.GetParam(0) // panics at runtime: "GetParam is unsupported for generic functions"
	_ = v
}
```

Nothing in the rule file, the hook function's own signature, or any build-time check stops this from compiling; the panic only fires when the instrumented `GenericFunc` is actually called at runtime.

### Expected behavior

At minimum, this should be documented clearly wherever hook-authoring is explained, so it isn't discovered only via a production panic. Ideally there's also some earlier signal — e.g. instrumentation-time validation that flags a hook function calling these methods when the target is generic, though this may be difficult given hook bodies aren't fully analyzed today.

### Relevant code path / file locations

- `tool/internal/instrument/trampoline.go:895-938` (`rewriteHookContextMethods`) — the genericity gate
- `tool/internal/instrument/trampoline.go:940-951` (`makeMethodPanic`) — the panic-body installer
- `tool/internal/instrument/trampoline.go:604-644` (`populateHookContext`) — confirms tool-generated code never calls these methods itself
- `tool/internal/instrument/trampoline.go:306-356` (`callBeforeHook`/`callAfterHook`) — confirms hook functions are called with positional args, not required to go through `ctx.GetParam`
- `tool/internal/instrument/testdata/golden/generic-functions/hook.go` — existing example hooks that happen to avoid this path, which is likely why it hasn't surfaced before

### Why this matters

This is a real, if narrow, correctness/DX gap: valid-looking hook code compiles cleanly and only fails at runtime, for a reason that isn't discoverable without already knowing this specific implementation detail. It's distinct from the already-tracked `HookContext` panic bugs (#916, #903 — both about `KeyData` type-assertion panics — and #726, about `SetParam(idx, nil)` silently dropping writes) — none of those concern the generic-target accessor-panic behavior.

### Suggested test cases

- A test confirming the current panic behavior is intentional and stable (i.e., a regression guard, not a "fix" test) — calling each of the four methods on a generic-target `HookContextImpl` and asserting the panic message.
- If documentation is added: no code test needed, but the golden testdata could include a comment noting why the example hooks avoid `ctx.GetParam`/etc.

### Possible direction (not a proposed implementation)

Not proposing a specific fix — supporting the accessor methods properly for generics would need real design work (the tool has no `go/types`-based mechanism today, per the earlier constraint-recovery investigation). The lower-cost option is documentation: state plainly, wherever hook-authoring conventions are documented, that these four methods panic for generic targets and hook authors should use the positional parameters instead.

darkraider01:
## Summary

Fixes #1174.

The Generic Functions limitation in the hook-authoring guide said we cannot use `HookContext` to *modify* parameters or return values, naming `SetParam` and `SetReturnVal`. Reading is equally unavailable: for a generic target `GetParam` and `GetReturnVal` are replaced with panicking bodies too, so a hook that only reads still panics at runtime. Nothing flags it at build time, which is what makes it hard to discover.

Docs-only, per y1yang0's confirmation on the issue that this should be documented:

- `docs/instrument-guide.md`: expands the limitation to name all four methods, notes that the failure only surfaces when the instrumented function actually runs, and shows reading the values from the hook's own parameters instead.
- `tool/internal/instrument/testdata/golden/generic-functions/hook.go`: short comment on why those example hooks read positionally, so the omission reads as deliberate rather than incidental. The issue suggested this.

No code changes. The panic generation itself (`rewriteHookContextMethods` / `makeMethodPanic` in `trampoline.go`) is already covered by the existing `generic-functions` golden test, so there is nothing to add there.

## Test plan

- [x] `npx markdownlint-cli -c .tools/markdownlint.yaml docs/instrument-guide.md` clean
- [x] `gofmt -l` clean on the touched fixture
- [x] `.github/scripts/license-check.sh` passes
- [x] Confirmed the fixture is consumed as a helper package (built by `buildTestcaseHelpers`, referenced from the rule YAML's `path:`) rather than copied into the instrumented source that golden output is compared against, so the added comment cannot shift the golden diff

### DIFF
--- a/docs/instrument-guide.md
+++ b/docs/instrument-guide.md
@@ -154,7 +154,24 @@ When implementing hooks, we must adhere to certain limitations:
 
    Importing other third-party libraries is not allowed.
 
-2. **Generic Functions**: If the target function is generic, we cannot use `HookContext` APIs to modify parameters or return values (e.g., `SetParam`, `SetReturnVal`).
+2. **Generic Functions**: If the target function or method receiver is generic, we cannot use the `HookContext` parameter and return value APIs.
+
+   `GetParam`, `SetParam`, `GetReturnVal`, and `SetReturnVal` are each replaced with a body
+   that panics, so reading panics just as writing does. Instead, we read the values from the
+   hook's own parameters, which already receive them positionally:
+
+   ```go
+   // Target: func GenericFunc[T any](p1 T, p2 int) (T, error)
+
+   // We read p1 and p2 from the hook's parameters, never through ictx.GetParam.
+   func GenericFuncBefore(ictx hook.HookContext, p1 interface{}, p2 int) {}
+
+   // The after hook likewise reads the return values from its own parameters.
+   func GenericFuncAfter(ictx hook.HookContext, r1 interface{}, r2 error) {}
+   ```
+
+   Nothing catches this at build time: a hook that calls one of these four methods on a
+   generic target compiles cleanly, and only panics once the instrumented function runs.
 
 ### GLS Operation for OTel SDK Instrumentation
 
--- a/tool/internal/instrument/testdata/golden/generic-functions/hook.go
+++ b/tool/internal/instrument/testdata/golden/generic-functions/hook.go
@@ -9,6 +9,10 @@ import (
 	"go.opentelemetry.io/otelc/pkg/hook"
 )
 
+// These hooks read their values from their own parameters rather than through
+// ctx.GetParam/GetReturnVal: those methods panic for generic targets, so the
+// positional parameters below are the only way to reach the values.
+
 func GenericFuncBefore(ctx hook.HookContext, p1 interface{}, p2 int) {}
 
 func GenericFuncAfter(ctx hook.HookContext, r1 interface{}, r2 error) {}


darccio (approved):
LGTM

PR #1195, opened 2026-08-21
feat: add ArgumentOfType and CallArgument to wrap_call

hannahkm:
## Description

1. Further expands the wrap_call rule to support FuncArgumentOfType and CallArgument N (which gets the Nth parameter to the wrapped function call, not including the receiver.
2. ONLY does the above for the inject raw code rule.

## Motivation

Further extend support for template variables.

---

## Checklist

- [ ] Tests pass: `make test`

[^I-know-my-stuff]:


tool/internal/instrument/call_template.go
```diff
@@ -180,10 +227,7 @@ func (t *callTemplate) compileExpression(node dst.Expr, enclosing *dst.FuncDecl)
 	}
 
 	// Replace placeholder with the actual node
-	result, replaced := replacePlaceholder(exprStmt.X, node)
-	if !replaced {
-		return nil, ex.New("template output did not contain placeholder expression")
-	}
+	result, _ := replacePlaceholder(exprStmt.X, node)
```
y1yang0: So besides adding new placeholder variables, this patch also changes the CallRule behavior? IIUC, previously the template always replaced the old code, while now we can conditionally replace it. If so, we should update rules.md and the PR description as well.

y1yang0: And remove return-value if not needed

tool/internal/instrument/call_template.go
```diff
@@ -255,6 +310,41 @@ func parseSnippetFuncDecl(src, label string) (*dst.FuncDecl, error) {
 	return funcDecl, nil
 }
 
+// exprSourceText renders expr back into Go source text.
+func exprSourceText(expr dst.Expr) (string, error) {
```
y1yang0: Move instrument.nodeSourceText and instrument.exprSourceText to ast.RenderNode and ast.RenderExpr respectively, as they are purely ast-manipulation functions and are likely used elsewhere in the future

tool/internal/instrument/call_template.go
hannahkm: That's a good point. I think it's better to keep the original behavior. I had originally removed this since adding more template variables enables the user to use other expressions and optionally leave `{{ . }}` out entirely. Instead, I'll just add an extra check to make sure the template actually contains `{{ . }}` before erroring.

hannahkm: Addressed in [b21bfa2](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1195/commits/b21bfa246757fe888f4bfd7697d9242407f3f384)

tool/internal/instrument/call_template.go
hannahkm: Done in [b20a983](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1195/commits/b20a98385c423d71b3e97df2d303644a653a15e7)

tool/internal/ast/render.go
```diff
@@ -0,0 +1,50 @@
...
+	restorer := decorator.NewRestorer()
+	if _, err := restorer.RestoreFile(synthetic); err != nil {
+		return "", ex.Wrapf(err, "failed to restore expression to source")
+	}
+	return RenderNode(restorer, cloned)
+}
+
+// RenderNode looks up node's restored counterpart in restorer and renders it
+// back to Go source text.
+func RenderNode(restorer *decorator.Restorer, node dst.Node) (string, error) {
```
y1yang0: They can be moved to shared.go

y1yang0 (approved):
Otherwise LGTM, thanks for doing this.

tool/internal/ast/render.go
hannahkm: Done in 0a41660!

PR #1197, opened 2026-08-21
fix(tool): reject invalid manifest rule targets

### LINKED ISSUE #1196: Manifest generator accepts invalid rule targets
### Description

Manifest generation accepts malformed, non-empty `rule.target` values that are rejected during normal rule loading.

For example:

```yaml
broken:
  target: example.com/[svc
  version: v1.0.0
```

Manifest generation accepts this target and may write it to `instrumentation-manifest.json`.

However, normal rule loading rejects the same target through `rule.ValidateTarget`.

### Expected behavior

Manifest generation should validate non-empty `rule.target` values using the same target validation as normal rule loading and return an error for malformed targets.

### Actual behavior

Manifest generation accepts malformed, non-empty targets and includes them in the generated manifest.

### Proposed fix

Validate non-empty manifest-side `rule.target` values while loading manifest entries, preferably by reusing `rule.ValidateTarget`.

Add focused regression tests covering:

* Rejection of malformed targets
* Acceptance of valid exact targets
* Acceptance of valid glob targets
* Preservation of the existing empty-target omission behavior

The broader shared-validation refactor between setup and manifest loading can remain separate.

ADITYA-CODE-SOURCE:
## Summary

Validate non-empty rule targets during manifest generation using the same target validation used by normal rule loading.

Malformed targets now fail with rule-name and rule-file context instead of being written to `instrumentation-manifest.json`. Empty targets remain omitted from manifest entries.

Fixes #1196

## Testing

- `go test -count=1 ./tool/internal/manifest ./tool/internal/rule ./tool/internal/setup`
- `make verify-manifest`
- `make lint/license-header`
- `./.bin/golangci-lint run --config .tools/golangci.yml ./tool/internal/manifest`
- `go list -deps ./tool/internal/manifest >/dev/null && go list -deps ./tool/internal/rule >/dev/null`
- `git diff --check`


tool/internal/manifest/manifest.go
```diff
@@ -154,3 +151,20 @@ func loadModuleEntries(moduleDir, modulePath string) (Manifest, error) {
 	}
 	return entries, nil
 }
+
```
y1yang0: Calling validatetarget in the caller function would be more simple and straightforward.

ADITYA-CODE-SOURCE: Updated. I moved `ValidateTarget` directly into `loadModuleEntries` before appending the manifest entry and removed the helper. I also flattened the existing directory branch to keep the function under the repo gocognit threshold without changing behavior, and added nested-directory coverage for both the non-module path and stat-error path.

ADITYA-CODE-SOURCE:
@y1yang0  all checks are green  Could you please take another look?

y1yang0 (approved):

PR #1092, opened 2026-08-13
fix(redis): redact AUTH/HELLO credentials in db.query.text

Anshumancanrock:
## Description

Redact AUTH / HELLO AUTH args in Redis `db.query.text`.

Handshake used to show up as `hello 3 auth default <password>`. Now it's `hello 3 auth ? ?`. Span is still created. SET/HSET unchanged.

## Motivation

go-redis sends `HELLO … AUTH <password>` through `ProcessHook` on connect. We were joining `cmd.Args()` with no masking, so the password landed in traces.

Fixes #1091

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [ ] Code formatted: `make format`
- [ ] Linters pass: `make lint`
- [ ] Tests pass: `make test`
- [x] Tests added for new functionality
- [x] Tests follow [testing guidelines](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/testing.md)
- [ ] Documentation updated (if applicable)
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation))
- [ ] This PR has content that I did not fully write myself.
 - [x] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md).
- [ ] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

PR #1201, opened 2026-08-22
fix(instrumentation/redis): redact command argument values from db.query.text by default

### LINKED ISSUE #1200: bug(instrumentation/redis): db.query.text captures full command argument values by default, leaking sensitive data
### Describe the bug

The Redis (`go-redis/v9`) instrumentation sets `db.query.text` on every client span to the **full, unredacted command, including argument values** — by default, with no opt-in and no way to disable it. This contradicts the OpenTelemetry Redis semantic conventions, which say query text "SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data" (https://opentelemetry.io/docs/specs/semconv/db/redis/), and it goes further than the general DB semconv guidance that only *parameterized* query text is safe to collect by default — parameter *values* require explicit opt-in (https://opentelemetry.io/docs/specs/semconv/db/sql/).

Redis commands don't separate "query" from "parameters" the way SQL does: the arguments to `SET`, `HSET`, `LPUSH`, etc. *are* the data being written. Capturing them verbatim means every value an application stores in Redis — session tokens, API keys, emails, free-text fields — is copied into every client span and shipped to whatever trace backend is configured, unconditionally.

This is a different, broader problem than #1091 (closed): #1091 fixed the special case of `AUTH`/`HELLO` connection-handshake credentials being leaked. The redaction added for that fix (`redisV9CredentialRedactRange` in `hook.go`) only recognizes the `auth` and `hello` command names — every other command's arguments, including the data values of ordinary write commands, are still serialized in full with no redaction and no way to turn it off.

### Environment

- OS: any (not OS-specific)
- Go Version: 1.25 (repo's `go.mod` minimum)
- Version: main @ `234ca1f`

### Evidence

`instrumentation/github.com/redis/go-redis/v9/semconv/client.go:27-33`:
```go
attrs := []attribute.KeyValue{
    semconv.DBSystemNameRedis,
    semconv.DBOperationName(req.FullName),
    semconv.ServerAddress(host),
    semconv.NetworkTransportTCP,
    semconv.DBQueryText(req.Statement),   // <- always included, unconditionally
}
```

`req.Statement` comes from `getRedisV9Statement` in `instrumentation/github.com/redis/go-redis/v9/hook.go:151-173`, which appends every command argument via `redisV9AppendArg` (`hook.go:219-267`) except the narrow `AUTH`/`HELLO AUTH` range computed by `redisV9CredentialRedactRange` (`hook.go:178-206`) — a switch on only `"auth"` and `"hello"`. There is no config surface (env var, `otelc.yaml` option, code constant) anywhere in the package that gates this — confirmed by grepping the package for `redact`/`sensitive`/`capture`/`config`.

### To Reproduce

```go
rdb := redis.NewClient(&redis.Options{Addr: "localhost:6379"})
rdb.AddHook(newOtelRedisHook(rdb.Options().Addr)) // done automatically by afterNewRedisClientV9

rdb.Set(ctx, "session:42", "s3cret-session-token", 0)
rdb.HSet(ctx, "user:42", "email", "alice@example.com")
```

Inspect the resulting client spans:

```
db.query.text = "set session:42 s3cret-session-token"
db.query.text = "hset user:42 email alice@example.com"
```

Both the session token and the user's email are now stored verbatim in the trace backend, for every `SET`/`HSET`/etc. call an application makes, with no way to prevent it short of not using this instrumentation.

### Expected behavior

`db.query.text` should not contain unredacted argument values by default. It should either be omitted/limited to the command name (and optionally the key) by default with a documented opt-in for full capture, or have values sanitized by default — consistent with the OTel Redis and DB semantic-convention guidance on when raw query text/parameters are safe to collect.

### Root cause

`getRedisV9Statement`/`RedisClientRequestTraceAttrs` treat "redact known credential commands" as equivalent to "safe to capture by default." They are not: the OTel Redis semantic conventions explicitly call for query text to be opt-in / sanitized by default, not blocklist-redacted for two command names. There is no default-off or default-sanitized behavior, and no configuration knob for users who need one.

### Impact

- **Security/privacy**: sensitive data (credentials, tokens, PII, arbitrary application values) is exported to trace backends by default, for every user of this instrumentation, with no indication in the README (there isn't one for this package) that this happens.
- Affects every command whose arguments carry data: `SET`, `SETEX`, `HSET`, `LPUSH`, `RPUSH`, `SADD`, `ZADD`, `GETSET`, `MSET`, pipelines containing any of these, etc. — not an edge case, this is normal Redis usage.
- Violates the semantic convention this instrumentation otherwise claims to implement (`db.query.text` per `go.opentelemetry.io/otel/semconv/v1.37.0`).

### Affected components

- `instrumentation/github.com/redis/go-redis/v9/hook.go` (`getRedisV9Statement`, `redisV9CredentialRedactRange`, `ProcessHook`, `ProcessPipelineHook`)
- `instrumentation/github.com/redis/go-redis/v9/semconv/client.go` (`RedisClientRequestTraceAttrs`)

### Acceptance criteria

- [ ] `db.query.text` is not populated with full argument values by default — either omitted by default and gated behind explicit opt-in (e.g. an env var consistent with this repo's `OTEL_GO_ENABLED_INSTRUMENTATIONS` conventions), or values are sanitized/redacted by default while the command name and key remain visible (matching how other mature Redis instrumentations approach this).
- [ ] The existing `AUTH`/`HELLO` credential redaction keeps working under whatever default is chosen (no regression of #1091).
- [ ] Behavior is documented (this package currently has no README — add one, or extend the closest existing docs) so users know what is captured by default and how to opt in/out.
- [ ] Unit tests cover: default behavior emits no sensitive argument values for a representative write command (`SET`/`HSET`), and the opt-in/sanitized path (whichever is chosen) is exercised explicitly.

Do not collect argument values in `db.query.text` by default. Keep the command name (`db.operation.name` already carries this) and, if desired, the key (first argument) — Redis commands are `<CMD> <key> [args...]`, and the key alone is far lower-risk than the values that follow it while still being useful for debugging ("what command touched what key"). Gate full-argument capture behind an explicit opt-in, following the pattern OTel semconv recommends for `db.query.parameter.<key>` (opt-in) versus the query "template" (default-on). This is a bounded, well-scoped change: it touches `getRedisV9Statement`/`redisV9AppendArg` and the enabler/config plumbing already established elsewhere in the repo (`runtime.Instrumented`), plus tests asserting the redacted default and the opt-in path.

bhuvan-somisetty:
## Description

`db.query.text` on Redis client spans captured every command argument verbatim by default, including data values (`SET`'s value, `HSET`'s field/value pairs, `LPUSH`'s elements, etc.). Only `AUTH`/`HELLO` credentials were redacted (#1092/#1101); everything else - session tokens, PII, arbitrary application data - was shipped to the trace backend on every write, with no opt-out.

```go
rdb.Set(ctx, "session:42", "s3cret-session-token", 0)
// db.query.text = "set session:42 s3cret-session-token"
```

This fix redacts command argument values by default, keeping only the command name and (conventionally) the key, matching the OpenTelemetry [Redis semantic conventions](https://opentelemetry.io/docs/specs/semconv/db/redis/), which call for query text to be sanitized or opt-in rather than collected by default.

### Root cause

`getRedisV9Statement` redacted only the index range returned by `redisV9CredentialRedactRange`, which matched exclusively on the `auth`/`hello` command names. Every other command fell through to "not redacted."

### Fix

Redis commands don't separate a query template from bindable parameters the way SQL does, so there's no general way to know which argument is safe to show. `redisV9CredentialRedactRange` is renamed to `redisV9RedactRange` (it's no longer only about credentials) and gets a `default` case: for any command other than `auth`/`hello`, redact every argument after `args[1]` (`args[0]` is the command name, `args[1]` is conventionally the key). The existing `AUTH`/`HELLO` rule is unchanged.

This is a conservative, safe-by-default choice with two accepted trade-offs (documented in the code): it over-redacts a few non-sensitive trailing arguments (e.g. `EXPIRE`'s TTL), and for multi-key commands like `MSET` it treats keys past the first as if they were values. Both err toward never leaking a value.

## Motivation

Why is this change needed? What problem does it solve?

Sensitive Redis command values (credentials, tokens, PII, arbitrary application data) were exported to trace backends by default for every user of this instrumentation, for ordinary Redis usage (`SET`, `HSET`, `LPUSH`, ...), not an edge case. This violates the semantic convention this instrumentation otherwise implements.

Fixes #1200.

---

## Testing

- Added `TestGetRedisV9Statement_RedactsCommandValues` (`SET`, `HSET`, `LPUSH`, `GET`, `PING`) and `TestProcessHook_RedactsCommandValues`, the latter exercising the full `ProcessHook` path and asserting the exact reproduction case from #1200 (a session token) never reaches the span.
- Updated existing table tests whose expectations assumed the old unredacted behavior (`SET`, `HSET`, `DEL`, `EXPIRE`, bool args) to the new redacted output; `AUTH`/`HELLO` cases are untouched and still pass.
- Updated `TestRedisClient` (`test/integration/redis_client_test.go`), which asserted the literal leaked value (`"set testkey testvalue"`) - now asserts `"set testkey ?"`.
- `go test ./...` in `instrumentation/github.com/redis/go-redis/v9` and its `semconv` subpackage: pass.
- `go test -tags integration -run TestRedisClient ./integration/...`: pass, building a real `otelc`-instrumented binary against miniredis.
- Ran unit tests across every other `instrumentation/*` module (mirroring `make test-unit/instrumentation`) to confirm the regenerated bundle doesn't affect unrelated packages: all pass.
- `golangci-lint run --config .tools/golangci.yml ./...` inside the redis module: no new findings beyond this package's pre-existing lint debt (unrelated `interface{}`-vs-`any` and `testifylint` style issues that predate this change).
- `go build` / `go vet` for the module and for the `otelc` tool binary: clean.
- `tool/data/otelc-bundle.tgz` regenerated and verified with `git diff --exit-code`, matching CI's `verify-bundle` check (`make` wasn't available in my environment, so I ran the underlying `cp`/`go mod tidy`/bundle-tool steps directly).

## Impact

Default-behavior change: `db.query.text` for write commands (and any command with more than a key argument) no longer contains argument values, only the command name and key. This is a deliberate, spec-aligned fix for a data-leak default. Nothing structural changes (span names, `db.operation.name`, other attributes).

## Checklist

`make` was not available in my environment, so I ran the underlying `go build`/`go vet`/`go test`/`golangci-lint` commands directly, scoped to the changed files/module, rather than the full repo-wide `make lint`/`make test` targets - see Testing above for exactly what ran. `golangci-lint fmt --diff` on every changed file reports no diff, so formatting is checked; I'm leaving lint/test unchecked since I can't claim the full repo-wide targets passed, only their scoped equivalents.

- [ ] Tests pass: `make test`


instrumentation/github.com/redis/go-redis/v9/hook.go
```diff
@@ -172,10 +176,26 @@ func getRedisV9Statement(cmd redis.Cmder) string {
...
+// SET's value, HSET's field/value pairs, LPUSH's elements, and so on are the
+// application data being written, indistinguishable from the command shape
+// without a per-command grammar. args[0] is always the command name and
+// args[1] is conventionally the key, so this keeps "what command touched
+// what key" visible while never emitting a value. It also over-redacts
+// arguments that happen not to be sensitive (e.g. EXPIRE's TTL) and, for
+// multi-key commands like MSET, redacts keys past the first as if they were
+// values - both accepted trade-offs for never leaking one by default. See
+// https://opentelemetry.io/docs/specs/semconv/db/redis/, which calls for
+// query text to be sanitized or opt-in rather than captured by default.
```
darccio, line 197: This is verbose. Please try to summarize it a bit.

instrumentation/github.com/redis/go-redis/v9/hook.go
```diff
@@ -172,10 +176,26 @@ func getRedisV9Statement(cmd redis.Cmder) string {
...
+// credentials outright, and HELLO's AUTH username/password sub-arguments are
+// too (SETNAME and the protocol version are not).
+//
+// Every other command redacts everything after args[1]. Unlike SQL, a Redis
+// command's arguments are not a query template plus bindable parameters:
+// SET's value, HSET's field/value pairs, LPUSH's elements, and so on are the
+// application data being written, indistinguishable from the command shape
+// without a per-command grammar. args[0] is always the command name and
+// args[1] is conventionally the key, so this keeps "what command touched
+// what key" visible while never emitting a value. It also over-redacts
```
darccio, line 192: The boundary "keep the key, redact the values" matches the proposal in issue #1200, and I support it. No maintainer has agreed to the boundary yet, and the ecosystem is more conservative:

- The OpenTelemetry Redis spec shows `SET ? ?` as the sanitized example. The example redacts the key too.
- Elastic APM Go records only the command name. It captures no key.
- dd-trace-go delegates redaction to the agent. The agent option "remove_all_args" replaces every argument, including the key, with "?".

Two concerns remain:

1. A key can hold personal data by convention, for example `user:42:alice@example.com`.
2. Every key produces a new attribute value, so raw keys increase attribute cardinality in analytics backends.

Please agree the boundary with the maintainers in issue #1200 before the merge. Also document the boundary for the user. The acceptance criteria in issue #1200 ask for user-facing documentation, and this package has no README. A function comment does not reach the user.

instrumentation/github.com/redis/go-redis/v9/hook.go
```diff
@@ -185,6 +205,10 @@ func redisV9CredentialRedactRange(name string, args []interface{}) (start, end i
 		if i := redisV9HelloAuthIndex(args); i >= 0 {
 			return i + 1, min(i+1+redisHelloAuthArgN, len(args))
 		}
+	default:
+		if len(args) > redisValueArgsStart {
+			return redisValueArgsStart, len(args)
+		}
```
darccio, line 211: go-redis builds `EVAL` and `EVAL_RO` as `[eval, <script>, <numkeys>, <keys...>, <args...>]`. The Lua script sits at `args[1]`, not a key. go-redis itself confirms the layout: it sets the first key position of `eval` to 3, so the client treats `args[1]` as the script.

The `default` case redacts arguments from `args[2]`, so the script reaches `db.query.text` in full. Two consequences:

1. A Lua script can hold literal secrets. A team that writes an access token into a script ships that token to the trace backend on every call.
2. `getRedisV9Statement` puts no limit on the length of `db.query.text`. After this change, every other command produces a bounded statement. `EVAL` still produces a statement of unbounded length, because a script has no length limit.

Redact the script like a credential. Treat every argument after `args[0]` as sensitive:

```suggestion
	case "eval", "eval_ro":
		return 1, len(args)
	default:
		if len(args) > redisValueArgsStart {
			return redisValueArgsStart, len(args)
		}
```

If you prefer a constant, add `redisEvalCmd` to the const block and use it in the case label.

`EVALSHA` needs no change: it carries a SHA-1 hash at `args[1]`, and the hash is not data. `SLAVEOF` and `REPLICAOF` still show an internal host name at `args[1]`. I accept that disclosure. The host name carries no user data.

darccio (changes requested):

PR #1204, opened 2026-08-22
fix(instrumentation/redis): redact db.query.text argument values by default

### LINKED ISSUE #1200: bug(instrumentation/redis): db.query.text captures full command argument values by default, leaking sensitive data

shivv23:
## Description

`db.query.text` on Redis client spans currently captures full, unredacted command argument values by default — session tokens, API keys, emails, and arbitrary application data are serialized into every span and shipped to trace backends unconditionally.

This changes `getRedisV9Statement` to redact all arguments at index ≥ 2 (everything after the command name and key) by default, replacing them with `?`. Users who need full argument capture can opt in via `OTEL_REDIS_CAPTURE_FULL_STATEMENT=true`.

Existing `AUTH`/`HELLO` credential redaction (from #1091) continues to work under both modes.

## Motivation

The OTel Redis semantic conventions state that `db.query.text` "SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data." Redis commands don't separate "query" from "parameters" the way SQL does — the arguments to `SET`, `HSET`, `LPUSH`, etc. *are* the data being written. Capturing them verbatim leaks sensitive data into traces by default.

Fixes #1200

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### DIFF
--- a/instrumentation/github.com/redis/go-redis/v9/hook.go
+++ b/instrumentation/github.com/redis/go-redis/v9/hook.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"net"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -148,16 +149,25 @@ func (o *otelRedisHook) DialHook(next redis.DialHook) redis.DialHook {
 	}
 }
 
+// captureFullStatement reports whether db.query.text should include
+// full command argument values. When false (the default), only the
+// command name and key are captured; all subsequent arguments are
+// replaced with "?" to avoid leaking sensitive data into traces.
+func captureFullStatement() bool {
+	return strings.EqualFold(os.Getenv("OTEL_REDIS_CAPTURE_FULL_STATEMENT"), "true")
+}
+
 func getRedisV9Statement(cmd redis.Cmder) string {
 	args := cmd.Args()
-	redactStart, redactEnd := redisV9CredentialRedactRange(cmd.Name(), args)
+	credStart, credEnd := redisV9CredentialRedactRange(cmd.Name(), args)
+	valStart, valEnd := redisV9ValueRedactRange(args)
 
 	b := make([]byte, 0, 64)
 	for i, arg := range args {
 		if i > 0 {
 			b = append(b, ' ')
 		}
-		if i >= redactStart && i < redactEnd {
+		if (i >= credStart && i < credEnd) || (i >= valStart && i < valEnd) {
 			b = append(b, redisQueryTextRedact...)
 			continue
 		}
@@ -187,6 +197,17 @@ func redisV9CredentialRedactRange(name string, args []interface{}) (start, end i
 		}
 	}
 	return 0, 0
+}
+
+// redisV9ValueRedactRange returns a half-open index range [2, len(args)) of
+// arguments that carry data values and should be redacted by default. Returns
+// (0, 0) when OTEL_REDIS_CAPTURE_FULL_STATEMENT is "true" or the command
+// has no value arguments (index <= 1).
+func redisV9ValueRedactRange(args []interface{}) (start, end int) {
+	if len(args) <= 2 || captureFullStatement() {
+		return 0, 0
+	}
+	return 2, len(args)
 }
 
 func redisV9HelloAuthIndex(args []interface{}) int {
--- a/instrumentation/github.com/redis/go-redis/v9/hook_test.go
+++ b/instrumentation/github.com/redis/go-redis/v9/hook_test.go
@@ -36,24 +36,24 @@ func TestGetRedisV9Statement(t *testing.T) {
 		expected string
 	}{
 		{
-			name:     "GET command",
+			name:     "GET command with only key",
 			cmd:      redis.NewCmd(context.Background(), "get", "mykey"),
 			expected: "get mykey",
 		},
 		{
-			name:     "SET command with value",
+			name:     "SET command redacts value by default",
 			cmd:      redis.NewCmd(context.Background(), "set", "mykey", "myvalue"),
-			expected: "set mykey myvalue",
+			expected: "set mykey ?",
 		},
 		{
-			name:     "HSET command",
+			name:     "HSET command redacts field values by default",
 			cmd:      redis.NewCmd(context.Background(), "hset", "myhash", "field1", "value1"),
-			expected: "hset myhash field1 value1",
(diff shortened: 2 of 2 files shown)


Labreo:
Note to maintainers:This is a duplicate PR for #1200 and there already exists a PR #1201 that has been opened for the same issue by the original submitter.And as such should be ideally closed.

shivv23:
Closed in favour of #1201

PR #1208, opened 2026-08-22
fix(setup): quote the toolexec executable path so spaces survive go's splitting

### LINKED ISSUE #1207: bug(tool/internal/setup): -toolexec flag breaks when otelc's own executable path contains a space
### Describe the bug

`buildWithToolexec` builds the `-toolexec` flag by just string-concatenating otelc's own executable path with no quoting:

```go
// tool/internal/setup/setup.go:541-545
execPath, err := os.Executable()
insert := "-toolexec=" + execPath + " toolexec"
```

If `execPath` has a space in it, `go`'s own `-toolexec` parser splits on whitespace and only picks up the first chunk as the executable , so it tries to run something like `C:\Users\...\my` instead of `C:\Users\...\my tool\otelc.exe`. Build just breaks. Any otelc install path with a space in it hits this  `Program Files` on Windows being the obvious one, but any user dir with a space works too.

### Environment

 - OS: Windows 11
 - Go Version: not version-specific, this is about how `cmd/go` parses `-toolexec`, happens on any recent Go
 - Version: v1.0.1 

### To Reproduce

1. Install/run otelc from a path with a space in it, e.g. `C:\Program Files\otelc\otelc.exe`, or just `.../my tool/otelc.exe`
2. Run `otelc go build .` on any module
3. otelc constructs `-toolexec=<path with space> toolexec` and hands it to `go build`
4. `go` truncates at the space and fails

Minimal repro without even needing otelc installed anywhere weird,  just simulating the same flag construction:

```go
func TestToolexecBreaksOnSpacePath(t *testing.T) {
	tempDir := t.TempDir()
	os.WriteFile(filepath.Join(tempDir, "go.mod"), []byte("module testmod\n\ngo 1.22\n"), 0644)
	os.WriteFile(filepath.Join(tempDir, "main.go"), []byte("package main\nfunc main() {}\n"), 0644)

	execPath := filepath.Join(tempDir, "my tool", "otelc.exe")
	insert := "-toolexec=" + execPath + " toolexec"

	cmd := exec.CommandContext(context.Background(), "go", "build", "-n", insert, ".")
	cmd.Dir = tempDir
	out, err := cmd.CombinedOutput()
	t.Logf("Output:\n%s\nError: %v", string(out), err)
}
```

Output:

```
go: error obtaining buildID for go tool compile: exec: "C:\Users\...\my": executable file not found in %PATH%
```

### Expected behavior

`go build` should run fine regardless of whether otelc's own path has spaces in it,  the `-toolexec` value needs to be quoted so `go` doesn't split it on whitespace.

### Additional context

Fix is to quote `execPath` before it goes into the `-toolexec=` value.

Worth being careful about how it gets quoted though, `strconv.Quote` doesn't work here. `cmd/internal/quoted.Split` does no unescaping inside quotes, it just strips the matching outer pair and takes whatever's inside literally. So a Windows path comes back out with doubled backslashes (`C:\\Program Files\\otelc.exe`) and it's still broken.

The repo already has the right thing for this, `util.QuoteGoflagsToken` in `tool/util/go.go`. It does `quoted.Join` semantics, wraps in `'` or `"` depending on which one isn't already in the token, errors if both are.

Also there's two places building this flag, not one:
- `tool/internal/setup/setup.go:541-545` (`buildWithToolexec`), the CLI case
- `tool/internal/instrument/toolexec.go:472` (`EnableNestedToolexec`), the nested `GOFLAGS` case

The second one already quotes the whole flag for `GOFLAGS`, but `go` splits twice, once on `GOFLAGS` and then again on the `-toolexec` value after stripping the prefix. So the inner `execPath` needs quoting there too, otherwise nested `go list -export` calls break the same way.

One edge case that falls out of this: if the path has a single quote in it, the inner quoting has to use double quotes, and then the nested token has both kinds in it. `GOFLAGS` can't encode that at all since `quoted.Split` has no escape syntax. Erroring out and saying so seems better than emitting a flag `go` will quietly misparse.

yzxcj797:
Fixes #1207.

## Root cause

`buildWithToolexec` builds the flag by plain concatenation:

```go
insert := "-toolexec=" + execPath + " toolexec"
```

`cmd/go` splits the `-toolexec` value on whitespace (`internal/quoted`), so an install path with a space — `C:\Program Files\...` on Windows being the obvious case, any user dir with a space works too — makes `go` run only the first chunk, and the build fails with `executable file not found` for the truncated path (the issue's exact repro).

## Fix

Wrap the path in **plain double quotes** — not `strconv.Quote`: go's splitter does not process backslash escapes, so `strconv.Quote` would corrupt Windows paths by doubling every backslash (verified against the splitter semantics; a first draft using `strconv.Quote` produced doubled-backslash tokens). Like `go` itself, values containing a literal double quote cannot be represented.

## Test

`TestToolexecFlagQuotesPathWithSpaces` splits the constructed flag value the way cmd/go's `internal/quoted` does (double-quoted runs are one token) and asserts the executable token is the **whole** path, the ` toolexec` argument survives as its own token, and the unquoted construction demonstrably breaks at the space (pinning the bug).

Package passes with the fix; differential with `setup.go` reverted fails on the broken-split assertions. `gofmt`/`go vet` clean.

### DIFF
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -542,7 +542,10 @@ func buildWithToolexec(ctx context.Context, cmd *cli.Command, vendored bool) err
 	if err != nil {
 		return ex.Wrapf(err, "failed to get executable path")
 	}
-	insert := "-toolexec=" + execPath + " toolexec"
+	// cmd/go splits the -toolexec value on whitespace (internal/quoted), so
+	// a path containing spaces — "Program Files" being the obvious one —
+	// must be quoted or only the first chunk reaches the tool (#1207).
+	insert := "-toolexec=" + quoteForGo(execPath) + " toolexec"
 	const additionalCount = 2
 	newArgs := make([]string, 0, len(args)+additionalCount) // Avoid in-place modification
 	// Add "go build"
@@ -663,4 +666,14 @@ func runGoBuild(ctx context.Context, cmd *cli.Command) error {
 	}
 	logger.InfoContext(ctx, "Instrumentation completed successfully")
 	return nil
+}
+
+// quoteForGo quotes a value for cmd/go's internal/quoted splitting: the
+// -toolexec flag value is split on whitespace, so paths with spaces must be
+// wrapped in double quotes. cmd/go's splitter does not process backslash
+// escapes (unlike strconv.Quote), so a plain quote wrap is the correct form;
+// like go itself, values containing a literal double quote cannot be
+// represented (#1207).
+func quoteForGo(v string) string {
+	return `"` + v + `"`
 }
--- a/tool/internal/setup/toolexec_quote_test.go
+++ b/tool/internal/setup/toolexec_quote_test.go
@@ -1,0 +1,64 @@
+package setup
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// Regression (#1207): cmd/go splits the -toolexec value on whitespace
+// (internal/quoted), so an otelc executable installed under a path with
+// spaces — "Program Files" being the obvious one — must be quoted or `go`
+// runs only the first chunk. The constructed flag value must therefore
+// contain the whole executable path as a single whitespace-free token after
+// go's unquoting, and the " toolexec" argument must survive as its own token.
+func TestToolexecFlagQuotesPathWithSpaces(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Labreo:
This is a duplicate PR as PR #1211 has been opened by the @darkraider01 who created the original issue #1207 .

PR #1209, opened 2026-08-22
fix(instrumentation/database/sql): end spans when disabled mid-operation

### LINKED ISSUE #1206: bug(instrumentation/database/sql): spans leak when instrumentation is disabled mid-operation
## Describe the bug

The database/sql client instrumentation has the same span-lifecycle issue #1094 fixed for net/http, in a subsystem #1094 did not touch.

A span is created in the before hook and ended in the after hook, but the after path re-checks `Enable()` and returns before ending the span. If instrumentation is enabled when the before hook runs, so the span is created and stored, and then disabled before the after hook runs, the span is never ended and leaks.

The re-check sits in front of `span.End()` in two layers:

- `instrumentEnd` (`instrumentation/database/sql/client.go:633`) starts with `if !clientEnabler.Enable() { return }`, ahead of the `defer span.End()` on the stored span.
- every after hook that ends a span (`afterPingContextInstrumentation`, `afterExecContextInstrumentation`, `afterQueryContextInstrumentation`, `afterTxInstrumentation`, `afterTxCommitInstrumentation`, `afterTxRollbackInstrumentation`, and the rest) starts with the same `if !clientEnabler.Enable() { return }` before delegating to `instrumentEnd`.

Because `Instrumented()` reads `OTEL_GO_DISABLED_INSTRUMENTATIONS` fresh on every call, a mid-operation disable skips `span.End()` at both layers.

## Reproduce

1. Enable the db instrumentation and run a before hook (e.g. `beforeExecContextInstrumentation`) so a span is created and stored.
2. Set `OTEL_GO_DISABLED_INSTRUMENTATIONS=databasesql`.
3. Run the matching after hook.

The stored span is never ended. This mirrors the net/http case #1094 describes, where `AfterServeHTTP` re-checked enablement before retrieving the span so it never got ended.

## Expected behavior

The after path should end a span created by the before hook regardless of the current enablement state, the same way net/http does after #1094. `instrumentEnd` already guards on the stored data (`ictx.GetData() == nil` and `span == nil`), and the field-injecting after hooks (`afterTxInstrumentation`, `afterConnInstrumentation`) already guard on their stored `GetData()` payload, so the `Enable()` re-check is redundant with guards that are already present.

## Proposed fix

Remove the redundant `if !clientEnabler.Enable() { return }` from `instrumentEnd` and the after hooks, keeping it in the before hooks and `instrumentStart` where span creation is correctly gated. This matches #1094 for net/http.

## Note on testing

Unlike net/http, the database/sql instrumentation does not build standalone (it reads tool-injected fields such as `db.Endpoint`), so it has no hook unit tests today; coverage runs through the e2e db app harness. Happy to send the fix.

vignesh2027:
## Description

The database/sql client instrumentation had the same span-lifecycle issue #1094 fixed for net/http, in a subsystem #1094 did not touch.

A span is created in the before hook and ended in the after hook, but the after path re-checked `Enable()` and returned before ending the span. Because `Instrumented()` reads `OTEL_GO_DISABLED_INSTRUMENTATIONS` fresh on every call, a span created while instrumentation was enabled leaked if instrumentation was disabled before the operation completed.

The redundant re-check sat in front of `span.End()` in two layers:

- `instrumentEnd` started with `if !clientEnabler.Enable() { return }`, ahead of the `defer span.End()` on the stored span.
- every after hook that ends a span started with the same check before delegating to `instrumentEnd`.

This change removes the `Enable()` re-check from the after hooks and `instrumentEnd`, keeping it in the before hooks and `instrumentStart` where span creation is correctly gated. `instrumentEnd` already guards on the stored data (`ictx.GetData() == nil` and the span type assertion), and the field-injecting after hooks (`afterTxInstrumentation`, `afterConnInstrumentation`) already guard on their stored payload, so the re-check was redundant with guards that are already present. A disabled before hook stores nothing, so the after path is a safe no-op; a span that was created is always ended.

This mirrors the net/http fix in #1094, and the same lifecycle principle was reaffirmed by a maintainer on #1137.

## Motivation

A span that is never ended is never exported and holds its resources for the life of the process. Disabling an instrumentation at runtime should stop new spans, not strand in-flight ones.

## Note on testing

The database/sql instrumentation does not build standalone (it reads tool-injected fields such as `db.Endpoint`), so it has no hook unit tests. This change is validated by the existing database/sql e2e apps in CI. The removed re-check is redundant with nil-guards that remain in place, so no behavior changes when instrumentation is enabled throughout an operation.

Fixes #1206

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


Labreo (commented):
@vignesh2027 You will need to resolve the merge conflicts.

PR #1211, opened 2026-08-22
fix(tool): quote otelc path in -toolexec so spaced paths work

### LINKED ISSUE #1207: bug(tool/internal/setup): -toolexec flag breaks when otelc's own executable path contains a space

darkraider01:
## Description

`buildWithToolexec` built the `-toolexec` flag by concatenating otelc's own path in unquoted. `cmd/go` parses the `-toolexec` value with `cmd/internal/quoted.Split`, so an unquoted executable path containing a space was read as a path plus extra arguments. A build run from `C:\Program Files\otelc\otelc.exe` failed with `exec: "C:\Program": executable file not found in %PATH%`.

Added `util.BuildToolexecFlag`, which quotes `execPath` with the existing `QuoteGoflagsToken`, and used it at both construction sites. The nested `GOFLAGS` entry now quotes twice on purpose: the inner quote survives the `-toolexec` value split, the outer one survives the `GOFLAGS` split.

A path containing a single quote leaves both quote kinds in the nested token, which `GOFLAGS` cannot encode since `quoted.Split` has no escape syntax. That case now fails with an error naming the path as the cause rather than silently producing a flag the go command misparses.

## Motivation

Any otelc install path with a space in it hits this, `Program Files` on Windows being the obvious one, but any user directory with a space works too. Build just breaks with a confusing error pointing at a truncated path.

Fixes #1207

---

## Checklist

- [x] Tests pass: `make test`
- [x ] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:

---


Labreo:
Note for Maintainers.This the second PR for the same issue #1207.There is another PR for the same issue at #1208.But this one was opened by the issue creator @darkraider01 which I would personally favor and according to @darccio holds the right to open a PR for the issue that is created.

tool/internal/instrument/toolexec.go
```diff
@@ -548,9 +547,24 @@ func EnableNestedToolexec() error {
...
+		// Reachable whenever the finished flag holds both quote characters:
+		// the inner quoting supplies one kind to keep a spaced path together,
+		// and a quote already in execPath supplies the other. GOFLAGS is split
+		// by cmd/internal/quoted.Split, which has no escape syntax, so such a
+		// token simply cannot be represented there.
+		return ex.Wrapf(err, "otelc's own path %q cannot be passed to nested "+
+			"go commands through GOFLAGS, because quoting it there would need "+
+			"both quote characters and GOFLAGS has no way to escape either; "+
+			"reinstall otelc under a path without quote characters", execPath)
 	}
```
Labreo: `EnableNestedToolexec()` calls `os.Executable()` directly, so its error return branches aren't reached in unit tests. 
That is what is causing the codecov patch error to persists to fix this I would advise to split out a small helper.It could look smtg like :
```go
func enableNestedToolexec(execPath string) error { ... }

**Review by Labreo — COMMENTED** [review]
Just left a suggestion to fix the codecov patch error.

**Comment by darkraider01** [comment]
@kakkoyun hey!! sorry for the ping, #1210 has already been re-opened here, which is marked as salvage candidates.

File: tool/internal/setup/setup.go [hunk 2]
```diff
@@ -530,20 +530,28 @@ func extractBuildFlags(args []string) []string {
...
-	logger := util.LoggerFromContext(ctx)
+// toolexecInsertArg builds the -toolexec=... argument for execPath, wrapping
+// any error with the execPath that caused it. Split out from
+// buildWithToolexec so the error path (reachable only when execPath contains
+// both quote characters) is testable without depending on os.Executable.
+func toolexecInsertArg(execPath string) (string, error) {
+	insert, err := util.BuildToolexecFlag(execPath)
+	if err != nil {
+		return "", ex.Wrapf(err, "building -toolexec flag for %q", execPath)
+	}
```
darccio, line 560: The two -toolexec construction sites now follow different error-wrapping
conventions for the same failure. nestedToolexecGoflagsToken propagates
BuildToolexecFlag's error unwrapped, on the grounds that it already names
execPath (the QuoteGoflagsToken message embeds the token with %q, and the
wrap adds the "quoting otelc executable path for -toolexec" context), per
#1231. toolexecInsertArg wraps it a third time with %q of execPath, so the
final message names the path twice. Dropping the redundant wrap here would
match the nested path's convention and the #1231 cleanup.

```suggestion
	insert, err := util.BuildToolexecFlag(execPath)
	if err != nil {
		// BuildToolexecFlag's own error already names execPath, so there
		// is nothing to add here, matching nestedToolexecGoflagsToken
		// and the #1231 convention.
		return "", err
	}
```

darccio (changes requested):

tool/internal/setup/setup.go
darkraider01: also added a unit test in setup_test.go, so the test description and assertions match the unwrapped error behavior

darccio (approved):

darkraider01:
sorry auto merge got disabled, it showed there's resolve conflicts in tool/util/go.go

PR #1213, opened 2026-08-22
fix(instrumentation): detect SSE content types case-insensitively

### LINKED ISSUE #1212: bug(instrumentation/openai): detect SSE content types case-insensitively
### Describe the bug

The OpenAI instrumentation identifies streaming responses with:

```go
  strings.HasPrefix(resp.Header.Get("Content-Type"), "text/event-stream")
```

This comparison is case-sensitive, although HTTP media types are case-insensitive. A valid SSE response such as `Content-Type: Text/Event-Stream` is therefore treated as a non-streaming response.

This affects the OpenAI Go SDK v1, v2, and v3 instrumentation.

### Environment

  - OS: Windows 11
  - Go Version: go version go1.25.5 windows/amd64
  - Version: current main

### To Reproduce

  1. Build an application using an instrumented OpenAI Go SDK client.
  2. Send a chat-completions request to an OpenAI-compatible endpoint that returns an SSE response.
  3. Return an SSE response with this header:
```http
     Content-Type: Text/Event-Stream
```
  4. Read the response as a stream.

The middleware does not recognize the response as streaming because the header value does not start with the lowercase string `text/event-stream`.

### Expected behavior

Any valid `text/event-stream` media type should be recognized as an SSE response regardless of casing, including:

  ```http
  Content-Type: text/event-stream
  Content-Type: text/event-stream; charset=utf-8
  Content-Type: Text/Event-Stream
  ```

The response body should be wrapped by the streaming reader so the span remains open for stream consumption and can record streaming response attributes.

### Additional context

The affected check appears in:

  - `instrumentation/github.com/openai/openai-go/middleware.go`
  - `instrumentation/github.com/openai/openai-go/v2/middleware.go`
  - `instrumentation/github.com/openai/openai-go/v3/middleware.go`

A small fix would be to parse `Content-Type` with Go’s standard-library `mime.ParseMediaType` and compare the parsed media type case-insensitively. Existing tests cover lowercase `text/event-stream`, but not a mixed-case media type.

Parse the response Content-Type with Go’s standard-library mime.ParseMediaType instead of checking the raw header with a case-sensitive prefix. Treat the parsed media type as streaming when it equals text/event-stream, while preserving existing handling for parameters such as charset=utf-8.

This is a small, shared change across the OpenAI Go SDK v1, v2, and v3 middleware paths. Add regression coverage for lowercase, parameterized, and mixed-case SSE media types.

yzxcj797:
## Description

The OpenAI (v1/v2/v3) and Anthropic middlewares decided whether a response was an SSE stream with a case-sensitive prefix check:

```go
strings.HasPrefix(resp.Header.Get("Content-Type"), "text/event-stream")
```

HTTP media types are case-insensitive (RFC 9110 §8.3.1), and servers behind OpenAI-compatible gateways do send e.g. `Text/Event-Stream; charset=utf-8`. Such responses fell into the non-streaming path: `gen_ai.request.is_stream` was never set and the body was buffered instead of being finalized incrementally by the streaming reader.

## Fix

Parse the header with `mime.ParseMediaType` (which also strips parameters) and compare the resulting media type with `strings.EqualFold`. Applied consistently in:

- `instrumentation/github.com/openai/openai-go` (v1)
- `instrumentation/github.com/openai/openai-go/v2`
- `instrumentation/github.com/openai/openai-go/v3`
- `instrumentation/github.com/anthropics/anthropic-sdk-go`

## Testing

- New integration test in the v1 package sends `Content-Type: Text/Event-Stream; charset=utf-8` and asserts `gen_ai.request.is_stream=true` plus the streamed response attributes — fails on the old prefix check, passes with the fix.
- Table-driven `TestIsSSEContentType` unit tests in all four packages cover mixed case, parameters, `text/event-streamx`, empty, and malformed-parameter headers.
- `go test ./...` passes in all four module directories.

Fixes #1212

### DIFF
--- a/instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
+++ b/instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"mime"
 	"net/http"
 	"strings"
 	"time"
@@ -158,7 +159,7 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 		// still answers with SSE, end the span without response attributes
 		// rather than hold it open on a body we do not accumulate yet.
 		contentType := resp.Header.Get("Content-Type")
-		if strings.HasPrefix(contentType, "text/event-stream") {
+		if isSSEContentType(contentType) {
 			span.SetAttributes(semconv.GenAIRequestIsStream(true))
 			span.End()
 			return resp, nil
@@ -168,6 +169,18 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 
 		return resp, nil
 	}
+}
+
+// isSSEContentType reports whether a Content-Type header names the SSE media
+// type. HTTP media types are case-insensitive and may carry parameters, so the
+// raw prefix check used previously misclassified headers like
+// "Text/Event-Stream; charset=utf-8".
+func isSSEContentType(contentType string) bool {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(mediaType, "text/event-stream")
 }
 
 func handleNonStreamingResponse(resp *http.Response, span trace.Span) {
--- a/instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go
+++ b/instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go
@@ -653,3 +653,24 @@ func assertBoolAttribute(t *testing.T, attrs []attribute.KeyValue, key string, e
 	require.True(t, ok, "attribute %s not found", key)
 	assert.Equal(t, expected, val.AsBool(), "attribute %s", key)
 }
+
+func TestIsSSEContentType(t *testing.T) {
+	tests := []struct {
+		name        string
+		contentType string
+		want        bool
+	}{
+		{"lowercase", "text/event-stream", true},
+		{"mixed case", "Text/Event-Stream", true},
+		{"uppercase with parameter", "TEXT/EVENT-STREAM; charset=utf-8", true},
+		{"json", "application/json", false},
+		{"empty", "", false},
+		{"prefix only", "text/event-streamx", false},
+		{"malformed parameter", "text/event-stream; ::", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, isSSEContentType(tt.contentType))
+		})
+	}
+}
--- a/instrumentation/github.com/openai/openai-go/middleware.go
+++ b/instrumentation/github.com/openai/openai-go/middleware.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"mime"
 	"net/http"
 	"strconv"
 	"strings"
@@ -187,7 +188,7 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 		}
 
 		contentType := resp.Header.Get("Content-Type")
-		isStreaming := strings.HasPrefix(contentType, "text/event-stream")
+		isStreaming := isSSEContentType(contentType)
 
 		if isStreaming {
 			span.SetAttributes(semconv.GenAIRequestIsStream(true))
@@ -198,6 +199,18 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 
 		return resp, nil
 	}
+}
+
+// isSSEContentType reports whether a Content-Type header names the SSE media
+// type. HTTP media types are case-insensitive and may carry parameters, so the
+// raw prefix check used previously misclassified headers like
+// "Text/Event-Stream; charset=utf-8".
+func isSSEContentType(contentType string) bool {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(mediaType, "text/event-stream")
 }
 
 func handleNonStreamingResponse(
--- a/instrumentation/github.com/openai/openai-go/middleware_test.go
+++ b/instrumentation/github.com/openai/openai-go/middleware_test.go
@@ -128,3 +128,24 @@ func TestParseChatResponse_Valid(t *testing.T) {
 	assert.Equal(t, int64(20), resp.Usage.CompletionTokens)
 	assert.Equal(t, "stop", resp.Choices[0].FinishReason)
 }
+
+func TestIsSSEContentType(t *testing.T) {
+	tests := []struct {
+		name        string
+		contentType string
+		want        bool
+	}{
+		{"lowercase", "text/event-stream", true},
+		{"mixed case", "Text/Event-Stream", true},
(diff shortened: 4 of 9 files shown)


Labreo:
Just a suggestion for the future.Please follow the [pull request template](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/.github/PULL_REQUEST_TEMPLATE.md).It helps contributors and maintainers get a better idea for the pull request.And allows them to manage and merge your pull request faster.You need to sign the EasyCLA aswell as it is one of the required checks.

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -200,6 +201,18 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
...
+// raw prefix check used previously misclassified headers like
+// "Text/Event-Stream; charset=utf-8".
+func isSSEContentType(contentType string) bool {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(mediaType, "text/event-stream")
+}
+
```
Labreo, line 215: If mime.ParseMediaType runs into a malformed parameter (like text/event-stream; ::), returning false pushes the response into handleNonStreamingResponse. That causes it to stall reading up to the 4 MB limit instead of streaming, and gen_ai.request.is_stream gets lost. Falling back to the base type before ; on error keeps streaming working for those cases.

instrumentation/github.com/openai/openai-go/v3/middleware.go
```diff
@@ -200,6 +201,18 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
...
+// raw prefix check used previously misclassified headers like
+// "Text/Event-Stream; charset=utf-8".
+func isSSEContentType(contentType string) bool {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(mediaType, "text/event-stream")
+}
+
```
Labreo, line 215: If mime.ParseMediaType runs into a malformed parameter (like text/event-stream; ::), returning false pushes the response into handleNonStreamingResponse. That causes it to stall reading up to the 4 MB limit instead of streaming, and gen_ai.request.is_stream gets lost. Falling back to the base type before ; on error keeps streaming working for those cases.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -200,6 +201,18 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
...
+// type. HTTP media types are case-insensitive and may carry parameters, so the
+// raw prefix check used previously misclassified headers like
+// "Text/Event-Stream; charset=utf-8".
+func isSSEContentType(contentType string) bool {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(mediaType, "text/event-stream")
+}
```
Labreo, line 214: If mime.ParseMediaType runs into a malformed parameter (like text/event-stream; ::), returning false pushes the response into handleNonStreamingResponse. That causes it to stall reading up to the 4 MB limit instead of streaming, and gen_ai.request.is_stream gets lost. Falling back to the base type before ; on error keeps streaming working for those cases.

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go
```diff
@@ -170,6 +171,18 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
...
+// raw prefix check used previously misclassified headers like
+// "Text/Event-Stream; charset=utf-8".
+func isSSEContentType(contentType string) bool {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(mediaType, "text/event-stream")
+}
+
```
Labreo, line 185: If mime.ParseMediaType runs into a malformed parameter (like text/event-stream; ::), returning false pushes the response into handleNonStreamingResponse. That causes it to stall reading up to the 4 MB limit instead of streaming, and gen_ai.request.is_stream gets lost. Falling back to the base type before ; on error keeps streaming working for those cases.

instrumentation/github.com/openai/openai-go/v2/middleware_test.go
```diff
@@ -128,3 +128,24 @@ func TestParseChatResponse_Valid(t *testing.T) {
...
+		contentType string
+		want        bool
+	}{
+		{"lowercase", "text/event-stream", true},
+		{"mixed case", "Text/Event-Stream", true},
+		{"uppercase with parameter", "TEXT/EVENT-STREAM; charset=utf-8", true},
+		{"json", "application/json", false},
+		{"empty", "", false},
+		{"prefix only", "text/event-streamx", false},
+		{"malformed parameter", "text/event-stream; ::", false},
```
Labreo, line 144: With the fallback check above, this should expect true so malformed parameters don't break streaming detection.

instrumentation/github.com/openai/openai-go/v3/middleware_test.go
```diff
@@ -128,3 +128,24 @@ func TestParseChatResponse_Valid(t *testing.T) {
...
+		contentType string
+		want        bool
+	}{
+		{"lowercase", "text/event-stream", true},
+		{"mixed case", "Text/Event-Stream", true},
+		{"uppercase with parameter", "TEXT/EVENT-STREAM; charset=utf-8", true},
+		{"json", "application/json", false},
+		{"empty", "", false},
+		{"prefix only", "text/event-streamx", false},
+		{"malformed parameter", "text/event-stream; ::", false},
```
Labreo, line 144: With the fallback check above, this should expect true so malformed parameters don't break streaming detection.

instrumentation/github.com/openai/openai-go/middleware_test.go
```diff
@@ -128,3 +128,24 @@ func TestParseChatResponse_Valid(t *testing.T) {
...
+		contentType string
+		want        bool
+	}{
+		{"lowercase", "text/event-stream", true},
+		{"mixed case", "Text/Event-Stream", true},
+		{"uppercase with parameter", "TEXT/EVENT-STREAM; charset=utf-8", true},
+		{"json", "application/json", false},
+		{"empty", "", false},
+		{"prefix only", "text/event-streamx", false},
+		{"malformed parameter", "text/event-stream; ::", false},
```
Labreo, line 144: With the fallback check above, this should expect true so malformed parameters don't break streaming detection.

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go
```diff
@@ -653,3 +653,24 @@ func assertBoolAttribute(t *testing.T, attrs []attribute.KeyValue, key string, e
...
+		contentType string
+		want        bool
+	}{
+		{"lowercase", "text/event-stream", true},
+		{"mixed case", "Text/Event-Stream", true},
+		{"uppercase with parameter", "TEXT/EVENT-STREAM; charset=utf-8", true},
+		{"json", "application/json", false},
+		{"empty", "", false},
+		{"prefix only", "text/event-streamx", false},
+		{"malformed parameter", "text/event-stream; ::", false},
```
Labreo, line 669: With the fallback check above, this should expect true so malformed parameters don't break streaming detection.

Issue #549, opened 2026-05-29
feat(tool/rule): glob matching on target for package selection


kakkoyun:
Follow-up from #377 (OP4 stack). Tracks the deferred "target globbing" item and supersedes the dropped `import_path` glob filter (#382).

## Context

Rule-to-package matching is currently exact-match: `runMatch` looks up `rulesByTarget[dep.ImportPath]` (a plain map keyed by the rule's `target`). There is no way to target a family of packages.

PR #382 (OP4.4, branch `kakkoyun/op4/4-import-path`) implemented a separate `import_path` glob filter. During #377 review the schema was narrowed (3-tier → 2-tier), and `import_path` was dropped from the stack to avoid two overlapping package-selection mechanisms. The agreed direction is to fold glob support into the existing `target` selector instead.

## Problem

Instrumentation rules cannot target package families (e.g. all of `example.com/foo/...`), forcing one rule per exact import path.

## Proposed work

- Support `*` (one path segment) and `**` (zero-or-more segments) on `target`.
- Split matching into exact rules (map lookup, unchanged fast path) and glob rules (evaluated against every dependency import path).
- `target` remains the single package-selection surface; no separate `import_path` field.

## Reference implementation

PR #382 implemented the glob engine and the `matchDeps` exact/glob split. **Caveats for porting:** it targets the old `tool/internal/filter` package layout (now `tool/internal/setup/filter.go`) and predates the upstream merge and the 2-tier schema, so it needs reworking — treat it as a design reference, not a cherry-pick.

## References

- #377 deferred item "Target globbing"
- Closed PR #382 (reference implementation)
- Umbrella #346

PR #1217, opened 2026-08-23
feat(tool/rule): add exclude_targets for package-level exclusion

### LINKED ISSUE #545: feat(tool/rule): support target exclusion for call-site instrumentation
Follow-up from #377 (review-summary question from @amazingakai).

## Context

Call-site instrumentation often wants to instrument *all* packages that call a given API, except a specific helper package that should be left alone. Concrete example: instrument all gRPC client call sites except `go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc`, which already handles its own tracing.

Today, `target` is exact-match and lives outside `where`. There is no way to express "all targets except X" without enumerating every other target — and once target globbing lands, there is still no negation operator.

## Problem

Without exclusion support, every call-site instrumentation rule must duplicate logic to skip known-instrumented packages, or risk recursive instrumentation loops.

## Proposed work

Two options to evaluate:

1. **Move `target` into `where`** so it can participate in `not:` / `one-of:`. Cleaner long-term but a breaking schema change.
2. **Add a top-level `exclude_targets:`** (or `target_excludes:`) list that is OR'd against `target` at match time. Less elegant, no schema break.

Decision needed: which option, and how it interacts with target globbing (which is also a deferred follow-up).

## References

- #377 review summary from @amazingakai

Harsh63870:
## Description

Adds optional top-level `exclude_targets` on instrumentation rules. After `target` (exact or glob) matches a dependency import path, any listed exclude entry skips the rule for that package at setup match time.

- `InstBaseRule.ExcludeTargets` + `GetExcludeTargets()` on `InstRule`
- Load-time validation via `ValidateExcludeTargets` (same glob engine as `target`)
- Setup `runMatch` filters excluded deps before version/AST matching
- Documented in `docs/rules.md`

Option 2 from #545 (non-breaking). Does not move `target` into `where`.

## Motivation

Call-site rules often need to match a broad `target` (including globs) but skip packages that already self-instrument (e.g. `otelgrpc`). Without exclusion, authors must duplicate rules or risk double instrumentation. Target globbing (#549) is already landed; this adds the complementary skip list.

Fixes #545

---

## Checklist

- [ ] Tests pass: `make test`


Labreo:
The branch is several commits behind `main` .A rebase is necessary.I checked your local fork and you need to sync it to upstream and then pull those changes so that your local repository can have the latest changes that have been added.

Labreo:
Also just a suggestion for the future.Please follow the [pull request template](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/.github/PULL_REQUEST_TEMPLATE.md).It helps contributors and maintainers get a better idea for the pull request.And allows them to manage and merge your pull request faster.

Harsh63870:
thanks @Labreo, rebased and pushed, should be in sync now..

Harsh63870:
@kakkoyun @amazingakai @darccio, all checks are passing, please review this pr...

tool/internal/setup/match.go
```diff
@@ -96,22 +93,48 @@ func parseRuleFromYaml(content []byte) ([]rule.InstRule, error) {
...
+	filtered := rules[:0]
+	for _, r := range rules {
+		if rule.MatchesExcludeTargets(importPath, r.GetExcludeTargets()) {
+			continue
+		}
+		filtered = append(filtered, r)
+	}
+	return filtered
+}
+
```
Labreo: Wont doing rules[:0] modifies the slice in exactRules in place, which is supposed to be read-only across workers?

We could only create a new slice when we actually find an excluded match then we can keep it zero-allocation for normal cases and avoid modifying the map's slice directly.

Labreo (commented):
Left a suggestion other than that looks good to me.

tool/internal/setup/match.go
Harsh63870: good catch, rules[:0] was aliasing the exactRules backing array, updated to only allocate when an exclusion actually matches; otherwise the original slice is returned unchanged...

Labreo:
Hey @Harsh63870 .According to codecov your PR is still 161 commits behind main.That needs to be rectified.

darccio:
@Harsh63870 Ack. Just a note: please, don't ping maintainers unless they are already participating in the PR.

Harsh63870:
everything is clean now...

Harsh63870:
@y1yang0, just a quick follow-up on the pr, could you please take a look when you get a chance...

tool/internal/rule/exclude_targets.go
```diff
@@ -0,0 +1,37 @@
...
+	"go.opentelemetry.io/otelc/tool/ex"
+)
+
+// ValidateExcludeTargets rejects malformed exclude_targets entries at load time.
+func ValidateExcludeTargets(excludes []string) error {
+	for i, pattern := range excludes {
+		if strings.TrimSpace(pattern) == "" {
+			return ex.Newf("exclude_targets[%d] is empty", i)
+		}
+		if err := ValidateTarget(pattern); err != nil {
```
ADITYA-CODE-SOURCE, line 18: `ValidateTarget` accepts `$root`, but `MatchesExcludeTargets` only handles literal paths and glob patterns. As a result, `$root` passes validation but never matches the resolved root module. Should `$root` be rejected specifically for `exclude_targets`, or should it be expanded during setup in the same way as a positive `$root` target?

ADITYA-CODE-SOURCE (changes requested):

tool/internal/rule/exclude_targets.go
Harsh63870: good catch, you're right.
`ValidateExcludeTargets` reuses `ValidateTarget`, so `$root` passes load-time validation, but `MatchesExcludeTargets` only compares against concrete import paths and globs. Since no dependency has import path `"$root"`, an exclude entry of `$root` is currently a silent no-op,
for positive `target`, setup already expands `$root` to `rootModule/**` in `matchDeps` before matching, exclusion doesn't go through that path today,
I'd lean toward **expanding `$root` during setup** (same as positive targets) so `exclude_targets` stays consistent with the documented "same syntax as `target"` semantics. Concretely: resolve `rootModulePaths` in setup and expand each `$root` exclude entry to `root + "/**"` before calling `MatchesExcludeTargets`,
the alternative is to **reject `$root` explicitly in `ValidateExcludeTargets`** with a load-time error, since exclusion operates on resolved import paths rather than setup selectors. That's a smaller change but means `$root` is not supported in `exclude_targets` even though the field docs mention the same syntax as `target`.
I'll add a test either way, either `$root` expands and excludes packages under the root module, or `$root` is rejected at parse time, so this can't regress silently.

which direction do you prefer?

ADITYA-CODE-SOURCE: @Harsh63870   I’d go with expanding `$root` during setup, since that keeps it consistent with positive targets and the documented syntax.

One case worth covering is `target: "**"` with `exclude_targets: ["$root"]`: root-module resolution should still happen when only the exclusion uses `$root`, with a clear error if no root module is found.

Could you add a setup-level test showing that the root package and its subpackages are excluded, while an unrelated dependency still matches? Doing the expansion before concurrent matching would also keep the shared rules read-only.

Harsh63870: done, `$root` in `exclude_targets` is now expanded during setup the same way as positive targets, root module resolution runs even when only the exclusion uses `$root`, and fails with a clear error if no root module is found,
added `TestMatchDeps_ExcludeRootTargetExpandsAtSetup` for `target: "**"` + `exclude_targets: ["$root"]`: root package and subpackages are excluded, unrelated dependency still matches, expansion happens once in `matchDeps` before concurrent matching so shared rules stay read-only...

PR #1220, opened 2026-08-23
fix(tool/internal/ast): handle import alias collisions in ImportAliasMap

### LINKED ISSUE #1219: bug(tool/internal/ast): ImportAliasMap overwrites map entries on colliding default import aliases
### Describe the bug

In `tool/internal/ast/typename.go`, `ImportAliasMap` parses package imports and maps local package identifiers to their full import paths. The docstring notes that it disambiguates:

> *"distinct import paths that happen to share a last path segment (e.g. `text/template` vs `html/template`, both conventionally `template`)"*

However, the implementation currently overwrites existing keys in `aliases` directly:

```
```go
// tool/internal/ast/typename.go:161-179
aliases := make(map[string]string, len(specs))
for _, imp := range specs {
	...
	alias := defaultImportAlias(path)
	if imp.Name != nil {
		alias = imp.Name.Name
	}
	...
	aliases[alias] = path // Overwrites previous entry if alias is identical
}
```

If a source file imports both text/template and html/template without explicit aliases, aliases["template"] will be overwritten by whichever import appears last in file.Imports, discarding the first entry and breaking type matching for that package.

**Environment**

- OS: macOS / Linux / Windows
- Go Version: not version-specific
- Version: main

**To Reproduce**

1. Parse a source file that imports two unaliased packages with identical path tails (e.g., text/template and html/template).
2. Pass the file to ImportAliasMap(file).
3. Notice that aliases["template"] only contains the second package path, overwriting the first.

**Expected behavior**
When multiple unaliased imports collapse to the same default alias, ImportAliasMap should detect the collision and remove the ambiguous alias from the map (unless one of the imports uses an explicit alias).

**Additional context**
Related files:

- tool/internal/ast/typename.go:160-181 (ImportAliasMap)
- tool/internal/ast/typename_test.go:324-360

vedant21-oss:
Fixes #1219

### Description
This PR resolves issue #1219 in `ImportAliasMap` where distinct unaliased imports sharing a last path segment (e.g. `text/template` and `html/template`, both conventionally defaulting to `template`) would silently overwrite existing map entries.

### Changes Made
1. Updated `ImportAliasMap` in `tool/internal/ast/typename.go` to track explicit vs default import aliases.
2. If two unaliased imports produce colliding default package names, the ambiguous alias key is deleted from `aliases` instead of overwriting the previous entry.
3. Explicit aliases (`import txttemplate "text/template"`) continue to take precedence over default package names.
4. Added comprehensive unit test coverage in `tool/internal/ast/typename_test.go`.

### Verification
- All AST package unit tests (`go test -v ./tool/internal/ast/...`) passed cleanly (100%).

### DIFF
--- a/tool/internal/ast/typename.go
+++ b/tool/internal/ast/typename.go
@@ -158,6 +158,7 @@ func ImportAliasMap(file *dst.File) map[string]string {
 	}
 
 	aliases := make(map[string]string, len(specs))
+	explicit := make(map[string]bool, len(specs))
 	for _, imp := range specs {
 		if imp.Path == nil {
 			continue
@@ -167,15 +168,31 @@ func ImportAliasMap(file *dst.File) map[string]string {
 			continue
 		}
 		alias := defaultImportAlias(path)
+		isExplicit := false
 		if imp.Name != nil {
 			alias = imp.Name.Name
+			isExplicit = true
 		}
 		// Blank and dot imports don't introduce a qualified identifier that a
 		// type reference could use, so they can't participate in matching.
 		if alias == "" || alias == "_" || alias == "." {
 			continue
 		}
-		aliases[alias] = path
+
+		if existing, exists := aliases[alias]; exists && existing != path {
+			if isExplicit {
+				aliases[alias] = path
+				explicit[alias] = true
+			} else if !explicit[alias] {
+				// Disambiguate: two distinct default aliases collide; remove ambiguous key.
+				delete(aliases, alias)
+			}
+		} else {
+			aliases[alias] = path
+			if isExplicit {
+				explicit[alias] = true
+			}
+		}
 	}
 	return aliases
 }
--- a/tool/internal/ast/typename.go
+++ b/tool/internal/ast/typename.go
@@ -159,6 +159,8 @@ func ImportAliasMap(file *dst.File) map[string]string {
 
 	aliases := make(map[string]string, len(specs))
 	explicit := make(map[string]bool, len(specs))
+	ambiguous := make(map[string]bool, len(specs))
+
 	for _, imp := range specs {
 		if imp.Path == nil {
 			continue
@@ -179,18 +181,17 @@ func ImportAliasMap(file *dst.File) map[string]string {
 			continue
 		}
 
-		if existing, exists := aliases[alias]; exists && existing != path {
-			if isExplicit {
-				aliases[alias] = path
-				explicit[alias] = true
-			} else if !explicit[alias] {
-				// Disambiguate: two distinct default aliases collide; remove ambiguous key.
-				delete(aliases, alias)
-			}
-		} else {
+		if isExplicit {
 			aliases[alias] = path
-			if isExplicit {
-				explicit[alias] = true
+			explicit[alias] = true
+			delete(ambiguous, alias)
+		} else if !explicit[alias] {
+			if existing, exists := aliases[alias]; exists && existing != path {
+				// Disambiguate: two or more distinct default aliases collide; remove key and track as ambiguous.
+				delete(aliases, alias)
+				ambiguous[alias] = true
+			} else if !ambiguous[alias] {
+				aliases[alias] = path
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


tool/internal/ast/typename.go
```diff
@@ -167,15 +168,31 @@ func ImportAliasMap(file *dst.File) map[string]string {
...
 		}
-		aliases[alias] = path
+
+		if existing, exists := aliases[alias]; exists && existing != path {
+			if isExplicit {
+				aliases[alias] = path
+				explicit[alias] = true
+			} else if !explicit[alias] {
+				// Disambiguate: two distinct default aliases collide; remove ambiguous key.
+				delete(aliases, alias)
```
darkraider01: This only actually disambiguates for an even number of colliding default aliases. Once this delete fires for the second collision, there's no record left that "template" was ever contested, so a third unaliased import that also guesses "template" just hits the exists==false branch below and gets written back in as if it were the first, unambiguous sighting.

I checked this against the branch directly. Two colliding defaults (example.com/foo/util, example.com/bar/util) correctly end with imports["util"] absent. Add a third (example.com/baz/util) and imports["util"] comes back as "example.com/baz/util", picked arbitrarily.

This isn't that contrived either, defaultImportAlias is documented right above as a syntactic guess off the path tail, and the doc comment even calls out go-redis declaring package "redis" as an example of the guess being wrong. Three unrelated, perfectly valid third-party imports landing on the same guessed tail is realistic. Since ImportAliasMap feeds the type matching in apply_call.go, apply_directive.go, apply_lit.go and shared.go, a wrong resolution here means a rule can silently match against the wrong package's type.

Would probably want to track "this alias is ambiguous" as its own state rather than deleting the map entry, so a later collision doesn't un-ambiguous it by accident.

tool/internal/ast/typename_test.go
```diff
@@ -324,6 +324,92 @@ func f(x *foo.T, y *bar.T) {}
 		assert.Equal(t, "github.com/b/bar/v2", imports["bar"])
 	})
 
+	t.Run("unaliased colliding default aliases are removed as ambiguous", func(t *testing.T) {
```
darkraider01, line 327: The two-way case here is covered well, along with both explicit-overrides-default orderings and the identical-duplicate case. A three-way collision case sitting next to this one would've caught the bug I left a comment on above, right now nothing exercises what happens after the map entry gets deleted once and a third collider shows up.

vedant21-oss: Good catch, @darkraider01! Thanks for the sharp review.

I've updated `ImportAliasMap` to track ambiguous alias keys in an `ambiguous := make(map[string]bool)` state map instead of just deleting the entry. When a 2nd, 3rd, or N-th default alias collision occurs, the key remains recorded in `ambiguous` so subsequent unaliased imports cannot re-introduce the alias by accident. If an explicit alias is encountered, it overrides the ambiguous state as expected.

I also added a 3-way collision test case (`three-way_unaliased_colliding_default_aliases_remain_ambiguous`) in `typename_test.go` covering `foo/util`, `bar/util`, and `baz/util`.

tool/internal/ast/typename.go
```diff
@@ -167,15 +170,30 @@ func ImportAliasMap(file *dst.File) map[string]string {
...
 			continue
 		}
-		aliases[alias] = path
+
+		if isExplicit {
+			aliases[alias] = path
+			explicit[alias] = true
+			delete(ambiguous, alias)
+		} else if !explicit[alias] {
+			if existing, exists := aliases[alias]; exists && existing != path {
```
darccio, line 189: Removing the alias from `aliases` does not prevent a later type match from
guessing the same path tail. `parsedTypeName.matches` treats a missing alias as
missing import context and falls back to `defaultImportAlias` at lines 61–68.

As a result, after this branch removes an ambiguous `util` alias, `util.T` can
still match each `.../util.T` target by tail. The ambiguity handling therefore
does not prevent the false match it intends to prevent.

Use the tail fallback only when `imports == nil`. When a non-nil import map does
not contain the qualifier, return `false`. Please also add a regression test
that calls `matches` with an `ImportAliasMap` result for colliding path tails
whose packages have different declared names.

darccio (changes requested):

vedant21-oss:
Closing this in favour of #1274, which covers the same collision fix plus the strict import matching. Keeping the discussion in one place.

PR #1226, opened 2026-08-24
refactor(tool): reuse AstParser.ParseSnippet for snippet parsing in call_template

### LINKED ISSUE #1223: Resue AstParser.ParseSnippet in call_template
While reviewing another PR, I found that `parseSnippetFuncDecl` is basically a duplicate of `AstParser.ParseSnippet`. It could probably be replaced by the latter with moderate changes.

bhuvan-somisetty:
## Description

`parseSnippetFuncDecl` in `tool/internal/instrument/call_template.go` wrapped a source
snippet in `package _\nfunc _() { ... }`, parsed it with `go/parser`, and decorated it into
a `dst.File` to get at the parsed statements. This duplicated `AstParser.ParseSnippet` in
`tool/internal/ast/parser.go`, which does the same wrap/parse/decorate for the same purpose.
`compileExpression` in the same file had its own third copy of this logic inlined.

This PR:

- Replaces `parseGoExpression` and `parseGoTypeExpression`'s use of `parseSnippetFuncDecl`
  with `AstParser.ParseSnippet` directly, and removes `parseSnippetFuncDecl`.
- Replaces `compileExpression`'s inline `go/parser` + `decorator` call with `AstParser.ParseSnippet`.
- Preserves existing error-message behavior: an empty/whitespace-only result still reports
  "function body is empty" (handled explicitly, since `ParseSnippet` would otherwise report
  the less specific "empty source"), and the existing single-expression-statement /
  single-declaration-statement checks are kept.
- As a side effect of routing through `ParseSnippet`, `parseGoExpression` and
  `parseGoTypeExpression` now explicitly reject multi-statement input (e.g. `"a(); b()"`)
  instead of silently keeping only the first statement and discarding the rest — the
  previous implementation never checked for more than one statement at these two call sites.

No behavior change for any existing call site under `tool/internal/instrument`
(`apply_call.go`, `apply_lit.go`, `apply_decl.go`).

## Motivation

Fixes #1223, which asks to deduplicate `parseSnippetFuncDecl` against the existing
`AstParser.ParseSnippet`, and (per the issue's follow-up comment) do the same for the
snippet-parsing code inlined in `compileExpression`.

Fixes #1223

---

## Checklist

- [x] Code formatted: `gofmt` (ran `go build`, `go vet`, `go test` on the `tool` module; `make format`/`make lint`/`make test` require Docker/Make tooling not available in this environment)
- [x] Linters pass: `golangci-lint run` on the changed packages (`tool/internal/instrument`, `tool/internal/ast`) reports no new issues
- [x] Tests pass: `go test ./tool/...` (full module)

[^I-know-my-stuff]:


tool/internal/instrument/call_template.go
```diff
@@ -138,45 +134,26 @@ func (t *callTemplate) compileExpression(node dst.Expr, enclosing *dst.FuncDecl)
...
+	// An empty template result would otherwise surface as an opaque "empty
+	// source" error from ParseSnippet; call it out explicitly instead.
+	if strings.TrimSpace(userResult) == "" {
+		return nil, ex.New("function body is empty")
 	}
 
-	// Convert ast.File to dst.File
-	dec := decorator.NewDecorator(fset)
-	dstFile, err := dec.DecorateFile(file)
+	stmts, err := toolast.NewAstParser().ParseSnippet(userResult)
```
darkraider01, line 195: Good cleanup, and nice catch on the multi statement case, the old parseSnippetFuncDecl silently dropping everything after the first statement was a real bug.

One regression though. ParseSnippet builds its wrapper as "package main; func _() {" + source + "}" with no newline before the closing brace, whereas the code being replaced here wrapped with "...\n}\n". So if a call template renders to something ending in a // line comment, the comment eats the closing brace and parsing fails with "expected '}', found 'EOF'".

I checked this on both branches with the template wrapper({{ . }}) // wrap the call. On main it compiles fine, on this branch it errors. Same thing happens via parseGoExpression and parseGoTypeExpression, so all three call sites here are affected. Nothing in the current rule set ends a template in a comment so CI stays green, but it's a trap for whoever writes one next.

Probably worth fixing in ParseSnippet itself rather than here, since apply_directive.go and apply_raw.go call it too and carry the same risk. Changing the wrapper to + "\n}" was enough to fix all three cases for me, and the ast and instrument suites still pass with it.

bhuvan-somisetty: Good catch, thanks for tracking that down across all three call sites. Fixed it in ParseSnippet itself (added the newline before the closing brace) in 41e13d1, plus a regression test for a snippet ending in a line comment. ast and instrument suites both pass locally.

tool/internal/instrument/call_template.go
```diff
@@ -138,45 +134,26 @@ func (t *callTemplate) compileExpression(node dst.Expr, enclosing *dst.FuncDecl)
...
-	fset := token.NewFileSet()
-	file, err := parser.ParseFile(fset, "", []byte(wrapped), parser.ParseComments)
-	if err != nil {
-		// Format the error with the generated code for debugging
-		formatted, _ := format.Source([]byte(wrapped))
-		return nil, ex.Wrapf(err, "failed to parse generated code\nGenerated code:\n%s", formatted)
+	// An empty template result would otherwise surface as an opaque "empty
+	// source" error from ParseSnippet; call it out explicitly instead.
+	if strings.TrimSpace(userResult) == "" {
+		return nil, ex.New("function body is empty")
```
y1yang0: You can fold into ParseSnippet.

tool/internal/instrument/call_template.go
```diff
@@ -138,45 +134,26 @@ func (t *callTemplate) compileExpression(node dst.Expr, enclosing *dst.FuncDecl)
...
-
-	if funcDecl.Body == nil || len(funcDecl.Body.List) == 0 {
+	if len(stmts) == 0 {
 		return nil, ex.New("function body is empty")
 	}
-	if len(funcDecl.Body.List) != 1 {
-		return nil, ex.Newf("expected single expression statement, got %d statements", len(funcDecl.Body.List))
+	if len(stmts) != 1 {
+		return nil, ex.Newf("expected single expression statement, got %d statements", len(stmts))
 	}
```
y1yang0: Only preserve `len(stmts)!=1` check

tool/internal/instrument/call_template.go
```diff
@@ -138,45 +134,26 @@ func (t *callTemplate) compileExpression(node dst.Expr, enclosing *dst.FuncDecl)
...
-	dstFile, err := dec.DecorateFile(file)
+	stmts, err := toolast.NewAstParser().ParseSnippet(userResult)
 	if err != nil {
-		return nil, ex.Newf("failed to decorate AST")
-	}
-
-	// Extract the expression from the function body
-	if len(dstFile.Decls) == 0 {
-		return nil, ex.New("no declarations found in generated code")
+		return nil, ex.Wrapf(err, "failed to parse generated code\nGenerated code:\n%s", userResult)
```
y1yang0, line 197: I think you can simply return err here and wrap err in ParseSnippet, something like

```
// ParseSnippet parses the AST from incomplete source code snippet.
func (ap *AstParser) ParseSnippet(source string) ([]dst.Stmt, error) {
	...
	if err != nil {
		return nil, ex.Wrap(err, "can not parse snippet "+ source) <---- wrap here
	}
	funcDecl := util.AssertType[*dst.FuncDecl](file.Decls[0])
	return funcDecl.Body.List, nil
}
```

y1yang0: I also noticed a minor issue. Wrap and Wrapf are somewhat overused right now, Tracked in #1231

tool/internal/ast/parser.go
```diff
@@ -62,7 +62,7 @@ func (ap *AstParser) ParseSnippet(source string) ([]dst.Stmt, error) {
 	if source == "" {
 		return nil, ex.New("empty source")
 	}
-	snippet := "package main; func _() {" + source + "}"
+	snippet := "package main; func _() {" + source + "\n}"
```
y1yang0, line 66: Good catch 👍

tool/internal/instrument/call_template.go
bhuvan-somisetty: Done, moved the source context wrapping into ParseSnippet itself (it now does ex.Wrapf(err, "can not parse snippet %s", source)), so compileExpression/parseGoExpression/parseGoTypeExpression just return nil, err on parse failure instead of re-wrapping. Also dropped the len(stmts)==0 check and the pre-check for empty template output, kept only len(stmts) != 1 since ParseSnippet already rejects whitespace only source with its own "empty source" error. Updated the three tests that asserted the old messages. Pushed in ba55f73, ast and instrument suites are green.

y1yang0 (approved):

y1yang0:
Can you resolve the conflicts? Then we can merge it. Thanks.

bhuvan-somisetty:
Hey, sorry for the delay. Conflicts are resolved and pushed now, should be good to merge.

PR #1228, opened 2026-08-24
fix(instrumentation): end linodego and k8s spans when disabled mid-operation

### LINKED ISSUE #1227: bug(instrumentation): linodego and k8s.io/client-go leak spans when disabled mid-operation
## Describe the bug

Two more instrumentations have the span-lifecycle bug #1094 fixed for net/http and #1209 fixes for database/sql: an after hook re-checks `Enable()` and returns before ending a span the before hook already created. Because `Instrumented()` reads `OTEL_GO_DISABLED_INSTRUMENTATIONS` fresh on every call, a span created while enabled leaks if instrumentation is disabled before the operation completes.

### linode/linodego

- `BeforeDoRequest` starts and stores a span (`hook.go`).
- `AfterDoRequest` (`hook.go`) starts with `if !enabler.Enable() { return }` ahead of retrieving the span and its `defer span.End()`.
- `afterAPICall` (`public_method.go`) has the same shape.

### k8s.io/client-go

- `beforeProcessDeltas` and `beforeProcessDeltasInBatch` start and store a span (`hook.go`).
- `afterProcessDeltas` and `afterProcessDeltasInBatch` start with `if !k8SEnabler.Enable() { return }` ahead of the span retrieval and `defer span.End()`.

In every case the `!ok || span == nil` guard sits after the `Enable()` re-check, so the guard already covers the disabled-at-before case (no span was stored), which makes the `Enable()` re-check redundant with a check that is already present.

## Reproduce

1. Enable the instrumentation and run the before hook so a span is created and stored.
2. Set `OTEL_GO_DISABLED_INSTRUMENTATIONS` to that instrumentation.
3. Run the matching after hook.

The stored span is never ended. `k8s.io/client-go`s `TestAfterProcessDeltas` "instrumentation disabled" case already sets this exact scenario up (a span in hook data, env disabled) and currently asserts 0 ended spans, documenting the leak.

## Expected behavior

The after hook should end a span created by the before hook regardless of the current enablement state, matching net/http after #1094. Remove the redundant `Enable()` re-check from the after hooks, keeping it in the before hooks where span creation is correctly gated.

Both packages build and test standalone, so this can carry a mid-operation-disable regression test, like `TestAfterRoundTrip_DisabledAfterStart_Regression` in net/http.

Happy to send the fix.

vignesh2027:
## Description

Two more instrumentations had the span-lifecycle bug #1094 fixed for net/http: an after hook re-checked `Enable()` and returned before ending a span the before hook already created. Because `Instrumented()` reads `OTEL_GO_DISABLED_INSTRUMENTATIONS` fresh on every call, a span created while enabled leaked if instrumentation was disabled before the operation completed.

- **linodego**: `AfterDoRequest` (`hook.go`) and `afterAPICall` (`public_method.go`).
- **k8s.io/client-go**: `afterProcessDeltas` and `afterProcessDeltasInBatch` (`hook.go`).

In every case the `!ok || span == nil` guard sat after the `Enable()` re-check, so the guard already covered the disabled-at-before case (no span stored). This removes the redundant re-check from the after hooks, keeping it in the before hooks where span creation is gated. A disabled before hook stores nothing, so the after path is a safe no-op; a span that was created is always ended.

This mirrors the net/http fix in #1094 and the sibling database/sql fix in #1209.

## Tests

Both packages build and test standalone, so this adds mid-operation-disable regression tests:

- linodego: `TestAfterDoRequest_DisabledAfterStart_Regression`, `TestPublicMethodHooks_DisabledAfterStart_Regression`.
- k8s.io/client-go: the `TestAfterProcessDeltas` and `TestAfterProcessDeltasInBatch` "disabled after span started" cases now assert the span is ended (they previously asserted the leak).

Verified each regression test fails on the pre-fix code and passes with the fix.

Fixes #1227

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


Labreo:
LGTM.Nice PR @vignesh2027 !

om7057 (commented):
Went ahead and pulled this one down to actually run it rather than just read through it. Full suite plus `go vet` came back clean for both linodego v2 and k8s.io/client-go, no surprises there.

Wanted to be sure the leak you're describing is real and not just something that reads plausible on paper, so I swapped `hook.go` back to whatever's currently on main while keeping your new tests in place, and ran them against that. Both new cases failed exactly the way you'd expect, span count came back 0 instead of 1 in both `TestAfterProcessDeltas` and `TestAfterProcessDeltasInBatch`. Then put the fixed version back and they passed. So the regression tests are actually catching the bug, not just asserting something trivially true.

The reasoning behind the fix also checks out once you look at it closely. The disabled-at-before-hook case was already handled by the `!ok || span == nil` guard sitting right there in the after hook, so that extra `Enable()` re-check wasn't protecting anything, it was just quietly causing the exact leak you're fixing whenever instrumentation got disabled mid-flight. Removing it is the correct call here, not a band-aid.

Good, clean mirror of #1094/#1209, nothing to push back on.

PR #1229, opened 2026-08-24
fix(tool): handle generic receiver type parameters

### LINKED ISSUE #1218: Generic receiver type arguments are not passed to generated trampoline calls
While investigating generic receiver handling, I found a case where the generated AfterTrampoline call cannot infer the receiver's type parameter.

For example:

type Box[T any] struct {
    value T
}

func (b *Box[T]) Reset() error {
    return nil
}

The trampoline generation correctly recovers T from the receiver and generates something equivalent to:

func OtelAfterTrampoline_Reset[T any](
    hookContext *HookContext,
    ret0 *error,
)

However, createTJumpIf uses funcDecl.Type.TypeParams when generating the call. Since the receiver's T isn't stored there for a method, the generated call becomes:

defer OtelAfterTrampoline_Reset(hookContext, &ret0)

There is no T in either argument, so Go cannot infer it. I reproduced this with the example above and the generated code fails with:

cannot infer T (in call to OtelAfterTrampoline_Reset)

Interestingly, the existing generic method test returns T, so that case passes because Go can infer T from the *T return value argument. A method returning error, bool, etc. exposes the problem.

I also traced this through the existing generic receiver implementation/history. The trampoline declaration already recovers the receiver type parameters, but the call site generation still uses funcDecl.Type.TypeParams.

I wanted to raise this separately first to confirm whether the receiver type arguments are expected to be passed explicitly at the trampoline call site,  i would like to be assigined for fixing this !

SSOHEB:
## Motivation

Fixes #1218

When instrumenting methods on generic types, the generated trampoline functions correctly include the receiver's type parameters, but the trampoline calls were not passing those type arguments.

This change resolves the generic type parameters from the method receiver and uses them when generating the trampoline calls. It also adds unit test coverage for methods with both single and multiple receiver type parameters.

The existing behavior for regular functions, regular methods, and top level generic functions remains unchanged.

## Checklist

- [x] Code formatted
- [x] Linters pass
- [x] Tests pass

### DIFF
--- a/tool/internal/instrument/apply_func.go
+++ b/tool/internal/instrument/apply_func.go
@@ -165,7 +165,7 @@ func createTrampArgs(names []string) []dst.Expr {
 	return exprs
 }
 
-func createTJumpIf(t *rule.InstFuncRule, funcDecl *dst.FuncDecl,
+func createTJumpIf(file *dst.File, t *rule.InstFuncRule, funcDecl *dst.FuncDecl,
 	args, retVals []string,
 ) *dst.IfStmt {
 	funcSuffix := t.Identity()
@@ -175,8 +175,9 @@ func createTJumpIf(t *rule.InstFuncRule, funcDecl *dst.FuncDecl,
 	argsToAfter = append([]dst.Expr{argHookContext}, argsToAfter...)
 	beforeCallName := makeName(t, funcDecl, true)
 	afterCallName := makeName(t, funcDecl, false)
-	beforeCall := ast.CallTo(beforeCallName, funcDecl.Type.TypeParams, argsToBefore)
-	afterCall := ast.CallTo(afterCallName, funcDecl.Type.TypeParams, argsToAfter)
+	typeParams := findTargetGenericType(file, funcDecl)
+	beforeCall := ast.CallTo(beforeCallName, typeParams, argsToBefore)
+	afterCall := ast.CallTo(afterCallName, typeParams, argsToAfter)
 	tjumpInit := ast.DefineStmts(
 		ast.Exprs(
 			ast.Ident(trampolineHookContextName+funcSuffix),
@@ -271,7 +272,7 @@ func (ip *instrumentPhase) insertTJump(t *rule.InstFuncRule, funcDecl *dst.FuncD
 	// the context, handles exceptions, etc, and ultimately jumps to the real
 	// hook code. By inserting trampoline-jump-if at the target function entry,
 	// we can intercept the original function and execute before/after hooks.
-	tjump := createTJumpIf(t, funcDecl, args, retVals)
+	tjump := createTJumpIf(ip.target, t, funcDecl, args, retVals)
 
 	// Record the trampoline-jump-if as they can be optimized later, they are
 	// performance-critical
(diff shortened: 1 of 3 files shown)


tool/internal/instrument/apply_func.go
```diff
@@ -175,8 +175,9 @@ func createTJumpIf(t *rule.InstFuncRule, funcDecl *dst.FuncDecl,
 	argsToAfter = append([]dst.Expr{argHookContext}, argsToAfter...)
 	beforeCallName := makeName(t, funcDecl, true)
 	afterCallName := makeName(t, funcDecl, false)
-	beforeCall := ast.CallTo(beforeCallName, funcDecl.Type.TypeParams, argsToBefore)
-	afterCall := ast.CallTo(afterCallName, funcDecl.Type.TypeParams, argsToAfter)
+	typeParams := findTargetGenericType(file, funcDecl)
```
darkraider01, line 178: This is the actual fix, swapping funcDecl.Type.TypeParams for findTargetGenericType(file, funcDecl). Makes sense since the trampoline declaration itself already gets its type params from findTargetGenericType (see buildTrampSignature in trampoline.go), so before this the call site and the declaration could disagree whenever the generics lived on the receiver instead of the func. That mismatch is exactly what issue 1218 describes.

tool/internal/instrument/apply_func.go
```diff
@@ -271,7 +272,7 @@ func (ip *instrumentPhase) insertTJump(t *rule.InstFuncRule, funcDecl *dst.FuncD
 	// the context, handles exceptions, etc, and ultimately jumps to the real
 	// hook code. By inserting trampoline-jump-if at the target function entry,
 	// we can intercept the original function and execute before/after hooks.
-	tjump := createTJumpIf(t, funcDecl, args, retVals)
+	tjump := createTJumpIf(ip.target, t, funcDecl, args, retVals)
```
darkraider01, line 275: Minor thing, not a blocker. findTargetGenericType does a dst.Clone on every call, and it gets called again here for every rule that matches the same funcDecl (insertTJump can run more than once per function per the comment above about multiple func rules). Since this only runs at codegen time on a handful of AST nodes it's not going to matter in practice, just noting it in case it ever gets called somewhere hotter.

tool/internal/instrument/testdata/golden/generic-functions/generic_functions.main.go.golden
```diff
@@ -21,9 +21,9 @@ type GenStruct[T any] struct {
 
 func (g *GenStruct[T]) GenericMethod(p1 T, p2 string) (_unnamedRetVal_1139503255_0 T, _unnamedRetVal_1139503255_1 error) {
 	//line <generated>:1
-	if hookContext1139503255, _ := OtelBeforeTrampoline_GenericMethod1139503255(&g, &p1, &p2); false {
+	if hookContext1139503255, _ := OtelBeforeTrampoline_GenericMethod1139503255[T](&g, &p1, &p2); false {
```
darkraider01, line 24: Confirmed this locally, built the branch and ran the instrument test suite plus go vet, all green, and the golden output here matches what I'd expect. Type arg [T] now shows up on both the before and after trampoline calls, which lines up with the trampoline declaration already carrying [T any]. Good fix.

Labreo:

SSOHEB:
> Just a suggestion for the future.Please follow the [pull request template](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/.github/PULL_REQUEST_TEMPLATE.md).It helps contributors and maintainers get a better idea for the pull request.And allows them to manage and merge your pull request faster.

thanks for reminding me of this, i ve edited my PR description i hope it follows the template rightly now

tool/internal/instrument/testdata/golden/generic-functions/generic_functions.main.go.golden
SSOHEB: Thanks for the detailed review and for testing this locally,  I ll keep the performance note in mind. Really appreciate it 🙌

y1yang0 (approved):

PR #1230, opened 2026-08-24

### LINKED ISSUE #766: fix(ast): stripGenericTypes fatals on qualified pointer receivers

VedantMadane:
## Summary

## Changes

- SelectorExpr *pkg.Type aborted via Unimplemented.
- Strip qualifier; unknown shapes are non-matches.

Fixes #766

### DIFF
--- a/tool/internal/ast/shared.go
+++ b/tool/internal/ast/shared.go
@@ -51,32 +51,55 @@ func stripGenericTypes(recvTypeExpr dst.Expr) string {
 	switch expr := recvTypeExpr.(type) {
 	case *dst.StarExpr: // func (*Recv)T or func (*Recv[T])T
 		// Check if X is an Ident (non-generic) or IndexExpr/IndexListExpr (generic)
+		// or a qualified SelectorExpr: *bufio.Writer
 		switch x := expr.X.(type) {
 		case *dst.Ident:
 			// Non-generic pointer receiver: *MyStruct
 			return "*" + x.Name
+		case *dst.SelectorExpr:
+			// Qualified pointer receiver: *pkg.Type — match on the type name only
+			if x.Sel != nil {
+				return "*" + x.Sel.Name
+			}
 		case *dst.IndexExpr:
 			// Generic pointer receiver with single type param: *GenStruct[T]
+			// or *pkg.GenStruct[T]
 			if baseIdent, ok := x.X.(*dst.Ident); ok {
 				return "*" + baseIdent.Name
+			}
+			if sel, ok := x.X.(*dst.SelectorExpr); ok && sel.Sel != nil {
+				return "*" + sel.Sel.Name
 			}
 		case *dst.IndexListExpr:
 			// Generic pointer receiver with multiple type params: *GenStruct[T, U]
 			if baseIdent, ok := x.X.(*dst.Ident); ok {
 				return "*" + baseIdent.Name
 			}
+			if sel, ok := x.X.(*dst.SelectorExpr); ok && sel.Sel != nil {
+				return "*" + sel.Sel.Name
+			}
 		}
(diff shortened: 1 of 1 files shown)


tool/internal/ast/shared.go
```diff
@@ -101,11 +124,10 @@ func findFuncDecl(root *dst.File, funcName, recv string) *dst.FuncDecl {
 		baseType := stripGenericTypes(recvTypeExpr)
 
 		if baseType == "" {
-			msg := fmt.Sprintf("unexpected receiver type: %T", recvTypeExpr)
-			util.Unimplemented(msg)
-		}
+		return false
```
darkraider01, line 127: Previously an unhandled receiver type would panic via util.Unimplemented, which made it obvious when a new type shape wasn't supported. Now it just returns false silently. Was this intentional? If a future receiver shape isn't handled by stripGenericTypes, we'll just get a quiet "no matching decl found" instead of a clear error pointing at the actual cause. Might be worth keeping some kind of loud failure here, or at least a log line.

Labreo:

dobbydobap:
Heads up, this is the third PR for #766 and #786 is still open on the same
function, so there are two live right now.

More to the point: the qualified receiver shape this handles is not reachable
from source that compiles. func (p.T) M() and func (*p.T) P() are both rejected
by the compiler with "undefined: p" on Go 1.25.12; go/parser accepts them, which
is where the confusion comes from. otelc only parses dependency source out of
the module cache, so it should never see a SelectorExpr receiver.

The shapes that do compile and still fall through are the parenthesised ones,
which @vignesh2027 measured on the issue. His suggestion there, returning a
no-match instead of calling util.Unimplemented, covers those and this one
together, and is still waiting on a maintainer answer.

Worth settling that on #766 before landing another shape-specific fix. Also
this changes shared.go with no test, and the existing behaviour is pinned by
TestStripGenericTypes.

vignesh2027:
Agreeing with @dobbydobap here, and thanks @VedantMadane for digging into this.

I checked the qualified receiver shape again to be sure: the compiler rejects both `func (m sync.Mutex) Foo()` and the pointer form with "cannot define new methods on non-local type", so a `SelectorExpr` receiver can only come from source that does not build. otelc only parses compiling dependency source out of the module cache, so it should not meet that shape in practice. go/parser accepts it, which is where it looks reachable.

The shapes that do compile and still fall through `stripGenericTypes` are the parenthesised ones (`(r (T))`, `(r (*T))`, `(r (G[P]))`, and `*(T)` reaching the StarExpr case). Those are the real gap, though gofmt rewrites most of them away, so they are rare in formatted cache source.

Both the qualified shape and the parenthesised ones are covered by one change: having the no-match path return empty instead of calling `util.Unimplemented`, so an unrecognised receiver means "the rule does not apply here" rather than ending the build. That closes the whole class in one place instead of a branch per shape. I raised it on #766 and it is still waiting on a maintainer call (@NameHaibinZhang @amazingakai), so it is probably worth settling there before landing another shape-specific fix. Happy to put that consolidated change up if it is wanted.

darccio:
Closed in favour of #1257

PR #1236, opened 2026-08-25
fix(tool): reject separated flags without values

### LINKED ISSUE #1235: bug(tool): separated-value flags without values pass setup parsing
### Description

`splitBuildTargets` skips the argument following recognized separated-value flags such as `-exec`, `-run`, and `-o`, but it does not verify that a value is present before advancing the argument index.

As a result, malformed commands such as:

```text
otelc go test ./pkg -exec
otelc go test ./pkg -run
otelc go build ./pkg -o
```

pass setup argument parsing. Setup may then perform unnecessary package-loading and instrumentation work before the underlying Go command rejects the missing value.

For example, the Go tool reports:

```text
flag needs an argument: -run
```

### Expected behavior

`splitBuildTargets` should return an early, focused error when a recognized separated-value flag is the final argument.

### Proposed fix

* Verify that a following argument exists before incrementing the parser index.
* Return a clear missing-value error when no value is present.
* Add table-driven regression cases to `TestSplitBuildTargets` for trailing `-exec`, `-run`, and `-o`.

ADITYA-CODE-SOURCE:
## Description

Reject recognized separated-value flags when no value follows them during build-target parsing.

Focused table-driven cases cover trailing `-exec`, `-run`, and `-o` flags.

## Motivation

Malformed commands such as `otelc go test ./pkg -exec` previously passed setup argument parsing. Setup could perform package-loading and instrumentation work before the Go command eventually reported the missing value.

Fixes #1235

---

## Testing

- `go test -count=1 ./tool/internal/setup -run TestSplitBuildTargets`
- `go test -count=1 ./tool/internal/setup`
- `make verify-manifest`
- `make lint/license-header`
- `./.bin/golangci-lint run --config .tools/golangci.yml ./tool/internal/setup` (reports existing findings on unchanged lines only)
- `git diff --check`

### DIFF
--- a/tool/internal/setup/setup.go
+++ b/tool/internal/setup/setup.go
@@ -219,6 +219,9 @@ func splitBuildTargets(args []string) ([]string, []string, error) {
 
 		if strings.HasPrefix(arg, "-") {
 			if !strings.Contains(arg, "=") && (flagsWithPathValues[arg] || testFlagsWithValues[arg]) {
+				if i+1 >= len(args) {
+					return nil, nil, ex.Newf("flag %q requires a value", arg)
+				}
 				i++ // skip this flag's separate value
 			}
 			continue
--- a/tool/internal/setup/setup_test.go
+++ b/tool/internal/setup/setup_test.go
@@ -156,6 +156,7 @@ func TestSplitBuildTargets(t *testing.T) {
 		fileTargets   []string
 		notPkgTargets []string // must NOT be parsed as packages (e.g. flag values)
 		expectError   bool
+		wantErr       string
 	}{
 		{
 			name:        "all package targets",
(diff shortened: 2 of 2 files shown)


ADITYA-CODE-SOURCE:
The Windows unit-test failure appears unrelated to this change. The new `TestSplitBuildTargets` cases passed on Windows; the pre-existing `TestAcquireBuildLockExcludes` failed while reopening its lock file with `Access is denied`. The macOS unit-test job was then cancelled by matrix fail-fast.

I do not have permission to rerun repository workflows. Could a maintainer rerun the failed unit-test workflow? @y1yang0, once CI is green, could you also review this change?

y1yang0 (approved):

PR #1245, opened 2026-08-26
fix(instrumentation/github.com/anthropics/anthropic-sdk-go): detect SSE responses case-insensitively

RedZapdos123:
## Description:

Anthropic middleware currently checks SSE responses with a case-sensitive `text/event-stream` prefix match. Valid mixed case headers like `Text/Event-Stream` miss the streaming fallback path, so the span is treated as non-streaming and loses `gen_ai.request.is_stream`.

This PR lowercases the response `Content-Type` before the SSE check and adds a regression test for mixed case headers.

## Verifications:
`cd instrumentation/github.com/anthropics/anthropic-sdk-go && go test . -run "TestOtelMiddleware_(SSEResponseFallback|SSEResponseFallbackMixedCaseContentType)" -count=1 -v`

- [x] `cd instrumentation/github.com/anthropics/anthropic-sdk-go && go test ./... -count=1`
- [x] `make package`


om7057 (commented):
Pulled this and ran it, full suite plus `go vet` pass clean, and the mixed-case regression test does what it says. Good, minimal fix for the actual bug.

One thing worth mentioning since it's not this PR's job to fix but is worth knowing about: the exact same case-sensitive `strings.HasPrefix(contentType, "text/event-stream")` check exists untouched in all three openai copies (v1/v2/v3 middleware.go). Same bug; the same fix would apply. 
Just flagging it so it doesn't get missed; I'll open a follow-up for those separately.
cc: @y1yang0

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go
```diff
@@ -505,6 +505,39 @@ func TestOtelMiddleware_SSEResponseFallback(t *testing.T) {
 	assert.Equal(t, sse, string(got))
 }
 
+func TestOtelMiddleware_SSEResponseFallbackMixedCaseContentType(t *testing.T) {
```
atharvamhaske, line 508: This test is a copy of `TestOtelMiddleware_SSEResponseFallback`. The only change is **Content-Type: Text/Event-Stream.** 
A table on the existing test (text/event-stream and Text/Event-Stream) would keep both cases in one place.

That is the useful note. The one-line fix is fine. The new test is a full duplicate.
Can you please fix it ?

RedZapdos123:
Superseded by #1320, which fixes case insensitive exact SSE media type detection and rejects prefix false positives across Anthropic and OpenAI instrumentations.

instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go
RedZapdos123: Thanks for the review. Though this PR was superseded by #1320.

PR #1250, opened 2026-08-26
feat(instrumentation): add compile-time instrumentation for gofiber/fiber/v2 (#1248)

### LINKED ISSUE #1248: New Instrumentation Request for github.com/gofiber/fiber/v2

### Package

github.com/gofiber/fiber/v2

### Version

v2.52.6

### Environment

Go 1.22+ on Linux / macOS / Windows

### Use-case

Building high-performance HTTP web servers and REST APIs using the GoFiber framework.

### Telemetry

HTTP Server spans with route pattern attributes (`http.route`), enriched span names (`METHOD /route/pattern`), and HTTP status code tracking.

### Verification / Test Proof

Unit tests for `gofiber/fiber/v2` instrumentation running against Fiber context hooks:

```text
=== RUN   TestFiberBeforeAndAfterNext
--- PASS: TestFiberBeforeAndAfterNext (0.00s)
=== RUN   TestFiberAfterNext_ErrorStatus
--- PASS: TestFiberAfterNext_ErrorStatus (0.00s)
PASS
ok  	go.opentelemetry.io/otelc/instrumentation/github.com/gofiber/fiber/v2	0.901s


### Tip

### PR DESCRIPTION
**Pull request by vedant21-oss**
## Description

Adds compile-time zero-code instrumentation for the GoFiber web framework (`github.com/gofiber/fiber/v2`).

- Adds AST rule definition `otelc.yaml` targeting `(*Ctx).Next()` in `github.com/gofiber/fiber/v2`.
- Implements `BeforeNext` and `AfterNext` context hooks for `*fiber.Ctx` to enrich HTTP server spans with route names (`METHOD /route`) and HTTP status code tracking.
- Adds unit test suite in `context_hook_test.go`.
- Adds E2E test application in `test/apps/fiberserver/`.
- Adds integration test suite in `test/integration/fiberserver_test.go`.

## Motivation

Fiber (`github.com/gofiber/fiber/v2`) is a widely used high-performance web framework in Go. Adding compile-time instrumentation enables automatic span creation and HTTP attribute enrichment for Fiber servers compiled with `otelc`.

Fixes #1248

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### CODE REVIEW

File: instrumentation/github.com/gofiber/fiber/v2/context_hook.go [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+
+	c.Locals(nextDepthKey, 0)
+
+	span := trace.SpanFromContext(c.UserContext())
+	if !span.IsRecording() {
+		return
+	}
+
+	status := c.Response().StatusCode()
+	if status >= 400 {
```
darkraider01: The status code attribute only gets set when status is >= 400, so successful requests never get an `http.response.status_code` attribute at all. Looks like the SetAttributes call ended up nested inside the error check by mistake.

```go
status := c.Response().StatusCode()
if status >= 500 {
    span.SetStatus(codes.Error, "")
}
span.SetAttributes(semconv.HTTPResponseStatusCodeKey.Int(status))
```

That way the status code gets recorded on every response, and only the error status gets set for 5xx, matching how the other instrumented libraries here do it.

vedant21-oss:
Thanks for catching that! Updated AfterNext so http.response.status_code gets recorded on every response, and codes.Error status is set for 5xx responses.

instrumentation/github.com/gofiber/fiber/v2/context_hook.go
vedant21-oss: This is already fixed, SetAttributes is outside the status check now so every response gets http.response.status_code, and only 5xx sets the error status. Also merged main in to fix the bundle conflict.

vedant21-oss:
Heads up, I've changed the approach in this PR rather than patching it, because the integration test was failing for a reason I couldn't fix at the old hook point.

What was wrong. The hooks were on (*fiber.Ctx).Next and they only decorated a span — trace.SpanFromContext(c.UserContext()), then bail if !span.IsRecording(). That assumes something upstream already started a server span. Nothing does. Fiber runs on fasthttp, so net/http's server instrumentation (which hooks net/http.serverHandler.ServeHTTP) never sees the request, and there's no fasthttp instrumentation in the repo. So SpanFromContext handed back the no-op span, IsRecording() was false, and both hooks returned on the first branch every time. TestFiberServer was failing with timeout waiting for 1 span(s), collector has 0.

I reproduced it outside CI to be sure: built test/apps/fiberserver through otelc (the rule matched fine, 4 injections in the debug log), ran it against a local collector. Request returned 200, /v1/metrics arrived, /v1/traces never did.

There was a second problem underneath. (*Ctx).Next is also the wrong place to hang route naming. Fiber calls a matched route's handler directly from (*App).next (router.go:143-145), so a route with one handler and no middleware never calls c.Next(). The test app has no middleware, so that hook would have seen no traffic even if there had been a span to decorate.

What it does now. Both hooks moved to (*App).next, which (*App).handler calls exactly once per request, and the span is created there as a real SpanKindServer span with W3C context extracted from the incoming headers. Three things worth flagging:

The route pattern doesn't exist yet when the hook starts — (*App).next assigns c.route only once it has matched. So the span opens named for the method and is renamed to {method} {route} in the after-hook, which is the shape the HTTP conventions ask for.
The router re-enters itself when middleware calls c.Next(), so a depth counter in Locals keeps the outermost frame as the span's owner. Without it you get one span per middleware layer — there's a test that fails if the guard is removed.
Status comes from the returned error, not from c.Response().StatusCode(). (*App).next hands its error up to (*App).handler, which only then runs the error handler and writes a status, so inside this hook the response still reads 200. Deriving it from the error keeps 404, 405 and handler failures from being reported as successes.
No new dependencies — the propagation carrier goes through *fiber.Ctx rather than naming *fasthttp.RequestHeader, so fasthttp stays indirect.

Unit tests cover the plain route, the middleware case, handler errors, *fiber.Error status mapping, the instrumentation being disabled, context propagation in, and the span being visible to the handler. I checked they actually bite: removing the depth guard fails TestMiddlewareProducesExactlyOneSpan, and taking the status from the response instead of the error fails TestFiberErrorKeepsItsOwnStatus.

govulncheck also flagged five advisories reachable from the pinned versions — four in fiber and one in the fasthttp it pulls in — so I've bumped to fiber v2.52.13 / fasthttp v1.70.0. (*App).next is unchanged there, and I re-verified the span after the bump. Integration tests now pass on all four platforms.

Happy to split this into a separate PR if reviewers would rather see the approach change on its own.

PR #1252, opened 2026-08-26
chore(tool/internal/instrument): resolve TODO on HookContextImpl meth…

darkraider01:
This closes out the TODO from [#1173](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/1173): it looked like the four panicking HookContextImpl methods (GetParam/SetParam/GetReturnVal/SetReturnVal) should get removed once a generic target has no After hook. Turns out they can't, the Before hook takes its context as a hook.HookContext interface, so the concrete type has to keep implementing all four or the generated code won't compile. For generics they already panic unconditionally regardless of After, on purpose, as a guard.

So: swapped the TODO for a comment explaining why, and added a golden test (generic-before-only) covering that exact case.

No behavior change, just a comment + a regression test.


tool/internal/instrument/optimize.go
```diff
@@ -364,7 +364,19 @@ func (ip *instrumentPhase) optimizeTJumps() error {
...
+		// generic target where rewriteHookContextMethods already makes all four
+		// panic unconditionally (regardless of whether an After hook exists).
+		// *HookContextImpl{suffix} is passed to the Before hook as a
+		// hook.HookContext interface value (see checkHookDecl), so it must keep
+		// satisfying the full interface or the generated code stops compiling.
+		// The panicking stubs also double as an intentional runtime guard: a
+		// hook author who calls one of these methods on a generic target gets
+		// an immediate, clear panic instead of silently wrong behavior. See
+		// testdata/golden/generic-before-only/hook.go for the documented
+		// workaround (read values from the hook's own positional parameters).
```
y1yang0: I suggest simply removing TODO without other comments

y1yang0 (approved):
Otherwise LGTM, thanks for adding extra test

tool/internal/instrument/optimize.go
darkraider01: Done, dropped the TODO with no replacement comment in cc78cb3.

PR #1253, opened 2026-08-26
fix(instrumentation/openai): streaming spans report gen_ai.usage.* = 0 when the response carries no usage object

### LINKED ISSUE #1251: bug(instrumentation/openai): streaming spans report gen_ai.usage.* = 0 when the response carries no usage object
### Describe the bug

`finalize` in the shared streaming reader always sets `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens` and `gen_ai.usage.total_tokens` on the span. The fields behind them only get assigned in `processChatChunk` and `processCompletionChunk`, and only when a chunk actually carries a `usage` object. When none does, all three sit at zero and get published anyway:

```go
r.span.SetAttributes(
	genAIResponseFinishReasonsKey.StringSlice(r.reasons),
	genAIUsageInputTokensKey.Int64(r.inputTokens),
	genAIUsageOutputTokensKey.Int64(r.outputTokens),
	genAIUsageTotalTokensKey.Int64(r.totalTokens),
)
if r.id != "" {
	r.span.SetAttributes(genAIResponseIDKey.String(r.id))
}
```

The same function already knows how to leave an attribute off when it has no value for it. `gen_ai.response.id` and `gen_ai.response.model` are both guarded that way, a few lines down. The token counts aren't, and those are the ones people add up on a cost dashboard.

v1, v2 and v3 all share this reader, so it affects all three.

### To Reproduce

Worth saying up front that this is the default path, not an edge case. OpenAI doesn't put `usage` in streaming chunks unless the request sets `stream_options: {"include_usage": true}`, and nothing in this repo sets it. `grep -rn "include_usage\|stream_options" .` comes back empty, and `parseChatRequest` never looks at either field.

1. Call `client.Chat.Completions.NewStreaming(...)` with default options.
2. The response comes back as `text/event-stream`, so `OtelMiddleware` wraps the body in a `StreamingReader`.
3. None of the chunks carry `usage`.
4. The span ends with all three token counts at 0.

Any OpenAI-compatible gateway that doesn't echo `usage` back hits the same thing.

`TestStreamingReader_EmptyStream` and `TestStreamingReader_CloseBeforeRead` already exercise this path, but they only assert that one span was recorded and never look at its attributes.

### Expected behavior

The three attributes should be left off when no usage object showed up. A zero gets averaged into cost-per-request panels without anyone noticing. A missing attribute is obvious.

The semconv version this repo pins (v1.37.0) has these at `requirement_level: recommended` and describes `gen_ai.usage.input_tokens` as the number of prompt tokens *as reported in the usage `prompt_tokens` property of the response*. With no usage property there's nothing to report.

One thing worth being careful about: the check needs to be whether a usage object arrived, not `if tokens > 0`. A response that genuinely reports `completion_tokens: 0` is still a measurement and should be recorded. Guarding on `> 0` just swaps one version of the absent-vs-zero mixup for another.

`gen_ai.response.finish_reasons` has a smaller version of the same problem, since an empty stream publishes an empty slice. Probably worth fixing in the same change.

Anshumancanrock:
## Description

`finalize` set `gen_ai.usage.input_tokens`, `output_tokens` and `total_tokens` on every streaming span, but those fields only get assigned when a chunk carries a `usage` object. When none did, all three were published as zero.

They're now only set when usage actually arrived. `processChatChunk` and `processCompletionChunk` take `usage` as a pointer so an absent object stays distinguishable from one reporting real zeros, and both go through a small `recordUsage` helper. The guard is "did usage arrive" rather than `if tokens > 0`, so a response genuinely reporting `completion_tokens: 0` is still recorded.

Two small things in the same function: `gen_ai.response.finish_reasons` is skipped when there are no reasons, and the attributes are built into one slice so the span is touched once instead of four times.

Tests cover a stream with no `usage` field, `"usage":null`, the completion variant, a usage object reporting a real zero, and a partial usage object not erasing earlier counts. The bundle is regenerated because it embeds `instrumentation/`.

## Motivation

OpenAI doesn't send `usage` on a streaming response unless the request sets `stream_options: {"include_usage": true}`, and nothing here sets it. So on the default streaming path every span reported zero tokens. A zero gets quietly averaged into cost-per-request dashboards; a missing attribute is visible.

`finalize` already skipped `gen_ai.response.id` and `gen_ai.response.model` when it had no value for them, so this brings the token counts in line with the rest of the function. Semconv has these at `requirement_level: recommended`, defined as what the response reported.

Fixes #1251 

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


om7057 (commented):
I was working on a similar issue. I ran this properly, though. Full suite plus `go vet` came back clean.

Since the whole point of this PR is a "was it there or wasn't it" distinction, I wanted to see it fail before I trusted it. Swapped `streaming.go` and `attrs.go` back to what's currently on main while keeping the new tests, and `TestStreamingReader_UsageOmittedWhenNotReported` failed on all three of its cases exactly as expected; attributes came back as 0 instead of getting omitted. Put the fix back, and it passed cleanly. So this is a real regression test, not just something asserting what the new code already does.

The `sawUsage` flag plus switching `Usage` to a pointer is the right way to tell "the field showed up reporting zero" apart from "the field never showed up," and it's applied the same way in both the chat and completion paths, which matters since this is the shared reader both go through. The partial-usage-merge behavior (`if input > 0` before overwriting) was already there before this PR and isn't something this change touches, so no new risk there.

Solid fix.

PR #1255, opened 2026-08-26
fix(pkg/runtime): ignore parsed-empty enabled instrumentation list

### LINKED ISSUE #1254: bug(pkg/runtime): parsed-empty OTEL_GO_ENABLED_INSTRUMENTATIONS disables all instrumentation
## Description:

`pkg/runtime.Instrumented` checks `OTEL_GO_ENABLED_INSTRUMENTATIONS` with a raw `!= ""` test before parsing it. Values that parse to an empty allowlist after trimming and dropping empty entries still activate allowlist mode, so every instrumentation returns `false`.

## Environment:

- OS: Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 GNU/Linux
- Go Version: go1.26.4 linux/amd64
- Version: `main` at `4873b48`

## Steps to Reproduce:

1. Check out `main` at `4873b48`.
2. From WSL in the repository root, run:
   ```bash
   cat >/tmp/repro-empty-allowlist.go <<'"'"'EOGO'"'"'
   package main

   import (
       "fmt"
       "os"

       runtime "go.opentelemetry.io/otelc/pkg/runtime"
   )

   func main() {
       cases := []struct {
           name string
           val  *string
       }{
           {name: "unset", val: nil},
           {name: "spaces", val: strPtr("   ")},
           {name: "commas", val: strPtr(",, ,")},
           {name: "normal allowlist", val: strPtr("nethttp")},
       }

       for _, tc := range cases {
           os.Unsetenv("OTEL_GO_ENABLED_INSTRUMENTATIONS")
           os.Unsetenv("OTEL_GO_DISABLED_INSTRUMENTATIONS")
           if tc.val != nil {
               _ = os.Setenv("OTEL_GO_ENABLED_INSTRUMENTATIONS", *tc.val)
           }
           fmt.Printf("%s: env=%q nethttp=%v grpc=%v\n", tc.name, os.Getenv("OTEL_GO_ENABLED_INSTRUMENTATIONS"), runtime.Instrumented("nethttp"), runtime.Instrumented("grpc"))
       }
   }

   func strPtr(s string) *string { return &s }
   EOGO

   cd pkg/runtime
   go run /tmp/repro-empty-allowlist.go
   ```

## Actual Behavior:

Whitespace-only or commas-only values disable every instrumentation instead of behaving like an empty allowlist input.

Example output:
```text
unset: env="" nethttp=true grpc=true
spaces: env="   " nethttp=false grpc=false
commas: env=",, ," nethttp=false grpc=false
normal allowlist: env="nethttp" nethttp=true grpc=false
```

## Expected Behavior:

Values that parse to an empty allowlist should behave like unset, so whitespace-only or commas-only `OTEL_GO_ENABLED_INSTRUMENTATIONS` should not disable all instrumentations.

## Notes:

This comes from `Instrumented` checking the raw env var before calling `parseInstrumentationList`:

- `enabledList := os.Getenv("OTEL_GO_ENABLED_INSTRUMENTATIONS")`
- `if enabledList != "" { ... }`

`parseInstrumentationList` already trims whitespace and drops empty elements, so inputs like `"   "` and `",, ,"` become an empty slice after parsing, but allowlist mode stays active.

The current docs describe `OTEL_GO_ENABLED_INSTRUMENTATIONS` as a comma-separated allowlist in `docs/configuration.md`, and the existing `pkg/runtime` tests cover case-insensitive names and spaces around valid entries but not parsed-empty inputs.

Overlap check: `gh search issues` / `gh search prs` did not find a direct existing report or PR for this parsed-empty allowlist behavior.

RedZapdos123:
## Description:

Issue #1254 reported that parsed empty `OTEL_GO_ENABLED_INSTRUMENTATIONS` values like whitespace only or commas only still activated allowlist mode and disabled every instrumentation.

Current `main` already trims and drops empty entries in `parseInstrumentationList` but `Instrumented` checked the raw env var before parsing it, so parsed empty inputs were still treated as a set allowlist.

This PR parses the enabled list before gating, adds regression tests for whitespace only and commas only inputs, and keeps `TestInstrumented` isolated from ambient `OTEL_GO_*` env vars.

Closes #1254.

## Verifications:
- [x] `cd pkg/runtime && go test . -run TestInstrumented -count=1 -v`
- [x] `cd pkg/runtime && go test . -count=1`
- [x] `cd pkg/runtime && go test ./... -count=1`
- [x] `make package`


y1yang0 (approved):

dobbydobap:
Rebasing will not clear this one, so worth stopping before another round.

I unpacked your committed archive against a freshly generated one on your current head. The
content is byte identical, including your fix in pkg_temp/runtime/otel_setup.go, and so are the
entry names, sizes, mtimes and gzip framing. The only difference is 458 tar headers: every regular
file is -rwxr-xr-x in the committed archive and -rw-r--r-- when regenerated on ext4.

That is your own #1269, so make package cannot produce a matching archive from that checkout no
matter how many times you run it. #1272 is the fix and it has not run CI once since the 28th,
every workflow on it is still action_required. That is what needs unsticking, not this PR.

PR #1256, opened 2026-08-26
fix(instrumentation/net/http): redact sensitive query parameters from url.full and url.query

### LINKED ISSUE #1243: bug(instrumentation/net/http): url.full and url.query export sensitive query parameters unredacted
## Describe the bug

The net/http instrumentation exports sensitive query-parameter values verbatim, which the OpenTelemetry URL semantic conventions say should be redacted by default.

- Client spans set `url.full` from `req.URL.String()` (`instrumentation/net/http/semconv/client.go`). The code already strips userinfo (`req.URL.User`), but the query string is included unredacted, so signing keys and tokens in the query survive.
- Server spans set `url.query` directly from `req.URL.RawQuery` (`instrumentation/net/http/semconv/server.go`), with no redaction at all.

The [URL semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/) state that query string values for a known set of keys "SHOULD be redacted by default and replaced by the value `REDACTED`", that key matching "SHOULD be case-sensitive", and that "when a query string value is redacted, the query string key SHOULD still be preserved". The default set is:

- `X-Amz-Signature`
- `X-Amz-Credential`
- `X-Amz-Security-Token`
- `sig`
- `X-Goog-Signature`

These appear in presigned URLs (AWS S3, Google Cloud Storage) and OAuth-style flows, so an application that calls a presigned URL through an instrumented `http.Client`, or a server that receives one, ships the signing credential straight to its trace backend today.

## Evidence

`client.go`, url.full construction:
```go
userinfo := req.URL.User
req.URL.User = nil
u = req.URL.String()   // query string still raw, e.g. ...&X-Amz-Signature=<real signature>
req.URL.User = userinfo
```

`server.go`:
```go
if req.URL != nil && req.URL.RawQuery != "" {
    attrs = append(attrs, semconv.URLQuery(req.URL.RawQuery))  // raw, unredacted
}
```

A grep of the package for `redact`/`sensitive`/`scrub` finds only the userinfo handling above; there is no query-parameter redaction anywhere.

## Expected behavior

`url.full` and `url.query` should have the values of the sensitive keys above replaced with `REDACTED`, case-sensitive, with the keys and all other parameters preserved, matching the URL semantic conventions and the behavior of opentelemetry-go-contrib`s otelhttp. For example `?color=blue&sig=abc123` becomes `?color=blue&sig=REDACTED`.

This is net/http only and self-contained; it does not change any other instrumentation. Happy to send the fix, with tests for both the client and server paths.

vignesh2027:
## Description

The net/http instrumentation exported sensitive query-parameter values verbatim. Client spans set `url.full` from `req.URL.String()` and server spans set `url.query` from `req.URL.RawQuery`, both with the query string unredacted. The client path already stripped userinfo, but not the query string, so signing credentials and tokens carried in query parameters survived.

The [URL semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/) name a default set of query parameter keys whose values SHOULD be redacted and replaced with `REDACTED`, with case-sensitive key matching and the key preserved. These appear in presigned URLs (AWS S3, Google Cloud Storage) and similar flows, so an application calling a presigned URL through an instrumented client, or a server receiving one, shipped the signing credential to its trace backend today.

This change adds a `redactQuery` helper in the shared semconv `util.go` that replaces the values of the sensitive keys with `REDACTED` while preserving key order and every other parameter, and applies it to both `url.full` (client) and `url.query` (server). A query string with no sensitive keys is returned unchanged, so the common path allocates nothing.

Redacted keys, matching the conventions: `X-Amz-Signature`, `X-Amz-Credential`, `X-Amz-Security-Token`, `sig`, `X-Goog-Signature`.

Example: `?color=blue&sig=abc123` becomes `?color=blue&sig=REDACTED`.

## Motivation

Credentials and tokens in query strings should not be copied into spans and shipped to trace backends by default. This aligns net/http with the URL semantic conventions and with opentelemetry-go-contrib's otelhttp, and it is self-contained to net/http (no other instrumentation changes).

## Tests

- `TestRedactQuery` covers each sensitive key, non-sensitive preservation, key-order preservation, case-sensitivity, bare keys, empty values, and repeated keys.
- `TestHTTPClientRequestTraceAttrs` and `TestHTTPServerRequestTraceAttrs` gain cases asserting `url.full` and `url.query` are redacted end to end.

Fixes #1243

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


instrumentation/net/http/semconv/util.go
```diff
@@ -217,3 +217,46 @@ func SpanMethod(method string) string {
...
+// with no sensitive keys is returned unchanged so the common case allocates
+// nothing.
+func redactQuery(rawQuery string) string {
+	if rawQuery == "" {
+		return rawQuery
+	}
+	parts := strings.Split(rawQuery, "&")
+	redacted := false
+	for i, part := range parts {
+		key, _, found := strings.Cut(part, "=")
```
ADITYA-CODE-SOURCE, line 249: Could we percent-decode the parameter name before checking `sensitiveQueryParams`? For example, `s%69g=secret` is equivalent to `sig=secret`, but comparing the raw key would leave the sensitive value unredacted. We could decode the key only for matching while preserving its original form in the output, and add a regression test for this case.

vignesh2027: Good catch, and it is reachable. Go's own parser reads `s%69g=secret&X-Amz-Signature=abc` as `map[X-Amz-Signature:[abc] sig:[secret]]`, so the raw key comparison was letting an encoded key carry its value through untouched.

Fixed the way you suggested. The key is decoded with `url.QueryUnescape` for the lookup only, and the original spelling is written back, so `s%69g=secret` becomes `s%69g=REDACTED` and nothing but the value changes. A key that fails to decode is compared as written rather than dropped.

Added five regression cases: a partially encoded key, an encoded AWS key, a fully encoded one, an undecodable key alongside a real one, and `S%49G` to confirm decoding does not quietly make matching case insensitive.

I checked the tests actually catch it rather than assuming: reverting to raw matching fails the first three with the secret still in place, for example `expected: "s%69g=REDACTED", actual: "s%69g=secret"`.

PR #1257, opened 2026-08-26
fix(tool/internal/ast): match parenthesized receivers and stop aborting on unrecognized ones

### LINKED ISSUE #766: fix(ast): stripGenericTypes fatals on qualified pointer receivers

vignesh2027:
## Description

This is the consolidated fix for #766 discussed on the issue: instead of handling one receiver shape at a time, it recognizes the reachable parenthesized forms and makes the matcher degrade to no match on anything still unrecognized, in one place.

`stripGenericTypes` recognized plain, pointer, and generic receivers, but returned `""` for parenthesized forms like `(r (T))`, `(r (*T))`, and `*(T)`. These compile (verified on go1.25) and appear in unformatted dependency source, which is what otelc parses out of the module cache. `findFuncDecl` then treated an empty result as `util.Unimplemented`, an `ex.Fatalf` that ended the user's whole build over a receiver an unrelated rule was not even targeting.

Two changes in `tool/internal/ast/shared.go`:

1. `stripGenericTypes` now recurses through `StarExpr` and has a `ParenExpr` case, so `(r (T))`, `(r (*T))`, `(r (G[P]))`, and `*(T)` resolve to the same base type name as their unparenthesized equivalents. Existing behavior for plain, pointer, and generic receivers is unchanged (pinned by `TestStripGenericTypes`).
2. `findFuncDecl` returns no match instead of calling `util.Unimplemented` when a shape is still unrecognized. A receiver the matcher does not understand is one the rule is not selecting, so it should not end the build.

Together this covers the parenthesized forms (the reachable gap), the qualified receiver named in #766 (which does not compile and so only ever comes from non-building source), and any future shape, without a branch per shape. It supersedes the shape-specific approaches in #786 and #1230.

## Tests

- `TestStripGenericTypes` gains the parenthesized value, pointer, generic, and pointer-to-parenthesized cases, plus pointer-to-unrecognized yielding empty.
- `TestFindFuncDeclParenthesizedReceiver` asserts `(r (T))` and `(r (*T))` now match their base type end to end.
- `TestFindFuncDeclUnrecognizedReceiverNoMatch` asserts an unrecognized receiver is a no match rather than a build abort.

Fixes #766

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


Labreo:
Note for maintainers: This PR supersedes #786 (inactive) and #1230 by implementing the consolidated 'return no-match + ParenExpr' approach discussed in issue #766.

dobbydobap:
Went through this properly since it is the file I worked on in #887. The parenthesised
forms do compile on go1.25.12, the branch merges clean into main at 403d04b and builds,
and the ast package is green.

Two things.

The design question you flagged on #766 may already be answered. #957 made this exact
change in the sibling matcher on 5 Aug: parsedTypeName.matches used to call
util.Unimplemented in its default branch and now returns false, on the argument in #954
that a matcher should report no match rather than end the build. After this PR, shared.go
holds the last util.Unimplemented in the package, so this reads as finishing that job
rather than opening a new question.

The other is TestFindFuncDeclUnrecognizedReceiverNoMatch. It does not reach the guard.
parseSharedFixture only contains func (s *MyStruct) Method(), so baseType is never empty
and the test passes through the ordinary no-match path. Put util.Unimplemented back on the
line the test is meant to pin and it still passes. That is also the line codecov flags as
the one miss.

Parsing a source with an unrecognised receiver does pin it. go/parser accepts a qualified
receiver even though the compiler rejects it, which makes it the easiest shape to reach:

func TestFindFuncDeclUnrecognizedReceiverNoMatch(t *testing.T) {
	file, err := NewAstParser().ParseSource(`package main

func (r pkg.T) Method() error { return nil }
`)
	require.NoError(t, err)
	assert.Nil(t, findFuncDecl(file, "Method", "T"))
}

That passes on your branch, and with util.Unimplemented restored it exits the test binary
at shared.go:107, which is the behaviour worth locking in.

TestFindFuncDeclParenthesizedReceiver is fine as is, it genuinely fails without the fix.

om7057 (commented):
I was looking into the same bug and came across this PR, so I went through `e08227f` this commit and the related code. `go test ./tool/internal/ast/...` passes, and the `stripGenericTypes` change looks good. Handling `StarExpr` and `ParenExpr` also makes the receiver matching simpler.

I noticed a few things that I was thinking to implement in my approach :

**1. The new test doesn't hit the guard it's testing.**

`TestFindFuncDeclUnrecognizedReceiverNoMatch` still returns false through the normal `baseType == recv` check because the fixture uses `*MyStruct`. To cover `if baseType == ""`, it needs a hand-built `dst.FuncDecl` with a `*dst.SelectorExpr` receiver, similar to the approach in #766.

**2. Is silent no-match intentional?**

The guard should only be reachable for invalid/unrecognized receiver syntax now, so returning no-match seems reasonable. But it does mean a rule can silently stop matching. Is that the intended behavior, or should this be surfaced somehow?

om7057:
@dobbydobap already covered both points, and the #957 / #954 history answers my question 2, so I'll fold this in rather than repeat it.

Two things I can add:

I confirmed the parser-based test reaches the guard. `func (r pkg.T) Method()` takes `stripGenericTypes` to `""`, and `shared.go:106` goes from 0 hits to 1, while the current `TestFindFuncDeclUnrecognizedReceiverNoMatch` leaves it at 0 even with the whole `internal/ast` suite running.

On the safety of the no-match: `stripGenericTypes` returning `""` can only make the predicate false, and a real match still requires `baseType == recv`, so an unrecognized receiver can never produce a wrong match, only a skipped rule. That supports treating it as no-match rather than a build abort.

darccio:
Reviewing.

tool/internal/ast/shared_test.go
```diff
@@ -884,3 +909,44 @@ func TestRenderNode(t *testing.T) {
...
+// TestFindFuncDeclUnrecognizedReceiverNoMatch covers the graceful degradation: a
+// receiver shape stripGenericTypes does not recognize returns no match rather
+// than ending the build. A qualified receiver does not compile and so never
+// reaches the tool from real source, but the guard must still not abort.
+func TestFindFuncDeclUnrecognizedReceiverNoMatch(t *testing.T) {
+	file := parseSharedFixture(t)
+
+	// Drive the guard directly with a synthetic qualified receiver, the shape
+	// that used to trigger util.Unimplemented.
+	fnDecl := findFuncDecl(file, "Method", "pkg.MyStruct")
```
darccio: `TestFindFuncDeclUnrecognizedReceiverNoMatch` uses the existing fixture method whose receiver is `*MyStruct`. Therefore, `stripGenericTypes` returns `"*MyStruct"`, not `""`. The test returns `nil` only because it compares `"*MyStruct"` with `"pkg.MyStruct"`.

As a result, the test never reaches the new guard at `tool/internal/ast/shared.go:106` and would still pass if the fatal behavior were restored.

Construct or parse a function with an unrecognized receiver AST, such as a `dst.SelectorExpr`, before calling `findFuncDecl`. For example:

```go
file, err := NewAstParser().ParseSource(`package main

func (r pkg.T) Method() error { return nil }
`)
require.NoError(t, err)

assert.Nil(t, findFuncDecl(file, "Method", "T"))
```

darccio (changes requested):

dobbydobap:
One more datum now this is being reviewed, on whether the guard closes the class or only the three
shapes it enumerates.

I put a fuzzer over findFuncDecl, feeding arbitrary source against arbitrary Func, Recv and Param
values. On current main it dies on the second seed before the fuzzer finishes gathering baseline
coverage:

    [0] Unimplemented: unexpected receiver type: *dst.ParenExpr
    [2] tool/internal/ast/shared.go:108  findFuncDecl.func1

With this branch merged into main at 434baec it runs 467,179 executions with no build-ending path
and no crashers. A second target over MatchesTypeName came back clean at 419,119, which is the
sibling #957 already closed the same way.

Happy to send the target as a follow-up once this lands. It cannot go in before, since on main it
is just a failing test. docs/testing.md asks for this invariant by name, and it would need one line
in tool/cmd/check-test-names/allowlist.go the way directive_fuzz_test.go has.

tool/internal/ast/shared_test.go
vignesh2027: You are right, and thanks for catching it. The test drove `findFuncDecl` against the fixture method, whose receiver is `*MyStruct`. `stripGenericTypes` recognises that shape and returns `*MyStruct`, so the call returned nil from the name comparison against `pkg.MyStruct` and never reached the guard. It would have passed with the fatal behaviour still in place, which makes it worthless as a regression test.

Rewritten along the lines you suggested: it now parses a qualified receiver, a shape `stripGenericTypes` does not recognise, and asks for the function name and receiver base name that declaration would match on, so nil can only come from the guard rather than from a mismatch.

I also checked it is no longer vacuous rather than assuming. Temporarily putting `util.Unimplemented` back in place of the guard makes the new test fail with `Unimplemented: unexpected receiver type`; with the guard it passes.

Issue #1242, opened 2026-08-25
Compile-time performance: several redundant subprocess spawns and re-parsing in the setup/toolexec pipeline


darkraider01:
I went through the `otelc` setup and toolexec pipeline (`tool/internal/{setup,instrument,pkgload,ast}`) looking for compile-time performance issues, and found a handful of spots where we're doing more subprocess spawning and re-parsing than we need to. None of these are correctness bugs, the tool works fine, but they add real wall-clock time, especially on larger projects and in CI. Opening this to get feedback on priority before I send PRs, since a couple of these touch performance-sensitive code.

1. `pkgload.ResolvePackageName` (`tool/internal/pkgload/pkgload.go`) spawns an uncached `go list` subprocess for every unaliased import, and `imports.FindNew` → `AddToFile` (`tool/internal/imports/imports.go`) end up parsing the same file's imports twice, plus another subprocess per newly-added import in `createSpec`. A file with 15 imports can trigger 30-50 subprocess spawns. Fix: cache by import path and build flags, and pass the already-parsed import mapping from `FindNew` into `AddToFile` instead of re-parsing.

2. `ast.ParseFileFast` (`tool/internal/ast/parser.go`) still runs full `dst.DecorateFile` even though the setup-phase rule matching (`tool/internal/setup/match.go`) only reads the AST and never writes it back. Fix: a read-only matcher over plain `go/ast` for the matching path, saving `dst` decoration for where code actually gets rewritten.

3. `findHookFile` and `getHookFunc` (`tool/internal/instrument/trampoline.go`) re-list and re-parse the hook package directory with zero caching, called once per trampoline and again during `optimizeTJumps`. `materializeTemplate` also re-parses the static `impl.tmpl` string from scratch for every single func rule instead of once.

4. Every `otelc toolexec compile` subprocess (one per package in the build) fully reads and JSON-unmarshals `matched.json` before checking if it even applies, even though most packages in a typical build aren't instrumented. There's already a `// TODO: shared memory across sub-processes is possible` comment acknowledging this.

5. On a project's first `otelc setup`, `findDeps` (a `go build -a -x -n` dry run) runs twice back to back because of how `generatePinnedProjects` intentionally returns an empty result after `syncDeps`. Cold-start only, not per-build, but it adds up in CI without cached state.

[...]

PR #1262, opened 2026-08-27
refactor(tool/internal/setup): batch resolveRulePaths into one call per module dir

darkraider01:
Item 6 from #1242.

resolveRulePaths was calling packages.Load once per import path referenced by matched rules, in a loop. Now it collects all the unique paths up front and does one batched packages.Load call per module dir instead.

Benchmarked with 15 distinct hook paths in one module dir: 3.08s before, 246ms after, about 12.5x. Added a test with two distinct paths to cover the batching (the old test only had one path used by two rules, so it never actually exercised more than one path at a time).


tool/internal/setup/store.go
```diff
@@ -21,70 +21,105 @@ import (
 // has enough context to resolve import paths (module directories). The resolved paths
 // are embedded into the rules and consumed directly during instrumentation.
 func resolveRulePaths(ctx context.Context, matched []*rule.InstRuleSet, moduleDirs map[string]bool) error {
```
y1yang0, line 20: One more observation: I took a quick look at the entire modulesDir flow, and changing it to []string might be better than the current map[string]bool, since all call sites seem to convert it to []string before processing it anyway. Maybe you can address in this PR or in a follow-up PR.

tool/internal/setup/store.go
```diff
@@ -21,70 +21,105 @@ import (
...
-	cache := make(map[string]string)
-
 	dirs := make([]string, 0, len(moduleDirs))
 	for dir := range moduleDirs {
 		dirs = append(dirs, dir)
 	}
 	slices.Sort(dirs)
```
y1yang0: dirs := slices.Sorted(maps.Keys(moduleDirs))

tool/internal/setup/store.go
```diff
@@ -21,70 +21,105 @@ import (
...
 
-	resolve := func(goPath string) (string, error) {
-		if dir, ok := cache[goPath]; ok {
-			return dir, nil
+	pending := make(map[string]bool)
```
y1yang0: Maybe simpler? Something like this

```
func resolveRulePaths(...) {
	paths := make([]string)
	for _, ruleset := range matched {
		for _, fileRule := range ruleset.FileRules {
			paths = append(paths, fileRule.Path)
		}
		for _, funcRule := range ruleset.AllFuncRules() {
			paths = append(paths, funcRule.Path)
		}
	}
	slices.Sort(paths)
	unresolved := slices.Compact(paths)
	resolved := make(map[string]string)
	for _, moduleDir := range slices.Sorted(maps.Keys(moduleDirs)) {
		.... //packages.load
	}

	if len(unresolved) > 0 {
		return ex.Newf...
	}

	for _, ruleset := range matched {
		for _, fileRule := range ruleset.FileRules {
			fileRule.ResolvedPath = resolved[fileRule.Path]
		}
		for _, funcRule := range ruleset.AllFuncRules() {
			funcRule.ResolvedPath = resolved[funcRule.Path]
		}
	}
	return nil
}
```

tool/internal/setup/store.go
darkraider01: Agreed, leaving this as a follow-up since it touches call sites outside this PR.

tool/internal/setup/store.go
darkraider01: Done, matches how add.go already does this elsewhere.

tool/internal/setup/store.go
darkraider01: Done, sort+compact instead of a map, cleaner since we need it sorted anyway.

darkraider01:
i'll wait this to be merged then open a follow up regarding modulesDir flow to avoid rebasing

tool/internal/setup/store.go
```diff
@@ -21,70 +22,97 @@ import (
...
-	resolve := func(goPath string) (string, error) {
-		if dir, ok := cache[goPath]; ok {
-			return dir, nil
+	var pending []string
+	for _, ruleset := range matched {
+		for _, fileRule := range ruleset.FileRules {
+			pending = append(pending, fileRule.Path)
 		}
-
-		var lastErr error
```
y1yang0, line 37: As I commented earlier, I don't think we need to build pending in resolveRulePaths, pass it into resolveImportPaths, construct remaining there, and then iterate. Why not iterate the deduplicated pending directly in resolveRulePaths?

darkraider01: Done, inlined resolveImportPaths directly into resolveRulePaths

tool/internal/setup/store.go
```diff
@@ -21,70 +22,86 @@ import (
...
+			Dir:     moduleDir,
+		}, pending...)
+		if loadErr != nil {
+			lastErr = loadErr
+			continue
+		}
+
+		next := pending[:0]
+		for _, p := range pending {
+			pkg, err := findPackage(pkgs, p)
```
y1yang0, line 72: I'm not sure findPackages adds much value here. Would you mind reverting to the original approach?

darkraider01: The old code called packages.Load with one pattern at a time, so pkgs[0] was trivially the result, there was nothing to order. Once we batch multiple patterns into one call, that stops being safe to assume. packages.Load doesn't guarantee the output is ordered to match the input patterns, and under GOPACKAGESDRIVER (Bazel-style setups route through a custom driver) that's even less guaranteed. If we indexed positionally and that assumption broke, it wouldn't error, it'd silently resolve one rule's path to a different rule's directory.

So findPackage isn't really extra  over the original approach, it's what makes batching safe in the first place. The original approach didn't need it because it never batched.

y1yang0: Sorry, I mean we can inline findPackages

tool/internal/setup/store.go
```diff
@@ -20,71 +21,91 @@ import (
 // This must be done during the setup phase because the instrument phase no longer
 // has enough context to resolve import paths (module directories). The resolved paths
 // are embedded into the rules and consumed directly during instrumentation.
+//
+// All pending paths are collected up front and resolved together, batching
+// them into a single packages.Load call per module dir instead of issuing
+// one call per path.
```
y1yang0: I suggest using simpler comment:
```
// resolveRulePaths resolves the import paths referenced by function and file rules
// to absolute filesystem paths.
```

tool/internal/setup/store.go
darkraider01: oh sure, absolutely

darkraider01: Turned it into a local closure inside resolveRulePaths instead of a fully flat block , a fully inlined version pushed resolveRulePaths to gocognit complexity 34, over this repo's 30 limit. The closure keeps it out of the package as a standalone symbol while resetting the nesting depth enough to land at exactly 30.

y1yang0 (approved):

PR #1263, opened 2026-08-27
refactor(tool/internal/instrument): stop re-parsing hook files and the trampoline template

darkraider01:
Item 3 from #1242.

Two things were getting re-parsed for no reason:

getHookFunc re-scanned and re-parsed the whole hook directory for every rule, with no caching at all. First pass at this cached by {resolved path, hook name}, but that only helped when the exact same rule got looked up twice (it's called once building the trampoline, once in optimizeTJumps). Real instrumentation packages don't work that way, database/sql alone has ~18 rules that all point at the same hook file with different hook names, so that cache barely did anything. Switched to caching the parsed file itself, keyed by path, so any rule sharing a file only pays for the parse once.

materializeTemplate re-parsed the same static impl.tmpl string from scratch on every func rule. Now it's parsed once per process and cloned per use (has to be cloned, not shared directly, since the caller mutates names on the clone to make each trampoline unique). Checked first whether cloning is actually cheaper than reparsing since that wasn't obvious, it is, about 14x (82µs vs 1.18ms in a quick benchmark).

Added a test that corrupts the hook file after the first lookup and confirms a second lookup for a different rule sharing that file still works, which only passes if it's served from the cache.

### DIFF
--- a/tool/internal/instrument/trampoline.go
+++ b/tool/internal/instrument/trampoline.go
@@ -9,6 +9,7 @@ import (
 	"go/token"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/dave/dst"
 
@@ -94,14 +95,23 @@ func (ip *instrumentPhase) ensureUnsafeImport() {
 	ip.target.Decls = append([]dst.Decl{unsafeImport}, ip.target.Decls...)
 }
 
+// parsedTemplateImpl parses the static impl.tmpl source once per process;
+// materializeTemplate runs once per instrumented func rule and clones this
+// prototype instead of re-parsing, since cloning is significantly cheaper.
+//
+//nolint:gochecknoglobals // memoized parse of an embedded constant, not mutable state
+var parsedTemplateImpl = sync.OnceValues(func() (*dst.File, error) {
+	return ast.NewAstParser().ParseSource(templateImpl)
+})
+
 func (ip *instrumentPhase) materializeTemplate() error {
 	// Read trampoline template and materialize before and after function
 	// declarations based on that
-	p := ast.NewAstParser()
-	astRoot, err := p.ParseSource(templateImpl)
+	proto, err := parsedTemplateImpl()
 	if err != nil {
 		return err
 	}
+	astRoot := util.AssertType[*dst.File](dst.Clone(proto))
 
 	ip.varDecls = make([]dst.Decl, 0)
 	ip.hookCtxMethods = make([]*dst.FuncDecl, 0)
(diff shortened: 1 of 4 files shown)


tool/internal/instrument/trampoline.go
```diff
@@ -181,42 +191,55 @@ func isHookDefined(root *dst.File, rule *rule.InstFuncRule) bool {
...
-func findHookFile(rule *rule.InstFuncRule) (string, error) {
+// parseHookFileCached parses file once per process and reuses the result
+// across every hook lookup that touches it, whether for the same func rule
+// looked up twice (createTrampoline and optimizeTJumps) or a different rule
+// whose hook happens to live in a file already scanned for another rule.
+func (ip *instrumentPhase) parseHookFileCached(file string) (*dst.File, error) {
+	if root, ok := ip.parsedHookFiles[file]; ok {
+		return root, nil
+	}
+	root, err := ast.ParseFile(file)
```
ADITYA-CODE-SOURCE: Since the cached AST is only used to inspect hook declarations and function bodies, could we continue using `ParseFileFast` here? It skips object resolution while preserving the information these consumers need. Given that this PR focuses on performance, it may be worth benchmarking both approaches and running the hook-signature and trampoline-optimization tests to verify the behavior.

darkraider01: Checked and the cached hook file is only ever read (declarations and param types), never written back, so skipping object resolution is safe here. Also grepped the codebase and nothing touches the legacy Obj/Scope/Unresolved fields that flag disables.

Benchmarked both on a real 297 line hook file, 3 runs of 500 iterations each: ParseFile averaged ~2.78ms, ParseFileFast ~2.59ms, so about 7% faster and more consistent run to run.

tool/internal/instrument/trampoline.go
```diff
@@ -94,14 +95,23 @@ func (ip *instrumentPhase) ensureUnsafeImport() {
...
+
 func (ip *instrumentPhase) materializeTemplate() error {
 	// Read trampoline template and materialize before and after function
 	// declarations based on that
-	p := ast.NewAstParser()
-	astRoot, err := p.ParseSource(templateImpl)
+	proto, err := parsedTemplateImpl()
 	if err != nil {
 		return err
 	}
```
darccio, line 115: The parse of `templateImpl` is memoized via sync.OnceValues over a `//go:embed`'d
constant, so this error branch is unreachable in practice: the same embedded
source either parses on every call or the binary itself was built broken. This
looks like one of the two lines Codecov flags as uncovered.

Keeping the branch is fine — the compiler requires it and swallowing the error
would be worse. But it may be worth a brief comment noting it is effectively
dead (defensive only), so a future reader doesn't add a test fixture trying to
exercise it, or so coverage tooling can be annotated deliberately.

darccio (changes requested):

darccio (approved):

PR #1266, opened 2026-08-27
fix(tool): remove unused HookContextImpl methods when After hook is absent

### LINKED ISSUE #1264: Optimization: Remove unused HookContextImpl methods when After hook is absent
During the instrumentation optimization phase in `optimizeTJumps()`, the `otelc` tool removes the generated calls to the `After` trampoline if a rule doesn't define an `After` hook (`rule.After == ""`). However, there is an unresolved `TODO` to also remove the corresponding `HookContextImpl` methods (such as `SetReturnVal` and `GetReturnVal`) that are only relevant when an `After` hook is executed. By leaving these methods in the generated `HookContextImpl` type for every instrumented function that lacks an `After` hook, we increase the size of the injected AST, which leads to unnecessary compilation overhead and slightly larger final binaries.

Extend `optimizeTJumps()` to locate the `HookContextImpl` type declaration associated with the trampoline. When `rule.After == ""` is true, the tool should actively remove the AST nodes for methods like `SetReturnVal` and `GetReturnVal`. This will keep the generated hook context as lean as possible and resolve the pending technical debt.

An alternative is leaving it as-is (the current behavior), which doesn't break functionality but leaves technical debt and incurs a slight performance penalty during compilation due to serializing/compiling unused AST nodes.

**Additional context**
Code pointer: `tool/internal/instrument/optimize.go` inside `func (ip *instrumentPhase) optimizeTJumps() error` (around line 367)

```go
// TODO: Remove corresponding HookContextImpl methods
removedOnExit := false
rule := tjump.rule
if rule.After == "" {
```

vishalmore90:
## Description

This PR extends the `optimizeTJumps()` logic to locate the `HookContextImpl` type declaration associated with a given trampoline and actively remove the AST nodes for `SetReturnVal`, `GetReturnVal`, and `GetReturnValCount` when the rule doesn't define an `After` hook (`rule.After == ""`).

## Motivation

Previously, `HookContextImpl` methods related to return values were generated for every instrumented function, even if the instrumentation lacked an `After` hook. By removing these methods when `rule.After == ""`, we resolve the pending technical debt (as noted in an inline `TODO`), keep the generated hook context leaner, and avoid unnecessary compilation overhead from serializing/compiling unused AST nodes.

Fixes #1264

---

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/tool/internal/instrument/optimize.go
+++ b/tool/internal/instrument/optimize.go
@@ -198,6 +198,28 @@ func removeAfterTrampolineDecl(targetFile *dst.File, tjump *tJump) error {
 	return ex.Newf("can not remove After trampoline function")
 }
 
+func removeUnusedHookContextImplMethods(targetFile *dst.File, tjump *tJump) error {
+	structName := trampolineHookContextImplType + tjump.rule.Identity()
+	var newDecls []dst.Decl
+	for _, decl := range targetFile.Decls {
+		if funcDecl, ok := decl.(*dst.FuncDecl); ok && ast.HasReceiver(funcDecl) {
+			t := util.AssertType[*dst.StarExpr](funcDecl.Recv.List[0].Type)
+			t2 := util.AssertType[*dst.Ident](t.X)
+			if t2.Name == structName {
+				methodName := funcDecl.Name.Name
+				if methodName == trampolineGetReturnValName ||
+					methodName == trampolineSetReturnValName ||
+					methodName == "GetReturnValCount" {
+					continue
+				}
+			}
+		}
+		newDecls = append(newDecls, decl)
+	}
+	targetFile.Decls = newDecls
+	return nil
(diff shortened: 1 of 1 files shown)


Labreo (commented):
Recent PR #1252 looked into this exact TODO. The Before hook takes a hook.HookContext interface. Because of that, HookContextImpl must provide all interface methods to compile in Go, even without an After hook.

Also, running go test ./tool/... currently fails on functions that use value receivers instead of pointer receivers.

Because Go requires the full interface implementation here, these methods need to stay on the struct.I think issue #1264 may be based on a misread of the TODO. The TODO was resolved by #1252 with a comment explaining why removal is not possible. It might be worth closing this PR and asking a maintainer to close the issue as well, since the constraint is intentional.

PR #1270, opened 2026-08-28
feat(tool): support minimum otelc version in rule files

### LINKED ISSUE #1261: feat: Support top-level `version` (min otelc version) field in `otelc.yaml` and define multi-file behavior
## Description
As brought up during the migration to the contrib repo (see open-telemetry/opentelemetry-go-compile-contrib#5), we need a way to specify the required `otelc` version in our `otelc.yaml` rule files. Adding a top-level `version` field (representing the **minimum `otelc` version** required) will allow us to evolve the compiler capabilities and schemas safely without breaking older environments.

Currently, the `otelc.yaml` file structure defines rules directly at the root level (where keys are rule names). Adding a `version` field will require updating the schema parser (either by reserving the `version` key or wrapping rules under a `rules:` block).

Additionally, since an instrumentation package is allowed to have multiple rule files (e.g., `otelc.yaml`, `http.otelc.yaml`, `client.otelc.yaml`), we need to define the semantics of this version requirement across these files.

## Open Questions / Design Decisions
1. **Schema Structure:** Should we reserve `version` as a special top-level key (meaning no rule can be named `version`), or should we change the schema to require a `rules:` block?
   ```yaml
   # Option A (Reserved key)
   version: "v1.1.0"
   my_rule_1:
     target: ...
   
   # Option B (Nested block)
   version: "v1.1.0"
   rules:
     my_rule_1:
       target: ...
   ```
2. **Multi-file versioning:** If a package contains multiple `*.otelc.yaml` files, how is the minimum version evaluated?
   - Is the version parsed **per-file**?
   - Or must all files in a package share the same version? If so, must they all declare it, or can a primary `otelc.yaml` declare the version for the entire package?
3. **Default behavior:** If a rule file does not contain a `version` field, what is the fallback behavior? (e.g., assume an older version, or throw a validation error?)

## Tasks
- [ ] Agree on the schema structure for the `version` field.
- [ ] Define and document the behavior for multiple `*.otelc.yaml` files in a single package.
- [ ] Update the `otelc.yaml` schema parser and validation logic to support the new field and check it against the running compiler version.
- [ ] Update all existing `otelc.yaml` files in the repository to include the version field.

ADITYA-CODE-SOURCE:
## Description

Adds support for declaring the minimum required `otelc` version in each rule file using a reserved top-level key:

```yaml
version: "v1.0.0"
```

This PR:

* Adds a shared parser for separating file-level version metadata from rule entries.
* Validates that minimum versions are released semantic versions.
* Rejects rule files that require a newer released `otelc` version.
* Skips compatibility checks for development and pseudo-version compiler builds.
* Treats unversioned rule files as requiring `v1.0.0` and emits a warning.
* Applies version handling to runtime rule loading, automatic pinning, manifest generation, and test utilities.
* Updates existing rule files to declare `version: "v1.0.0"`.
* Updates the documentation and embedded instrumentation bundle.
* Adds tests covering explicit versions, legacy files, malformed requirements, unsupported versions, and development builds.

## Motivation

Rule files can evolve to use features that older `otelc` releases do not understand. Without an explicit compatibility requirement, an older compiler may attempt to process a newer rule file and fail with an unclear parsing error or unexpected behavior.

A per-file minimum version provides users with an early, actionable compatibility error while keeping rule files self-contained. It also preserves compatibility with existing unversioned files through the `v1.0.0` fallback.

Fixes #1261


tool/internal/rulefile/rulefile.go
```diff
@@ -0,0 +1,66 @@
...
+	if doc.Kind != yaml.MappingNode {
+		return Document{}, ex.New("rule file root must be a mapping")
+	}
+
+	result := Document{MinimumVersion: LegacyVersion, Legacy: true, Rules: make(map[string]yaml.Node)}
+	for i := 0; i < len(doc.Content); i += 2 {
+		name := doc.Content[i].Value
+		value := doc.Content[i+1]
+		if name != "version" {
+			result.Rules[name] = *value
```
darkraider01: Should we reject duplicate top-level keys here? Since this is now parsed through yaml.Node, a duplicate rule name is silently overwritten in result.Rules. Previously the YAML map unmarshal rejected duplicate keys, so this could make a rule disappear without any error.

instrumentation/github.com/sirupsen/logrus/otelc.yaml
```diff
@@ -1,3 +1,5 @@
+version: "v1.0.0"
+
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
```
amazingakai: The version needs to go after the copyright header

instrumentation/log/slog/otelc.yaml
```diff
@@ -1,3 +1,5 @@
+version: "v1.0.0"
+
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
```
amazingakai: Here as well.

instrumentation/log/otelc.yaml
```diff
@@ -1,3 +1,5 @@
+version: "v1.0.0"
+
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
```
amazingakai: Also here

tool/internal/setup/match.go
```diff
@@ -459,6 +482,9 @@ func loadCustomRules(ruleConfig string) ([]rule.InstRule, error) {
 			if err != nil {
 				return nil, ex.Wrapf(err, "failed to read %s from -rules flag", file)
 			}
+			if err = checkRuleFileVersion(file, content, warn); err != nil {
+				return nil, err
+			}
 
 			var rules []rule.InstRule
 			rules, err = parseRuleFromYaml(content)
```
amazingakai: We're parsing the same yaml twice, can we do this in one go in parseRuleFromYaml directly?

tool/internal/rulefile/rulefile.go
```diff
@@ -0,0 +1,72 @@
...
+		result.Legacy = false
+	}
+
+	return result, nil
+}
+
+func CheckVersion(current, required string) error {
+	if current == "" || current == "(devel)" || current == "v0.0.0" || module.IsPseudoVersion(current) {
+		return nil
+	}
```
amazingakai: This is slightly fragile, consider using `semver.IsValid` instead:

```suggestion
	if !semver.IsValid(current) || module.IsPseudoVersion(current) || current == "v0.0.0" {
		return nil // Skip check for custom/dev builds
	}
```

tool/internal/rulefile/rulefile.go
```diff
@@ -0,0 +1,72 @@
+// Copyright The OpenTelemetry Authors
```
amazingakai: IMO this should be part of `rule` package and also consider unexporting variables/functions that aren't needed to be exported

docs/adr/0003-structured-rule-schema.md
```diff
@@ -40,9 +40,12 @@ decision so future contributors do not re-litigate the question.
 
 Adopt a **2-tier `where`/`do` schema** for all instrumentation rules.
 
-A rule has exactly the following top-level shape:
+A rule file reserves the top-level `version` key for its minimum required `otelc`
+version. Each remaining top-level entry is a rule with the following shape:
 
 ```yaml
+version: "v1.0.0"
+
```
amazingakai: ADR's shouldn't be edited I think

tool/internal/rulefile/rulefile.go
```diff
@@ -0,0 +1,72 @@
...
+
+	doc := root.Content[0]
+	if doc.Kind != yaml.MappingNode {
+		return Document{}, ex.New("rule file root must be a mapping")
+	}
+
+	result := Document{MinimumVersion: LegacyVersion, Legacy: true, Rules: make(map[string]yaml.Node)}
+	seen := make(map[string]struct{})
+	for i := 0; i < len(doc.Content); i += 2 {
+		name := doc.Content[i].Value
```
amazingakai: `.Value` would be `""` for a complex key (like `[a, b]`), consider adding a proper guard for those keys.

amazingakai (changes requested):

amazingakai:
Please also rebase to resolve merge conflicts

ADITYA-CODE-SOURCE:
Thanks for the review @darkraider01  @amazingakai. I’ve addressed all the feedback, rebased onto the latest main, and the relevant checks are passing. Could you please take another look?

darkraider01:
One thing I want a second opinion on: doesn't this feature break itself on rollout? The old `parseRuleFromYaml` unmarshals the whole file into `map[string]map[string]any`, so the moment any rule file has a top-level `version: "v1.0.0"`, any `otelc` built before this PR merges will hard-fail on it with a raw YAML type error, not a helpful one. And since this PR adds `version:` to every bundled rule file in the same commit, the first release with this feature is also the first rule-file set that breaks anyone who hasn't upgraded yet. Is that an accepted tradeoff, or should older otelc get a compat patch first so it can at least ignore an unrecognized `version:` key before the bundled files start using it?

 Also found two smaller things while reading through:
- `gen-manifest`'s `parseRuleEntries` parses the file's `version:` key but never actually checks it against the current otelc version, unlike the runtime loader and the pinning path, which both do. Low real-world impact today since `go run` reports `(devel)` as the version, but seems like a gap versus what the PR description says it covers.
 - `pin.go`'s `loadModuleRules` reimplements `checkRuleFileVersion`'s body inline instead of just calling it, since they're in the same package.

@amazingakai curious what you think, especially on the rollout question.

ADITYA-CODE-SOURCE:
> One thing I want a second opinion on: doesn't this feature break itself on rollout? The old `parseRuleFromYaml` unmarshals the whole file into `map[string]map[string]any`, so the moment any rule file has a top-level `version: "v1.0.0"`, any `otelc` built before this PR merges will hard-fail on it with a raw YAML type error, not a helpful one. And since this PR adds `version:` to every bundled rule file in the same commit, the first release with this feature is also the first rule-file set that breaks anyone who hasn't upgraded yet. Is that an accepted tradeoff, or should older otelc get a compat patch first so it can at least ignore an unrecognized `version:` key before the bundled files start using it?
> 
> Also found two smaller things while reading through:
> 
> * `gen-manifest`'s `parseRuleEntries` parses the file's `version:` key but never actually checks it against the current otelc version, unlike the runtime loader and the pinning path, which both do. Low real-world impact today since `go run` reports `(devel)` as the version, but seems like a gap versus what the PR description says it covers.
> * `pin.go`'s `loadModuleRules` reimplements `checkRuleFileVersion`'s body inline instead of just calling it, since they're in the same package.
> 
> @amazingakai curious what you think, especially on the rollout question.

Good catch thanks @darkraider01. Existing embedded bundles are safe, but older otelc versions would fail on newer external rule files.
 @amazingakai would you prefer a staged rollout where this PR adds parser support but leaves existing rule files unversioned? I’ll also add the missing manifest check, reuse the existing version-check helper, and rebase afterward.

tool/internal/setup/pin.go
```diff
@@ -210,7 +226,10 @@ func loadModuleRules(moduleDir, module string, loaded map[string][]yamlRule) err
 	})
 }
 
-func loadMinimalRules(rulesRoot string) (map[string][]yamlRule, error) {
+func loadMinimalRules(
+	rulesRoot, currentVersion string,
+	warn func(string, ...any),
```
amazingakai: Instead of passing `warn func(string, ...any)` to all these functions, can we pass ctx and use `util.LoggerFromContext`?

amazingakai (commented):
This looks mostly good to me @ADITYA-CODE-SOURCE 

Older releases of `otelc` use the `otelc-bundle.tgz` that is statically compiled into their binary at release time. Because those older bundles are frozen in time, older compilers will never see the new `version:` keys we are adding to the in-tree files, so they won't crash. (If someone manually passes a new custom rule file with `version:` to an old compiler via `--rules`, it will fail with a YAML error, but that's an expected outcome when using new schema features on old tooling).

IMO you can keep all the `version:` additions to the existing rule files in this PR!

tool/internal/setup/pin.go
ADITYA-CODE-SOURCE: @amazingakai  Updated. The rule-file loading helpers now accept context.Context and use util.LoggerFromContext(ctx).WarnContext(...) directly instead of threading warning callbacks through the call chain. I also reused the shared version-check helper in pinning and added the missing minimum-version validation to manifest generation.

ADITYA-CODE-SOURCE:
@amazingakai Addressed the feedback and rebased onto the latest main. All checks are running again. Could you please take another look?

amazingakai (approved):
LGTM, please rebase/fix conflicts

darccio (approved):

PR #1272, opened 2026-08-28
fix(tool): normalize bundle file modes

### LINKED ISSUE #1269: bug(tool): make package depends on host execute bits, so Verify Bundle differs between /mnt/c and ext4
## Description:

On latest `main`, `make package` produces different `tool/data/otelc-bundle.tgz` bytes depending on whether the repository is built from a `/mnt/c` checkout or a native Linux/ext4 checkout.

Under `/mnt/c`, DrvFS reports regular source files as executable (`0777`), so the bundle tool writes many archive entries as `0755`. The same commit built from ext4 writes those same files as `0644`. The tracked source content is identical, but the archive differs, so `Verify Bundle` fails.

## Steps to Reproduce:

1. Check out `main` at `790631ccc`.
2. From PowerShell in the repository root, run:
   ```powershell
   $repo = (wsl wslpath -a ((Get-Location).Path -replace '\', '/')).Trim()

   $script = @"
   set -euo pipefail

   cd '$repo'
   make package >/dev/null
   echo 'drvfs hash:'
   sha256sum tool/data/otelc-bundle.tgz
   echo 'drvfs sample:'
   tar tvzf tool/data/otelc-bundle.tgz | sed -n '1,4p'

   rm -rf /tmp/otelci-bundle-mode-repro
   git clone https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation /tmp/otelci-bundle-mode-repro >/dev/null 2>&1
   cd /tmp/otelci-bundle-mode-repro
   git checkout 790631ccc >/dev/null 2>&1
   make package >/dev/null
   echo 'ext4 hash:'
   sha256sum tool/data/otelc-bundle.tgz
   echo 'ext4 sample:'
   tar tvzf tool/data/otelc-bundle.tgz | sed -n '1,4p'

   echo 'source mode on drvfs:'
   stat -c '%a %n' '$repo/pkg/runtime/otel_setup.go'
   echo 'source mode on ext4:'
   stat -c '%a %n' /tmp/otelci-bundle-mode-repro/pkg/runtime/otel_setup.go
   "@

   wsl bash -lc $script
   ```

## Actual Behavior:

The repro prints different bundle hashes and different file modes for the same tracked file content.

Example output:
```text
drvfs hash:
240fdf34f9f90eb19e80e5160dad7a139e82f936c892a71dcf19efd6b98e3cd2  tool/data/otelc-bundle.tgz
drvfs sample:
drwxr-xr-x 0/0               0 1970-01-01 00:00 instrumentation_temp/
drwxr-xr-x 0/0               0 1970-01-01 00:00 instrumentation_temp/database/
drwxr-xr-x 0/0               0 1970-01-01 00:00 instrumentation_temp/database/sql/
-rwxr-xr-x 0/0           17703 1970-01-01 00:00 instrumentation_temp/database/sql/client.go

ext4 hash:
3a47482852281260c0425e216cb47f85f5cf490d598c196eeed4a0812a99f18d  tool/data/otelc-bundle.tgz
ext4 sample:
drwxr-xr-x 0/0               0 1970-01-01 00:00 instrumentation_temp/
drwxr-xr-x 0/0               0 1970-01-01 00:00 instrumentation_temp/database/
drwxr-xr-x 0/0               0 1970-01-01 00:00 instrumentation_temp/database/sql/
-rw-r--r-- 0/0           17703 1970-01-01 00:00 instrumentation_temp/database/sql/client.go

source mode on drvfs:
777 /mnt/c/.../pkg/runtime/otel_setup.go
source mode on ext4:
644 /tmp/otelci-bundle-mode-repro/pkg/runtime/otel_setup.go
```

So the bundle depends on the checkout filesystem’s execute-bit behavior, not just on the tracked source.

## Expected Behavior:

`make package` should produce the same `tool/data/otelc-bundle.tgz` bytes for the same commit regardless of whether the checkout is on `/mnt/c` or on native Linux/ext4.

A Windows-mounted DrvFS checkout should not cause ordinary source files to be archived as `0755` when the same files from ext4 are archived as `0644`.

## Notes:

- `.tools/bundle/main.go` currently normalizes regular-file mode based on `hdr.Mode&0o111`.
- On `/mnt/c`, `tar.FileInfoHeader(...)` sees many regular source files as executable because DrvFS exposes them as `0777`.
- That causes `normalizeHeader` to write `0755` into the archive for ordinary `.go`, `go.mod`, `go.sum`, `yaml`, and `md` files.
- This makes `Verify Bundle` fail for contributors who regenerate `tool/data/otelc-bundle.tgz` from a `/mnt/c` checkout, even when source content is otherwise correct.
- This is separate from #1123, which was about untracked junk files like `.DS_Store` entering the archive.

RedZapdos123:
## Description:

Issue #1269 reported that `make package` could produce different `tool/data/otelc-bundle.tgz` bytes for the same sources depending on whether the checkout lived on WSL `/mnt/c` or native Linux/ext4.

Current `main` normalized regular file modes from host execute bits so DrvFS-reported `0777` source files were archived as `0755` while the same files from ext4 were archived as `0644`. This PR normalizes regular files to `0644` unconditionally, keeps directories at `0755`, and adds a regression test covering executable host bits.

Closes #1269.

## Verifications:

- [x] `cd .tools/bundle && go test . -count=1 -v`.
- [x] `make package`.
- [x] `./.github/scripts/verify-bundle.sh`.


dobbydobap:
This is the right fix and it is blocking real work, so worth pulling in.

I hit the same thing from the other side. #1255 fails Verify Bundle with content that is
completely correct. I unpacked its committed archive against a regenerated one and the entry
names, sizes, mtimes and gzip framing all match, as does the fix itself inside
pkg_temp/runtime/otel_setup.go. The only difference is 458 tar headers reading -rwxr-xr-x
instead of -rw-r--r--, which is exactly what #1269 describes.

Two things I checked that might save a reviewer time.

Nothing in the bundle wants the execute bit. All 229 regular entries in main's archive are
already 0644 and there are no scripts or binaries in it, so dropping the branch loses nothing.
The fix does not change the committed archive. I applied it to main and ran make package and
the sha256 stayed 3a474828, so this PR is right not to ship a regenerated bundle with it.
What is actually holding it up looks procedural rather than technical: every workflow here is
still action_required from the 28th, so CI has never run on it.

y1yang0 (approved):

PR #1274, opened 2026-08-28
fix(tool/ast): handle import alias collisions and enforce strict import matching (#1271)

### LINKED ISSUE #1271: bug(tool/ast): handle import alias collisions gracefully and enforce strict import matching in `typename.go`
### Problem Description

In `tool/internal/ast/typename.go`, package import resolution and AST type matching exhibit two edge-case issues:

1. **Import Alias Collisions:** `ImportAliasMap` maps default package aliases to import paths. When a source file contains multiple unaliased imports that share a default alias (such as `import "text/template"` and `import "html/template"`), `aliases[alias] = path` silently overwrites earlier entries with whichever import appears last. Consequently, calls to `MatchesTypeName` for types in the overwritten package fail or match against the wrong import path.
2. **Fallback Matching under Import Context:** In `parsedTypeName.matches` (`*dst.SelectorExpr`), when `imports` context is provided (`imports != nil`), failing to find `ident.Name` in `imports` falls through to path-tail matching (`defaultImportAlias(t.importPath) == ident.Name`). When explicit import context is available, missing a key in `imports` indicates that `ident.Name` is not that imported package, so falling back to tail matching can cause false-positive matches on unimported packages or local variables.

### Proposed Fix / Solution

- **Disambiguate Collisions in `ImportAliasMap`:** Track whether an import alias is explicitly set (`imp.Name != nil`). If multiple unaliased imports collide on a default alias, remove the ambiguous default alias from `aliases` so unaliased references do not resolve to an arbitrary import. Explicit aliases take precedence.
- **Strict Import Context Checking:** In `parsedTypeName.matches`, if `imports != nil` and `ident.Name` is not present in `imports`, return `false` immediately. Preserve path-tail fallback only when `imports == nil` (for synthetic/test AST nodes).

### Reproducer / Unit Tests

We have implemented and verified tests in `tool/internal/ast/typename_test.go`:
- `TestImportAliasMap_CollidingDefaultAliases`
- `TestTypeNameMatches_StrictImportContext`

vedant21-oss:
## Description

Fixes #1271.

This PR addresses two edge-case issues in `tool/internal/ast/typename.go`:

1. **Import Alias Collisions:** `ImportAliasMap` previously mapped default package aliases to import paths without checking for collisions. When a source file contained multiple unaliased imports sharing a default alias (e.g. `import "text/template"` and `import "html/template"`), `aliases[alias] = path` silently overwrote earlier entries. We now track explicit vs default aliases and exclude ambiguous default aliases from the map so unaliased references do not resolve to an arbitrary import.
2. **Strict Import Context Matching:** In `parsedTypeName.matches` (`*dst.SelectorExpr`), when `imports` context is provided (`imports != nil`), failing to locate `ident.Name` in `imports` now returns `false` immediately instead of falling through to path-tail matching on unimported packages or local variables.

## Verification / Unit Tests

Added unit tests in `tool/internal/ast/typename_test.go`:
- `TestImportAliasMap_CollidingDefaultAliases`
- `TestTypeNameMatches_StrictImportContext`

All tool package tests pass cleanly (`go test ./tool/...`).

### DIFF
--- a/tool/internal/ast/typename.go
+++ b/tool/internal/ast/typename.go
@@ -58,8 +58,11 @@ func (t parsedTypeName) matches(node dst.Expr, imports map[string]string) bool {
 			// Populated by a resolving decorator; already the real import path.
 			return t.importPath == ident.Path && t.name == n.Sel.Name
 		}
-		if resolved, importOk := imports[ident.Name]; importOk {
-			return t.importPath == resolved && t.name == n.Sel.Name
+		if imports != nil {
+			if resolved, importOk := imports[ident.Name]; importOk {
+				return t.importPath == resolved && t.name == n.Sel.Name
+			}
+			return false
 		}
 		// No import context at all (imports == nil, e.g. hand-built AST nodes in
 		// tests with no backing *dst.File): compare against importPath's last
@@ -158,6 +161,9 @@ func ImportAliasMap(file *dst.File) map[string]string {
 	}
 
 	aliases := make(map[string]string, len(specs))
+	explicit := make(map[string]bool, len(specs))
+	collided := make(map[string]bool)
+
 	for _, imp := range specs {
 		if imp.Path == nil {
 			continue
@@ -166,8 +172,9 @@ func ImportAliasMap(file *dst.File) map[string]string {
 		if err != nil {
 			continue
 		}
+		isExplicit := imp.Name != nil
 		alias := defaultImportAlias(path)
-		if imp.Name != nil {
+		if isExplicit {
 			alias = imp.Name.Name
 		}
 		// Blank and dot imports don't introduce a qualified identifier that a
@@ -175,8 +182,33 @@ func ImportAliasMap(file *dst.File) map[string]string {
 		if alias == "" || alias == "_" || alias == "." {
 			continue
 		}
-		aliases[alias] = path
+		if existingPath, exists := aliases[alias]; exists {
+			if existingPath != path {
+				if isExplicit && !explicit[alias] {
+					// New explicit alias overrides previous default alias
+					aliases[alias] = path
+					explicit[alias] = true
+					delete(collided, alias)
+				} else if !isExplicit && explicit[alias] {
+					// Previous explicit alias wins over new default alias
+					continue
+				} else {
+					// Collision between two default aliases (or two conflicting explicit aliases)
+					collided[alias] = true
+				}
+			}
+		} else {
+			aliases[alias] = path
+			if isExplicit {
+				explicit[alias] = true
+			}
+		}
 	}
+
+	for c := range collided {
+		delete(aliases, c)
+	}
+
 	return aliases
 }
 
--- a/tool/internal/ast/typename.go
+++ b/tool/internal/ast/typename.go
@@ -59,10 +59,8 @@ func (t parsedTypeName) matches(node dst.Expr, imports map[string]string) bool {
 			return t.importPath == ident.Path && t.name == n.Sel.Name
 		}
 		if imports != nil {
-			if resolved, importOk := imports[ident.Name]; importOk {
-				return t.importPath == resolved && t.name == n.Sel.Name
-			}
-			return false
+			resolved, importOk := imports[ident.Name]
+			return importOk && t.importPath == resolved && t.name == n.Sel.Name
 		}
 		// No import context at all (imports == nil, e.g. hand-built AST nodes in
 		// tests with no backing *dst.File): compare against importPath's last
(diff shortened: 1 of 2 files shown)


tool/internal/ast/typename_test.go
```diff
@@ -522,3 +522,60 @@ func TestMatchesTypeName_UnsupportedNodeDoesNotMatch(t *testing.T) {
...
+
+func f() {}
+`)
+		require.NoError(t, err)
+
+		imports := ImportAliasMap(file)
+		assert.Equal(t, "net/http", imports["http"])
+		assert.NotContains(t, imports, "template")
+	})
+
```
darkraider01, line 546: The alias names here (template and htmltemplate) don't collide, so this test doesn't actually hit the override logic. Using template "html/template" instead will properly test the collision override

tool/internal/ast/typename.go
```diff
@@ -166,17 +172,43 @@ func ImportAliasMap(file *dst.File) map[string]string {
...
+		if isExplicit {
 			alias = imp.Name.Name
 		}
 		// Blank and dot imports don't introduce a qualified identifier that a
 		// type reference could use, so they can't participate in matching.
 		if alias == "" || alias == "_" || alias == "." {
 			continue
 		}
-		aliases[alias] = path
+		if existingPath, exists := aliases[alias]; exists {
```
darkraider01: If existingPath == path and isExplicit is true (e.g. import "net/http" followed by import http "net/http"), explicit[alias] isn't set to true. We should update explicit[alias] = true here as well.

tool/internal/ast/typename.go
```diff
@@ -136,28 +137,35 @@ func MatchesTypeName(node dst.Expr, typeStr string, imports map[string]string) (
...
+		for _, spec := range genDecl.Specs {
+			if importSpec, isImport := spec.(*dst.ImportSpec); isImport {
+				specs = append(specs, importSpec)
 			}
 		}
 	}
+	return specs
+}
+
+func ImportAliasMap(file *dst.File) map[string]string {
```
Labreo: When you pulled out collectImportSpecs, the original doc comment for ImportAliasMap ended up sitting above collectImportSpecs instead. You might want to shift that comment down so godoc and IDE hover text attach cleanly to ImportAliasMap. While you are there, adding a quick nil check at the start of collectImportSpecs could be good for extra safety.

tool/internal/ast/typename_test.go
Labreo, line 1: To fix the codecov coverage.You could throw in a quick subtest for colliding explicit aliases (for example, two different packages using the same alias name like tpl) to confirm they get excluded as well.

Labreo (commented):
I have left two small notes about moving the doc comment down to ImportAliasMap and adding an extra subtest for conflicting explicit aliases. Overall, changes look solid to me!

vedant21-oss:
Thanks for catching that! I've moved the doc comment down so it attaches directly to ImportAliasMap and added the if file == nil check at the start of collectImportSpecs.

dobbydobap:
Had a look since I was in typename.go earlier this week. The fix is worth landing, but the
collision being tested cannot happen, and that applies to the suggestions above too.

Two imports introducing the same identifier in one file is a compile error. On go1.25 all three
of these fail with "redeclared in this block":

  import ("html/template"; "text/template")
  import (template "html/template"; "text/template")
  import (tpl "html/template"; tpl "text/template")

go/parser accepts them, which is why they look reachable, but otelc only parses compiling
dependency source out of the module cache. Same shape as #766.

The collision that does happen comes from defaultImportAlias guessing. Its own comment says the
default name is a syntactic guess from the path tail, so two imports whose tails match but whose
real package names differ collide in the map and still compile:

  example.com/m/bar        package bar
  example.com/m/other/bar  package qux

Both unaliased, that builds. main gives map["bar"] = "example.com/m/other/bar", so the second
import silently wins and bar.A no longer matches example.com/m/bar.A. This PR returns an empty
map instead, which is the right direction. It is the fixture that wants swapping.

vedant21-oss:
Thanks for the thorough explanation @dobbydobap! That makes total sense — I've updated the test fixtures in `typename_test.go` in commit 25c638e to use valid compiling unaliased import paths with matching path tails (`example.com/m/bar` vs `example.com/m/other/bar`).

dobbydobap:
Thanks for turning that around so fast. I checked the four fixtures against a real module where
example.com/m/bar is package bar and example.com/m/other/bar is package qux.

The first three compile now, so they pin real behaviour.

The last one still does not. Two explicit aliases with the same name is a redeclaration
regardless of what they point at:

    tpl "example.com/m/bar"
    tpl "example.com/m/other/bar"
    -> tpl redeclared in this block

Swapping the paths did not change that, because the clash is in the identifier rather than the
import path. Worth deciding whether that branch is defensive handling or can go, but either way
the subtest is describing source that cannot reach the tool.

vedant21-oss:
Merged main in and fixed the CI failure.

TestApplyCallRule_ConditionalReplace has three fixtures that declare a context.Context parameter without importing context. They passed before because matching fell back to comparing an import path's last segment when the identifier wasn't in the file's import map.

With strict matching that no longer resolves, which is the point of #1271 — a type reference should only match the package the file actually imports. Real Go can't use context.Context without importing context, so I made the fixtures valid instead of relaxing the matching. Happy to go the other way if you'd rather.

Also worth flagging: #1312 touches the same function and will conflict with this one, so whichever lands first, I'll rebase.

tool/internal/ast/typename.go
vedant21-oss: This is already the case on the branch — collectImportSpecs starts with a nil check and the doc comment sits with ImportAliasMap. Thanks for flagging it.

tool/internal/ast/typename_test.go
vedant21-oss: Added — two imports both aliased tpl pointing at different packages, asserting neither ends up in the map.

Issue #1260, opened 2026-08-27
[Epic] Migrate instrumentation/ and parts of pkg/ to Contrib Repository


amazingakai:
Following up on our community discussion, we are migrating our instrumentations to the new `open-telemetry/opentelemetry-go-compile-contrib` repository. This will enable registry-based discoverability (see #849) without temporary manifests in the long run.

During this migration, we will **freeze PRs** modifying the `instrumentation/` and `pkg/` trees (with the exception of `pkg/hook`). Once the migration is complete, authors of open PRs will be asked to move their contributions to the new repo.

**Tasks:**
- [ ] Ensure all critical bug fixes are merged before the freeze.
- [ ] Initial repository setup: https://github.com/open-telemetry/opentelemetry-go-compile-contrib/issues/4
- [ ] Migrate `instrumentation/` and `pkg/`: https://github.com/open-telemetry/opentelemetry-go-compile-contrib/issues/5
- [ ] Migrate CI, E2E, and Integration Tests: https://github.com/open-telemetry/opentelemetry-go-compile-contrib/issues/6
- [ ] First Contrib Release and otelc Integration: https://github.com/open-telemetry/opentelemetry-go-compile-contrib/issues/7
- [ ] Remove local bundle replace logic.

darkraider01:
i'd like to take this if its available!

amazingakai:
@darkraider01 This is a big issue, so PTAL at the sub-issues and see if there's a specific one you'd like to take on. I wouldn't recommend trying to take the whole thing at once 😅

darkraider01:
ah right, so firstly i'd like to start from initial repo setup !

amazingakai:
@darkraider01 sounds good

Labreo:
@amazingakai I would like to take up migrating the CI, integration, and E2E test suites (contrib#6) once the initial repo setup lands.

amazingakai:
@Labreo Sounds good, I'm currently working on contrib#5, I think we can do this after that. 

In the meantime, that issue has a big scope and should be broken down in sub-issues, you can take care of that if you want

Labreo:
> issue has a big scope and should be broken down in sub-issues, you can take care of that if you want

Noted.

Harsh63870:
@amazingakai, I traced the otelc-side coupling for the last epic task (remove local
bundle replace / contrib #7):

[...]

PR #1277, opened 2026-08-29
fix(pkg/runtime): treat a parses-to-empty OTEL_GO_ENABLED_INSTRUMENTATIONS as no allowlist

### LINKED ISSUE #1254: bug(pkg/runtime): parsed-empty OTEL_GO_ENABLED_INSTRUMENTATIONS disables all instrumentation

vignesh2027:
## Description

`Instrumented` gated the enabled-list check on the raw environment string being non-empty, then compared the target instrumentation against the parsed names:

```go
enabledList := os.Getenv("OTEL_GO_ENABLED_INSTRUMENTATIONS")
if enabledList != "" {
    enabled := parseInstrumentationList(enabledList)
    if !slices.Contains(enabled, name) {
        return false
    }
}
```

A value that is non-empty but parses to nothing, for example `","`, `",,"`, or `"  "`, makes `enabledList != ""` true while `parseInstrumentationList` returns an empty slice (it drops empty and whitespace-only items). `slices.Contains(nil, name)` is then false for every instrumentation, so the function returns false and disables all instrumentation.

The fix gates on the parsed list length instead of the raw string, so an allowlist only restricts when it actually names something. A malformed-but-nonempty value is treated the same as an unset one, and instrumentation stays enabled.

```go
enabled := parseInstrumentationList(os.Getenv("OTEL_GO_ENABLED_INSTRUMENTATIONS"))
if len(enabled) > 0 && !slices.Contains(enabled, name) {
    return false
}
```

The disabled-list path already behaves correctly for the same input (an empty parsed list means `slices.Contains` is false, so nothing is disabled), so it is unchanged.

## Tests

`TestInstrumented` gains two cases: a comma-only value and a whitespace-only value both keep instrumentation enabled. I verified they fail against current `main` and pass with this change.

Fixes #1254

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


dobbydobap:
Two things worth knowing before this goes further.

pkg/runtime is inside the freeze from #1260, and #1255 has been on the same fix for #1254 since
the 26th, with Verify Bundle red. Same change in Instrumented, same two cases. I have asked on
the epic whether a bug like this is meant to land here or move to contrib, since that decides
what happens to both.

The one real difference is worth keeping either way. #1255 also drops the if tt.enabledList != ""
guard around t.Setenv in TestInstrumented. Without it, the cases that expect an unset variable
inherit whatever is already in the environment:

    OTEL_GO_ENABLED_INSTRUMENTATIONS=somethingelse go test ./pkg/runtime -run TestInstrumented
    --- FAIL: TestInstrumented/default_enabled
            expected: true
            actual  : false

Main fails the same way so it is not something you introduced, but a test for env var handling
that reads ambient env vars is worth closing while the file is open.

vignesh2027:
Thanks @dobbydobap, you are right on both counts. #1255 has the same fix for #1254 and predates this by two days, so this is a duplicate and I am closing it in favor of #1255. I should have found it before opening this. The point about dropping the `if tt.enabledList != ""` guard around `t.Setenv` in TestInstrumented is a good one and worth keeping in #1255, since a test for env-var handling should not read ambient env. Understood on the pkg/ freeze from #1260 as well. Sorry for the extra noise.

dobbydobap:
Thanks for closing it out so quickly, and no need to apologise. Easy to miss with both landing
in the same week.

PR #1278, opened 2026-08-29
fix(tool): close the output file via defer in ast and importcfg writeFile

### LINKED ISSUE #1275: fix(tool/ast): ensure file handle closure on panic in ast.WriteFile and importcfg.WriteFile
### Problem Description

In `tool/internal/ast/parser.go` (`ast.writeFile`) and `tool/internal/imports/importcfg.go` (`ImportConfig.writeFile`), file writing functions invoke `w.Close()` after completing write operations without using `defer`.

If an unhandled panic occurs during writing (for instance, if an injected AST node is malformed or an internal formatting failure occurs), execution escapes before `w.Close()` is reached. This leaves open file descriptors on the operating system until garbage collection runs.

### Proposed Fix / Solution

Refactor both `ast.writeFile` and `ImportConfig.writeFile` to defer `w.Close()`:
- Use `defer` immediately upon entry so `w.Close()` is guaranteed to execute under all circumstances (normal return, error return, or panic).
- Capture close errors on successful writes and preserve original write errors when both fail.

### Reproducer / Unit Tests

Added panic-safety unit tests in `parser_test.go` and `importcfg_test.go` (`TestWriteFile_PanicSafety`) using mock writers that panic on `Write`, confirming `Close()` is executed under panic conditions.

vignesh2027:
## Description

`ast.writeFile` (`tool/internal/ast/parser.go`) and `ImportConfig.writeFile` (`tool/internal/imports/importcfg.go`) called `Close()` only after the write finished:

```go
func writeFile(w writeCloser, ...) error {
    if err := <write>(w, ...); err != nil {
        _ = w.Close()
        return ex.Wrapf(err, "failed to write ...")
    }
    if err := w.Close(); err != nil {
        return ex.Wrapf(err, "failed to close ...")
    }
    return nil
}
```

If the write panics (a malformed injected AST node, an internal formatting failure), execution escapes before either `Close()` runs, leaking the file descriptor until GC.

Both functions now close via `defer` with a named return, so the descriptor is released on a normal return, an error return, or a panic. A close error is reported only when the write itself succeeded, so the original write error is preserved. Behavior is unchanged on the normal, write-error, and close-error paths.

## Tests

Each package gains `TestWriteFile_PanicSafety`, using the existing `mockWriteCloser` extended to panic on `Write` and record whether `Close` ran. The test asserts the file is closed even when writing panics. I verified both fail against current `main` and pass with this change.

This is tool-only; no bundle regeneration is required.

Fixes #1275

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### DIFF
--- a/tool/internal/ast/parser.go
+++ b/tool/internal/ast/parser.go
@@ -102,14 +102,18 @@ func WriteFile(filePath string, root *dst.File) error {
 	return writeFile(file, filePath, root)
 }
 
-func writeFile(w io.WriteCloser, filePath string, root *dst.File) error {
+func writeFile(w io.WriteCloser, filePath string, root *dst.File) (retErr error) {
+	// The deferred close runs on every return path, including a panic during the write.
+	// The deferred close reports an error only when the write succeeds, so the write error takes priority.
+	defer func() {
+		if closeErr := w.Close(); closeErr != nil && retErr == nil {
+			retErr = ex.Wrapf(closeErr, "failed to close file %s", filePath)
+		}
+	}()
+
 	r := decorator.NewRestorer()
 	if err := r.Fprint(w, root); err != nil {
-		_ = w.Close()
 		return ex.Wrapf(err, "failed to write to file %s", filePath)
-	}
-	if err := w.Close(); err != nil {
-		return ex.Wrapf(err, "failed to close file %s", filePath)
 	}
 	return nil
 }
--- a/tool/internal/imports/importcfg.go
+++ b/tool/internal/imports/importcfg.go
@@ -112,13 +112,17 @@ func (r *ImportConfig) WriteFile(filename string) error {
 	return r.writeFile(file, filename)
 }
 
-func (r *ImportConfig) writeFile(w io.WriteCloser, filename string) error {
+func (r *ImportConfig) writeFile(w io.WriteCloser, filename string) (retErr error) {
+	// The deferred close runs on every return path, including a panic during the write.
+	// The deferred close reports an error only when the write succeeds, so the write error takes priority.
+	defer func() {
+		if closeErr := w.Close(); closeErr != nil && retErr == nil {
+			retErr = ex.Wrapf(closeErr, "failed to close file %s", filename)
+		}
+	}()
+
 	if err := r.write(w); err != nil {
-		_ = w.Close()
 		return ex.Wrapf(err, "failed to write to file %s", filename)
-	}
-	if err := w.Close(); err != nil {
-		return ex.Wrapf(err, "failed to close file %s", filename)
 	}
 	return nil
 }
--- a/tool/internal/ast/parser_test.go
+++ b/tool/internal/ast/parser_test.go
@@ -163,11 +163,16 @@ func TestWriteFile_CreateError(t *testing.T) {
 }
 
 type mockWriteCloser struct {
-	writeErr error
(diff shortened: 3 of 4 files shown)


tool/internal/ast/parser.go
darccio: Main replaced this local `writeCloser` interface with the standard `io.WriteCloser` type in PR #1316. That change conflicts with this PR on the same lines. Delete this interface, and change the `writeFile` parameter type at line 110 to `io.WriteCloser`.

```suggestion
```

tool/internal/imports/importcfg.go
darccio: Main replaced this local `writeCloser` interface with the standard `io.WriteCloser` type in PR #1316. That change conflicts with this PR on the same lines. Delete this interface, and change the `writeFile` parameter type at line 120 to `io.WriteCloser`.

```suggestion
```

tool/internal/ast/parser.go
```diff
@@ -107,15 +107,20 @@ func WriteFile(filePath string, root *dst.File) error {
 	return writeFile(file, filePath, root)
 }
 
-func writeFile(w writeCloser, filePath string, root *dst.File) error {
+func writeFile(w writeCloser, filePath string, root *dst.File) (retErr error) {
+	// Close via defer so the file descriptor is released on a normal return, an
+	// error return, or a panic during writing. A close error is reported only
+	// when the write itself succeeded, so the original write error is preserved.
```
darccio: The three-line comment restates the code below it at more length than the logic needs. A tighter version keeps the same why without extra words.

```suggestion
	// The deferred close runs on every return path, including a panic during the write.
	// The deferred close reports an error only when the write succeeds, so the write error takes priority.
```

tool/internal/imports/importcfg.go
```diff
@@ -117,14 +117,19 @@ func (r *ImportConfig) WriteFile(filename string) error {
 	return r.writeFile(file, filename)
 }
 
-func (r *ImportConfig) writeFile(w writeCloser, filename string) error {
+func (r *ImportConfig) writeFile(w writeCloser, filename string) (retErr error) {
+	// Close via defer so the file descriptor is released on a normal return, an
+	// error return, or a panic during writing. A close error is reported only
+	// when the write itself succeeded, so the original write error is preserved.
```
darccio: The three-line comment restates the code below it at more length than the logic needs. A tighter version keeps the same why without extra words.

```suggestion
	// The deferred close runs on every return path, including a panic during the write.
	// The deferred close reports an error only when the write succeeds, so the write error takes priority.
```

darccio (changes requested):

darccio (approved):

vignesh2027:
Thanks for the review. All four points are addressed and the branch is rebased onto main, so the conflict is gone.

- Removed the local `writeCloser` interface in both files and changed the `writeFile` parameter to `io.WriteCloser`, matching what #1316 landed on main.
- Applied your shorter two line comment in both places.

After the rebase the only change left on top of main is the panic safety one: `writeFile` now closes through a `defer`, so the file is closed on every return path including a panic inside the restorer, and a close error is reported only when the write itself succeeded so the write error keeps priority. `TestWriteFile_PanicSafety` in each package covers that. The rest of what this PR originally carried is already on main.

CI is green.

PR #1279, opened 2026-08-30
fix(tool/internal/setup): recognize quoted -mod tokens in GOFLAGS

### LINKED ISSUE #1267: bug(tool/internal/setup): quoted -mod=readonly in GOFLAGS is overridden by forceModMod
### Describe the bug

`forceModMod()` in `tool/internal/setup/vendor.go` parses `GOFLAGS` with `strings.Fields()` before checking whether a `-mod` flag is already present.

Go accepts quoted `GOFLAGS` tokens, including a value such as:

```text
'-mod=readonly'
```

However, `strings.Fields()` leaves the quote characters attached to the token. As a result, `forceModMod()` does not recognize `'-mod=readonly'` as a `-mod` token and concludes that no module-mode flag was supplied.

It then appends `-mod=mod`, producing:

```text
'-mod=readonly' -mod=mod
```

Because repeated Go flags are last-value-wins, this changes the effective setting from `readonly` to `mod`.

This also conflicts with the function's documented behavior: `-mod=readonly` is intentionally supposed to be preserved because it already disables automatic vendoring while respecting the user's module-mode choice.

### Environment

- OS: Linux 7.1.9-arch1-2 x86_64 GNU/Linux
- Go Version: go1.27.0 linux/amd64
- Version: current `main` at 790631ccc042ba719239f3d0f89a2f3efb2ea6ef

### To Reproduce

1. Confirm that the Go command accepts the quoted token:

```bash
GOFLAGS="'-mod=readonly'" go list ./...
```

2. Add this temporary regression test in package `setup`:

```go
func TestQuotedReadonlyGoflagsRepro(t *testing.T) {
    assert.Equal(
        t,
        "'-mod=readonly'",
        forceModMod("'-mod=readonly'"),
    )
}
```

3. Run:

```bash
go test ./tool/internal/setup -run TestQuotedReadonlyGoflagsRepro -count=1 -v
```

4. The test fails because the current result is:

```text
'-mod=readonly' -mod=mod
```

instead of:

```text
'-mod=readonly'
```

### Expected behavior

Quoted `GOFLAGS` tokens should be interpreted using the same tokenization semantics as the Go command.

A quoted `-mod=readonly` token should be detected as an existing module-mode flag and left unchanged, just like an unquoted `-mod=readonly` token.

Similarly, quoted forms of supported `-mod` values should behave consistently with their unquoted forms.

### Additional context

The repository already contains quote-aware `GOFLAGS` tokenization in `tool/util/go.go` (`splitGoflags`), used by `StripToolexecFromGoflags`.

A possible direction would be to factor out or reuse the quote-aware tokenization and matching outer-quote removal already used by StripToolexecFromGoflags, instead of using strings.Fields() in forceModMod(), with regression tests for quoted -mod tokens.

I searched the repository's existing issues and pull requests for this quoted `GOFLAGS` / `forceModMod` behavior and did not find a matching report.

vignesh2027:
## Description

`forceModMod` in `tool/internal/setup/vendor.go` split `GOFLAGS` with `strings.Fields` and matched the raw tokens. The go command accepts quoted `GOFLAGS` tokens such as `'-mod=readonly'`, and the surrounding quotes stayed attached to the token, so `isModToken` did not recognize it as a `-mod` flag.

`forceModMod` then concluded no module-mode flag was present and appended `-mod=mod`, producing:

```text
'-mod=readonly' -mod=mod
```

Because Go applies last-value-wins for a repeated flag, the effective mode flipped from `readonly` to `mod`, overriding the user's choice and defeating the function's own documented intent to preserve `-mod=readonly` (which already disables vendoring).

The fix matches on the unquoted form of each token, via a small `unquoteGoflagsToken` helper that strips one pair of surrounding matching quotes the way the go command does, while keeping the original token in the output. A quoted `-mod` flag is now recognized and preserved verbatim, and a quoted `-mod=vendor` is still rewritten to `-mod=mod`.

## Tests

`TestForceModMod` gains cases for single- and double-quoted `-mod=readonly` (preserved), a quoted `-mod` among other flags (preserved), and a quoted `-mod=vendor` (still overridden). I verified they fail against current `main` and pass with this change. This is tool-only; no bundle regeneration is required.

Fixes #1267

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:

### DIFF
--- a/tool/internal/setup/vendor.go
+++ b/tool/internal/setup/vendor.go
@@ -67,14 +67,20 @@ func isModToken(tok string) bool {
 // vendoring, so we respect the user's intent). -mod=mod is appended only when
 // no -mod/--mod token is present at all.
 func forceModMod(goflags string) string {
-	fields := strings.Fields(goflags)
+	// SplitGoflags keeps quoted tokens (such as -tags='foo bar') intact, unlike
+	// strings.Fields which would split them on the inner space.
+	fields := util.SplitGoflags(goflags)
 	hasMod := false
 	for i, f := range fields {
-		switch {
-		case isModVendorToken(f):
+		// The go command strips one layer of surrounding quotes from a GOFLAGS
+		// token before interpreting it, so a quoted flag such as '-mod=readonly'
+		// is a -mod flag. Match on the unquoted form, but keep the original
+		// token in the output so an existing -mod choice is preserved verbatim.
+		switch tok := util.UnquoteGoflagsToken(f); {
+		case isModVendorToken(tok):
 			fields[i] = modMod
 			hasMod = true
-		case isModToken(f):
+		case isModToken(tok):
 			hasMod = true
 		}
 	}
--- a/tool/internal/setup/vendor_test.go
+++ b/tool/internal/setup/vendor_test.go
@@ -91,6 +91,10 @@ func TestForceModMod(t *testing.T) {
 		// following -mod flag is still recognized and not clobbered.
 		{"spaced quoted flag before quoted mod", "-tags='foo bar' '-mod=readonly'", "-tags='foo bar' '-mod=readonly'"},
 		{"spaced quoted flag no mod appends once", "-tags='foo bar'", "-tags='foo bar' -mod=mod"},
+		// The go command strips only a surrounding quote pair, so -mod='vendor' keeps
+		// its quoted value and is not a -mod=vendor token. It stays as written and go
+		// rejects the value later.
+		{"value-quoted vendor preserved; go rejects it later", "-mod='vendor'", "-mod='vendor'"},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/tool/internal/setup/vendor_test.go
+++ b/tool/internal/setup/vendor_test.go
@@ -78,6 +78,19 @@ func TestForceModMod(t *testing.T) {
 		{"double-dash mod left unchanged, no append", "--mod=mod", "--mod=mod"},
 		{"double-dash readonly left unchanged, no append", "--mod=readonly", "--mod=readonly"},
 		{"bare double-dash mod left as is, no append", "--mod", "--mod"},
+		// The go command accepts quoted GOFLAGS tokens. A quoted -mod flag is
+		// still a -mod flag, so it must be recognized and not have -mod=mod
+		// appended after it, which would win by last-value.
+		{"single-quoted readonly left unchanged", "'-mod=readonly'", "'-mod=readonly'"},
+		{"double-quoted readonly left unchanged", `"-mod=readonly"`, `"-mod=readonly"`},
+		{"quoted readonly among flags left unchanged", "-trimpath '-mod=readonly'", "-trimpath '-mod=readonly'"},
+		{"quoted vendor overridden", "'-mod=vendor'", "-mod=mod"},
+		{"quoted double-dash readonly left unchanged", "'--mod=readonly'", "'--mod=readonly'"},
+		{"quoted double-dash vendor overridden", "'--mod=vendor'", "-mod=mod"},
+		// A quoted flag whose value contains a space stays one token, so the
+		// following -mod flag is still recognized and not clobbered.
+		{"spaced quoted flag before quoted mod", "-tags='foo bar' '-mod=readonly'", "-tags='foo bar' '-mod=readonly'"},
+		{"spaced quoted flag no mod appends once", "-tags='foo bar'", "-tags='foo bar' -mod=mod"},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/tool/util/go.go
+++ b/tool/util/go.go
@@ -92,12 +92,12 @@ func IsCgoCommand(line string) bool {
 		!strings.Contains(line, "-dynimport")
 }
 
-// splitGoflags splits a GOFLAGS value into tokens like the go command does
+// SplitGoflags splits a GOFLAGS value into tokens like the go command does
 // (https://cs.opensource.google/go/go/+/master:src/cmd/internal/quoted/quoted.go)
 //
 // space-separated, but a token starting with a quote runs to the matching close quote.
 // Quotes are kept so tokens re-join verbatim.
-func splitGoflags(goflags string) []string {
+func SplitGoflags(goflags string) []string {
 	isSpace := func(c byte) bool {
 		return c == ' ' || c == '\t' || c == '\n' || c == '\r'
 	}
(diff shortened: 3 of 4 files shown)


tool/internal/setup/vendor.go
```diff
@@ -70,11 +70,15 @@ func forceModMod(goflags string) string {
 	fields := strings.Fields(goflags)
```
darkraider01: Can we use splitGoflags from tool/util here instead of strings.Fields? strings.Fields splits on spaces inside quoted flags like '-tags=foo bar', whereas splitGoflags handles quotes properly.

tool/internal/setup/vendor.go
```diff
@@ -84,6 +88,22 @@ func forceModMod(goflags string) string {
 	return strings.Join(fields, " ")
 }
 
+// unquoteGoflagsToken removes one pair of surrounding matching quotes from a
+// GOFLAGS token, mirroring how the go command interprets it. A token without a
+// matching surrounding quote pair is returned unchanged.
+// quotedTokenMinLen is the shortest token that can hold a surrounding quote
+// pair: one opening quote and one closing quote.
+const quotedTokenMinLen = 2
```
darkraider01: We have the same unquoting logic in util.StripToolexecFromGoflags. Could we export a helper in tool/util and reuse it in both places?

tool/internal/setup/vendor.go
```diff
@@ -70,11 +70,15 @@ func forceModMod(goflags string) string {
...
+		// is a -mod flag. Match on the unquoted form, but keep the original
+		// token in the output so an existing -mod choice is preserved verbatim.
+		switch tok := unquoteGoflagsToken(f); {
+		case isModVendorToken(tok):
 			fields[i] = modMod
 			hasMod = true
-		case isModToken(f):
+		case isModToken(tok):
 			hasMod = true
 		}
```
darkraider01, line 85: how about we also add test cases for quoted --mod (e.g. "'--mod=readonly'") and flags with spaces (e.g. "-tags='foo bar' '-mod=readonly'"?

vignesh2027:
Thanks @darkraider01, all three make it better and they are in the latest push:

- forceModMod now uses SplitGoflags instead of strings.Fields, so a quoted flag with an inner space like -tags='foo bar' stays a single token.
- Pulled the unquoting into an exported util.UnquoteGoflagsToken and reused it in StripToolexecFromGoflags, so the two no longer carry duplicate logic.
- Added cases for quoted --mod (readonly and vendor) and for a spaced quoted flag before a quoted -mod.

golangci-lint and the setup and util packages are green locally.

tool/internal/setup/vendor_test.go
```diff
@@ -78,6 +78,19 @@ func TestForceModMod(t *testing.T) {
...
+		{"single-quoted readonly left unchanged", "'-mod=readonly'", "'-mod=readonly'"},
+		{"double-quoted readonly left unchanged", `"-mod=readonly"`, `"-mod=readonly"`},
+		{"quoted readonly among flags left unchanged", "-trimpath '-mod=readonly'", "-trimpath '-mod=readonly'"},
+		{"quoted vendor overridden", "'-mod=vendor'", "-mod=mod"},
+		{"quoted double-dash readonly left unchanged", "'--mod=readonly'", "'--mod=readonly'"},
+		{"quoted double-dash vendor overridden", "'--mod=vendor'", "-mod=mod"},
+		// A quoted flag whose value contains a space stays one token, so the
+		// following -mod flag is still recognized and not clobbered.
+		{"spaced quoted flag before quoted mod", "-tags='foo bar' '-mod=readonly'", "-tags='foo bar' '-mod=readonly'"},
+		{"spaced quoted flag no mod appends once", "-tags='foo bar'", "-tags='foo bar' -mod=mod"},
```
darccio, line 93: Add this:

```suggestion
		{"spaced quoted flag no mod appends once", "-tags='foo bar'", "-tags='foo bar' -mod=mod"},
		{"value-quoted vendor preserved; go rejects it later", "-mod='vendor'", "-mod='vendor'"},
```

tool/util/go.go
```diff
@@ -151,6 +147,23 @@ func StripToolexecFromGoflags(goflags string) string {
...
 
+// quotedTokenMinLen is the shortest GOFLAGS token that can hold a surrounding
+// quote pair: one opening quote and one closing quote.
+const quotedTokenMinLen = 2
+
+// UnquoteGoflagsToken removes one pair of surrounding matching quotes from a
+// GOFLAGS token, mirroring how the go command interprets it before acting on
+// the flag. A token without a matching surrounding quote pair is returned
+// unchanged. The token should already have been produced by SplitGoflags.
+func UnquoteGoflagsToken(token string) string {
```
darccio, line 158: `UnquoteGoflagsToken` has no direct test. `TestForceModMod` and
`TestStripToolexecFromGoflags` cover the helper only through its consumers.
No test pins the contract directly: a mismatched pair (`'x"`) returns the
token unchanged, a one-character token returns unchanged, and a double
layer (`"'x'"`) strips one layer only.

A table test in `tool/util/go_test.go` would protect that contract:

```go
func TestUnquoteGoflagsToken(t *testing.T) {
	tests := []struct {
		name  string
		token string
		want  string
	}{
		{"unquoted token unchanged", "-mod=mod", "-mod=mod"},
		{"single-quoted token stripped", "'-mod=readonly'", "-mod=readonly"},
		{"double-quoted token stripped", `"-mod=readonly"`, "-mod=readonly"},
		{"mismatched pair unchanged", "'-mod=readonly\"", "'-mod=readonly\""},
		{"double layer strips one layer", `"'-mod=readonly'"`, "'-mod=readonly'"},
		{"one-character token unchanged", "'", "'"},
		{"empty token unchanged", "", ""},
	}
	for _, tt := range tests {
			if got := UnquoteGoflagsToken(tt.token); got != tt.want {
				t.Errorf("UnquoteGoflagsToken(%q) = %q, want %q", tt.token, got, tt.want)
			}
		})
	}
}
```

darccio (changes requested):

vignesh2027:
Thanks, both added.

- `TestUnquoteGoflagsToken` now pins the contract directly rather than only through its callers, covering the mismatched pair, the one-character and empty tokens, and the doubly quoted token losing one layer only. I used the table style already in `go_test.go` with testify, to match the file.
- Added the `-mod='vendor'` case. It stays as written because the go command strips only a surrounding quote pair, so the value keeps its quotes and the token is not `-mod=vendor`.

The earlier points from @darkraider01 were already in: `forceModMod` uses `util.SplitGoflags`, and the unquoting logic is the exported `util.UnquoteGoflagsToken` shared with `StripToolexecFromGoflags`.

tool/internal/setup/vendor_test.go
vignesh2027: Added. The spaced quoted case was already in the table, so I added the value-quoted one alongside it:

```go
{"value-quoted vendor preserved; go rejects it later", "-mod='vendor'", "-mod='vendor'"},
```

with a short note on why it is not a `-mod=vendor` token: the go command strips only a surrounding quote pair, so the value keeps its quotes and go rejects it later rather than us rewriting it.

tool/util/go.go
vignesh2027: Added `TestUnquoteGoflagsToken` in `tool/util/go_test.go`, covering all seven cases you listed: the unquoted token, both quote styles, the mismatched pair, the double layer stripping one layer only, the one-character token and the empty token.

One deviation from your snippet: I used the table plus testify form the rest of that file already uses rather than `t.Errorf`, so it reads the same as its neighbours like `TestQuoteGoflagsToken`. Happy to switch it if you would rather keep the plain form.

darccio (approved):

PR #1284, opened 2026-08-30
fix(tool/internal/instrument): stop GetParam/GetReturnVal panicking for ordinary params on generic functions

### LINKED ISSUE #1280: bug(tool/internal/instrument): ctx.GetParam/GetReturnVal panics when the target function is generic
### Describe the bug

If you write a rule that hooks a generic function (one with a type parameter, like `func Query[T any](...)`), calling `ctx.GetParam()`, `ctx.SetParam()`, `ctx.GetReturnVal()`, or `ctx.SetReturnVal()` inside your hook panics at runtime, no matter what index you pass. These four methods seem to be hardcoded to panic for any generic target instead of actually doing anything.

### Environment

- OS: Windows 11
- Version: current `main` branch

### To Reproduce

1. Add a generic function to instrument:
```go
func Query[T any](ctx context.Context, query string) (T, error) {
    var zero T
    return zero, nil
}
```
2. Add a hook rule targeting it:
```yaml
query_hook:
  target: mypkg
  where:
    func: Query
  do:
    - inject_hooks:
        after: AfterQuery
        path: "myinstrumentation"
```
3. In the hook, try to read the return value the normal way:
```go
func AfterQuery(ctx hook.HookContext, ret error) {
    err, _ := ctx.GetReturnVal(1).(error)
    if err != nil {
        // record error on span
    }
}
```
4. Build and run the app, and call `Query`.

### Expected behavior

Expected `GetReturnVal`/`GetParam` to just work and give me the value, same as it does for non-generic functions.

### Additional context

Actual output:
```
panic: GetReturnVal is unsupported for generic functions
```

There's a workaround: skip `GetReturnVal`/`GetParam` and just add the value as a plain parameter on the hook function itself (it comes through as `interface{}`). That works, but it's not how any other hook in this repo is written, they all use `GetParam`/`GetReturnVal`, so this is easy to trip over.

Looks like this comes from `tool/internal/instrument/trampoline.go`, in `rewriteHookContextMethods` , when the target function is generic, it swaps all 4 methods for a hardcoded panic instead of generating real code. Might only be worth fixing for the simple case (return/param type is the plain type parameter itself, not something like `[]T` or `map[string]T`) since that part looks a lot trickier.

darkraider01:
## Description

Fixes #1280.

`rewriteHookContextMethods` blanket-panicked all four `HookContextImpl` accessor methods (`GetParam`, `SetParam`, `GetReturnVal`, `SetReturnVal`) whenever the target function had a type parameter, even for parameters/return values whose own type has nothing to do with the generic. For example:

```go
func Query[T any](ctx context.Context, query string) (T, error)
```

Calling `ctx.GetReturnVal(1)` to read the `error` panicked with `"GetReturnVal is unsupported for generic functions"`, even though index 1 is an ordinary `error`.

## Fix

Only the slots whose type actually mentions the type parameter (bare `T`, or nested in something like `*GenStruct[T]`) now panic, per-index, with a message naming that index. Every other slot generates the exact same accessor code as a non-generic function.

This is intentionally scoped to the common case: a parameter/return value whose type mentions a type parameter still isn't readable via `GetParam`/`GetReturnVal` (that would need type information that isn't in scope in `HookContextImpl`'s own, non-generic methods), so those indices still panic — just individually instead of taking every other parameter down with them.

## Testing

- Added `TestReferencesTypeParameter` covering the new helper (bare type param, pointer/slice/generic-instantiated composites, unrelated concrete types, nil inputs).
- Updated the `generic-functions` and `generic-before-only` golden fixtures to the new expected output.
- `go test ./tool/...` passes.

## Checklist

- [x] Tests pass: `make test`

### DIFF
--- a/demo/app/basic/instrumentation/basic_hook.go
+++ b/demo/app/basic/instrumentation/basic_hook.go
@@ -101,6 +101,29 @@ func MyHookGenericAfter(ictx hook.HookContext, _ interface{}) {
 	ictx.SetReturnVal(0, 999)
 }
 
+// MyHookQueryBefore/MyHookQueryAfter target Query[T any](ctx context.Context,
+// query string) (T, error): a generic function whose ctx/query params and
+// error return don't mention T. GetParam/GetReturnVal must work on those
+// ordinary slots instead of panicking just because the function is generic,
+// while the slot that does mention T (return index 0) must still panic.
+func MyHookQueryBefore(ictx hook.HookContext, _ context.Context, _ string) {
+	println("Query before hook")
+	fmt.Printf("[Query] GetParam(0): %v\n", ictx.GetParam(0))
+	fmt.Printf("[Query] GetParam(1): %v\n", ictx.GetParam(1))
+}
+
+func MyHookQueryAfter(ictx hook.HookContext, _ interface{}, _ error) {
+	println("Query after hook")
+	fmt.Printf("[Query] GetReturnVal(1): %v\n", ictx.GetReturnVal(1))
+
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Printf("[Query] GetReturnVal(0) panic (expected): %v\n", r)
+		}
+	}()
+	ictx.GetReturnVal(0)
+}
+
 func BeforeUnderscore(ictx hook.HookContext, _ int, _ float32) {
 	println("Underscore")
 }
--- a/demo/app/basic/instrumentation/otelc.yaml
+++ b/demo/app/basic/instrumentation/otelc.yaml
@@ -122,6 +122,16 @@ hook_generic:
         after: MyHookGenericAfter
         path: "go.opentelemetry.io/otelc/demo/app/basic/instrumentation"
 
+hook_query:
+  target: main
+  where:
+    func: Query
+  do:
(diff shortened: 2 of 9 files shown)


om7057 (commented):
Read through `trampoline.go` on `main` around the changed hunks and checked it against #1122. The implementation looks correct: the per-index `panicClause` keeps ordinary `int`/`error` slots reachable, while `referencesTypeParameter` only affects slots that actually reference a type parameter. The golden output also matches the expected `switch idx` pattern and compiles.

Noticed a few things:

**1. The fixed path isn't covered at runtime.**

The golden tests only compile the generated code, and `verifyGenericHookContextLogs` still exercises only the panic path. The generic fixtures don't currently have ordinary params/results to verify.

It would be useful to add something like `Query[T any](ctx context.Context, query string) (T, error)` from #1280 and assert `GetParam` and `GetReturnVal`. This would also cover the return-position case from the original issue.

**2. `referencesTypeParameter`'s comment isn't quite accurate.**

It says the function mirrors `replaceTypeParamsWithAny` "exactly", but it also recurses into `InterfaceType` methods, which the latter doesn't. The behavior looks fine, but the comment should reflect that difference.

Issue #1183, opened 2026-08-20
bug: k8s.informer span duration does not reflect actual handler execution time


om7057:
### Describe the bug

The `k8s.informer.<kind>.add` / `.update` / `.delete` spans created by the k8s.io/client-go instrumentation do not reflect the actual execution time of the registered event handler. The span wraps a call into the informer's own dispatch code, which only enqueues the notification onto each listener's channel. The real handler runs later, in a separate goroutine, outside the span entirely. I confirmed this by giving a handler a deliberate 500ms delay and the span duration did not change.

### Environment

- OS: Ubuntu 22.04 (Jammy), x86_64
- Go Version: 1.26.0
- Version: built from main, commit 128eb42 (run `./otelc version` for the exact string if needed)

### To Reproduce

Small app using `k8s.io/client-go`'s fake clientset, a shared informer for Pods, and two `ResourceEventHandler`s registered with `AddEventHandler`:

cat /tmp/k8s-repro/baseline/main.go

<img width="891" height="781" alt="Image" src="https://github.com/user-attachments/assets/f21d0263-8402-44da-814e-5c4871105b9b" />

1. Build the app with otelc and run it, triggering a single Pod Add through the fake clientset:

```bash
OTEL_TRACES_EXPORTER=console otelc go build -o app .
OTEL_TRACES_EXPORTER=console ./app
```

Baseline run, no artificial delay:

<img width="1799" height="707" alt="Image" src="https://github.com/user-attachments/assets/719567f1-7dd0-4417-817d-c4b7a4cd311e" />

2. Note the StartTime/EndTime on the printed k8s.informer.pod.add span.
3. Add a time.Sleep(500 * time.Millisecond) inside one handler before it does its real work:

<img width="895" height="813" alt="Image" src="https://github.com/user-attachments/assets/f6f82a16-81a8-4e5a-af44-6f449b422a23" />

4. Rebuild and run again:

<img width="1838" height="507" alt="Image" src="https://github.com/user-attachments/assets/91661d8a-6aa3-4b7c-820a-81a77ef076b4" />

5. Compare the span duration between the two runs. It does not change.

<img width="1180" height="105" alt="Image" src="https://github.com/user-attachments/assets/e9a10498-bacb-41a6-94fc-ba749a4346a6" />

I have a small repro app I can share if useful.

Expected behavior
[...]

PR #1288, opened 2026-08-31
refactor(instrumentation/k8s.io/client-go): cache GVK lookup per object type

om7057:
## Description

`getObjectInfo` calls `scheme.Scheme.ObjectKinds` on every `OnAdd`/`OnUpdate`/`OnDelete` event to resolve the object's GroupVersionKind. That's a reflection-based registry scan, running in the informer's hottest path, even though a given informer only ever watches a single Go type for its lifetime, so the result never changes between events for that informer.

This caches the GVK lookup per `reflect.Type`, computed once and reused for every subsequent event of that type.

Suggested by @amazingakai on #1183, independent of the open question in that thread about moving span creation to hook `AddEventHandler` instead of the internal `processDeltas`/`processDeltasInBatch` functions. This is scoped to only the caching improvement so it doesn't need to wait on that decision.

## Changes

- `lookupGVK` caches the result of `scheme.Scheme.ObjectKinds` in a `sync.Map`, keyed by `reflect.Type`.
- Both hits and misses are cached. An unregistered type resolves the same way on every call, so caching the miss avoids re-scanning, and re-logging, for a type that will never resolve.
- `objectKindsFunc` indirects the real `scheme.Scheme.ObjectKinds` so tests can substitute a counting stub, proving the cache actually avoids repeat lookups rather than only asserting the returned values are correct.

## Testing

- `TestLookupGVK_CachesPerType`: two different instances of the same type only trigger one real lookup.
- `TestLookupGVK_CachesDifferentTypesSeparately`: two distinct types each trigger their own lookup, and don't share cached results.
- `TestLookupGVK_CachesMisses`: an unregistered type's failed lookup is also cached, not retried every call.
- Full existing package test suite passes unchanged, `getObjectInfo`'s returned values are identical to before, this only changes how often the underlying scan runs.

## Checklist

- [x] Tests pass: `go test ./...` in `instrumentation/k8s.io/client-go`
- [ ] Documentation updated (if applicable) (n/a, internal implementation detail)
- [ ] OpenTelemetry Registry updated (if applicable) (n/a, no new instrumentation)


instrumentation/k8s.io/client-go/k8s_otel_handler.go
```diff
@@ -89,6 +91,48 @@ func getSpanName(kind, action string) string {
...
+}
+
+// objectKindsFunc is scheme.Scheme.ObjectKinds, indirected so tests can
+// substitute a counting stub to verify the cache actually avoids repeat
+// lookups rather than just asserting the returned values are correct.
+var objectKindsFunc = scheme.Scheme.ObjectKinds
+
+func lookupGVK(runtimeObj runtime.Object) gvkLookup {
+	t := reflect.TypeOf(runtimeObj)
+	if cached, hit := gvkCache.Load(t); hit {
```
amazingakai, line 119: Using `reflect.TypeOf` isn't a good idea because two different CRDs using dynamic informers may be both of same type (`unstructured.Unstructured`)

amazingakai (changes requested):
I was thinking of doing this after (or together with) we migrate to `AddEventHandler` because we can simply tie each handler to a wrapper and cache the gvk within the wrapper struct, that completely avoids the `unstructured` bug as well and we wouldn't need a `sync.Map`. WDYT?

om7057:
> I was thinking of doing this after (or together with) we migrate to `AddEventHandler` because we can simply tie each handler to a wrapper and cache the gvk within the wrapper struct, that completely avoids the `unstructured` bug as well and we wouldn't need a `sync.Map`. WDYT?

Thanks for taking a look at this.
Sure, that's why I was a bit hesitant to open the PR considering the current priorities, but I thought it was a standalone fix, so I opened it after I completed verifying the results locally. No worries if I need to hold this off till the migration to `AddEventHandler` is completed, which will then be much cleaner to add.

cc: @amazingakai

instrumentation/k8s.io/client-go/k8s_otel_handler.go
om7057: Sure, will use the wrapper-struct approach after the migration completes.

amazingakai:
My suggestion is to fix this after/with the AddEventHandler migration

om7057:
> My suggestion is to fix this after/with the AddEventHandler migration

@amazingakai Works! So should I keep this in draft or let it stay open?

om7057:
Converting to draft since this is on hold for the AddEventHandler migration, will mark ready again once that lands or if we decide to take it standalone.

om7057:
Marking this ready again. @amazingakai answered on #1260 that the freeze hasn't actually started yet (it kicks in once contrib#9 merges), and that bug fixes are still wanted merged before that happens. This is small and self-contained, so figured it's worth another look now rather than waiting.

PR #1291, opened 2026-08-31
docs(schemas): backfill missing coverage rows for aws and linodego

darkraider01:
`schemas/otelc/README.md` says a missing row in the coverage table means an undeclared instrumentation. `aws.yaml` and `linodego.yaml` both existed with no row, and `linodego.yaml` was missing from the tree diagram too. This adds both, so the table actually matches what's in `schemas/otelc/groups/`.


Labreo (approved):
lgtm.

y1yang0 (approved):

PR #1294, opened 2026-09-01
chore(tool): drop redundant comments on rule struct fields

### LINKED ISSUE #1290: Remove comments on rule fields
This is a simple cleanup task.

Since rules.md already precisely defines the usage and meaning of each rule field, repeating them in the code is redundant. We could consider removing them:
```
type InstCallRule struct {
    InstBaseRule `yaml:",inline"`
    ImportPath string `json:"import-path" yaml:"-"`
    FuncName string `json:"func-name" yaml:"-"`
    Replace string `json:"replace" yaml:"replace"`
    AppendArgs []string `json:"append_args" yaml:"append_args"`
    VariadicType string `json:"variadic_type" yaml:"variadic_type"`
}
```
Also, the fields don't need the extra blank lines between them.

darkraider01:
Closes #1290

rules.md already spells out what every rule field means, so the per-field comments on the InstXxxRule structs were just repeating it, plus some had gone stale (the has_directive comment claimed it wasn't executed yet, but it is).

Dropped those comments and the blank lines that were only there to separate them, so the structs read as compact tables. The struct-level doc comments with the YAML examples are untouched, those still earn their keep.

A few fields (ImportPath, FuncName, ResolvedPath, TypeName) are derived internally rather than part of the YAML schema, so rules.md doesn't cover them. Kept a short trailing note on those instead of dropping them silently.

### DIFF
--- a/tool/internal/rule/base.go
+++ b/tool/internal/rule/base.go
@@ -33,41 +33,15 @@ type InstRule interface {
 // support remains intentionally narrow: simple leaf predicates are supported,
 // while qualifier composition is validated but not yet executed.
 type FilterDef struct {
-	// AllOf, OneOf, and Not are the boolean combinators. Each composes nested
-	// FilterDefs so predicates can be combined. A node may use at most one
-	// combinator, and a combinator may not be mixed with sibling leaf predicates
-	// on the same node; both rules are enforced at build time.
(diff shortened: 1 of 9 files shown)


y1yang0 (approved):
Thanks for doing this.

PR #1296, opened 2026-09-01
feat(instrumentation): adds zerolog hook injecting trace & span ids

1000p:
## Description

Add compile-time instrumentation for [zerolog](https://github.com/rs/zerolog) package to automatically inject trace_id and span_id into log output when an active span exists in GLS.

## Motivation

Fixes #

---

## Checklist

- [x] Tests pass: `make test`

[^I-know-my-stuff]:


test/integration/logs_test.go
```diff
@@ -77,3 +77,37 @@ func TestLogsLogrus(t *testing.T) {
...
+
+func TestLogsZerolog(t *testing.T) {
+	t.Parallel()
+	testutil.Build(t, "", "logszerolog", "go", "build", "-a")
+
+	f := testutil.NewTestFixture(t, testutil.WithoutCollector())
+	output := f.Run("logszerolog")
+
+	zerologMessages := []string{
+		"trace message",
```
darkraider01, line 117: This trace message check is going to fail. zerolog's default global level is Debug, and Trace sits below that, so `logger.Trace()` never actually emits anything unless `zerolog.SetGlobalLevel(zerolog.TraceLevel)` gets called somewhere. Might be worth adding that call in `test/apps/logszerolog/main.go`, or just dropping trace message from this list.

instrumentation/github.com/rs/zerolog/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+The instrumentation is enabled by default and can be configured with the `logs/zerolog` key in `OTEL_GO_ENABLED_INSTRUMENTATIONS` or `OTEL_GO_DISABLED_INSTRUMENTATIONS`.
+
+## How trace and span IDs are attached
+
+The hook calls `runtime.GetTraceAndSpanID()`, which returns the current goroutine's active trace and span IDs from goroutine-local storage (GLS), or two empty strings if none is available. If the trace ID is empty, the event is left unchanged. The span ID is added only when it is non-empty.
+
+**`go.opentelemetry.io/otel/sdk/trace` must already be part of the application's build dependency graph (as seen by `go list -deps` / `go build -a -x -n`) for this instrumentation to inject `trace_id`/`span_id`.**
+
+Instrumentation imports declared in an `otel.instrumentation.go` file are tracked in `go.mod` but are intentionally **not** added to the build's dependency graph — that file is marked with the `//go:build tools` build tag specifically so instrumentation imports do not become build dependencies themselves. So relying on `logrus`'s own instrumentation module to pull in `sdk/trace` does not work.
```
darkraider01: Looks like this got copied from the logrus README and the find and replace missed a spot, it still says logrus here instead of zerolog.

instrumentation/github.com/rs/zerolog/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+In practice, all that's needed is a plain blank import of the package somewhere in your own application source, outside of any `//go:build tools`-tagged file:
+
+```go
+import _ "go.opentelemetry.io/otel/sdk/trace"
+```
+
+That's enough to put it in the build graph — you don't need to actually construct or use a `TracerProvider` yourself. Without it, spans are still created and exported correctly, but `logrus` output will not carry `trace_id`/`span_id`.
```
darkraider01: Same thing here, this says logrus output instead of zerolog output.

instrumentation/github.com/rs/zerolog/zerolog_hook.go
```diff
@@ -0,0 +1,50 @@
...
+		return
+	}
+
+	event.Str(traceIDKey, traceID)
+	if spanID != "" {
+		event.Str(spanIDKey, spanID)
+	}
+}
+
+func AfterZerologNew(ictx hook.HookContext, logger zerolog.Logger) {
```
darkraider01: Small note, this only hooks zerolog.New, so a logger built any other way (zero value Logger, zerolog.Nop, etc) won't get the trace_id/span_id fields. Probably fine for most cases since New is the common path, just flagging it since logrus needed a few hook points to cover something similar.

schemas/otelc/README.md
```diff
@@ -59,7 +59,7 @@ signals an undeclared instrumentation.
 | `go.opentelemetry.io/otel`                          | `otel-sdk.yaml`     | nothing — guards the global tracer provider                            |
 | `go.opentelemetry.io/otel/sdk/trace`                | `otel-sdk.yaml`     | nothing — maintains the GLS span chain                                 |
 | `go.opentelemetry.io/otel/trace`                    | `otel-sdk.yaml`     | nothing — GLS fallback for `SpanFromContext`                           |
-| `log`, `log/slog`, `github.com/sirupsen/logrus`     | `logs.yaml`         | nothing — injects `trace_id`/`span_id` into the application's own logs |
+| `log`, `log/slog`, `github.com/rs/zerolog`, `github.com/sirupsen/logrus` | `logs.yaml`         | nothing — injects `trace_id`/`span_id` into the application's own logs |
```
darkraider01: Minor, every other row in this table is padded so the columns line up, this new row is longer and throws off the alignment when you look at the raw file. Doesn't change how it renders, just easy to fix while it's fresh.

instrumentation/github.com/rs/zerolog/zerolog_hook.go
```diff
@@ -0,0 +1,50 @@
...
+	"go.opentelemetry.io/otelc/pkg/runtime"
+)
+
+const (
+	instrumentationKey = "logs/zerolog"
+	traceIDKey         = "trace_id"
+	spanIDKey          = "span_id"
+)
+
+type logEnabler struct{}
```
darkraider01, line 18: Not really specific to this PR since log, slog and logrus all do the same thing, but this struct is just wrapping one function call, could be a plain function instead. Flagging it since it's now repeated in four places.

instrumentation/github.com/rs/zerolog/zerolog_hook.go
```diff
@@ -0,0 +1,50 @@
...
+func (l logEnabler) Enable() bool {
+	return runtime.Instrumented(instrumentationKey)
+}
+
+var enabler = logEnabler{}
+
+type traceHook struct{}
+
+func (traceHook) Run(event *zerolog.Event, level zerolog.Level, message string) {
+	if !enabler.Enable() || event == nil {
```
darkraider01, line 43: Also not new to this PR, but heads up that Enable() re-reads and re-parses both env vars from scratch on every single log call, no caching involved. Same as logrus already does, just compounds a bit more now that a 4th instrumentation does it too.

darkraider01 (commented):
Went through this one closely, really nice work overall. Left a few inline notes, one of them (the trace message check) will actually fail as written, the rest are smaller stuff.

instrumentation/github.com/rs/zerolog/zerolog_hook.go
darkraider01: maybe we can open a follow up pr touching this part and the comment below, @NameHaibinZhang , what do you think?

test/integration/logs_test.go
1000p: https://github.com/rs/zerolog#:~:text=Default%20level.,messages%20are%20suppressed.

Looks like this is not the case, and integration tests prove it.
Anyway, adding that explicitly makes sence so I have added it.

instrumentation/github.com/rs/zerolog/zerolog_hook.go
1000p: Great catch.
I added two new rules.
- `add_struct_fields` - adds a `sync.Once`;
- `newEvent` func hook - looks like this func is the one `zerolog` always uses for creating its events;

Removed the now redundant `AfterZerologNew` rule.

schemas/otelc/README.md
1000p: Formatted correctly, 10x!

darkraider01 (commented):
Really nice work addressing the last round, the trace level fix, the README typos, the table formatting, and especially catching every logger construction path via newEvent instead of just New, all solid. One new thing came up from that last change though, left details inline.

instrumentation/github.com/rs/zerolog/zerolog_hook.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+}
+
+func BeforeZerologNewEvent(_ hook.HookContext, logger *zerolog.Logger, _ zerolog.Level, _ func(string)) {
+	if !enabler.Enable() || logger == nil {
+		return
+	}
+
+	logger.OtelcInitOnce.Do(func() {
+		*logger = logger.Hook(traceHook{})
```
darkraider01: Heads up, this fix works fine for the actual logging calls like Trace, Debug, Info, the sync.Once takes care of those. But zerolog loggers get copied around a lot, things like With(), Level(), Output(), Sample() all just copy the whole logger with no locking involved. So if one goroutine is logging for the first time, which is what triggers this `*logger = logger.Hook(...)` line, at the same exact moment another goroutine copies that same logger through one of those other methods, you've got one thing writing to it while another reads it with nothing protecting either side. Best case nothing happens, worst case it crashes or the logger ends up in a broken state. Won't show up in a single threaded test, but could show up once this is actually used in something like a web server where the same logger gets shared across a bunch of requests at once.

Didn't just want to say this without proof, so I built a tiny standalone repro with the exact same shape (a value type struct with a sync.Once field, one pointer method doing `Once.Do` plus an in place mutation, one value method just reading/copying) and ran it under `go run -race`. It caught it right away:

```
WARNING: DATA RACE
Read at 0x00c0000ba000 by goroutine 8:
  main.main.func2()
      /tmp/racerepro/main.go:37 +0x84

Previous write at 0x00c0000ba000 by goroutine 7:
  main.main.func1.(*Logger).newEvent.1()
      /tmp/racerepro/main.go:23 +0xd6
  sync.(*Once).doSlow()
      /usr/lib/go-1.25/src/sync/once.go:78 +0xd1
  sync.(*Once).Do()
      /usr/lib/go-1.25/src/sync/once.go:69 +0x44
  main.(*Logger).newEvent()
      /tmp/racerepro/main.go:22 +0xb4
  main.main.func1()
      /tmp/racerepro/main.go:33 +0x72

Goroutine 8 (running) created at:
  main.main()
      /tmp/racerepro/main.go:35 +0x1e8

Goroutine 7 (finished) created at:
  main.main()
      /tmp/racerepro/main.go:31 +0x146
==================
Found 1 data race(s)
```

The write is the `Once.Do` closure (same as `BeforeZerologNewEvent` here), the read is a `With()`-style value copy happening concurrently. Same shape as what's in this file, just stripped down.

schemas/otelc/groups/logs.yaml
```diff
@@ -1,11 +1,12 @@
...
 #
-# Covers the three logging-library instrumentations:
+# Covers the four logging-library instrumentations:
 #
-#   log         instrumentation/log/log_hook.go            (BeforeLogOutput)
-#   log/slog    instrumentation/log/slog/slog_hook.go      (AfterSlogNewRecord)
+#   log             instrumentation/log/log_hook.go                           (BeforeLogOutput)
+#   log/slog        instrumentation/log/slog/slog_hook.go                     (AfterSlogNewRecord)
 #   sirupsen/logrus instrumentation/github.com/sirupsen/logrus/logrus_hook.go (traceHook.Fire)
+#   rs/zerolog      instrumentation/github.com/rs/zerolog/zerolog_hook.go     (AfterZerologNew)
```
darkraider01: Small thing, this comment still says AfterZerologNew but that function got removed in the rework, it's BeforeZerologNewEvent now.

instrumentation/github.com/rs/zerolog/otelc.yaml
```diff
@@ -0,0 +1,26 @@
...
+    struct: Logger
+  do:
+    - add_struct_fields:
+        new_field:
+          - name: OtelcInitOnce
+            type: "sync.Once"
+  imports:
+    sync: "sync"
+
+hook_zerolog_new_event:
```
darkraider01: Noticed the version constraint got dropped here too. Might be worth adding one back, especially now since this hooks newEvent which is unexported, internal to zerolog. Public API is covered by their semver promises, internal stuff like this isn't, so a future zerolog release could rename or restructure it without warning and nothing here would catch that.

1000p: Added version constraint starting from where all public event finalizer methods are available.

instrumentation/github.com/rs/zerolog/zerolog_hook.go
1000p: Very good catch.
I'll just attach a hook for each public finalizer method of `zerolog.Event`.

1000p:
Hey @darkraider01 , can you take another look please 🙏

darkraider01:
@1000p hey!! Yea i did take a look, looks okay to me !!! Now let us wait for the maintainers to have a look on this PR

PR #1312, opened 2026-09-02
feat: get real import aliases instead of guessing

hannahkm:
## Description

On startup, otelc loads each dependency once and builds a table that maps its import path to its declared package name. otelc passes this table to every rule that needs it instead of guessing the name from the import path.

When a target file already imports a path under an alias a rule does not expect, otelc rewrites the rule's generated code to use the file's alias. otelc no longer fails the build over an alias mismatch.

If the table is unavailable, such as after an older setup run, otelc reverts to the previous guess.

## Motivation

A user can leave an import unaliased. For an unaliased import, otelc previously guessed the package name from the import path's last segment. For example, an unaliased import of `github.com/redis/go-redis/v9` declares the package name `redis`. The import path's last segment is `go-redis` instead. A rule that targets redis never matched this import. otelc silently skipped instrumentation for the dependency, with no error.

A rule also assumes its own alias for each import it uses. A target file can already import the same path under a different alias. otelc previously failed the build with an alias mismatch error in that case.


tool/internal/instrument/imports.go
```diff
@@ -98,3 +99,110 @@ func (ip *instrumentPhase) addRuleImports(
...
+	aliasOverrides := resolveAliasOverrides(ruleImports, existingAliases)
+	return importAliases, aliasOverrides
+}
+
+// resolveAliasOverrides reports the alias to substitute for each rule
+// import already present in the target file under a different alias.
+// Substituting the file's alias into generated code, instead of the
+// rule's alias, avoids a build failure.
+//
+// existingAliases must resolve an unaliased import to its real name,
```
vedant21-oss, line 127: The comment here says existingAliases has to be the real name and not a guess, since a guess can be an illegal identifier. But `resolveImportOverrides` passes `ast.ImportAliasMap(root, ip.importNames)`, and that still falls back to `defaultImportAlias` when a path isn't in the table, which the PR description says can happen after an older setup run.

I tried that case locally with an unaliased `github.com/redis/go-redis/v9` and got:

```
file aliases: {"go-redis": "github.com/redis/go-redis/v9"}
overrides:    {"redis": "go-redis"}
```

so the generated code would end up as `go-redis.Nil`, which doesn't compile because of the hyphen. Before this PR that case just didn't match, so it was a missed instrumentation but the build still worked.

Would it make sense to skip an override when the existing alias isn't a valid identifier (`token.IsIdentifier`), or to have `ImportAliasMap` tell you which names are real and which are guessed? Maybe I'm missing something and the table is always there by then, let me know.

tool/internal/ast/typename.go
```diff
@@ -167,6 +165,9 @@ func ImportAliasMap(file *dst.File) map[string]string {
 			continue
 		}
 		alias := defaultImportAlias(path)
+		if name, ok := resolvedNames[path]; ok && name != "" {
```
vedant21-oss, line 168: Small thing, just to be sure I understand the scope: this fixes the unaliased name, but two different paths that declare the same package name still overwrite each other on `aliases[alias] = path` below. I checked with `text/template` + `html/template` and even with the table I get only `{"template": "text/template"}`, so `html/template` is still lost.

That's the case in #1219 / #1271, and I have fixes open for it in #1220 and #1274 which touch this same function, so they conflict with this PR right now. Happy to rebase mine on top once this lands, just wanted to flag it so we don't both change the same thing.

vedant21-oss (commented):
This is a really nice change, the guessing has bitten a few rules already. I built it and ran the tool tests locally and everything passes. Two things I ran into while playing with it:

om7057 (commented):
This is a substantially larger change than the others I've reviewed recently, roughly thirty files, most of it test coverage, so I gave it a correspondingly closer look before commenting.

Pulled the branch down, and it builds and vets cleanly. I did hit test failures on the first run, but traced each one back to my own environment rather than the PR. The `tool/internal/setup` failures come from real `go mod tidy` invocations hitting a corporate proxy that isn't reachable from here, and a couple of `pkgload` path-comparison tests failed due to a `GOTMPDIR` I had set with mixed slash conventions. I confirmed both by running the same tests against unpatched `main` and getting identical failures there, so neither is attributable to this change. With the environment corrected, `ast`, `instrument`, and `pkgload` all pass without issue.

Rather than take the motivating bug on faith, I wrote a small standalone test against `ImportAliasMap` using an unaliased `github.com/redis/go-redis/v9` import. Without the resolved-names table, it guesses `go-redis`, the incorrect name described in the PR. Supplying the resolved name correctly returns `redis`. That confirms the core fix behaves as claimed.

The overall design holds together well. `CollectPackageNames` builds the import-path-to-package-name table once during setup by walking the dependency graph, persists it to `import_names.json`, and the instrument phase reads it back without needing a second `go/packages` load, which the code comments correctly note would be too expensive to repeat per compile process. The `resolveAliasOverrides`/`replaceQualifierAliases` handling for a file that already imports a path under a different alias is a genuine improvement as well; that case previously failed the build outright and now rewrites the generated code to match. The relocation of `usedRuleImports` from `apply_call.go` into the shared `imports.go` is a straightforward deduplication with no behavioral change.

tool/internal/instrument/imports.go
ADITYA-CODE-SOURCE: I agree that this is blocking. When `import_names.json` is missing after a setup run from an older version, `ImportAliasMap` may return a path-derived guess such as `go-redis`.

This PR then uses that guess as a source qualifier through `resolveAliasOverrides`, producing invalid generated code such as `go-redis.Nil`. The previous fallback could cause instrumentation to be missed, but it did not introduce a compilation failure.

Could we distinguish explicit or resolved aliases from path-derived guesses and avoid generating an override from a guessed name? A regression test with a missing name table and an unaliased `github.com/redis/go-redis/v9` import would cover this case.

ADITYA-CODE-SOURCE (changes requested):

tool/internal/instrument/imports.go
hannahkm: Thanks for the comments, and good points! I've just opened a follow up [PR](https://github.com/hannahkm/opentelemetry-go-compile-instrumentation/pull/7) to address this issue, since I was worried about the size of it. I can cherry pick the changes into this PR or merge it into the main later.

vedant21-oss: Nice, splitting it into ResolvedImportAliasMap for generated code and keeping the guess for matching is cleaner than what I suggested. Looks right to me

Issue #1314, opened 2026-09-02
bug(instrumentation/openai): requests over 1 MB produce no span at all


OfficialAbhinavSingh:
### Describe the bug

`OtelMiddleware` reads a bounded copy of the request body to extract the model and request attributes:

```go
const maxRequestBodySize = 1 << 20 // 1 MB
...
bodyBytes, err := io.ReadAll(io.LimitReader(tee, maxRequestBodySize))
```

A request larger than 1 MB is truncated before parsing. The truncated bytes fail JSON unmarshalling, so `parseChatRequest` returns an empty model, and the middleware takes this path:

```go
if model == "" {
    return next(req)
}
```

That `return` happens **before `tracer.Start`**. No span is created, no `gen_ai.client.operation.duration` is recorded, and nothing is logged. The operation disappears from telemetry entirely.

The API call itself is unaffected — the body is reassembled with `io.MultiReader` and reaches the SDK intact. This is purely a telemetry loss.

A 1 MB chat request is not exotic: a RAG context, a long conversation history, or a single base64-encoded image will exceed it.

### Environment

- OS: any
- Go version: 1.25.6
- Version: 907ac54

### To Reproduce

Drive `OtelMiddleware()` end to end against a real `tracetest.SpanRecorder` — no mocking — with a valid `chat/completions` body whose `messages` content pushes it past 1 MB:

```
prompt=1000     bytes  requestBody=1076     bytes  spans=1  model="gpt-4"
prompt=2097152  bytes  requestBody=2097228  bytes  spans=0  model=""
```

In both cases the full body reaches `next` intact (1076 and 2097228 bytes respectively), so the request is sent correctly either way.

### Expected behavior

A large but valid request should still produce a span with at least `gen_ai.operation.name`, `gen_ai.request.model`, `gen_ai.provider.name` and the duration metric. Losing optional request attributes on a huge body is reasonable; losing the entire operation is not.

### Additional context

This is the request-side counterpart of #1309, and it fails harder. That issue covers `maxResponseBodySize` (4 MB), where the span still exists and only response attributes go missing. Here the cap is 1 MB — four times easier to reach — and the span is never created, so the call is invisible rather than merely incomplete.

[...]

PR #1315, opened 2026-09-03
fix(instrumentation/openai): stop silently dropping telemetry on oversized bodies

### LINKED ISSUE #1309: bug: openai instrumentation silently drops response telemetry on large responses
### Describe the bug

The OpenAI instrumentation reads response bodies through a fixed 4MB cap (`maxResponseBodySize` in `middleware.go`). A large but completely normal response- a batch embeddings call over many inputs is the easiest way to hit this - gets truncated by that cap before parsing. The truncated bytes fail JSON unmarshalling, and the parse function just returns. No error, no log, nothing.

The span still gets created with correct request-side attributes and ends with status Unset, looking like a totally successful call. But none of the response data, model, token usage, or finish reasons ever gets attached. It just silently isn't there.

The actual API call to the SDK caller is unaffected; the full response body is untouched and reassembled correctly. This is purely a gap in what gets recorded for telemetry.

### Environment

- OS: any
- Go version: any supported version
- Version: 58f1bf0

### To Reproduce

Send a batch embeddings request that returns a response bigger than 4MB, easy to hit with a few hundred `text-embedding-3-large` vectors in one batch. I verified this by driving the real `OtelMiddleware()` function end-to-end in a test: a real 200 OK response, a real span recorder, no mocking:

legitimate response size: `36912998` bytes (35.2 MB)
span name: `embeddings text-embedding-3-large.`
span status: `{Unset}`
span attributes: `map[gen_ai.operation.name:embeddings gen_ai.provider.name:openai gen_ai.request.model:text-embedding-3-large gen_ai.system:openai] has gen_ai.response.model:false has gen_ai.usage.total_tokens:false`

Image

Request-side attributes are all there. `gen_ai.response.model` and `gen_ai.usage.total_tokens` are simply missing.

### Expected behavior

A large but valid response shouldn't silently drop all of its telemetry with no indication that anything went wrong. At minimum, this should be visible somehow: a log line, an attribute flagging truncation, something, rather than a span that looks fully successful while missing everything response-related.

### Additional context

Checked this against #866 and #970 first since both touch similar territory. Neither is the same bug. #866 is the request side (`maxRequestBodySize`), and only fires when the model can't be identified at all; in this case, the model is known from the small request body, so that guard never triggers. #970 is about the streaming path failing on a real network error mid-stream. This is the non-streaming path, on a fully successful response, no network failure involved at all, purely our own read cap discarding a valid payload.

The same mechanism applies to `parseChatResponse`/`parseCompletionResponse` too; a chat completion with a large `n` or a long completion can hit the same cap, just less commonly than embeddings.

rupeshkumar92a-arch:
## Summary

Fixes #1309 and #1314.

Both request and response paths in `instrumentation/github.com/openai/openai-go/middleware.go` (and its `v2`/`v3` copies) read the HTTP body through a bounded `io.LimitReader`. When a body exceeds the cap (e.g. a large-but-normal batch embeddings response, or a large chat request), the read is silently cut off, `json.Unmarshal` fails on the truncated bytes, and the parse functions quietly return nothing.

The result: a span that looks fully successful (correct request-side attributes, status `Unset`) while every response attribute (`gen_ai.response.model`, `gen_ai.usage.*`, etc.) is simply missing, with no error, log, or indication that anything was cut off.

## Changes

- Added a `readBounded` helper that reads one byte past the cap to detect truncation, while still reassembling the complete, untruncated body for the SDK/HTTP caller either way — no data is lost for the actual API call, only the telemetry parse is bounded.
- When a response body is truncated, skip the doomed parse attempt and log a warning (operation + cap size) instead of silently falling through.
- When a request body is truncated, same treatment: pass through un-instrumented (as before) but now with a logged reason instead of silence.
- Applied identically to the `v1`, `v2`, and `v3` copies of the instrumentation to keep them in sync.

No new span attributes were added — this repo governs `gen_ai.*` attributes through a Weaver semconv registry (`schemas/otelc/`), so a new attribute would need a registry change out of scope here. A log line satisfies the issue's own stated minimum bar.

## Test plan

- [x] `go build ./...`, `go vet ./...` pass for all three modules (v1/v2/v3)
- [x] Full existing test suite passes for all three modules
- [x] Added `TestOtelMiddleware_ResponseBodyTruncated` and `TestOtelMiddleware_RequestBodyTruncated` (mirrored across all three modules) asserting: the span still carries request-side attributes but no response attributes, the caller still receives the complete untruncated body, and the warning is logged
- [x] `gofmt -l` reports no formatting issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/instrumentation/github.com/openai/openai-go/v2/middleware.go
+++ b/instrumentation/github.com/openai/openai-go/v2/middleware.go
@@ -108,6 +108,28 @@ func operationName(op operationType) string {
 	}
 }
 
+// readBounded reads up to limit bytes from body for attribute parsing while
+// preserving the full stream for downstream consumers (the SDK or the HTTP
+// caller). truncated reports whether body held more than limit bytes, in
+// which case bodyBytes is a cut-off prefix and must not be treated as
+// complete JSON.
+func readBounded(body io.ReadCloser, limit int64) (bodyBytes []byte, truncated bool, reassembled io.ReadCloser, err error) {
+	var buf bytes.Buffer
+	tee := io.TeeReader(body, &buf)
+	// Read one byte past the limit so a full body can be told apart from one
+	// that was cut short. That extra byte is still captured by the tee, so
+	// reassembled below reproduces the stream in full either way.
+	read, err := io.ReadAll(io.LimitReader(tee, limit+1))
+	reassembled = struct {
+		io.Reader
+		io.Closer
+	}{io.MultiReader(&buf, body), body}
+	if int64(len(read)) > limit {
+		return read[:limit], true, reassembled, err
+	}
+	return read, false, reassembled, err
+}
+
 // OtelMiddleware returns an HTTP middleware that creates spans for OpenAI API
 // calls following GenAI semantic conventions.
 func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, error)) (*http.Response, error) {
@@ -126,15 +148,14 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 		opName := operationName(op)
 
 		// Read a bounded copy for attribute parsing, but preserve the full body for the SDK.
-		var buf bytes.Buffer
-		tee := io.TeeReader(req.Body, &buf)
-		bodyBytes, err := io.ReadAll(io.LimitReader(tee, maxRequestBodySize))
-		// Reassemble: buffered bytes + remaining unread body.
-		req.Body = struct {
-			io.Reader
-			io.Closer
-		}{io.MultiReader(&buf, req.Body), req.Body}
+		bodyBytes, truncated, reassembledBody, err := readBounded(req.Body, maxRequestBodySize)
+		req.Body = reassembledBody
 		if err != nil {
+			return next(req)
+		}
+		if truncated {
+			logger.Warn("openai request body exceeded size cap, skipping request attributes",
+				"operation", opName, "cap_bytes", maxRequestBodySize)
 			return next(req)
 		}
 
@@ -238,15 +259,14 @@ func handleNonStreamingResponse(
 	defer span.End()
 
 	// Read a bounded preview for parsing, but reassemble the full body for callers.
-	var buf bytes.Buffer
-	tee := io.TeeReader(resp.Body, &buf)
-	bodyBytes, err := io.ReadAll(io.LimitReader(tee, maxResponseBodySize))
-	// Reassemble: preview bytes + remaining unread body.
-	resp.Body = struct {
-		io.Reader
-		io.Closer
-	}{io.MultiReader(&buf, resp.Body), resp.Body}
+	bodyBytes, truncated, reassembledBody, err := readBounded(resp.Body, maxResponseBodySize)
+	resp.Body = reassembledBody
 	if err != nil {
+		return
+	}
+	if truncated {
+		logger.Warn("openai response body exceeded size cap, skipping response attributes",
+			"operation", operationName(op), "cap_bytes", maxResponseBodySize)
 		return
 	}
 
--- a/instrumentation/github.com/openai/openai-go/middleware.go
+++ b/instrumentation/github.com/openai/openai-go/middleware.go
@@ -108,6 +108,28 @@ func operationName(op operationType) string {
 	}
 }
 
+// readBounded reads up to limit bytes from body for attribute parsing while
+// preserving the full stream for downstream consumers (the SDK or the HTTP
+// caller). truncated reports whether body held more than limit bytes, in
+// which case bodyBytes is a cut-off prefix and must not be treated as
+// complete JSON.
+func readBounded(body io.ReadCloser, limit int64) (bodyBytes []byte, truncated bool, reassembled io.ReadCloser, err error) {
+	var buf bytes.Buffer
+	tee := io.TeeReader(body, &buf)
+	// Read one byte past the limit so a full body can be told apart from one
+	// that was cut short. That extra byte is still captured by the tee, so
+	// reassembled below reproduces the stream in full either way.
+	read, err := io.ReadAll(io.LimitReader(tee, limit+1))
+	reassembled = struct {
+		io.Reader
+		io.Closer
+	}{io.MultiReader(&buf, body), body}
+	if int64(len(read)) > limit {
+		return read[:limit], true, reassembled, err
+	}
+	return read, false, reassembled, err
+}
+
 // OtelMiddleware returns an HTTP middleware that creates spans for OpenAI API
 // calls following GenAI semantic conventions.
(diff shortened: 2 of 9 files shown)


om7057 (commented):
I went ahead and pulled this down to check it against #1309 myself since I filed that one. When I ran the full suite plus `go vet` for v1/v2/v3, all green, and re-ran the new truncation test with a body sized roughly like the one from my original repro (35MB embeddings response). 
Confirmed the three things that actually matter here: 
- the caller still gets the full untruncated body no matter what, 
- the response attributes are correctly left off instead of getting parsed from garbage partial JSON, and
- the drop now shows up as a warning log with the operation and cap size instead of just vanishing.

Checked the `readBounded` logic too, reading one byte past the limit to tell a truncated body apart from one that just happens to land exactly at the cap. That's correct, no off-by-one there.

I'm unclear and want an opinion on: this fixes the silent-drop symptom with a log line, but the actual size cap is still there, so a big response still loses its attributes, just loudly now instead of quietly. A streaming decoder would let parsing work regardless of size. For this PR's scope, I think the log is the right call; ripping out the cap feels like a separate, bigger decision.

Nice fix either way; appreciate someone picking this up.
cc: @darccio @rupeshkumar92a-arch

SSOHEB (commented):
I inspected the body reading and reassembly, including the size limit cases and read errors. The limit+1 approach correctly detects when a body is too large, while still keeping the full body available to the caller. I also checked v1, v2, and v3, and the tests are passing. good 😊

you need to sign the CLA, please do it

SSOHEB:
> I went ahead and pulled this down to check it against #1309 myself since I filed that one. When I ran the full suite plus `go vet` for v1/v2/v3, all green, and re-ran the new truncation test with a body sized roughly like the one from my original repro (35MB embeddings response). Confirmed the three things that actually matter here:
> 
> * the caller still gets the full untruncated body no matter what,
> * the response attributes are correctly left off instead of getting parsed from garbage partial JSON, and
> * the drop now shows up as a warning log with the operation and cap size instead of just vanishing.
> 
> Checked the `readBounded` logic too, reading one byte past the limit to tell a truncated body apart from one that just happens to land exactly at the cap. That's correct, no off-by-one there.
> 
> I'm unclear and want an opinion on: this fixes the silent-drop symptom with a log line, but the actual size cap is still there, so a big response still loses its attributes, just loudly now instead of quietly. A streaming decoder would let parsing work regardless of size. For this PR's scope, I think the log is the right call; ripping out the cap feels like a separate, bigger decision.
> 
> Nice fix either way; appreciate someone picking this up. cc: @darccio @rupeshkumar92a-arch

I think keeping the size cap makes sense for this PR. The main issue was that the truncation was happening silently, and now it at least shows up with a warning. A streaming decoder could be useful, but I think that’s better as a separate discussion.

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -126,17 +148,16 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
...
-		}{io.MultiReader(&buf, req.Body), req.Body}
+		bodyBytes, truncated, reassembledBody, err := readBounded(req.Body, maxRequestBodySize)
+		req.Body = reassembledBody
 		if err != nil {
 			return next(req)
 		}
+		if truncated {
+			logger.Warn("openai request body exceeded size cap, skipping request attributes",
+				"operation", opName, "cap_bytes", maxRequestBodySize)
+			return next(req)
```
ADITYA-CODE-SOURCE, line 159: The bounded read and body reconstruction look correct, including the one-byte overflow detection. However, the request-truncation branch currently bypasses instrumentation entirely.

Exceeding the body-size limit should suppress only the attributes that require parsing the request body, not the span itself or independently available response telemetry.

Could we preserve the normal span and response-processing path for oversized requests and update the mirrored regression tests in v1, v2, and v3?

ADITYA-CODE-SOURCE (changes requested):

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -108,6 +108,28 @@ func operationName(op operationType) string {
...
+// caller). truncated reports whether body held more than limit bytes, in
+// which case bodyBytes is a cut-off prefix and must not be treated as
+// complete JSON.
+func readBounded(body io.ReadCloser, limit int64) (bodyBytes []byte, truncated bool, reassembled io.ReadCloser, err error) {
+	var buf bytes.Buffer
+	tee := io.TeeReader(body, &buf)
+	// Read one byte past the limit so a full body can be told apart from one
+	// that was cut short. That extra byte is still captured by the tee, so
+	// reassembled below reproduces the stream in full either way.
+	read, err := io.ReadAll(io.LimitReader(tee, limit+1))
```
vedant21-oss, line 122: Checked the boundary by hand and it's right: with a limit of 10, a 9-byte body gives truncated=false, exactly 10 gives false, 11 gives true, and the reassembled body is full length in all three.

Thing is, nothing pins that. Both new tests use bodies far past the cap, so they'd still pass if limit+1 here became limit in a later refactor, which is exactly where an off-by-one lives. Could we add a case with a body exactly at the cap asserting it isn't treated as truncated?

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -108,6 +108,28 @@ func operationName(op operationType) string {
...
+	reassembled = struct {
+		io.Reader
+		io.Closer
+	}{io.MultiReader(&buf, body), body}
+	if int64(len(read)) > limit {
+		return read[:limit], true, reassembled, err
```
vedant21-oss, line 128: Small one: both callers return early when truncated is true, so this prefix is never used. Returning nil would make it impossible for a future caller to accidentally parse half a JSON body.

instrumentation/github.com/openai/openai-go/v2/middleware.go
```diff
@@ -108,6 +108,28 @@ func operationName(op operationType) string {
 	}
 }
 
+// readBounded reads up to limit bytes from body for attribute parsing while
+// preserving the full stream for downstream consumers (the SDK or the HTTP
+// caller). truncated reports whether body held more than limit bytes, in
+// which case bodyBytes is a cut-off prefix and must not be treated as
+// complete JSON.
+func readBounded(body io.ReadCloser, limit int64) (bodyBytes []byte, truncated bool, reassembled io.ReadCloser, err error) {
```
vedant21-oss, line 116: Not necessarily for this PR, but readBounded is now identical in v1, v2 and v3, and all three already depend on internal/streaming. Worth putting it there so there's one copy? I'm working on #1370 which is largely about removing this kind of duplication, happy to do it separately.

Issue #1318, opened 2026-09-03
bug(instrumentation/openai): SSE response detection is case-sensitive, same as anthropic's #1245


om7057:
### Describe the bug

`OtelMiddleware` in the openai instrumentation checks for a streaming SSE response with a case-sensitive prefix match:

```go
contentType := resp.Header.Get("Content-Type")
if strings.HasPrefix(contentType, "text/event-stream") {
```

A server returning a mixed-case value like `Text/Event-Stream` misses this check, so the response is treated as non-streaming and the span never gets `gen_ai.request.is_stream`.

This is the same bug as the anthropic instrumentation had, fixed in #1245 by lowercasing the content type before the prefix check. The openai copies were never touched by that PR and still have the original bug, identically in all three versioned copies:

- `instrumentation/github.com/openai/openai-go/middleware.go`
- `instrumentation/github.com/openai/openai-go/v2/middleware.go`
- `instrumentation/github.com/openai/openai-go/v3/middleware.go`

### Expected behavior

Same fix as #1245: lowercase the `Content-Type` value before the `text/event-stream` prefix check, applied to all three versioned copies, with a mixed-case regression test mirroring `TestOtelMiddleware_SSEResponseFallbackMixedCaseContentType`.

### Additional context

Found while reviewing #1245.

om7057:
@darccio @y1yang0 flagging this one since it's the direct sibling of #1245, 
I've reproduced it locally; once you approve, I'll open the PR.

om7057:
Just a small reminder on this one @darccio @y1yang0.

RedZapdos123:
Fixed by #1320. OpenAI v1, v2, and v3 now use exact case insensitive SSE media type matching and no longer treat `text/event-streaming` as SSE. 

This issue should be closed as resolved.

om7057:
Thanks for picking this up and for the tighter exact-match fix, appreciate it. Closing this out.

PR #1320, opened 2026-09-03
fix(instrumentation): avoid false SSE detection

RedZapdos123:
## Description:

The Anthropic and OpenAI middleware use `strings.HasPrefix` for SSE response detection. A distinct media type such as `text/event-streaming` therefore follows the streaming response path even though it is not SSE.

This PR separates the type component from response parameters, trims it, and compares it case-insensitively. Only an exact `text/event-stream` type uses the SSE path. Malformed parameter syntax such as `text/event-stream; badparam` still preserves SSE detection.

## Verifications:

- [x] `cd instrumentation/github.com/anthropics/anthropic-sdk-go && go test ./... -count=1 && go vet ./...`
- [x] `cd instrumentation/github.com/openai/openai-go && go test ./... -count=1 && go vet ./...`
- [x] `cd instrumentation/github.com/openai/openai-go/v2 && go test ./... -count=1 && go vet ./...`
- [x] `cd instrumentation/github.com/openai/openai-go/v3 && go test ./... -count=1 && go vet ./...`
- [x] `make build`
- [x] `make package && .github/scripts/verify-bundle.sh`


RedZapdos123:
While analyzing the SSE response detection related to #1245 and issue #1318, I found a separate false positive in the same `strings.HasPrefix` check: `text/event-streaming` is treated as SSE.

I opened this to ensure only the exact `text/event-stream` media type uses the SSE path.

instrumentation/github.com/openai/openai-go/middleware.go
```diff
@@ -208,8 +209,8 @@ func OtelMiddleware() func(*http.Request, func(*http.Request) (*http.Response, e
 			return resp, nil
 		}
 
-		contentType := resp.Header.Get("Content-Type")
-		isStreaming = strings.HasPrefix(contentType, "text/event-stream")
+		mediaType, _, err := mime.ParseMediaType(resp.Header.Get("Content-Type"))
```
darccio: mime.ParseMediaType fixes the text/event-streaming false positive. It also
creates a narrower false negative. When a Content-Type parameter has no equals
sign, for example text/event-stream; badparam, ParseMediaType returns an
error. The middleware then treats the response as non-streaming, but the base
media type is still text/event-stream. The previous strings.HasPrefix check
accepted this header.

Verified locally:

| Content-Type | old (`HasPrefix`) | new (`ParseMediaType`) |
|---|---|---|
| `text/event-stream; badparam` | stream | **non-stream, `err != nil`** |

If Anthropic and OpenAI servers never send malformed parameters, this
behavior does not matter in practice. The new check fails in the opposite
direction from the original bug: it treats some malformed SSE responses as
non-streaming. Confirm that this trade-off is intentional.

As an alternative, split the header on the first semicolon and compare only
the media type with strings.EqualFold. This approach keeps the exact-match
fix and does not reject on parameter syntax:

```go
strings.EqualFold(strings.TrimSpace(strings.SplitN(contentType, ";", 2)[0]), "text/event-stream")
```

The same pattern appears in:
- `instrumentation/github.com/openai/openai-go/v2/middleware.go:212`
- `instrumentation/github.com/openai/openai-go/v3/middleware.go:212`
- `instrumentation/github.com/anthropics/anthropic-sdk-go/middleware.go:178` (early-return shape)

darccio (changes requested):

instrumentation/github.com/openai/openai-go/middleware_test.go
```diff
@@ -76,7 +76,8 @@ func TestOtelMiddleware_RecordsDuration(t *testing.T) {
 	next := func(r *http.Request) (*http.Response, error) {
 		return &http.Response{
 			StatusCode: http.StatusOK,
-			Header:     http.Header{"Content-Type": []string{"application/json"}},
+			// A distinct media type with the SSE prefix must remain non-streaming.
+			Header: http.Header{"Content-Type": []string{"text/event-streaming"}},
```
darccio: `TestOtelMiddleware_RecordsDuration` now carries two responsibilities. It
asserts duration recording, and it also guards the `text/event-streaming`
false-positive regression, through the fixture change on line 80. The
function name states only the first responsibility. A reader who scans test
names to find SSE-detection coverage will not find it here.

The inline comment on line 79 explains the intent, so this does not block
the PR. A split into a dedicated test, for example
`TestOtelMiddleware_NonStreamingContentTypeWithSSEPrefix`, would make the
regression guard discoverable by name and keep this test focused on
duration recording alone.

The same pattern repeats in:
- `instrumentation/github.com/openai/openai-go/v2/middleware_test.go:66-80`
- `instrumentation/github.com/openai/openai-go/v3/middleware_test.go:66-80`
- `instrumentation/github.com/anthropics/anthropic-sdk-go/middleware_test.go:260-277` (`TestOtelMiddleware_Messages`)

instrumentation/github.com/openai/openai-go/middleware.go
RedZapdos123: Applied in 1bb5ce8. SSE detection now compares the trimmed portion before the first semicolon with EqualFold. Text/Event-Stream; badparam remains streaming, while text/event-streaming remains non-streaming. Both cases are covered in Anthropic and OpenAI v1/v2/v3.

instrumentation/github.com/openai/openai-go/middleware_test.go
RedZapdos123: Applied in 1bb5ce8. The generic duration and Messages fixtures are restored. Added TestOtelMiddleware_NonStreamingContentTypeWithSSEPrefix in Anthropic and OpenAI v1/v2/v3 so the false-positive regression is directly discoverable.

RedZapdos123: The generic duration and Messages tests are restored.

Added `TestOtelMiddleware_NonStreamingContentTypeWithSSEPrefix` in Anthropic and OpenAI v1/v2/v3 so the regression is directly discoverable.

instrumentation/github.com/openai/openai-go/middleware.go
RedZapdos123: SSE detection now splits before the first `;`, trims the type, and uses `EqualFold`.

`Text/Event-Stream; badparam` remains streaming while `text/event-streaming` remains non-streaming. Tests cover Anthropic and OpenAI v1/v2/v3.

darccio (approved):
LGTM

PR #1203, opened 2026-08-22
fix(instrumentation/net/http): guard nil request, URL, and Header in client BeforeRoundTrip

Labreo:
## Description

This PR adds nil guards in `instrumentation/net/http/client/client_hook.go`:
- Early returns when `req == nil` before dereferencing `req.Context()`.
- Guards `req.URL.String()` in debug logging when `req.URL` is nil.
- Ensures `req.Header` is initialized before passing to `propagator.Inject(ctx, propagation.HeaderCarrier(req.Header))` to prevent `panic: assignment to entry in nil map`.
- Adds unit tests in `client_hook_test.go` covering nil `req`, nil `req.Header`, and nil `req.URL`.
- Regenerates and verifies `tool/data/otelc-bundle.tgz`.

## Motivation

In standard Go, manually constructed `&http.Request{Method: "GET", URL: u}` structs with nil `Header` maps are legal inputs to `(*http.Transport).RoundTrip`. Previously, `BeforeRoundTrip` attempted to inject trace headers into `req.Header` without checking if it was nil, causing a runtime panic on nil map assignment.

Fixes #1202 

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format`
- [x] Linters pass: `make lint`
- [x] Tests pass: `make test`
- [x] Tests added for new functionality
- [x] Tests follow [testing guidelines](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/testing.md)
- [ ] Documentation updated (if applicable)
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation))
- [x] This PR has content that I did not fully write myself.
  - [x] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

PR #1323, opened 2026-09-04
fix(net/http): stop client instrumentation panicking on nil req, Header, or URL

### LINKED ISSUE #1202: [Bug]: net/http client instrumentation panics on nil request or nil Request.Header in (*Transport).RoundTrip
### Describe the bug

In `instrumentation/net/http/client/client_hook.go`, `BeforeRoundTrip` intercepts `(*http.Transport).RoundTrip(req *http.Request)` but makes two unsafe nil assumptions:
1. `req.Context()` is invoked without checking if `req == nil`, causing an immediate nil-pointer dereference panic if `req` is nil.
2. `propagator.Inject(ctx, propagation.HeaderCarrier(req.Header))` is called directly on `req.Header`. If `req.Header` is `nil` (which is standard and legal in Go when constructing requests like `&http.Request{Method: http.MethodGet, URL: u}` without calling `http.NewRequest`), `HeaderCarrier.Set` attempts `h[key] = []string{value}` on a nil map, causing:
   ```text
   panic: assignment to entry in nil map
   ```
3. `req.URL.String()` in debug logging also panics if `req.URL` is nil.

In standard library Go, calling `(*http.Transport).RoundTrip` with a manually initialized `&http.Request{Method: "GET", URL: u}` and nil `Header` works properly. Compile-time instrumentation should not introduce panics on valid stdlib usage.

### Environment

 - OS: macOS / Linux / Windows
 - Go Version: 1.24+
 - Version: main

### To Reproduce

Steps to reproduce:
1. Instrument an application using `otelc`.
2. Issue an HTTP request via `http.DefaultTransport.RoundTrip(&http.Request{Method: "GET", URL: u})` with a `nil` `Header`.
3. The hook executes `BeforeRoundTrip`, attempting `propagator.Inject` on `propagation.HeaderCarrier(nil)`.
4. Observe panic: `panic: assignment to entry in nil map`.

### Expected behavior

`BeforeRoundTrip` should safely handle nil `req`, guard `req.URL` in debug logging, and ensure `req.Header` is initialized (`if req.Header == nil { req.Header = make(http.Header) }`) before injecting trace context headers.

ipsitapp8:
## Description

`BeforeRoundTrip` in `instrumentation/net/http/client/client_hook.go` assumed a fully populated `*http.Request`. It panicked in three cases that are legal input to `(*http.Transport).RoundTrip`:

- `req == nil`
- `req.Header == nil` (a manually built request, not via `http.NewRequest`, has a nil Header by default) - injecting trace context called `HeaderCarrier.Set` on that nil map: `panic: assignment to entry in nil map`
- `req.URL == nil` - the debug log called `req.URL.String()` directly, which panics on a nil `*url.URL`

Fix: return early on a nil req, initialize `req.Header` when nil before injecting trace context, and guard the debug-log URL string.

## Motivation

Compile-time instrumentation should not introduce panics on requests the standard library accepts. `http.DefaultTransport.RoundTrip(&http.Request{Method: "GET", URL: u})` with a nil `Header` works fine uninstrumented, but panicked once `otelc` was compiled in.

Added regression tests in `client_hook_test.go` covering all three nil cases. Confirmed they fail against the pre-fix code with the exact panics described in the issue, and pass against the fix.

Fixes #1202


instrumentation/net/http/client/client_hook.go
```diff
@@ -59,6 +59,15 @@ func BeforeRoundTrip(ictx hook.HookContext, transport *http.Transport, req *http
 		return
 	}
 
+	// A manually built http.Request (not http.NewRequest) is legal input to
+	// RoundTrip and may arrive here with a nil Header, or, if the caller
+	// skipped validation entirely, a nil req. Neither should crash the
+	// caller's program just because instrumentation is compiled in.
+	if req == nil {
+		logger.Debug("BeforeRoundTrip: nil request, skipping instrumentation")
+		return
```
darkraider01, line 68: Heads up, I don't think this early-return actually stops the crash it's meant to stop.

We never call `ictx.SetSkipCall(true)` here, so after this hook returns, the generated trampoline still goes ahead and calls the real `(*http.Transport).RoundTrip(nil)` with the original nil req. And that panics on its own, instrumentation or not. I checked locally:

```
$ http.DefaultTransport.RoundTrip(nil)
```

So end to end, a caller doing `transport.RoundTrip(nil)` on an instrumented binary still crashes — just one call frame later than before. The new test only calls `BeforeRoundTrip` directly, so it doesn't catch this.

`instrumentation/go.opentelemetry.io/otel/hook.go:18` does the right thing here (`ictx.SetSkipCall(true)`) when it wants to actually stop the wrapped call from running. Might be worth doing the same, or just being upfront that the nil-req case can't really be fixed from this hook alone since stdlib itself doesn't support it — only the Header/URL guards are actually load-bearing.

instrumentation/net/http/client/client_hook_test.go
```diff
@@ -430,3 +431,65 @@ func TestAfterRoundTrip_DisabledAfterStart_Regression(t *testing.T) {
...
+		req := &http.Request{Method: http.MethodGet, Header: http.Header{}}
+		require.Nil(t, req.URL)
+
+		mockCtx := hooktest.NewMockHookContext()
+		transport := &http.Transport{}
+
+		assert.NotPanics(t, func() {
+			BeforeRoundTrip(mockCtx, transport, req)
+		})
+		assert.Empty(t, sr.Ended(), "span should not be ended in Before hook")
```
darkraider01, line 493: Small nit: this subtest never checks that trace context still gets injected into `req.Header`, unlike the "nil header" subtest right above it which checks `newReq.Header.Get("traceparent")`. Wouldn't catch a regression that accidentally skips injection whenever URL is nil. Not blocking, just flagging.

Labreo:
The orignal issue has been fixed and this should probably be closed.

darccio:
Closing in favour of #1203.

@ipsitapp8 Please check if issues have linked PRs before starting a new one. The link appears automatically when the PR refers the issue.

ipsitapp8:
@darccio yaa my mistake.. next time I will take care

PR #1324, opened 2026-09-04
fix(database/sql): link Tx Commit/Rollback spans to the transaction's trace

### LINKED ISSUE #1187: [database/sql] Transaction Commit and Rollback spans are detached from distributed traces

ipsitapp8:
## Description

`beforeTxCommitInstrumentation` and `beforeTxRollbackInstrumentation` in `instrumentation/database/sql/client.go` hardcoded `context.Background()` as the parent context when starting the COMMIT/ROLLBACK span, instead of using the context the transaction was started with. That orphans every commit and rollback span onto its own new trace, disconnected from the trace `BeginTx` started.

Fix: added an `OtelCtx context.Context` field to the compile-time-injected `Tx` struct. `afterTxInstrumentation` now stores the span context `BeginTx`'s hook created onto `tx.OtelCtx`, the same place it already stores `Endpoint`/`DriverName`/`DSN`/`DbName`. `beforeTxCommitInstrumentation` and `beforeTxRollbackInstrumentation` read it back through a small `txContext` helper, falling back to `context.Background()` only if a `Tx` somehow never went through the `BeginTx` hook.

Added a regression assertion to the existing `TestDBClient/Transaction` integration test: commit's span must share `BeginTx`'s trace ID. Confirmed it fails against the old code (commit lands in its own trace) and passes with the fix.

## Motivation

Any application relying on this compile-time instrumentation for distributed tracing gets broken, disconnected traces for every database transaction: the `COMMIT`/`ROLLBACK` span shows up as an isolated single-span trace in the backend instead of completing the transaction's timeline.

Fixes #1187


instrumentation/database/sql/client.go
```diff
@@ -304,6 +304,14 @@ func afterTxInstrumentation(ictx hook.HookContext, tx *sql.Tx, err error) {
 	tx.DriverName = dbRequest.DriverName
 	tx.DSN = dbRequest.Dsn
 	tx.DbName = dbRequest.DbName
+
+	// Carry the caller's context onto the transaction so commit and rollback
+	// spans, created long after this hook returns, can still be linked into
+	// the trace that started the transaction instead of falling back to
+	// context.Background().
+	if ctx, ok := callData["ctx"].(context.Context); ok {
+		tx.OtelCtx = ctx
```
darkraider01, line 323: Nice catch on the DB.BeginTx path, but I think there's a twin bug still sitting right next to it.

`afterConnTxInstrumentation` is the after-hook for `(*sql.Conn).BeginTx` (wired up separately in `hook_conn_begin_tx` in otelc.yaml)  it's a totally different function from the one you patched. It never sets `tx.OtelCtx`, it just calls `instrumentEnd`. So if someone does `conn.BeginTx(...)` instead of `db.BeginTx(...)`, the tx they get back still has a nil `OtelCtx`, and `txContext(tx)` falls back to `context.Background()` again  same orphaned-trace bug as #1187, just on the Conn path instead of the DB path.

Worth setting `tx.OtelCtx` here too (basically the same 6 lines you added to `afterTxInstrumentation`). There's also no test anywhere that exercises `conn.BeginTx`, so this slipped through CI.

test/integration/db_client_test.go
```diff
@@ -124,6 +124,14 @@ func TestDBClient(t *testing.T) {
...
 		)
 		require.Equal(t, "COMMIT", commitSpan.Name())
+
+		// Regression for #1187: Commit must land in the same trace as BeginTx,
+		// not get orphaned onto its own trace by starting from
+		// context.Background(). (ExecContext's span isn't asserted here: it
+		// takes its own ctx argument from the caller, independent of the
+		// transaction's begin context, so it isn't guaranteed to match.)
+		require.Equal(t, beginSpan.TraceID(), commitSpan.TraceID(),
+			"commit span must share the transaction's trace, not start a new one")
```
darkraider01, line 134: Since we're touching this: rollback goes through the exact same `txContext(tx)` fix as commit, but only commit's trace ID gets checked here. Might be worth adding one line for rollback too, just so a future change that breaks it on that side doesn't sail through untested.

instrumentation/database/sql/otelc.yaml
```diff
@@ -45,6 +45,8 @@ add_new_field_tx:
             type: string
           - name: DSN
             type: string
+          - name: OtelCtx
```
darkraider01, line 48: Minor one: `OtelCtx` is typed `context.Context`, but this rule doesn't declare an `imports:` entry for it the way the tool supports (see `addRuleImports`). It compiles fine right now only because `database/sql`'s `sql.go` already imports `context` for its own stuff elsewhere in the file  not because this rule guarantees it. Not urgent, just a bit of luck holding this together rather than the rule being explicit about what it needs.

instrumentation/database/sql/client.go
ipsitapp8: Good catch, thanks. Same bug, same fix: pulled the endpoint/driver + OtelCtx copy into a shared `populateTxFromBeginHook(ictx, tx)` helper and call it from both `afterTxInstrumentation` and `afterConnTxInstrumentation` now, so `conn.BeginTx()` gets the same trace linkage (and the Endpoint/DriverName/DSN/DbName it was also missing) as `db.BeginTx()`. Fixed in 470589c.

test/integration/db_client_test.go
ipsitapp8: Added a separate TransactionRollback subtest (new doTxRollback op in the dbclient test app) that checks rollback's trace ID against begin's independently of commit, so a future regression that breaks one path but not the other gets caught. Fixed in 470589c.

instrumentation/database/sql/otelc.yaml
ipsitapp8: Fair, no reason to lean on that. Added an explicit imports: context: "context" on the rule so it declares what it needs instead of depending on sql.go's own unrelated import. Fixed in 470589c.

ADITYA-CODE-SOURCE (approved):

PR #1241, opened 2026-08-25
refactor(tool): drop redundant intermediate error wraps

Labreo:
## Description

This PR drops redundant intermediate `ex.Wrap` and `ex.Wrapf` calls across the internal tool packages (`tool/internal/setup`, `tool/internal/instrument`, `tool/internal/pkgload`, and `tool/internal/imports`). Intermediate internal functions now return errors directly (`return err`), while preserving wrapping at error origins (like stdlib/OS calls) and at locations providing unique contextual information. Error assertions in affected tests have been updated accordingly.

## Motivation

Errors in `otelc` are stackful (`ex.New`, `ex.Wrap`, `ex.Wrapf` capture full stack traces). Wrapping errors on intermediate internal call returns where the wrapper only restates the function's action adds redundant message frames without new context. Propagating the error directly avoids cluttering the error message while preserving full stack traces and origin context.

Fixes #1231

---

## Checklist

- [x] PR title follows [conventional commits](https://www.conventionalcommits.org/) format
- [x] Code formatted: `make format`
- [x] Linters pass: `make lint`
- [x] Tests pass: `make test`
- [x] Tests added for new functionality
- [x] Tests follow [testing guidelines](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/testing.md)
- [ ] Documentation updated (if applicable)
- [ ] OpenTelemetry Registry updated (if applicable, see [registry guide](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/instrument-guide.md#4-register-the-instrumentation))
- [x] This PR has content that I did not fully write myself.
  - [x] I used AI and I have read and followed the [Generative AI Contribution Policy](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/AI_POLICY.md).
- [x] I have the experience and knowledge necessary to understand, review, and validate all content in this PR.[^I-know-my-stuff]

[^I-know-my-stuff]:
    Yes, I can answer maintainer questions about the content of this PR, without using AI.

PR #1326, opened 2026-09-05
fix(tool): FuncArgumentOfType fails on aliased imports in call and raw rules

### LINKED ISSUE #1276: bug(tool): preserve import context in call and raw template helpers
### Describe the bug

`FuncArgumentOfType` and `FuncReturnOfType` do not reliably resolve types in call-rule and raw-rule templates when the target file uses aliased imports or when multiple packages share the same default package name.

The issue is caused by import context being dropped when creating `funcTemplateData`.

In `compileExpression`, the enclosing function is passed with `nil` imports:

```go
data.enclosing = newFuncTemplateData(enclosing, nil, nil, "")
```

Although `applyCallReplace` already has the target file's `importAliases`, they are not passed into `compileExpression`.

The same issue exists in `renderRawCode`, which also creates `funcTemplateData` with `nil` imports.

This differs from directive rules, which correctly pass `ast.ImportAliasMap(root)` into `newFuncTemplateData`.

As a result, `FuncArgumentOfType` / `FuncReturnOfType` may produce incorrect matches because `ast.MatchesTypeName` cannot resolve import aliases or distinguish packages with the same default package name.

For example:

```go
import althttp "net/http"

func Handler(r *althttp.Request) {
    http.Get("url")
}
```

With a call-rule template using:

```text
{{ .FuncArgumentOfType "*net/http.Request" }}
```

the expected result is `r`, but the current call-rule path can return an empty result because the `althttp` import alias is not available to the template data.

The same problem can cause incorrect matches when packages have the same final path component, such as `text/template` and `html/template`.

### Environment

* OS: Any
* Go Version: Any supported version
* OpenTelemetry Go compile-time instrumentation version: `release/main` (current development tree)

### To Reproduce

1. Create a target Go file that uses an aliased import:

```go
package main

import althttp "net/http"

func Handler(r *althttp.Request) {
    http.Get("url")
}
```

2. Create a call-rule with a `replace` template that uses `FuncArgumentOfType`:

```yaml
replace: 'traced({{ .FuncArgumentOfType "*net/http.Request" }}, {{ . }})'
```

3. Apply the instrumentation to the target file.

4. Observe that `FuncArgumentOfType` does not resolve `*net/http.Request` to the `r` parameter as expected.

The same behavior can be reproduced with `FuncReturnOfType` and with raw rules using these template helpers.

### Expected behavior

`FuncArgumentOfType` and `FuncReturnOfType` should resolve types using the import map of the target Go file, regardless of whether they are used through directive, call, or raw rules.

For the example above:

```text
{{ .FuncArgumentOfType "*net/http.Request" }}
```

should resolve to:

```text
r
```

Aliased imports should be resolved to their actual package paths, and packages with identical default names should be correctly distinguished.

### Additional context

The underlying type-matching logic already supports import resolution. Existing tests for `funcTemplateData` verify that aliased imports are handled correctly when the import map is provided.

The problem is in the wiring of the call-rule and raw-rule paths:

* `applyCallReplace` already has `importAliases`, but `compileExpression` does not receive them.
* `compileExpression` creates `funcTemplateData` with `nil` imports.
* `renderRawCode` also creates `funcTemplateData` with `nil` imports.
* Directive rules already pass the target file's import map correctly.

This suggests the fix should be localized to passing the target file's import context through the call/raw template paths.

Integration tests should cover at least:

* aliased imports such as `althttp "net/http"`
* packages with the same default package name, such as `text/template` vs `html/template`
* both `FuncArgumentOfType` and `FuncReturnOfType`
* call-rule and raw-rule template paths

atharvamhaske:
## Description

Import aliases are name tags. `althttp` is still `net/http`, you just called it something else. The type matcher already reads those tags. Call and raw templates just never handed them over.

`applyCallReplace` already had `importAliases` in its pocket. Then `compileExpression` built `funcTemplateData` with `nil`. `renderRawCode` did the same thing. Directive rules pass `ast.ImportAliasMap(root)` and never hit this. Yeah, that gap is the whole ticket.

This PR threads the target file's import map through those two paths. Call-rule templates also get `FuncReturnOfType` now (raw and directive already had it; call was the odd one out).

Tests hit both helpers on call and raw: `althttp "net/http"`, and the classic `text/template` vs `html/template` collision (same default name, different packages).

## Motivation

You write `{{ .FuncArgumentOfType "*net/http.Request" }}` and the param is `*althttp.Request`. You want `r`. You get `""`. That's the bug.

And if you import both `text/template` and `html/template`, they both look like `template` if you only stare at the last path segment. Without the real import map you can match the wrong one.

Fixes #1276


ipsitapp8 (commented):
Reviewed and tested this locally.

Verified the bug is real and the fix is necessary: reverted just the two threaded params (`compileExpression(call, enclosing, importAliases)` -> `nil` in apply_call.go, `ast.ImportAliasMap(root)` -> `nil` in apply_raw.go) and re-ran the new tests. Both the aliased-import and the text/template-vs-html/template cases fail without the fix, in a worse way than the issue's `""` example: with two type lookups going empty at once, the generated code doesn't even parse (`traced(, , _.PLACEHOLDER_0)`). Restored the fix, tests pass again.

Checked for other call sites that might have the same twin-bug shape (this is what got missed on my own #1324): `apply_decl.go` and `apply_lit.go` also call `compileExpression` with `nil` imports, but that's correct there, not a second instance of the bug — both pass `enclosing: nil` (package-level initializers and composite literal fields have no enclosing function), so `FuncArgumentOfType`/`FuncReturnOfType` are unreachable on those paths regardless of the imports value. `apply_directive.go` already threaded `ast.ImportAliasMap(root)` before this PR, matching the description. So this covers every path that actually needed it.

Test design is good, especially the second subtest: proving `text/template.Template` and `html/template.Template` don't cross-match is a real disambiguation case, not just a "does aliasing work" smoke test, and it exercises the negative match (asserting a literal `""` in the generated AST) rather than only the positive one.

`FuncReturnOfType` on call rules is additive beyond what #1276 strictly required (the issue is about `FuncArgumentOfType`), but it shares the exact same plumbing this PR is already touching (`compileExpression` -> `newFuncTemplateData`), so bundling it here instead of a separate PR seems reasonable rather than scope creep.

No CHANGELOG.md entry, consistent with how other internal tool/rule-engine fixes in this repo (e.g. #1078, #1241) that don't change end-user runtime behavior have handled it.

`go build ./tool/...`, `go vet ./tool/...`, `gofmt -l`, and the full `tool/internal/instrument`, `tool/internal/ast`, `tool/internal/rule` suites all pass clean on this branch.

Nothing blocking from me.

tool/internal/instrument/apply_raw.go
```diff
@@ -50,16 +50,18 @@ func renameReturnValues(funcDecl *dst.FuncDecl) {
 // FuncArgument N, FuncReturn N, ...) in raw code injected by a raw rule. Raw
 // code that does not contain "{{" is returned unchanged. hash salts synthetic
 // argument/return names the same way InstRawRule.Identity salts other rules'
-// trampoline/template names.
-func renderRawCode(raw string, decl *dst.FuncDecl, hash string) (string, error) {
+// trampoline/template names. imports is the target file's import alias map
+// (see ast.ImportAliasMap); FuncArgumentOfType / FuncReturnOfType need it to
+// resolve aliased imports and packages that share a default name.
```
y1yang0: I think you can use simple comment instead, e.g.

```
// renderRawCode renders the shared function template variables (FuncName,
// FuncArgument N, FuncReturn N, ...) in raw code injected by a raw rule.
```

tool/internal/instrument/apply_raw.go
```diff
@@ -50,16 +50,18 @@ func renameReturnValues(funcDecl *dst.FuncDecl) {
...
+func renderRawCode(raw string, decl *dst.FuncDecl, hash string, imports map[string]string) (string, error) {
 	if !strings.Contains(raw, "{{") {
 		return raw, nil
 	}
 	tmpl, err := rule.ParseFuncTemplate(raw)
 	if err != nil {
 		return "", ex.Wrap(err)
```
y1yang0: This should be `return "", err` instead.

y1yang0 (approved):
Otherwise LGTM, maybe @darccio or @kakkoyun could take a look as well

darccio (approved):

PR #1329, opened 2026-09-07
chore: remove CHANGELOG.md re-introduced in #1012

### LINKED ISSUE #1234: Clean up the repository root
Our root directory keeps growing. Some of it clearly belongs elsewhere, and moving it would simplify the directory structure.
- codecov.yml and .github/codeconv.yaml are duplicates. One of them should be eliminated and move to `.github`  eventually
- AI_POLICY.md/RELEASE.md/SECURITY.md can be moved to the docs directory. That is what the `docs` directory is for.
- CHANGELOG.md can be deleted. It is duplicated work and adds maintenance burden.
- `.adr-dir/.semconv-version` could be moved to `.github`, but this may not work out. It needs some extra work and investigation.

darkraider01:
Fixes #1234

removes CHANGELOG.md which got added back in #1012


darkraider01:
in draft because currently hitting max pr limit.

amazingakai:
@y1yang0 IMO we should keep the CHANGELOG, it can be useful for documenting breaking changes and other notable changes.

I noticed we did not mention a breaking change we had in v1.1.0 release, we had mentioned this in CHANGELOG.md but the CHANGELOG was removed without documenting the breaking change anywhere else:
```
- ⚠️ **Breaking Change:** Template variable `{{ FuncName }}` should now be accessed with `{{ .FuncName }}`. ([#729](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/729))
```

y1yang0:
@amazingakai First, semantic versioning already conveys the notion of breaking changes to some extent. Second, release notes are a better fit for this. A hand written changelog adds maintenance cost: every PR author needs to confirm with a maintainer whether an entry is required, and we still need to check for PRs that slipped through. Release notes already contain every patch, which makes it easy to filter for breaking changes.

darccio:
@amazingakai I agree with @y1yang0 here. Changelog are heavy, release notes are lightweight.

y1yang0 (approved):

PR #1333, opened 2026-09-08
feat(tool): write instrumentation diff under --debug

### LINKED ISSUE #1298: otelc diff: preview instrumentation changes without building
## Is your feature request related to a problem? Please describe.

Right now the only way to see what `otelc` will inject into a file is to run a real build and dig through the toolexec temp/build output afterward. There's no way to check upfront whether a given file will be instrumented, or what would actually change, you have to trust the build. This is the most common complaint with auto-instrumentation in general: it feels like a black box, and when a trace looks off later there's no quick way to confirm whether that's what actually got woven in versus something else going on.

## Describe the solution you'd like

Per the discussion below, not a separate `otelc diff <path>` subcommand, extend the existing build flow instead.

`otelc` already writes both the original and instrumented versions of every compiled file into `.otelc-build` on every build (`keepForDebug` / `writeInstrumented` in `tool/internal/instrument/apply_func.go`), unconditionally, not gated behind `--debug`. Since both versions are already sitting there when the file gets written, a flag can make otelc print or save a diff between them, no new command needed.

Still open: whether that flag is the existing `--debug` or a new `--diff`.

To actually see a diff, the target package needs to rebuild, so a stale/cached build won't produce one. That's expected: clear the cache and rebuild to get output.

The output should be descriptive per kakkoyun's note: which files changed, which rule triggered each change, and the order changes were applied in, important for troubleshooting.

## Describe alternatives you've considered

- Reading the generated code after a real build (current status quo). Works, but means digging through temp/build output after the fact rather than checking upfront.
- A separate dry-run subcommand that runs matching/apply outside a real build. Dropped in favor of reusing the existing build flow, since it's a smaller, simpler change and avoids a second parallel entry point into the matching/apply engine.

## Additional context

Follow-ups raised in the discussion, not in scope for this issue but worth tracking separately later: emitting the diff as a `.patch` file, and a mode for call-site instrumentation that injects generated code between pragma comments so it can be reviewed or cleaned up.

darkraider01:
## Description

Implements instrumentation diff reporting under `--debug` (`OTELC_DEBUG=1`), allowing users to inspect the exact Go source modifications and additions introduced by `otelc` into the compiler.

### Diff Categories

Diff reports cover four distinct categories of source transformations:

1. **Modified target source files (`<file>.diff`)**:
   - Captures per-rule AST snapshots before and after each rule is applied.
   - Generates labeled unified diffs in rule application order:
     `=== rule X/N: <rule-name> ===`
   - Generates an authoritative full diff capturing all transformations, imports, and post-processing optimizations:
     `=== full diff: <original> -> <compiled> ===`
2. **`InstFileRule` generated files (`otelc.<name>.go.diff`)**:
   - Added-source unified diffs against `/dev/null` with a header attributing the source to the responsible rule (`=== rule: <name> ===`).
3. **Globals file (`otelc.globals.go.diff`)**:
   - Added-source unified diffs against `/dev/null` with a header listing all contributing trampoline/hook context rules.
4. **Runtime file (`otelc.runtime.go.diff`)**:
   - Added-source unified diffs against `/dev/null` generated during setup for package-level runtime dependencies, linknames, and trampolines, listing contributing rules in deterministic order.

### Directory Layout & Collision Prevention

Debug artifacts are written under package-scoped directories:
- `.otelc-build/debug/<escaped-package>/`

Package import paths are sanitized using a consistent escaping scheme (e.g. `example.com/app/cmd/client` -> `example_com_app_cmd_client`). This guarantees that packages sharing identical filesystem directory basenames (such as `cmd/client` and `internal/client`) write to isolated directories without collision.

### Lifecycle & Stale Artifact Management

- **Wrapper mode (`otelc --debug go build/test/install`)**: The top-level wrapper process acquires the build lock and resets `.otelc-build/debug/` once at the parent lifecycle boundary before setup and compilation begin. Toolexec child processes never perform global cleanup.
- **Direct mode (`go build -toolexec otelc` / `GOFLAGS`)**: Toolexec children run independently and concurrently without inferring a build lifecycle. For fresh diagnostic reports in direct mode, run `otelc --debug setup` immediately before `OTELC_DEBUG=1 go build -a`. `otelc --debug setup` is the explicit boundary that cleans `.otelc-build/debug` and generates fresh setup runtime diffs; subsequent toolexec children only append their package-scoped artifacts.
- **Per-file stale removal**: If an instrumented file no longer matches any rules, its individual `.diff` file is removed, with diagnostic warnings surfaced if removal encounters an unexpected OS error.
- **Go build cache**: The Go toolchain only invokes compiler tools for packages that are actually recompiled. To force package recompilation and fresh compiler diffs, pass `-a` (e.g. `otelc --debug go build -a .` or `OTELC_DEBUG=1 go build -a .`).
- **Zero debug-off overhead**: When `OTELC_DEBUG` is not set, all AST rendering and diff generation paths are completely bypassed.

---

#### 1. Modified Source File (`roundtrip.go.diff`)

Ran a real `otelc go build --debug` against `test/apps/httpclient`. It matched the `net/http` client rule and wrote `.otelc-build/debug/net_http/roundtrip.go.diff`:

```diff
=== rule 1/1: client_hook ===
--- C:\Program Files\Go\src\net\http\roundtrip.go
+++ C:\Program Files\Go\src\net\http\roundtrip.go
@@ -26,10 +26,164 @@
 // Like the RoundTripper interface, the error types returned
 // by RoundTrip are unspecified.
-func (t *Transport) RoundTrip(req *Request) (*Response, error) {
+func (t *Transport) RoundTrip(req *Request) (_unnamedRetVal_..._0 *Response, _unnamedRetVal_..._1 error) {
+	if /* __TRAMPOLINE_JUMP_IF__ */ hookContext..., skip... := OtelBeforeTrampoline_RoundTrip...(&t, &req); skip... {
+		OtelAfterTrampoline_RoundTrip...(hookContext..., &_unnamedRetVal_..._0, &_unnamedRetVal_..._1)
+		return _unnamedRetVal_..._0, _unnamedRetVal_..._1
+	} else {
+		defer OtelAfterTrampoline_RoundTrip...(hookContext..., &_unnamedRetVal_..._0, &_unnamedRetVal_..._1)
+	}
 	if t == nil {
 		panic("transport is nil")
 	}
 	return t.roundTrip(req)
 }

+// ... HookContextImpl struct, trampoline funcs, //go:linkname decls ...

=== full diff: C:\Program Files\Go\src\net\http\roundtrip.go -> ...\roundtrip.go ===
--- C:\Program Files\Go\src\net\http\roundtrip.go
+++ ...\roundtrip.go (compiled)
@@ -26,10 +26,162 @@
-func (t *Transport) RoundTrip(req *Request) (*Response, error) {
+func (t *Transport) RoundTrip(req *Request) (_unnamedRetVal_..._0 *Response, _unnamedRetVal_..._1 error) {
+	if hookContext..., _ := OtelBeforeTrampoline_RoundTrip...(&t, &req); false {
+	} else {
+		defer OtelAfterTrampoline_RoundTrip...(hookContext..., &_unnamedRetVal_..._0, &_unnamedRetVal_..._1)
+	}
 	if t == nil {
 		panic("transport is nil")
 	}
 	return t.roundTrip(req)
 }
+// ...
```

The two sections differ exactly where `optimizeTJumps` runs after rule application: the per-rule section still has the dynamic `skip...` check and the `__TRAMPOLINE_JUMP_IF__` marker; the full-diff summary shows it simplified to a static `false`, matching what the compiler actually sees. That is why the full diff remains a separate, authoritative section rather than just concatenating the per-rule diffs.

#### 2. Generated Instrumentation File (`otelc.runtime.go.diff`)

Files introduced entirely by instrumentation are generated as added-source diffs against `/dev/null` with metadata headers:

```diff
=== generated instrumentation file: otelc.runtime.go ===
rules:
  - net_http_client
  - net_http_transport
--- /dev/null
+++ otelc.runtime.go
@@ -0,0 +1,14 @@
+// Code generated by otelc. DO NOT EDIT.

package http

import (
	_ "unsafe"
)

//go:linkname otelc_roundtrip_trampoline net/http.(*Transport).RoundTrip
func otelc_roundtrip_trampoline(t *Transport, req *Request) (*Response, error)
```

## Motivation

Related #1298
Fixes #1298


tool/internal/instrument/diff.go
```diff
@@ -0,0 +1,150 @@
...
+// default build path pays nothing extra.
+func (ip *instrumentPhase) applyRulesCapturingDiffs(
+	ctx context.Context,
+	rules []rule.InstRule,
+	root *dst.File,
+) (bool, []ruleChange, error) {
+	debug := diffDebugEnabled()
+	var prev []byte
+	if debug {
+		prev, _ = ast.RenderFile(root)
```
ADITYA-CODE-SOURCE: Could we handle the errors from both `RenderFile` calls instead of discarding them? If either call fails, `prev` or `after` may be nil or incomplete, yet still become the baseline for attributing this and subsequent changes to rules. The build could then succeed and report the diff while producing incomplete or misleading per-rule diagnostics. Could we propagate the error with the relevant rule and file context, or emit a warning and disable per-rule capture for that file?

darkraider01: ah right, ignoring that may cause causing the --debug report to attribute a whole-file addition/deletion to a rule or otherwise show a misleading per-rule diff. 
thanks

PR #1335, opened 2026-09-08
fix(ast): stripGenericTypes fatals on qualified pointer receivers

### LINKED ISSUE #766: fix(ast): stripGenericTypes fatals on qualified pointer receivers

ipsitapp8:
## Description

`stripGenericTypes` extracts the base type name from a method receiver so rules with a `recv` filter (like `recv: "*Writer"`) can match against it. Under the `*dst.StarExpr` case it only handled `*dst.Ident` (`*MyStruct`) and the generic index forms (`*GenStruct[T]`). It never handled `*dst.SelectorExpr`, which is what a qualified pointer receiver like `*bufio.Writer` or `*http.ResponseWriter` parses as.

So any rule targeting a method with a qualified pointer receiver hit the fallthrough at the bottom of the function, got an empty string back, and `findFuncDecl` called `util.Unimplemented` on it, which fatals:

```
```

This crashes `otelc setup` at load time, not just the one rule.

## Fix

Added a `*dst.SelectorExpr` case next to the existing `*dst.Ident` one, same shape, pulls `Sel.Name` off it so `*pkg.Type` resolves to `*Type`. Left the bare (non-pointer) `*dst.SelectorExpr` case alone, that one's a value receiver on a foreign type, which isn't valid Go, so it stays as returning empty like before.

## Tests

- Added a case to the existing `TestStripGenericTypes` table for `*bufio.Writer` -> `*Writer`.
- Added `TestFindFuncDeclForRule_QualifiedPointerReceiver`, an end to end test parsing `func (*bufio.Writer) Write(...)` and matching it with `recv: "*Writer"`, same pattern as `TestFindFuncDeclForRule_QualifiedParamType` right above it.

Checked both new tests fatal on main without the fix (stashed the change and reran).

No `instrumentation/` files touched, so the bundle doesn't need regenerating.

Fixes #766

---

### DIFF
--- a/tool/internal/ast/shared.go
+++ b/tool/internal/ast/shared.go
@@ -50,6 +50,7 @@ func FindFuncDeclWithoutRecv(root *dst.File, funcName string) *dst.FuncDecl {
 // - MyStruct -> MyStruct
 // - *GenStruct[T] -> *GenStruct
 // - GenStruct[T] -> GenStruct
+// - *pkg.Type -> *Type
 func stripGenericTypes(recvTypeExpr dst.Expr) string {
 	switch expr := recvTypeExpr.(type) {
 	case *dst.StarExpr: // func (*Recv)T or func (*Recv[T])T
@@ -68,6 +69,9 @@ func stripGenericTypes(recvTypeExpr dst.Expr) string {
 			if baseIdent, ok := x.X.(*dst.Ident); ok {
 				return "*" + baseIdent.Name
 			}
+		case *dst.SelectorExpr:
+			// Qualified pointer receiver: *pkg.Type
+			return "*" + x.Sel.Name
 		}
 	case *dst.Ident: // func (Recv)T
 		return expr.Name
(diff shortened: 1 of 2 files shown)


darkraider01:
there are already three PRs open against the same issue, #1257 , #1230 , #786

darccio:
Closed in favour of #1257

PR #1339, opened 2026-09-09
feat(instrumentation/zap): inject trace_id and span_id into zap logs

### LINKED ISSUE #1331: New Instrumentation Request for go.uber.org/zap

### Package

`go.uber.org/zap`

### Version

Current `go.uber.org/zap` releases. Match the version-range style of `github.com/sirupsen/logrus`.

### Environment

Go services built with `otelc`. Kubernetes Deployments and local binaries.

### Use-case

Apps that log with zap and already create traces through other `otelc` instrumentations. Operators want `trace_id` and `span_id` on zap lines so they can join logs and traces.

### Telemetry

In scope:

- When an active span is present, add `trace_id` and `span_id` to the zap record.
- Read the ids from `runtime.GetTraceAndSpanID()`, same as slog and logrus.
- Do not add fields when there is no active span.
- Follow `instrumentation/github.com/sirupsen/logrus`.
- Add unit tests and an integration test that asserts `trace_id` / `span_id` on the log line.

Out of scope:

- slog (already in `main`)
- Export of log records to an OTel `LoggerProvider`

### Anything else?

Replaces stale-closed #297. Related: #1319

atharvamhaske:
## Description

Add compile-time log correlation for `go.uber.org/zap`.

The hook runs on `(*zapcore.CheckedEntry).Write` in `go.uber.org/zap/zapcore`. When a span is active, it appends `trace_id` and `span_id` fields. `Logger` and `SugaredLogger` both go through that method.

This matches the existing `log`, `log/slog`, and `logrus` hooks. It does not export OTel log records and does not install a `LoggerProvider`.

Supported from zap `v1.21.0`. Disable with `OTEL_GO_DISABLED_INSTRUMENTATIONS=logs/zap`.

## Motivation

`go.uber.org/zap` is a common structured logger in Go. Without this hook, zap lines have no `trace_id` / `span_id`, so they cannot be joined to the request span. slog and logrus already have this.

Fixes #1331


instrumentation/go.uber.org/zap/zap_hook.go
```diff
@@ -0,0 +1,47 @@
...
+	if !enabler.Enable() || ce == nil {
+		return
+	}
+
+	traceID, spanID := runtime.GetTraceAndSpanID()
+	if traceID == "" {
+		return
+	}
+
+	fields = append(fields, stringField(traceIDKey, traceID))
```
darccio: `fields = append(fields, ...)` appends directly onto the caller-supplied variadic slice. When the caller passes a `[]zapcore.Field` that was itself built via `append` (leaving spare capacity), this can write into and alias the caller's backing array instead of allocating a fresh one — corrupting a slice the caller may reuse on a later log call. zap's own `Logger.Info`/`Warn`/etc. always pass fields straight from the call site, so this doesn't trigger from a plain `logger.Info("msg", zap.String(...))` — the risk is limited to app code that pre-builds and reuses a `[]zap.Field` across multiple log calls.

Consider a defensive copy before appending, e.g.:

```go
newFields := make([]zapcore.Field, len(fields), len(fields)+2)
copy(newFields, fields)
fields = newFields
```

atharvamhaske: ahh yes makes sense, doing it

darccio (commented):
Walked through the change area by area (hook implementation, unit tests, module wiring, manifest/bundle, docs/semconv, test app + integration test) against issue #1331's agreed scope. One inline finding posted below; everything else checks out — design matches the issue scope, version floor (v1.21.0) verified safe against zap's `Write` signature back to v1.20.0, manifest/bundle in sync, docs updated consistently, and CI is fully green.

atharvamhaske:
@darccio did the fix mentioned in comment, can be merged !

darccio (approved):

PR #1340, opened 2026-09-09
feat(instrumentation/elastic): add olivere/elastic v7 client spans

### LINKED ISSUE #1322: New Instrumentation Request for github.com/olivere/elastic/v7

- [X] I have searched the existing issues

### Package

github.com/olivere/elastic/v7

### Version

v7.0.32

### Environment

As Deployments on Kubernetes.

### Use-case

Services that read and write to Elasticsearch or OpenSearch through this client.

### Telemetry

Client spans for the request, with the operation, target index and outcome, so a search or index call is visible as a child of the inbound request span.

Today these calls are invisible: the client builds its own `*http.Client`, so `net/http/client` does not produce a span for them either, and a trace goes silent across the whole datastore hop.

### Anything else?

The package is archived upstream but still widely used; `github.com/opensearch-project/opensearch-go` is the maintained successor and has the same gap, so a shared approach may be worth considering.

atharvamhaske:
## Description

Add compile-time client spans for `github.com/olivere/elastic/v7`.

The hook runs on `(*Client).PerformRequest`. That is the shared HTTP path for Search, Index, Bulk, Delete, and the rest of the API. Each call becomes one CLIENT span named `{operation} {index}` (for example `search orders`).

The span records:

- `db.system.name=elasticsearch`
- `db.operation.name` (search, index, bulk, …)
- `db.collection.name` (target index, when the path has one)
- HTTP method, path, and `db.response.status_code`

The request context is updated so the span is a child of the inbound request. Inner `net/http` client spans are suppressed so the datastore hop is not also shown as a generic GET/POST. The host is chosen inside `PerformRequest` after the before hook, so this span does not set `server.address` or `url.full`.

`IgnoreErrors` is honored (Exists 404 / UpdateByQuery 409 stay `Unset`). Encoded paths are unescaped. Cluster-style ops use an allow-list so `/_tasks/{id}` is not a high-cardinality name.

Supported from elastic `v7.0.0` (covers the requested `v7.0.32`). Disable with `OTEL_GO_DISABLED_INSTRUMENTATIONS=elastic`.

The package is archived upstream but still widely used. This change does not add `opensearch-go`. That client is a separate gap.

Fixes #1322

## Test plan

- [x] Unit tests for path parse, IgnoreErrors, encoded/wildcard paths, disable
- [x] Integration: Search, Index, Delete, exact span counts
- [x] Integration: `OTEL_GO_DISABLED_INSTRUMENTATIONS=elastic`


instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+		Path:      opt.Path,
+		Operation: operation,
+		Index:     index,
+	}
+	ctx, span := tracer.Start(ctx,
+		esemconv.SpanName(operation, index),
+		trace.WithSpanKind(trace.SpanKindClient),
+		trace.WithAttributes(esemconv.ClientTraceAttrs(req)...),
+	)
+	ctx = runtime.SuppressHTTPClientInstrumentation(ctx)
```
darkraider01: critical: suppressing the http client span here kills url.full/server.address/error.type, which is the only place those currently show up. net loss of info vs before this PR. suggest not suppressing, let it nest under this span like linodego does.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+// parseElasticPath extracts the Elasticsearch operation and target index from
+// a REST method + path. The first path segment that does not start with '_'
+// is the index. The operation comes from a later '_'-prefixed action
+// (search, bulk, …) or, for document routes (/_doc, /_create), from the
+// HTTP method.
+func parseElasticPath(method, path string) (operation, index string) {
+	if i := strings.IndexByte(path, '?'); i >= 0 {
+		path = path[:i]
+	}
+	segs := splitPath(path)
```
darkraider01: critical: olivere percent-encodes path segments (uritemplates.Expand), never decoded here. multi-index search `orders,logs` -> `orders%2Clogs` in both span name and db.collection.name. same for wildcards (`logs-*` -> `logs-%2A`) and non-ascii. needs a url-decode on segments.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+	return methodFallback(method)
+}
+
+func lastKnownAction(segs []string) string {
+	var found string
+	for _, s := range segs {
+		if !strings.HasPrefix(s, "_") {
+			continue
+		}
+		if op, ok := knownActions[strings.TrimPrefix(s, "_")]; ok {
```
darkraider01: non-blocking: this just grabs the last `_`-prefixed segment anywhere in the path, ignoring method. so `DELETE /_search/scroll` (ClearScroll) resolves to operation `search`, same as an actual search. would rather see method factored in for cases like this.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+	if !ok || data == nil || data.span == nil {
+		logger.Debug("AfterPerformRequest: no span from before hook")
+		return
+	}
+	span := data.span
+	defer span.End()
+
+	if resp != nil && resp.StatusCode > 0 {
+		span.SetAttributes(semconv.HTTPResponseStatusCode(resp.StatusCode))
+		if resp.StatusCode >= 400 {
```
darkraider01: critical: olivere's IgnoreErrors (Exists/IndicesExists/Delete use 404, UpdateByQuery uses 409) makes those "errors" intentional non-errors. err is nil, resp.StatusCode is just informational. this marks the span Error for any status >= 400 regardless, so every "does this index exist" check on a missing index gets flagged as an error span. should check opt.IgnoreErrors before setting error status.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+func clusterStyleOp(segs []string) string {
+	if len(segs) == 0 || !strings.HasPrefix(segs[0], "_") {
+		return ""
+	}
+	head := strings.TrimPrefix(segs[0], "_")
+	if len(segs) > 1 {
+		next := segs[1]
+		next = strings.TrimPrefix(next, "_")
+		if next != "" {
+			return head + "." + next
```
darkraider01: critical: this blindly joins segs[0]+segs[1]. for `/_tasks/{task_id}` or `/_nodes/{node_id}/stats`, segs[1] is a random id, not a fixed word, so db.operation.name (and span name) becomes `tasks.<random-id>`, unbounded cardinality. should only join when segs[1] is a known literal, not any arbitrary segment.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	ctx, span := tracer.Start(ctx,
+		esemconv.SpanName(operation, index),
+		trace.WithSpanKind(trace.SpanKindClient),
+		trace.WithAttributes(esemconv.ClientTraceAttrs(req)...),
+	)
+	ctx = runtime.SuppressHTTPClientInstrumentation(ctx)
+	ictx.SetParam(ctxParamIndex, ctx)
+	ictx.SetData(&hookData{span: span})
+
```
darkraider01: non-blocking: other hooks in this repo (see linodego) use SetKeyData instead of SetData for this exact pattern, so a second rule hooking the same func doesn't stomp this data. worth matching that convention even though nothing else hooks PerformRequest today.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+	"go.opentelemetry.io/otel/trace"
+
+	esemconv "go.opentelemetry.io/otelc/instrumentation/github.com/olivere/elastic/v7/semconv"
+	"go.opentelemetry.io/otelc/pkg/hook"
+	"go.opentelemetry.io/otelc/pkg/runtime"
+)
+
+const (
+	instrumentationName = "go.opentelemetry.io/otelc/instrumentation/github.com/olivere/elastic/v7"
+	instrumentationKey  = "ELASTIC"
```
darkraider01: non-blocking: `ELASTIC` as the key is pretty broad. opensearch-go and elastic/go-elasticsearch instrumentations would want this space too. might be worth a more specific key now (e.g. OLIVERE_ELASTIC) so it doesn't collide later.

instrumentation/github.com/olivere/elastic/v7/semconv/client.go
```diff
@@ -0,0 +1,71 @@
...
+		attrs = append(attrs, semconv.DBOperationName(op))
+	}
+	if index := strings.TrimSpace(req.Index); index != "" {
+		attrs = append(attrs, semconv.DBCollectionName(index))
+	}
+	if method := strings.ToUpper(strings.TrimSpace(req.Method)); method != "" {
+		attrs = append(attrs, semconv.HTTPRequestMethodKey.String(method))
+	}
+	if path := normalizePath(req.Path); path != "" {
+		attrs = append(attrs, semconv.URLPath(path))
```
darkraider01: critical: ES semconv wants url.full (required), db.response.status_code + error.type (conditionally required), server.address (recommended). this emits url.path + http.response.status_code instead, not the ES-specific attrs. combined with suppressing the http span, none of the real url/server info survives anywhere in the trace.

instrumentation/github.com/olivere/elastic/v7/semconv/client.go
```diff
@@ -0,0 +1,71 @@
...
+	default:
+		return "elasticsearch"
+	}
+}
+
+// ClientTraceAttrs returns trace attributes for an Elasticsearch client request.
+func ClientTraceAttrs(req Request) []attribute.KeyValue {
+	attrs := []attribute.KeyValue{
+		semconv.DBSystemNameElasticsearch,
+		semconv.NetworkTransportTCP,
```
darkraider01: non-blocking: network.transport=tcp hardcoded with nothing else in the network.* namespace feels a little unmoored. other db instrumentations here (cassandra) don't set it without server address/port alongside it.

instrumentation/github.com/olivere/elastic/v7/README.md
```diff
@@ -0,0 +1,40 @@
...
+[`github.com/olivere/elastic/v7`](https://github.com/olivere/elastic).
+
+`(*Client).PerformRequest` is the shared HTTP path for Search, Index, Bulk,
+Delete, and the rest of the API. Each call becomes one CLIENT span.
+
+| Target | Span name | Attributes |
+| --- | --- | --- |
+| `(*Client).PerformRequest` | `{operation} {index}` | `db.system.name=elasticsearch`, `db.operation.name`, `db.collection.name` (index), HTTP method/path/status |
+
+Inner `net/http` client spans are suppressed so the datastore hop is not
```
darkraider01: non-blocking: worth calling out here that suppressing the http span also drops retry visibility. PerformRequest retries internally and this span only shows the final outcome, so multiple actual HTTP attempts collapse into one span with no trace of the retries.

instrumentation/github.com/olivere/elastic/v7/hook_test.go
```diff
@@ -0,0 +1,228 @@
...
+		{"GET", "/orders/_search?pretty=true", "search", "orders"},
+		{"PUT", "/orders/_doc/1", "index", "orders"},
+		{"POST", "/orders/_doc/", "index", "orders"},
+		{"GET", "/orders/_doc/1", "get", "orders"},
+		{"DELETE", "/orders/_doc/1", "delete", "orders"},
+		{"PUT", "/orders/_create/1", "create", "orders"},
+		{"POST", "/orders/_update/1", "update", "orders"},
+		{"POST", "/_bulk", "bulk", ""},
+		{"POST", "/orders/_bulk", "bulk", "orders"},
+		{"POST", "/orders,logs/_search", "search", "orders,logs"},
```
darkraider01: test gap: olivere can't actually produce `/orders,logs/_search` literally, it percent-encodes the comma to `%2C`. this test case exercises input that never happens in real life, which is exactly why the encoding bug wasn't caught. swap for the real encoded form, plus a wildcard and a unicode index case.

darkraider01 (commented):
Looking through the implementation, there are a few core behavior issues and semconv mismatches where we actually end up losing trace info compared to what we had before, plus several test gaps. Here’s a breakdown:


### Implementation & Semconv
* **Net loss of URL / Host info:** Suppressing the inner `net/http` client span kills `url.full`, `server.address`, and `error.type`. Because this hook only sees the relative `opt.Path`, none of the server/URL info survives on the trace. Additionally, ES semconv expects `url.full`, `server.address`, and `db.response.status_code` rather than `url.path` + `http.response.status_code`.
* **Path encoding:** `olivere` percent-encodes path segments (never decoded here). Multi-index searches like `orders,logs` become `orders%2Clogs` in both the span name and `db.collection.name`. The same applies to wildcards and unicode index names.
* **`opt.IgnoreErrors` not respected:** Calls like `Exists()`/`IndicesExists()` intentionally treat 404 as non-error, and `UpdateByQuery()` does the same for 409. The hook flags any status `>= 400` as `codes.Error`, causing routine exists-checks to show up as errors.
* **Unbounded span names / operations:** `clusterStyleOp` blindly joins `segs[0] + "." + segs[1]`. For paths like `/_tasks/{task_id}` or `/_nodes/{node_id}/stats`, `segs[1]` is a dynamic ID/UUID, leading to high-cardinality `db.operation.name` and span names.

### Test Gaps
* **HTTP span suppression check:** `WaitForSpans(2)` / `(4)` only verifies a lower bound. A leaked HTTP client span won't fail the test we should assert exact span counts.
* **Real-world path encoding:** The `orders,logs` test case can't actually occur as written because the client encodes the comma. We need tests for percent-encoded, wildcard, and unicode index paths.
* **Path cardinality regression:** No test verifying `/_tasks/{id}` or `/_nodes/{id}/stats` style paths keep low cardinality.
* **Ignored error status codes:** No test verifying status codes in `opt.IgnoreErrors` (404 on exists, 409 on update-by-query) avoid error status.
* **Failure modes:** No integration tests for Elasticsearch error responses, network/transport errors, or context cancellations.
* **API coverage:** Only `Search` and `Index` are covered end-to-end (missing `Bulk`, `Delete`, `Scroll`).
* **Standard client setup:** Only `NewSimpleClient` is tested. The standard `elastic.NewClient` runs sniffing and health checks that may bypass this hook and leak plain HTTP spans.
* **Disabled flag:** No test for `OTEL_GO_DISABLED_INSTRUMENTATIONS=elastic` to confirm the app still works and fallback `net/http` spans return.

instrumentation/github.com/olivere/elastic/v7/hook_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt.setupEnv(t)
+			assert.Equal(t, tt.want, enabler.Enable())
+		})
+	}
+}
+
+func TestParseElasticPath(t *testing.T) {
```
darkraider01: test gap: no case here for `/_tasks/{task_id}` or `/_nodes/{node_id}/stats` style paths, exactly the ones that blow up cardinality in clusterStyleOp. add a couple so a regression here fails a test.

instrumentation/github.com/olivere/elastic/v7/hook_test.go
```diff
@@ -0,0 +1,228 @@
...
+	require.Len(t, spans, 1)
+	got := spans[0]
+	assert.Equal(t, codes.Error, got.Status().Code)
+	attrs := attrMap(got.Attributes())
+	assert.Equal(t, "get", attrs["db.operation.name"])
+	assert.Equal(t, int64(404), attrs["http.response.status_code"])
+	require.NotEmpty(t, got.Events(), "expected RecordError event")
+}
+
+func TestBeforeAfterPerformRequest_HTTPStatusError(t *testing.T) {
```
darkraider01: test gap: no test with a status code that's in opt.IgnoreErrors (404 exists-check, 409 update-by-query). right now that path is completely unverified and it's the one that'll misfire in prod the most.

test/integration/elastic_client_test.go
```diff
@@ -0,0 +1,138 @@
...
+	f.Start("elasticclient",
+		fmt.Sprintf("-front-port=%d", frontPort),
+		"-es-url="+es.URL,
+	)
+	testutil.WaitForTCP(t, fmt.Sprintf("127.0.0.1:%d", frontPort))
+
+	searchBody := hitEndpoint(t, frontPort, "/search")
+	require.Contains(t, searchBody, "hits=")
+
+	f.WaitForSpans(2)
```
darkraider01: test gap: WaitForSpans(2) is a lower bound, not exact. if the http client span leaks through despite suppression, this wouldn't catch it. use an exact count or explicitly assert no http/url.full span exists.

test/integration/elastic_client_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+
+	"go.opentelemetry.io/otelc/test/testutil"
+)
+
+func TestElasticClient(t *testing.T) {
```
darkraider01: test gap: only happy-path search+index covered. nothing here hits an ES error response, a transport error, or a cancelled context, so span status/error recording is untested end-to-end. also no case with OTEL_GO_DISABLED_INSTRUMENTATIONS=elastic to confirm the app still works and the plain http span comes back.

test/apps/elasticclient/main.go
```diff
@@ -0,0 +1,83 @@
...
+
+var (
+	esURL     = flag.String("es-url", "http://localhost:9200", "Elasticsearch URL")
+	frontPort = flag.Int("front-port", 8080, "port for HTTP frontend")
+)
+
+func main() {
+	flag.Parse()
+
+	client, err := elastic.NewSimpleClient(elastic.SetURL(*esURL))
```
darkraider01: test gap: only NewSimpleClient is exercised. the more common elastic.NewClient path adds sniffing/healthchecks which bypass this hook entirely and emit their own plain http spans. worth at least one test app route on that path since the span-count behavior differs there.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	span := data.span
+	defer span.End()
+
+	if resp != nil && resp.StatusCode > 0 {
+		span.SetAttributes(semconv.HTTPResponseStatusCode(resp.StatusCode))
+		if resp.StatusCode >= 400 {
+			span.SetStatus(codes.Error, "")
+		}
+	}
```
darccio, line 134: `opt.IgnoreErrors` (`PerformRequestOptions.IgnoreErrors []int`) tells the client
which status codes are not errors — e.g. `Exists()` treats 404 as a normal
negative result, `UpdateByQuery()` treats 409 as expected. When the response
status matches one of those codes, `checkResponse` in the vendored library
returns `nil`, so `PerformRequest` returns `(resp, nil)`.

This code marks the span `codes.Error` purely from `resp.StatusCode >= 400`,
ignoring that `err == nil` already signals the caller didn't consider this an
error. Every ignore-listed exists-check or ignore-409 update shows up as a
span error.

Plumb `opt.IgnoreErrors` through to this hook (via hook data set in
`BeforePerformRequest`) and skip `SetStatus(codes.Error, "")` for status codes
in that list, or otherwise defer the error/non-error decision to `err` rather
than the raw status code.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+		next = strings.TrimPrefix(next, "_")
+		if next != "" {
+			return head + "." + next
+		}
+	}
+	if head != "" {
+		return head
+	}
+	return ""
+}
```
darccio, line 270: `clusterStyleOp` unconditionally joins `segs[0]` and `segs[1]` whenever
`segs[0]` starts with `_`. That's correct for `/_cluster/health` →
`cluster.health`, but for `/_tasks/{task_id}` or `/_nodes/{node_id}/stats`,
`segs[1]` is a dynamic identifier, not a fixed sub-resource — this produces
unbounded-cardinality `db.operation.name` values (and span names) like
`tasks.a1b2c3d4`.

Consider a fixed allow-list of known second-segment sub-resources (similar in
spirit to `knownActions`), falling back to just `head` when the second
segment isn't recognized, so dynamic IDs don't leak into span names/attribute
values.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+		path = path[:i]
+	}
+	segs := splitPath(path)
+	rest := segs
+	if len(segs) > 0 && !strings.HasPrefix(segs[0], "_") {
+		index = segs[0]
+		rest = segs[1:]
+	}
+	return operationFrom(method, rest), index
+}
```
darccio, line 170: `opt.Path` arrives already percent-encoded by `olivere`'s URI templating for
reserved characters. A multi-index call like `Index("orders", "logs")` builds
its path via `uritemplates.Expand("/{index}/_search", ...)`; the plain
`{index}` template (no `+`/`#` modifier) runs with `allowReserved=false`,
which percent-encodes `,`. So `opt.Path` is actually `/orders%2Clogs/_search`,
not `/orders,logs/_search`.

`parseElasticPath` never unescapes the path, so the encoded form leaks into
both the span name and `db.collection.name` (e.g. `search orders%2Clogs`).
The same applies to wildcards and non-ASCII index names. Consider running the
path (or just the extracted index segment) through `url.PathUnescape` before
using it in span name / attributes.

instrumentation/github.com/olivere/elastic/v7/hook.go
darccio, line 109: Suppressing the inner `net/http` client span here removes the only place that
captured `server.address`/`url.full`/`error.type` for the resolved connection.
`(*Client).PerformRequest` resolves the target host internally via `c.next()`,
inside the method body — after this hook runs and before `AfterPerformRequest`
runs — but the resolved host is never surfaced back out to this span. As
written, this span structurally cannot carry host/URL info even if
`ClientTraceAttrs` were extended to look for it, which is a net loss versus
the suppressed span.

Worth a design decision: either accept the trade-off explicitly (and document
it), find a hook point that sees the resolved connection so `server.address`/
`url.full` can be added here, or don't suppress the inner span and instead
adjust/reparent it under the client span.

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+	"go.opentelemetry.io/otelc/pkg/hook"
+	"go.opentelemetry.io/otelc/pkg/runtime"
+)
+
+const (
+	instrumentationName = "go.opentelemetry.io/otelc/instrumentation/github.com/olivere/elastic/v7"
+	instrumentationKey  = "ELASTIC"
+
+	// PerformRequest param indices (receiver is 0).
+	ctxParamIndex = 1
```
darccio: `ctxParamIndex = 1` hardcodes the assumption that `PerformRequest`'s signature
is `(ctx, opt)` with the receiver at index 0. That's true today, but if the
upstream signature ever reorders, `SetParam(ctxParamIndex, ctx)` would write
the new context into the wrong argument slot silently rather than failing at
compile/injection time.

Is there a check elsewhere (codegen, `otelc.yaml` validation, or a runtime
type assertion) that would catch a signature drift here, or does this rely
purely on the CI test suite failing?

instrumentation/github.com/olivere/elastic/v7/hook.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	switch method {
+	case "GET", "HEAD":
+		return "get"
+	case "DELETE":
+		return "delete"
+	default:
+		return "index"
+	}
+}
```
darccio, line 246: `docOperation` returns `"index"` as the default case for any method other
than GET/HEAD/DELETE reaching a `_doc` path — covering the two real cases
(PUT = full index, POST = index-with-auto-id). Is there a third method that
can reach this path with different semantics, or is the `default: return
"index"` intentionally the catch-all for anything else the client might send?

instrumentation/github.com/olivere/elastic/v7/semconv/client.go
```diff
@@ -0,0 +1,71 @@
...
+		attrs = append(attrs, semconv.DBCollectionName(index))
+	}
+	if method := strings.ToUpper(strings.TrimSpace(req.Method)); method != "" {
+		attrs = append(attrs, semconv.HTTPRequestMethodKey.String(method))
+	}
+	if path := normalizePath(req.Path); path != "" {
+		attrs = append(attrs, semconv.URLPath(path))
+	}
+	return attrs
+}
```
darccio, line 58: The Elasticsearch semantic convention
(https://opentelemetry.io/docs/specs/semconv/database/elasticsearch/) calls
for `url.full`, `server.address`, and `db.response.status_code`. This emits
`url.path` (relative, here) instead of `url.full`, never sets
`server.address`, and status is set as `http.response.status_code` in
`AfterPerformRequest` rather than `db.response.status_code`.

`server.address`/`url.full` need the resolved connection host, which
`(*Client).PerformRequest` only resolves internally via `c.next()` — not
available at this hook boundary today (see the `hook.go:98` comment on
suppressing the inner `net/http` span). But `db.response.status_code` doesn't
have that dependency and could replace `http.response.status_code` now.

Given `schemas/otelc/groups/elasticsearch.yaml` in this PR codifies this
attribute set as the contract, is this divergence from the ES convention an
intentional interim shape, or should `db.response.status_code` swap in before
merge?

test/integration/elastic_client_test.go
darccio, line 39: `f.WaitForSpans(2)` and `f.WaitForSpans(4)` wait for *at least* N spans, then
the test looks up specific spans by predicate (`IsServer`/`IsClient` +
attribute match). If suppression of the inner `net/http` client span
partially failed and leaked an extra span, both calls would still pass,
since the predicates find their target spans regardless of extras.

Consider asserting the exact span count at each checkpoint (e.g.
`require.Len(t, f.Traces()..., 2)` / `4`) so a suppression regression on this
path fails the test instead of passing silently.

The same applies to line 67.

test/integration/elastic_client_test.go
darccio, line 22: The PR description says the hook covers "Search, Index, Bulk, Delete, and
the rest of the API," but this test only exercises `Search` and `Index`
end-to-end. There's no integration coverage for `Bulk`/`Delete`/`Scroll`, and
the test app only uses `elastic.NewSimpleClient` (skips sniffing/healthcheck)
— the standard `elastic.NewClient` constructor path isn't exercised, so it's
untested whether sniffing/healthcheck requests interact with the hook
differently (see the cross-cutting question on whether `PerformRequest` is
really the single choke point for all client-issued requests).

Would be good to add at least one `Bulk` or `Delete` case, and a note (or
follow-up issue) tracking the standard-client gap.

darccio, line 22: The disabled path (`OTEL_GO_DISABLED_INSTRUMENTATIONS=elastic`) is covered at
the unit level (`TestBeforePerformRequest_Disabled` in `hook_test.go`), but
there's no integration-level test confirming the compiled binary still works
end-to-end with the instrumentation disabled — i.e. that disabling doesn't
silently break the app or leave an unexpected tracing gap, and that fallback
`net/http` client spans reappear as expected.

Worth adding an integration test variant that sets the disable env var and
checks the app still serves `/search`/`/index` successfully.

test/integration/elastic_client_test.go
```diff
@@ -0,0 +1,138 @@
...
+	require.Equal(t, searchServer.SpanID(), searchClient.ParentSpanID())
+	testutil.RequireElasticsearchClientSemconv(
+		t,
+		searchClient,
+		"search",
+		"orders",
+		"POST",
+		"/orders/_search",
+		200,
+	)
```
darccio, line 63: This asserts the `search` span's status via `RequireElasticsearchClientSemconv`
with a 200 status, and the unit tests (`hook_test.go`) cover the "should be
an error" cases (`_Error`, `_HTTPStatusError`, `_PlainError`). But there's no
test — unit or integration — for the negative-space case tied to the
`opt.IgnoreErrors` bug flagged on `hook.go:120-125`: a call where Elasticsearch/
the client itself doesn't consider the response an error (status in
`IgnoreErrors`, `err == nil`) should leave the span `codes.Unset`, not
`codes.Error`.

Once that hook.go fix lands, a test here (or in `hook_test.go`) asserting
`codes.Unset` for an ignored-status response would close the loop and prevent
a regression.

darccio (changes requested):
Is `(*Client).PerformRequest` actually the single choke point for every code
path that talks to Elasticsearch here, or are there client-issued requests
that bypass it?

Two concrete cases worth checking against the vendored `olivere/elastic`
source before relying on this hook as complete coverage:

- The standard `elastic.NewClient` constructor (not exercised anywhere in
  this PR — only `NewSimpleClient` is used, which skips sniffing/healthcheck)
  runs periodic sniffing and healthchecks. Do those go through
  `PerformRequest`, or a separate internal path that would leak plain
  `net/http` spans instead of a client span?
- The PR description claims coverage for "Search, Index, Bulk, Delete, and
  the rest of the API," but only `Search`/`Index` have any span-level test
  coverage. Has the full call graph actually been checked to confirm every
  service builder funnels through `PerformRequest`, or is broader coverage
  inferred from `Search`/`Index` working?

If either turns out not to route through `PerformRequest`, that's a silent
coverage gap (missing span, or a leaked HTTP span) rather than a
attribute-shape issue — worth settling with a source check rather than
closing on assumption.

atharvamhaske:
> Is `(*Client).PerformRequest` actually the single choke point for every code path that talks to Elasticsearch here, or are there client-issued requests that bypass it?
> 
> Two concrete cases worth checking against the vendored `olivere/elastic` source before relying on this hook as complete coverage:
> 
> * The standard `elastic.NewClient` constructor (not exercised anywhere in
>   this PR — only `NewSimpleClient` is used, which skips sniffing/healthcheck)
>   runs periodic sniffing and healthchecks. Do those go through
>   `PerformRequest`, or a separate internal path that would leak plain
>   `net/http` spans instead of a client span?
> * The PR description claims coverage for "Search, Index, Bulk, Delete, and
>   the rest of the API," but only `Search`/`Index` have any span-level test
>   coverage. Has the full call graph actually been checked to confirm every
>   service builder funnels through `PerformRequest`, or is broader coverage
>   inferred from `Search`/`Index` working?
> 
> If either turns out not to route through `PerformRequest`, that's a silent coverage gap (missing span, or a leaked HTTP span) rather than a attribute-shape issue — worth settling with a source check rather than closing on assumption.

Read client.go in source repo code. Public API goes through PerformRequest. sniffNode / healthcheck call c.c.Do.

instrumentation/github.com/olivere/elastic/v7/hook.go
atharvamhaske: Plumbed `opt.IgnoreErrors` through hook data. Ignored 404/409 with err == nil stay Unset. Added `TestBeforeAfterPerformRequest_IgnoredStatus`

instrumentation/github.com/olivere/elastic/v7/hook.go
atharvamhaske: Allow-list for the second segment. 
`/_tasks/{id}` → `tasks`, 
`/_nodes/{id}/stats` →` nodes.stats`

instrumentation/github.com/olivere/elastic/v7/hook.go
atharvamhaske: `url.PathUnescape `before span name / index. Tests use %2C and a unicode path

instrumentation/github.com/olivere/elastic/v7/hook.go
atharvamhaske: Kept suppress. Documented that next() runs after the before-hook, so no `server.address / url.full`

instrumentation/github.com/olivere/elastic/v7/hook.go
atharvamhaske: Comment only. Inject-time signature match fails the build if PerformRequest changes.

instrumentation/github.com/olivere/elastic/v7/hook.go
atharvamhaske: Comment. PUT/POST on /_doc are both index.

instrumentation/github.com/olivere/elastic/v7/semconv/client.go
atharvamhaske: `http.response.status_code `→` db.response.status_code.` Host attrs still omitted (same next() hole)

test/integration/elastic_client_test.go
atharvamhaske: done !

test/integration/elastic_client_test.go
atharvamhaske: Test `app /delete `→ `client.Delete().Index("orders").Id("1").` Integration asserts a `delete` CLIENT span. `NewClient/sniff` still only documented

atharvamhaske: `TestElasticClient_Disabled `sets `OTEL_GO_DISABLED_INSTRUMENTATIONS=elastic`. App still serves. 
No **`db.system.name=elasticsearch`**. Inner `net/http` client spans come back (exact count 4 after search+index).

test/integration/elastic_client_test.go
atharvamhaske: done !

atharvamhaske:
@darccio addressed all the comments and made changes accordingly, can be reviewed again 
thanks !

darccio (approved):

Issue #1265, opened 2026-08-27
chore(ci): keep golangci-lint version in sync with .tools/go.mod


Harsh63870:
### Describe the bug
CI and local golangci-lint versions have drifted. Local `make lint/go` builds golangci-lint from `.tools/go.mod` (v2.12.2). CI pins a separate, hardcoded version in `.github/workflows/lint-golang.yaml` (v2.11.3) via a `GOLANGCI_LINT_VERSION` env var, with an existing TODO:

```yaml
# lint-golang.yaml — Lines 11-12
# TODO: Configure dependabot or renovate to update this version.
GOLANGCI_LINT_VERSION: v2.11.3
```

Renovate already bumps `.tools/go.mod`, but `.github/**` is disabled in `.github/renovate.json5` (actions are ratchet-pinned), so the workflow env var never tracks the tools module. The path filter also omits `.tools/go.mod` / `.tools/go.sum`, so a tools-only bump doesn't even re-run lint. Net result: CI and local lint can disagree.

### Environment
- OS: N/A (CI/tooling issue, not platform-specific)
- Go Version: N/A
- Version: `.tools/go.mod` golangci-lint v2.12.2 vs CI-pinned v2.11.3

### To Reproduce
1. Bump `.tools/go.mod` to a newer golangci-lint version (Renovate does this automatically)
2. Run `make lint/go` locally — it builds/uses the new version
3. Push and let `.github/workflows/lint-golang.yaml` run — it still uses the stale `GOLANGCI_LINT_VERSION` pin
4. See version mismatch between local and CI lint results

### Expected behavior
CI should read the golangci-lint version from `.tools/go.mod` (the same source of truth local builds use) instead of a separately hardcoded env var, so the two never drift. A tools-only version bump should also trigger the lint workflow via the path filter.

### Additional context
Proposed fix:
1. Drop `GOLANGCI_LINT_VERSION`.
2. After `setup-go`, read the version from `.tools/go.mod` and pass it to `golangci-lint-action`.
3. Include `.tools/go.mod` / `.tools/go.sum` in the workflow path filter.
4. Remove the TODO.

Suggested workflow change:

```yaml
- uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c  # v6.1.0
  with:
    go-version-file: go.mod

- name: Read golangci-lint version
  id: golangci
  run: |
[...]

PR #1342, opened 2026-09-09
chore(ci): add .tools/go.mod to lint-actionlint path filter

### LINKED ISSUE #1334: chore(ci): add .tools/go.mod to lint-actionlint path filter
### Describe the bug

`.github/workflows/lint-actionlint.yaml` runs `make lint/action`, which builds both actionlint and ratchet from `.tools/go.mod`. But the workflow's `paths:` filter only watches workflow and action files, so a PR that bumps only actionlint or ratchet does not trigger the job at all.

`.github/workflows/lint-actionlint.yaml` lines 4-16:

```yaml
on:
  pull_request:
    paths:
      - ".github/workflows/*.yml"
      - ".github/workflows/*.yaml"
      - ".github/actions/**/action.yml"
      - ".github/actions/**/action.yaml"
  push:
    branches: [main]
    paths:
      - ".github/workflows/*.yml"
      - ".github/workflows/*.yaml"
      - ".github/actions/**/action.yml"
      - ".github/actions/**/action.yaml"
```

The job's only real step (line 33) is:

```yaml
      - run: make lint/action
```

And Makefile line 267 shows both tools come from the tools module:

```
lint/action: $(ACTIONLINT) ratchet/check
```

`.tools/go.mod` pins both:

```
github.com/rhysd/actionlint v1.7.12
github.com/sethvargo/ratchet v0.11.4
```

### Why this is a problem

When Renovate opens a PR bumping actionlint or ratchet in `.tools/go.mod`, no path in the filter matches, so *Check GitHub Actions* never runs on that PR. A tool version with new or changed rules merges without ever validating the repo's workflows against it. The breakage only surfaces later, on an unrelated PR that happens to touch a workflow file.

This is the same failure mode as #1265, which fixed the equivalent gap in `lint-golang.yaml`.

### Expected behavior

The filter should include the tools module, matching what the sibling workflows already do:

```yaml
    paths:
      - ".github/workflows/*.yml"
      - ".github/workflows/*.yaml"
      - ".github/actions/**/action.yml"
      - ".github/actions/**/action.yaml"
      - ".tools/go.mod"
      - ".tools/go.sum"
```

### Additional context

`lint-actionlint.yaml` is the only remaining Go-tool workflow with this gap. Both peers already watch the tools module:

- `.github/workflows/lint-golang.yaml` lines 34-35 — `.tools/go.mod`, `.tools/go.sum`
- `.github/workflows/check-makefile.yaml` lines 10-11 and 17-18 — `.tools/go.mod`, `.tools/go.sum`

Note this workflow uses top-level `paths:` rather than the `dorny/paths-filter` + done gate pattern used by `lint-golang.yaml`, so a non-match means the workflow is skipped entirely rather than short-circuited by a passing gate job.

**Effort:** low: two lines added to two filter blocks.

**Suggested labels:** `scope:chore`, `effort:low`

atharvamhaske:
## Description

Add `.tools/go.mod` and `.tools/go.sum` to both `paths:` filters in `.github/workflows/lint-actionlint.yaml`.

`make lint/action` builds actionlint and ratchet from the tools module. The filter previously watched only workflow and action files. A PR that only bumps those tools did not run Check GitHub Actions.

This matches `check-makefile.yaml`, which already watches the tools module.

## Motivation

When Renovate bumps actionlint or ratchet, the new or changed rules must run against the repo workflows on that PR. The breakage must not wait for a later PR that happens to touch a workflow file.

Same gap as #1265 for `lint-golang.yaml`.

Fixes #1334


darccio (approved):
LGTM

PR #1344, opened 2026-09-09
fix(tool/imports): write importcfg atomically instead of truncating in place

### LINKED ISSUE #1343: fix(tool/imports): write importcfg atomically instead of truncating in place
### Description

`ImportConfig.WriteFile` opens the target with `os.Create`:

```go
func (r *ImportConfig) WriteFile(filename string) error {
	file, err := os.Create(filename)
	if err != nil {
		return ex.Wrapf(err, "failed to create file %s", filename)
	}
	return r.writeFile(file, filename)
}
```

`os.Create` truncates immediately, before anything is written. If the write
doesn't finish (Ctrl-C mid-build, disk full, an NFS hiccup) the importcfg is left
truncated with no way to recover it.

The catch is whose file it is. Both callers pass the toolchain's own `-importcfg`
path — `updateImportConfig` (`toolexec.go:190`) and `interceptLink`
(`toolexec.go:343`) — so the failure surfaces as a Go toolchain error about
missing packages rather than anything pointing back at otelc.

`ast.WriteFile` also uses `os.Create`, but its call sites all write new files into
otelc's work dir, so nothing pre-existing is at risk there. This is the only place
we truncate a file the toolchain created and still needs.

### Location

`tool/internal/imports/importcfg.go:107` (`ImportConfig.WriteFile`)

### Proposed Solution

Render to a buffer and use the helper we already have:

```go
func (r *ImportConfig) WriteFile(filename string) error {
	var buf bytes.Buffer
	if err := r.write(&buf); err != nil {
		return ex.Wrapf(err, "failed to render importcfg for %s", filename)
	}
	return util.WriteFileAtomic(filename, buf.Bytes())
}
```

`util.WriteFileAtomic` (`tool/util/sys.go:149`) writes to a temp file and renames,
so the old file survives until the new one is complete. #935 made this same change
for `SetupPhase.store` and was closed as completed. (#1275 touched
`ImportConfig.writeFile` recently, but that was about fd leaks on panic, not this.)

importcfg files are small, so buffering is fine. The existing `writeFile` helper
can stay for the tests that use it directly.

Happy to put up a PR.

krsatyamthakur-droid:
## Description

`ImportConfig.WriteFile` used `os.Create`, which empties the file before writing anything
back to it. If the write doesn't finish, you're left with a truncated importcfg and no way to
recover it, since the only complete copy was in memory.

That hurts here because the file isn't ours. Both callers hand us the toolchain's own
`-importcfg` path (`toolexec.go:190` and `toolexec.go:343`), so the toolchain ends up reading
a file we emptied.

The fix renders the config into a buffer first, then lets `util.WriteFileAtomic` write it.
That writes a temp file next to the target and renames over it, so if anything fails the old
file is still there, untouched. Permissions are kept as well.

## One thing to know

Rename needs write access to the directory, while `os.Create` needed it on the file itself.
So a read-only importcfg now gets replaced instead of refused. The file lives in the
toolchain's `$WORK` and is never read-only in practice, but it is a real change so I'd rather
mention it.

It's also why two tests moved: `TestUpdateImportConfigWriteError` and
`TestInterceptLinkWriteError` used to make the target unwritable to force a failure, and now
make the directory unwritable instead. They still check the same thing.

## Tests

Added three: the old file is byte-for-byte unchanged when the write can't go through, a
`0640` file keeps its mode, and no temp file is left behind. `writeFile` itself is untouched
and keeps its existing tests, including the panic one from #1275.

`go test ./tool/...` passes, and the bundle isn't affected since this only touches `tool/`.

## Motivation

A truncated importcfg is short, not malformed, so the toolchain just says a package is
missing. That gives you nothing to trace back to otelc, which is a rough thing to debug for
how little code it takes to avoid.

#935 did the same thing for `SetupPhase.store` and was merged. `util.WriteFileAtomic` is
already used in `gen-manifest`, `setup/` and `ast/parser.go` — this was the last place we
still overwrote an existing file in place.

Fixes #1343


tool/internal/imports/importcfg.go
```diff
@@ -104,12 +106,18 @@ func parse(r io.Reader) (ImportConfig, error) {
...
 func (r *ImportConfig) WriteFile(filename string) error {
-	file, err := os.Create(filename)
-	if err != nil {
-		return ex.Wrapf(err, "failed to create file %s", filename)
+	var buf bytes.Buffer
+	if err := r.write(&buf); err != nil {
+		return ex.Wrapf(err, "failed to render importcfg for %s", filename)
 	}
-	return r.writeFile(file, filename)
+	return util.WriteFileAtomic(filename, buf.Bytes())
```
ADITYA-CODE-SOURCE, line 114: `WriteFileAtomic` states that replacement is not guaranteed to be atomic on non-Unix platforms, but `WriteFile` now appears to promise unconditional atomicity. Since Windows is supported, have we verified that replacing an existing toolchain `importcfg` works there, including when another process has the destination open without delete sharing? Could we either use an approach that is safe across all supported platforms or qualify the guarantee and add Windows-specific test coverage?

krsatyamthakur-droid: Good catch on both.

- You're right that the doc comment was overpromising. `util.WriteFileAtomic` only claims
  atomicity for same-filesystem renames on Unix, and I wrote it as if it were unconditional.
  Reworded it to say what this call site actually cares about — **the target never ends up
  half-written**.

- The Windows question turned up a bug in my own test, so thanks for asking. I'd made the
  *directory* read-only to force the failure, but `os.Chmod` on Windows just sets
  `FILE_ATTRIBUTE_READONLY`, and that doesn't stop you creating files inside a directory.
  **It would have failed on `windows-latest`.** Guarded it the same way `manifest_test.go`
  guards its chmod test.

- Also added `TestWriteFile_ReplacesExistingFile`, which rewrites an existing importcfg and
  parses it back. No permission assumptions in it, so **it actually runs on Windows** and
  covers the replace-over-existing path you were asking about.

- On the open-handle thing — `MoveFileEx` does need `FILE_SHARE_DELETE`, but `os.Create` had
  a similar problem via `FILE_SHARE_WRITE`, so the gap between old and new is pretty narrow.
  And where they do differ, **the new code fails cleanly instead of truncating**. As far as I
  can tell the go driver closes the importcfg before it invokes the tool through `-toolexec`,
  so nothing should be holding it open — but if you know of a case where something does, I'd
  rather handle it than guess.

> [!NOTE]
> I don't have a Windows machine, so `GOOS=windows go vet ./tool/...` is as far as I could get
> locally. CI's the real check here.

tool/internal/imports/importcfg.go
```diff
@@ -104,12 +106,24 @@ func parse(r io.Reader) (ImportConfig, error) {
...
+// temporary file and renamed over the target, so a failure leaves the existing
+// file untouched instead of truncating it. The callers pass the Go toolchain's
+// own -importcfg path, and the parsed copy that could restore it does not
+// outlive the process.
+//
+// Whether the replacement itself is atomic is left to util.WriteFileAtomic,
+// which documents it as atomic on Unix-like systems for same-filesystem
+// renames and unspecified elsewhere. This function does not depend on that
+// stronger guarantee: what it needs is that the target is never left
+// half-written, which holds on every supported platform.
```
y1yang0: I don’t think you need to write such verbose comments. Remove entirely

y1yang0 (commented):
Otherwise LGTM

tool/internal/imports/importcfg.go
krsatyamthakur-droid: done

y1yang0 (approved):

PR #1345, opened 2026-09-09
fix(tool/internal/setup): drop -json from the build plan dry run

rarguelloF:
## Description

Setup derives its build plan from a dry run of the go command. That dry run now drops `-json`. The real build still receives it, so `-json` output is unchanged.

## Motivation

`otelc go test -json ./pkg` built and ran the tests, but produced an uninstrumented binary. No rule was applied and nothing failed loudly.

**Why it happens.** Setup finds the packages to instrument by running the build as a dry run:

```
go test -a -x -n <user args>
```

`-x -n` makes the go command print the compile commands it would run to stderr, which otelc parses. `-json` sends that output to stdout as JSON events instead. stderr is left empty, so the plan parses to nothing, no dependency matches, and `matched.json` is written as `[]`. The instrument phase then has nothing to apply.

**How to reproduce.** In any module with an instrumentation rule:

```
otelc go test -json ./pkg    # .otelc-build/matched.json is []
otelc go test ./pkg          # .otelc-build/matched.json is populated
```

`go build -json` and `go install -json` fail the same way, as do `--json` and `-json=true`. Only `-json=false` was unaffected.

**Second effect.** The dry run's JSON also reached stdout alongside the real run's, so the stream carried a duplicate `start` and `pass` for each package, with the extra `pass` arriving before the tests ran. Consumers of the stream saw every package twice.

### DIFF
--- a/tool/internal/setup/find.go
+++ b/tool/internal/setup/find.go
@@ -75,6 +75,61 @@ func findCommands(buildPlanLog *os.File) ([]string, error) {
 	return commands, nil
 }
 
+// planIrrelevantFlags names flags that change how the go command formats its
+// own output without changing the build plan. listBuildPlan reads the plan from
+// the go command's stderr, and -json moves it to stdout as a stream of JSON
+// build-output events instead, so the plan parses empty and no dependency is
+// found. The real build keeps the flag; only the dry run drops it.
+//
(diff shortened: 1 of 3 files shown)


darccio (approved):
LGTM. I opened issue #1346 for an unrelated bug found during this review.

PR #1349, opened 2026-09-10
feat(instrumentation/amqp): add rabbitmq/amqp091-go producer and consumer spans

### LINKED ISSUE #1013: New Instrumentation Request for rabbitmq/amqp091-go

### Package

rabbitmq/amqp091-go

### Version

_No response_

### Environment

Local, in docker, or on kubernetes

### Use-case

Event-driven applications with RabbitMQ as the message queue. Very popular message queue and this is the Go client library for intergating with RabbitMQ.

### Telemetry

Traces: Tracing the full operation starting from when a message is consumed till it is acknowledged (or other indicators of processing being finished for this message).
Metrics: could be how many jobs were positively acknowledged, negatively acknowledged, drooped, or other. Also how many jobs were sent.

### Anything else?

Package's github: https://github.com/rabbitmq/amqp091-go

### Tip

atharvamhaske:
## Description

This PR adds compile-time producer and consumer spans for `github.com/rabbitmq/amqp091-go`.

The publish hook is `(*Channel).PublishWithDeferredConfirm`. `Publish`, `PublishWithContext`, and `PublishWithDeferredConfirmWithContext` all call this method. As a result, one send creates one `PRODUCER` span. The span name is `{exchange} send`. If the exchange is empty, the name uses `(default)`.

The library drops the context of `PublishWithContext` before that call. When a GLS span exists, the send span gets its parent from that span. Publisher confirms do not extend the span.

`Consume`, `ConsumeWithContext`, and `Get` start one `CONSUMER` span for each delivery. A manual-ack span uses the `process` operation. This span ends on `Delivery.Ack`, `Nack`, or `Reject`.
An auto-ack span uses the `receive` operation. When the consumer reads the delivery, this span ends. The after hook replaces the consume channel. If that channel closes, leftover process spans end.

The hook injects W3C `traceparent` into `Publishing.Headers`. Then the hook extracts `traceparent` from `Delivery.Headers`.

The routing key is only `messaging.rabbitmq.destination.routing_key`. The routing key is not in the span name. `(*Channel).connection` is unexported.
As a result, these spans do not set `server.address`, `Channel.Ack` and `Channel.Nack` bypass `Delivery`. As a result, they do not end these spans.

The minimum version is `v1.10.0` because that version adds `ConsumeWithContext`.
**You can disable the instrumentation with `OTEL_GO_DISABLED_INSTRUMENTATIONS=amqp`.**
This change is trace-only. Ack and nack metrics are later work.

## Motivation

RabbitMQ is a common message queue for Go.
Issue #1013 asked for traces from consume through ack, and for send counts. This PR adds the traces.
Each message gets one send span and one process or receive span. The headers link the two spans.

Fixes #1013

## Test plan

- [x] Unit tests for publish headers, errors, disable, process until Ack, auto-ack, Nack/Reject, channel close, Get, and send-to-consume parent
- [x] Integration: exact 2 spans. The send span is the parent of the process span through headers
- [x] Integration: `OTEL_GO_DISABLED_INSTRUMENTATIONS=amqp` emits 0 spans


instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+	tracer     trace.Tracer
+	propagator propagation.TextMapPropagator
+	initOnce   sync.Once
+
+	errDeliveryNotInitialized = errors.New("delivery not initialized")
+
+	// ponytail: one parent slot per *Channel. Concurrent PublishWithContext
+	// on the same Channel can swap parents; serialize publishes or use one
+	// Channel per goroutine if that matters.
+	publishParents sync.Map // *amqp.Channel -> context.Context
```
darccio: Concurrent PublishWithContext/PublishWithDeferredConfirmWithContext calls on
the same *Channel can race between the before-hook's Store and the publish
hook's LoadAndDelete, letting one call's span parent from another call's
context. The comment documents this as an accepted limitation ("serialize
publishes or use one Channel per goroutine"), which lines up with
amqp091-go's own guidance against concurrent publishes on a single Channel —
worth confirming that assumption explicitly (e.g. a doc-comment cross-link or
a mention in the package README) rather than leaving it only in this
internal comment, since it's the kind of constraint a caller integrating
this instrumentation needs to know about up front.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+	span, ok := ictx.GetData().(trace.Span)
+	if !ok || span == nil {
+		return
+	}
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+	}
+	span.End()
+}
```
darccio, line 197: AfterPublishWithDeferredConfirm ends the span on client hand-off, not broker
confirmation — documented here in prose, but there's no span attribute
distinguishing the two. kafka-go's producer instrumentation sets a
`messaging.kafka.async` attribute for the analogous caveat (async write
handed off vs. confirmed). Consider whether an equivalent attribute here
would help trace consumers avoid misreading this span's duration/status as
delivery confirmation, or whether the prose caveat is considered sufficient
given RabbitMQ's confirm model differs from Kafka's.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+	)
+}
+
+func cloneTable(t amqp.Table) amqp.Table {
+	out := make(amqp.Table, len(t))
+	for k, v := range t {
+		out[k] = v
+	}
+	return out
+}
```
darccio, line 98: cloneTable does a shallow copy of the header table before injecting
traceparent. Since only top-level keys are read/written here (the injected
key is added, existing values are never mutated), a shallow copy is
sufficient for this call site — nested Table/[]interface{} values legal in
the AMQP wire format are passed through untouched either way. Worth a short
comment noting that the copy is deliberately shallow and why that's safe,
so a future change that starts mutating a value in place doesn't silently
reintroduce aliasing with the caller's original headers.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+	pending := &pendingAcks{}
+	go func() {
+		defer close(out)
+		defer pending.endAll()
+		for d := range in {
+			out <- startDeliverySpan(ch, queue, autoAck, d, pending)
+		}
+	}()
+	return out
+}
```
darccio, line 283: The wrapped channel out has no buffer. The goroutine blocks on the send to out until the consumer reads the delivery. The original amqp091-go delivery channel may hold a buffer, set by the library's own internal queue size. If the original channel buffers deliveries and this wrapper does not, the wrapper changes how many deliveries the consumer can hold in flight before the client applies backpressure to the broker. Check the buffer size of the original delivery channel in amqp091-go, and state in a comment whether the unbuffered wrapper preserves the original flow-control behavior.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+		Queue:           queue,
+		Operation:       op,
+		MessageBodySize: len(d.Body),
+	}
+	parent := propagator.Extract(context.Background(), amqprop.NewTableCarrier(&d.Headers))
+	_, span := startSpan(parent, req, trace.SpanKindConsumer)
+	if autoAck {
+		span.End()
+		return d
+	}
```
darccio, line 303: For an auto-ack delivery, startDeliverySpan ends the span before it returns the delivery to the caller. The span duration covers only the time from message enqueue to message read. The span does not cover the caller's processing of the message. The operation attribute reads receive, with no other attribute that marks this limit. A trace consumer may read the span duration as the processing time, which it is not. Add a short note, either in the span description or in a dedicated attribute, that states the span ends at read time, not at the end of processing.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+}
+
+// AfterGet attaches a consumer span to a successful Get delivery.
+func AfterGet(ictx hook.HookContext, msg amqp.Delivery, ok bool, err error) {
+	data, got := ictx.GetData().(*consumeData)
+	if !got || data == nil || err != nil || !ok {
+		return
+	}
+	ictx.SetReturnVal(0, startDeliverySpan(data.ch, data.queue, data.autoAck, msg, nil))
+}
```
darccio, line 267: AfterGet calls startDeliverySpan with local set to nil. A Consume delivery registers its span in two maps: the call-local pending map and the channel-global channelAcks map. A Get delivery registers its span only in the channel-global map. The call-local map gives Consume a close-time safety net through pending.endAll, which ends a span the caller never acked when the delivery channel closes. Get has no equivalent event, since a single Get call has no delivery channel to close. If the caller never acks a Get delivery, and the Channel is never closed, the span stays open until the process exits. State this leak explicitly in the package documentation, next to the existing note about the channelAcks map, so a caller of Get sees the same warning that Consume already gets.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+// The Enable() check is omitted: if BeforeConsume was disabled, GetData is
+// empty. Re-checking here could skip the wrap if the flag flipped between
+// Before and After.
+func AfterConsume(ictx hook.HookContext, deliveries <-chan amqp.Delivery, err error) {
+	data, ok := ictx.GetData().(*consumeData)
+	if !ok || data == nil || err != nil || deliveries == nil {
+		return
+	}
+	ictx.SetReturnVal(0, wrapDeliveries(data.ch, data.queue, data.autoAck, deliveries))
+}
```
darccio, line 251: AfterConsume and AfterGet skip the Enable check and rely on GetData returning nil when the before-hook did not run. This matches the same pattern and the same stated reason in the kafka-go consumer hook. Confirm that BeforeConsume has no early-return path in amqp091-go itself that would leave data set but the wrap not needed, or leave data unset for a reason other than the disabled flag. If such a path exists, state it here, next to the existing comment, so a future reader does not assume the only cause of a nil consumeData is disablement.

instrumentation/github.com/rabbitmq/amqp091-go/internal/propagation/carrier.go
```diff
@@ -0,0 +1,58 @@
...
+func (c TableCarrier) Keys() []string {
+	if c.table == nil || *c.table == nil {
+		return nil
+	}
+	keys := make([]string, 0, len(*c.table))
+	for k := range *c.table {
+		keys = append(keys, k)
+	}
+	return keys
+}
```
darccio, line 60: Keys allocates a new slice and copies every key in the table on each call. The default TraceContext propagator never calls Keys, since it looks up traceparent and tracestate by name. The propagator that Inject and Extract use here comes from otel.GetTextMapPropagator, a process-global value that application code outside this instrumentation can set. If application code configures a composite propagator that calls Keys, such as one that also propagates Baggage or a vendor format, each call to Inject or Extract on a large header table pays a full scan and a new allocation. State this assumption in a comment on TableCarrier: the carrier is built for TraceContext, and Keys costs one allocation and one full scan per call.

instrumentation/github.com/rabbitmq/amqp091-go/semconv/client.go
```diff
@@ -0,0 +1,86 @@
...
+	RoutingKey      string
+	Queue           string
+	Operation       Operation
+	MessageBodySize int
+}
+
+// DefaultExchange is the span destination when the publish exchange is empty.
+// The default exchange is a real AMQP name (the empty string). Using this
+// token keeps the span name low-cardinality and readable.
+const DefaultExchange = "(default)"
```
darccio, line 38: DefaultExchange holds the token (default), used as the span destination when the publish exchange is the empty string. The real default exchange in the AMQP protocol has an empty name, not the name amq.default, which names a different, separate exchange. An exchange named literally (default), parentheses included, would produce a send span that a reader cannot tell apart from a true default-exchange publish. This trade-off already appears in the hook.go package doc and in the README, but not here. State the same trade-off in this doc comment, so a reader of this file alone sees the reasoning without a cross-reference to another file.

instrumentation/github.com/rabbitmq/amqp091-go/semconv/client.go
```diff
@@ -0,0 +1,86 @@
...
+	case OperationReceive:
+		attrs = append(attrs, semconv.MessagingOperationTypeReceive)
+	case OperationProcess:
+		attrs = append(attrs, semconv.MessagingOperationTypeProcess)
+	}
+	if key := strings.TrimSpace(req.RoutingKey); key != "" {
+		attrs = append(attrs, semconv.MessagingRabbitMQDestinationRoutingKey(key))
+	}
+	if q := strings.TrimSpace(req.Queue); q != "" && req.Operation != OperationSend {
+		attrs = append(attrs, semconv.MessagingDestinationSubscriptionName(q))
```
darccio, line 83: This condition adds the subscription-name attribute only when Operation is not OperationSend and Queue is non-empty. No current caller builds a Request with Operation set to OperationSend and Queue set to a non-empty value: BeforePublishWithDeferredConfirm never sets Queue, and startDeliverySpan sets Queue only for OperationReceive and OperationProcess. The req.Operation != OperationSend check is therefore unreachable with the two current call sites. State in a comment that the check guards against a future caller that builds a send Request with a Queue value, so a future reader does not assume a live test exercises this branch.

schemas/otelc/groups/rabbitmq.yaml
```diff
@@ -0,0 +1,40 @@
+groups:
+  # ---------------------------------------------------------------------------
+  # rabbitmq/amqp091-go instrumentation emission contract.
+  #
+  # Source of truth for this file:
+  #   instrumentation/github.com/rabbitmq/amqp091-go/hook.go
+  #   instrumentation/github.com/rabbitmq/amqp091-go/semconv/client.go
+  #
```
darccio, line 8: This header names hook.go and semconv/client.go as the source of truth for this schema. The schema attribute lists are hand-written prose that describes the current output of TraceAttrs, not a value generated from that function. A future change to TraceAttrs that adds or removes an attribute can drift from this schema, unless a check compares the two on every build. State whether a schema validator, a linter, or a test enforces this match today, in this repository or in the kafka.yaml schema that this file follows. If no such check exists, add one, or state in this comment that the match relies on reviewer diligence alone.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+// Consume and ConsumeWithContext return a delivery channel. The after hook
+// replaces that channel. Each Delivery starts a CONSUMER span. Manual-ack
+// spans use operation process and end on Delivery.Ack / Nack / Reject or
+// Channel.Ack / Nack / Reject (including multiple=true). Auto-ack spans use
+// operation receive and end when the Delivery is read: the server already
+// settled the message, so there is no later finish signal. Closing the
+// consume channel ends leftover process spans so they do not leak.
+//
+// (*Channel).connection is unexported, so these spans do not set
+// server.address. The instrumentation is trace-only.
```
darccio: This package doc comment states the context-drop caveat, the (default) sentinel choice, and the auto-ack timing rule. The module README states the same three points, in different words, for a different reader. A future edit to one of the two copies can drift from the other, since neither document points to the other. Pick one document as the canonical source for each point, and make the second document reference the first, rather than restate the point in its own words.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
```diff
@@ -0,0 +1,500 @@
...
+func BeforePublishWithDeferredConfirmWithContext(
+	_ hook.HookContext,
+	ch *amqp.Channel,
+	ctx context.Context,
+	_ /*exchange*/, _ /*key*/ string,
+	_ /*mandatory*/, _ /*immediate*/ bool,
+	_ amqp.Publishing,
+) {
+	stashPublishParent(ch, ctx)
+}
```
darccio, line 151: BeforePublishWithContext has a dedicated test, TestPublishWithContext_UsesAPIParent, that calls the function and asserts the stashed context wins as the send span's parent. BeforePublishWithDeferredConfirmWithContext calls the same stashPublishParent function with the same argument shape, but no test calls this function directly. Add a test for BeforePublishWithDeferredConfirmWithContext that mirrors TestPublishWithContext_UsesAPIParent, so a future change to stashPublishParent, or to this function's argument order, has a test on each of its two callers, not one.

instrumentation/github.com/rabbitmq/amqp091-go/README.md
```diff
@@ -0,0 +1,49 @@
...
+export OTEL_GO_ENABLED_INSTRUMENTATIONS=amqp
+export OTEL_GO_DISABLED_INSTRUMENTATIONS=amqp
+```
+
+Instrumentation key: `AMQP` (case-insensitive).
+
+## Supported versions
+
+- Module: `github.com/rabbitmq/amqp091-go`
+- Minimum bound: **v1.10.0** (`ConsumeWithContext`).
```
darccio: This line states v1.10.0 as the minimum bound because that version adds ConsumeWithContext. The rabbitmq/amqp091-go repository shows ConsumeWithContext present at tag v1.9.0 and absent at tag v1.8.1, so v1.9.0 introduces this method, not v1.10.0. Either lower the floor in otelc.yaml, in tool/data/instrumentation-manifest.json, and in this line to v1.9.0, or state the actual reason v1.9.0 does not work for this instrumentation, such as a bug fixed in a later release. As written, the stated reason for v1.10.0 does not match the method's real introduction version.

darccio (changes requested):
The integration suite in test/integration/amqp_client_test.go covers two scenarios: a manual-ack Consume happy path, and the disabled kill switch. It does not cover auto-ack, Get, or Nack and Reject, though unit tests cover all of those paths through hooktest.NewMockHookContext. The unit tests confirm the hook functions behave correctly in isolation. The integration suite confirms the hooks weave correctly into a real compiled build. Add at least one integration scenario for auto-ack or for Nack, so the suite that proves the real build also covers a code path other than the manual-ack happy path, or state why unit coverage of those paths is enough on its own.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: yes done

instrumentation/github.com/rabbitmq/amqp091-go/README.md
atharvamhaske: Floor is now v1.9.0 in otelc.yaml, go.mod, README, manifest.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: added a comment + README mention as well as is library consume is already unbuffered.

atharvamhaske:
@darccio i have addressed all comments, ptal once you get time !!

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: Moved this to the README. One stored context per Channel. `amqp091-go `already says not to publish
 concurrently on one Channel.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: added a comment on cloneTable. Inject only adds top-level string keys, so a shallow copy is enough

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: i have added `TestPublishWithDeferredConfirmWithContext_UsesAPIParent`

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: README is the source. The `hook.go `package comment points there.

schemas/otelc/groups/rabbitmq.yaml
atharvamhaske: `make lint-schema` checks Weaver registry shape. It does not compare this list to TraceAttrs. Same as
`kafka.yaml.`

instrumentation/github.com/rabbitmq/amqp091-go/semconv/client.go
atharvamhaske: The `DefaultExchange` doc comment now states that a real exchange named (default) looks the same as the empty default exchange.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: BeforeConsume only returns early when disabled. After also skips the wrap if Consume returns an error or a nil channel.

instrumentation/github.com/rabbitmq/amqp091-go/hook.go
atharvamhaske: receive is the marker. A comment at span.End and the README say the span ends when the Delivery is read

darccio (approved):

PR #1350, opened 2026-09-10
fix(tool/internal/setup): handle test flags and delimiters for go test

### LINKED ISSUE #1346: fix(tool/internal/setup): honor go test -test.* flags and -- delimiter
## Problem

`otelc go test` does not follow all of `cmd/go`'s argument-boundary rules during setup.

`splitBuildTargets` recognizes `-run TestName`, but not the equivalent
`-test.run TestName` form. It therefore interprets `TestName` as a package
target.

It also does not recognize the `--` delimiter. For `go test`, Go passes `--`
and every following argument to the test binary. `otelc` instead continues to
scan the tail as package or file targets.

## Impact

These valid invocations can fail before `go test` starts:

```sh
otelc go test -test.run TestName
otelc go test -- custom-test-binary-argument
```

When an explicit package is present, a test-binary argument ending in `.go`
can also be treated as a file build target. This can cause `otelc` to append
`otelc.runtime.go` after `--`, changing the test binary's argument list.

## Expected behavior

`otelc go test` should preserve Go's target parsing rules:

- Treat supported `-test.<name>` flags like their unprefixed forms.
- In test mode, stop target parsing at `--`, `-args`, and `--args`.
- Apply the same logic to setup target discovery and the phase-two
  file-target check.

`go build -- <packages>` must keep its existing behavior, so delimiter
handling should be specific to `go test`.

## Suggested tests

- `-test.run TestName ./pkg` identifies only `./pkg` as a target.
- `./pkg -- -test.run TestName test-binary-input.go` identifies only
  `./pkg`.
- No-explicit-package calls with `-test.run TestName` and `-- customArg`
  select the current package and do not fail setup.

darkraider01:
## Description

* Unify command-line argument parsing into a shared classifier (`classifyArgs`) across target discovery, build-flag extraction, build-plan handling, and vendoring rewrites.
* Align `go test` argument handling with `cmd/go`:

  * Keep package discovery open across recognized Go/test flags, so package targets can appear on both sides of flags like `-run`.
  * Unknown/custom flags close package discovery, while unjoined unknown flags still associate their following value correctly.
  * Recognize supported `-test.*` flags and their values without mistaking them for package targets or delimiters.
  * Handle `--`, `-args`, and `--args` as test-binary boundaries without treating delimiter-like flag values as actual delimiters.
* Keep test-binary arguments isolated so flags such as `-tags` or `-mod=vendor` after the test boundary do not leak into build-context extraction or vendoring rewrites.
* Preserve `go build` / `go install` `--` behavior, where `--` stops flag parsing and the remaining arguments are treated as targets.
* Ensure generated `otelc.runtime.go` is inserted before the test-argument delimiter for file-based `go test` invocations.
* Pass the active subcommand through setup/pinning paths so argument handling stays consistent across build, install, and test flows.

## Motivation

When running `go test`, package arguments, Go/test flags, and arguments intended for the test binary can be interspersed in ways that the previous parsing logic did not handle consistently.

This could cause flag values to be treated as package targets, `--` to be missed as a test delimiter, package targets after recognized flags to be skipped, or test-binary arguments such as `-tags` / `-mod=vendor` to leak into setup and build handling.

The parsing logic was also spread across target discovery, build-flag extraction, build-plan generation, and vendoring rewrites. Using a shared classifier keeps these paths consistent and avoids fixing the same argument-boundary cases independently in multiple places.

Fixes #1346

### DIFF
--- a/tool/internal/setup/args.go
+++ b/tool/internal/setup/args.go
@@ -1,0 +1,450 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package setup
+
+import (
+	"strings"
+)
+
+// ArgKind represents the semantic category of a command-line argument.
+type ArgKind int
+
+const (
+	// ArgTarget represents a package pattern or file target (e.g., "./...", "main.go").
+	ArgTarget ArgKind = iota
+
+	// ArgBuildFlag represents a recognized Go build/command flag (e.g., "-tags=foo", "-race").
+	ArgBuildFlag
+
+	// ArgBuildFlagValue represents the separated value argument for an ArgBuildFlag.
+	ArgBuildFlagValue
+
+	// ArgTestFlag represents a recognized go test flag (e.g., "-run", "-test.run", "-v").
+	ArgTestFlag
+
+	// ArgTestFlagValue represents the separated value argument for an ArgTestFlag.
+	ArgTestFlagValue
+
+	// ArgTestDelimiter represents "--", "-args", or "--args" in go test.
+	ArgTestDelimiter
+
+	// ArgTestBinary represents test-binary arguments (tokens following a test delimiter,
+	// or unknown/custom test flags and their values/positionals).
+	ArgTestBinary
+
+	// ArgFlagTerminator represents "--" in go build / go install.
+	ArgFlagTerminator
+)
+
+// ClassifiedArg contains the classification metadata for a single command-line argument token.
+type ClassifiedArg struct {
+	Index    int
+	Raw      string
+	Kind     ArgKind
+	FlagName string // normalized single-dash name, e.g. "-tags", "-run", "-test.run"
+	HasValue bool   // whether joined =value was present
+	Value    string // value from joined =value or separate token
+}
+
+// isBuildFlagWithValue reports whether name is a Go build flag that accepts a value.
+func isBuildFlagWithValue(name string) bool {
+	switch name {
+	case "-C", "-o", "-p", "-covermode", "-coverpkg",
+		"-asmflags", "-buildmode", "-buildvcs", "-compiler",
+		"-gccgoflags", "-gcflags", "-installsuffix", "-ldflags",
+		"-mod", "-modfile", "-overlay", "-pgo", "-pkgdir",
+		"-tags", "-toolexec":
+		return true
+	default:
+		return false
+	}
+}
+
+// isBuildBoolFlag reports whether name is a Go build boolean flag.
+func isBuildBoolFlag(name string) bool {
+	switch name {
+	case "-a", "-n", "-v", "-x", "-race", "-msan",
+		"-asan", "-cover", "-trimpath", "-work", "-linkshared", "-json":
+		return true
+	default:
+		return false
+	}
+}
+
+// isTestFlagWithValue reports whether name is a go test flag that accepts a value.
+func isTestFlagWithValue(name string) bool {
+	switch name {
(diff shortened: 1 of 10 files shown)


tool/internal/setup/args.go
```diff
@@ -0,0 +1,450 @@
...
+				Kind:  ArgTestDelimiter,
+			})
+			testState.testArgsFinalized = true
+			i++
+			continue
+		}
+
+		if strings.HasPrefix(arg, "-") {
+			if testState.inPkgList {
+				testState.inPkgList = false
```
ADITYA-CODE-SOURCE: A recognized test flag does not end `cmd/go`’s package list. For example, `go test fmt -run TestX math` tests both `fmt` and `math`.

However, clearing `inPkgList` here causes `math` to be classified as `ArgTestBinary`, so `otelc` discovers and instruments only `fmt`.

Could we keep package discovery open across recognized Go/test flags and add a regression test with package targets on both sides of `-run`?

om7057 (commented):
I pulled this into a worktree and ran the full suite (`go vet ./...`, `go build ./...`, `go test ./...` across the whole `tool` module), everything passes clean.

I also wanted to independently confirm the bug described in #1346 was real before trusting the PR description, so I wrote a standalone test calling `splitBuildTargets` directly on `main` with the two invocations from the issue:

```go
splitBuildTargets([]string{"-test.run", "TestName"})       // -> pkgs=[TestName], wrong
splitBuildTargets([]string{"--", "custom-test-binary-argument"}) // -> pkgs=[custom-test-binary-argument], wrong
```

Both misparse on `main` exactly as described (`testFlagsWithValues` has no `-test.run` entry, and there's no `--` delimiter handling at all, only `-args`). Running the same two cases against this branch's `splitBuildTargets`, both come back with no package targets, so the fix does what it says.

One thing worth flagging for other reviewers: the scope here is bigger than the issue title suggests. This isn't a small patch to `splitBuildTargets`, it's a full refactor that introduces a new `args.go` classifier (`classifyArgs`/`classifyBuildArgs`/`classifyTestArgs`) and threads a `subcommand` parameter through `find.go`, `pin.go`, `setup.go`, and `vendor.go` to use it. I read through the state machine in `classifyTestArgs` (the `testArgState` tracking `inPkgList`/`packageListEstablished`/`testArgsFinalized`, and the "optimistic" handling of a positional right after an unknown flag) and it matches `cmd/go`'s actual behavior as far as I can tell, but it's a meaningfully larger surface to review carefully than "handle two missed cases," especially the unknown-flag-value heuristic in `classifyTestFlagToken`/`classifyPositionalTestArg`.

Test coverage looks solid, the new `TestRewriteModVendor` cases specifically exercise the delimiter/flag-value distinctions that used to be missed (e.g. `-test.run` and `--`/`-args`/`--args` preserving a trailing `-mod=vendor`), and I don't see any leftover dead code from the old `flagsWithPathValues`/`testFlagsWithValues` maps or the old scanning logic in `find.go`/`setup.go`/`vendor.go`.

darkraider01:
about the scope i mentioned it in the issue, https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/1346#issuecomment-5638112248,but maybe yourw right i should update the pr description aswelel

tool/internal/setup/args.go
darkraider01: ty ! I'll update the state handling to preserve package discovery across recognized flags while keeping unknown flags and explicit delimiters as boundaries.

PR #1354, opened 2026-09-12
fix(instrumentation/net/http): guard nil request and URL in client AfterRoundTrip

### LINKED ISSUE #1353: bug(instrumentation/net/http): nil pointer dereference in client AfterRoundTrip debug logging
### Describe the bug

In `instrumentation/net/http/client/client_hook.go`, `AfterRoundTrip` logs request details when debug logging is enabled:

```go
if debugEnabled() {
    logger.Debug("AfterRoundTrip called",
        "method", res.Request.Method,
        "url", res.Request.URL.String(),
        "status_code", res.StatusCode,
        "duration_ms", time.Since(data.start).Milliseconds())
}
```

In Go standard library `net/http`, `http.Response.Request` is not guaranteed to be non-nil (for example, in mock transports like `httpmock`, custom `RoundTripper` implementations, or synthetic responses). When `res.Request == nil` or `res.Request.URL == nil` and debug logging is enabled (`OTEL_LOG_LEVEL=debug`), this causes an immediate runtime panic (`SIGSEGV`):

```text
go.opentelemetry.io/otelc/instrumentation/net/http/client.AfterRoundTrip(...)
    instrumentation/net/http/client/client_hook.go:164
```

### Environment

- OS: macOS / Linux / Windows
- Go Version: 1.25
- Version: main (commit 434baec)

### To Reproduce

1. Enable debug logging: `export OTEL_LOG_LEVEL=debug`
2. Perform an HTTP round trip that returns a response with `Request: nil` (or `Request.URL: nil`), such as with a mock transport or synthetic response.
3. Observe panic: `runtime error: invalid memory address or nil pointer dereference` at `client_hook.go:164`.

### Expected behavior

`AfterRoundTrip` should defensively guard `res.Request` and `res.Request.URL` before attempting to read `Method` and call `URL.String()`, logging empty values when absent rather than crashing the process.

Labreo:
## Description

Guards against nil pointer dereferences in `instrumentation/net/http/client`'s `AfterRoundTrip` hook when logging debug messages.

Specifically:
- Check if `res.Request != nil` before reading `res.Request.Method` and `res.Request.URL`.
- Check if `res.Request.URL != nil` before calling `res.Request.URL.String()`.
- Add test coverage in `client_hook_test.go` verifying that `res.Request == nil` and `res.Request.URL == nil` do not panic with debug logging enabled.
- Re-bundle `tool/data/otelc-bundle.tgz` via `make package`.

## Motivation

In Go `net/http`, `http.Response.Request` is not guaranteed to be populated on synthetic responses or custom `http.RoundTripper` implementations (such as test doubles or `httpmock`). When debug logging is enabled (`OTEL_LOG_LEVEL=debug`), calling `AfterRoundTrip` with `res.Request == nil` or `res.Request.URL == nil` panics with a segmentation violation.

Fixes #1353


instrumentation/net/http/client/client_hook_test.go
```diff
@@ -373,6 +426,59 @@ func TestAfterRoundTrip(t *testing.T) {
 	}
 }
 
+func TestAfterRoundTrip_NilRequestAndURL_DebugEnabled(t *testing.T) {
+	oldLogger := logger
+	logger = slog.New(slog.NewJSONHandler(io.Discard, &slog.HandlerOptions{Level: slog.LevelDebug}))
```
darkraider01, line 431: Could we use a bytes.Buffer here instead of io.Discard and assert that missing request data is logged with empty method / url values? NotPanics covers the regression, but checking the emitted fields would also lock in the intended fallback behavior.

instrumentation/net/http/client/client_hook_test.go
```diff
@@ -351,6 +353,57 @@ func TestAfterRoundTrip(t *testing.T) {
 				assert.Equal(t, 1, len(spans))
 			},
 		},
+		{
+			name: "response with nil Request",
```
vedant21-oss: Nice fix! One thing I noticed while testing locally: the two new table cases ("response with nil Request" / "response with nil Request URL") still pass if I revert the change in `client_hook.go`. The panic only happens inside `if debugEnabled()`, and the table tests run with the default logger, so debug is off and that branch never runs.

`TestAfterRoundTrip_NilRequestAndURL_DebugEnabled` is the test that actually catches the regression (it panics without the fix). Maybe drop the two table cases, or turn on debug logging for them, so they don't look like they cover the bug when they don't?

vedant21-oss (commented):
Tested locally, fix looks good. One note on the tests.

instrumentation/net/http/client/client_hook_test.go
Labreo: Updated to capture log output into a bytes.Buffer and asserted that "method":"", "url":"", and "status_code":200 are emitted when Request is nil, and "method":"GET" with "url":"" when Request.URL is nil.

instrumentation/net/http/client/client_hook_test.go
Labreo: Good catch! I dropped the two redundant table cases so TestAfterRoundTrip_NilRequestAndURL_DebugEnabled serves as the explicit regression test with debug logging enabled.

Labreo:
The required changes have been resolved and the merge conflicts have been resolved.

Issue #1347, opened 2026-09-10
bug(tool/internal/setup): where predicates are ignored for add_file rules


rarguelloF:
### Describe the bug

A `where` block on an `add_file` rule parses without error and is then ignored. The rule applies to every compile of the target package regardless of what the predicates say.

The `is_test` predicate itself is fine. It is built and evaluated for every other rule type. File rules never reach the filter.

In `tool/internal/setup/match.go`, file rules are collected before the filter loop runs:

```go
// Separate file rules from rules that need precise matching
preciseRules := make([]rule.InstRule, 0, len(filteredRules))
for _, r := range filteredRules {
	// If the rule is a file rule, it is always applicable
	if fr, ok := r.(*rule.InstFileRule); ok {
		set.AddFileRule(fr)
		sp.Info("Match file rule", "rule", fr, "dep", dep)
		continue
	}
	preciseRules = append(preciseRules, r)
}
```

The `where` predicates gate `matchOneRule`, which skips file rules:

```go
case *rule.InstFileRule:
	// Skip as it's already processed
```

So the filter built from `where` is never consulted for an `add_file` rule.

### Environment

 - OS: macOS 26.5.2 (darwin/arm64)
 - Go Version: 1.27.0
 - Version: `47b984aaa1fa` (`v1.1.1-0.20260910072957-47b984aaa1fa`, main at the time of filing)

### To Reproduce

Two modules, the rules module and an application that opts into it.

`myrules/go.mod`:

```
module example.com/myrules

go 1.26.0
```

`myrules/rules.go`:

```go
package myrules
```

`myrules/added.go`:

```go
//go:build ignore

package myrules

func init() {
	println("added init ran")
}
```

`myrules/otelc.yaml`:

```yaml
---
add_to_main:
  target: main
  where:
    file:
      is_test: true
  do:
    - add_file:
        file: added.go
        path: example.com/myrules
```

`app/go.mod`:

```
module example.com/app

go 1.26.0

require example.com/myrules v0.0.0

replace example.com/myrules => ../myrules
```

`app/main.go`:

```go
package main

func main() {
	println("main ran")
}
```

`app/main_test.go`:

```go
package main

import "testing"

func TestNothing(t *testing.T) {}
```

`app/otel.instrumentation.go`:

```go
//go:build tools

package tools

import (
	_ "example.com/myrules"
)
```

Then, from `app/`:

```
$ otelc go build ./... && ./app
added init ran
[...]

PR #1356, opened 2026-09-12
fix(tool/internal/instrument): add_file emits an unparseable package clause in test builds

### LINKED ISSUE #1348: bug(tool/internal/instrument): add_file writes an empty package clause for the generated test-main
### Describe the bug

An `add_file` rule with `target: main` also applies to the test-main compile unit that `go test` generates. otelc writes the added file there with an empty `package` clause, so the build fails with a syntax error.

`applyFileRule` rewrites the added file's package clause from the rule set's package name (`tool/internal/instrument/apply_file.go`):

```go
root.Name.Name = pkgName
```

`pkgName` comes from `InstRuleSet.PackageName`, which starts empty (`tool/internal/rule/base.go`):

```go
func NewInstRuleSet(importPath string) *InstRuleSet {
	return &InstRuleSet{
		PackageName: "",
```

and is only populated by `SetPackageName`, which asserts a non-empty name:

```go
func (irs *InstRuleSet) SetPackageName(name string) {
	util.Assert(name != "", "package name is empty")
	irs.PackageName = name
}
```

Both call sites in `tool/internal/setup/match.go` depend on the compile having listed sources:

```go
if !set.IsEmpty() && len(dep.Sources) > 0 {
	name, err := ast.ParsePackageName(dep.Sources[0])
	...
	set.SetPackageName(name)
}
```

```go
for _, source := range dep.Sources {
	...
	set.SetPackageName(tree.Name.Name)
```

The generated test-main compile unit has no listed sources, so neither runs, `PackageName` stays `""`, and the added file is written with a bare `package` keyword.

### Environment

 - OS: macOS 26.5.2 (darwin/arm64)
 - Go Version: 1.27.0
 - Version: `47b984aaa1fa` (`v1.1.1-0.20260910072957-47b984aaa1fa`, main at the time of filing)

### To Reproduce

Two modules, the rules module and an application that opts into it.

`myrules/go.mod`:

```
module example.com/myrules

go 1.26.0
```

`myrules/rules.go`:

```go
package myrules
```

`myrules/added.go`:

```go
//go:build ignore

package myrules

func init() {
	println("added init ran")
}
```

`myrules/otelc.yaml`:

```yaml
---
add_to_main:
  target: main
  do:
    - add_file:
        file: added.go
        path: example.com/myrules
```

`app/go.mod`:

```
module example.com/app

go 1.26.0

require example.com/myrules v0.0.0

replace example.com/myrules => ../myrules
```

`app/main.go`:

```go
package main

func main() {
	println("main ran")
}
```

`app/main_test.go`:

```go
package main

import "testing"

func TestNothing(t *testing.T) {}
```

`app/otel.instrumentation.go`:

```go
//go:build tools

package tools

import (
	_ "example.com/myrules"
)
```

Then, from `app/`:

```
$ otelc go build ./... && ./app
added init ran
main ran

$ otelc go test ./...
# example.com/app.test
$WORK/b001/otelc.added.go:5:1: syntax error: unexpected keyword func, expected name
FAIL	example.com/app [build failed]
```

The build succeeds and the test build fails. Keeping the work directory shows the added file written into the test-main compile unit with a bare `package` keyword and no name:

```go
// Copied into the target package by the add_file rule in otelc.yaml. otelc
// rewrites the package clause and drops the build constraint on the way in.
package

func init() {
	println("added init ran")
}
```

### Expected behavior

Either the package name is resolved for generated compile units, or a file rule that cannot resolve one is skipped instead of emitting a file that does not parse.

### Additional context

This blocks using `add_file` against `main` to inject a `func init()`, which is the only way to run code before other packages' inits.

Filed alongside the companion bug that `where` predicates are ignored for `add_file` rules. They surface together but are distinct: the other one is that a rule cannot be scoped away from test builds in the first place, this one is that once the rule does apply there, the emitted code does not compile.

krsatyamthakur-droid:
## Description

Turns out an `add_file` rule targeting `main` also lands on the test main that `go test`
generates. That compile unit shows up with no sources in the build-plan dry run, so setup
never calls `SetPackageName` and the rule set reaches us with `PackageName` empty.
`applyFileRule` wrote that straight into the package clause — hence the bare `package` and the
syntax error against generated code.

The compile command does list the sources though, so we can read the name off there instead. I
put it before `instrument()` runs so file rules and the globals file both pick it up, and
because `compileArgs` still only has the original sources at that point.

Also stuck a guard in `applyFileRule` — if the name still can't be worked out it errors with
the rule name rather than writing a file that won't compile.

## Tests

`TestResolvePackageName` covers the normal case, unreadable paths, no sources at all, and a
parse failure. `TestApplyFileRule_EmptyPackageName` checks the guard errors and leaves nothing
behind.

After review I also added `TestInterceptCompileFileRuleWithoutPackageName`, which goes through
`interceptCompile` itself with an empty `PackageName` — the same state setup leaves for the
test main. It fails on the original code, so it covers the actual gap rather than just the
pieces.

Ran the repro from the issue too — `otelc go test ./...` goes from the syntax error to
`ok example.com/app`, and `go build` is unchanged.

## Motivation

This currently makes `add_file` against `main` unusable in any module with tests, which is
awkward since it's the only way to run a `func init()` ahead of other packages' inits.

Worth noting the injected init now runs in the test binary too. Keeping it out of test builds
is #1347.

Fixes #1348


tool/internal/instrument/toolexec_test.go
```diff
@@ -681,6 +681,47 @@ func TestInterceptLinkAllImportsPresent(t *testing.T) {
...
+	t.Run("propagates a parse failure", func(t *testing.T) {
+		broken := filepath.Join(dir, "broken.go")
+		require.NoError(t, os.WriteFile(broken, []byte("not a go file"), 0o644))
+		ip := &instrumentPhase{compileArgs: []string{"compile", broken}}
+		_, err := ip.resolvePackageName()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "parsing package clause")
+	})
+}
+
```
darkraider01, line 726: Maybe we can add a regression test that exercises this through `interceptCompile` (or an integration `otelc go test` case) with an empty `PackageName` rule set. The original bug is caused by the gap between setup's dry-run sources and the actual test-main compile invocation, so testing `resolvePackageName` in isolation does not fully cover the failing path.

krsatyamthakur-droid: Good point, the isolated tests missed the actual gap.

Added `TestInterceptCompileFileRuleWithoutPackageName`, which goes through **`interceptCompile`** with an empty `PackageName` and a compile command listing `_testmain.go` — same state setup leaves for the test main. It **fails on the original code**, and checks the added file ends up as `package main`.

Skipped an integration test since this hits the same path without needing a test app, but happy to add one if you'd like.

ADITYA-CODE-SOURCE (approved):
Reviewed package-name recovery from the compile command, the empty-name guard, and the end-to-end interceptCompile regression test. The name is resolved before generated files are added, and the original test-main failure path is now covered. I did not find an additional issue.

PR #1358, opened 2026-09-12
fix(tool/internal/instrument): only strip ignore build constraints in add_file rules

### LINKED ISSUE #1357: bug(tool/internal/instrument): stripBuildIgnoreTag strips all build constraints and misses legacy // +build ignore
### Describe the bug

`stripBuildIgnoreTag` in `tool/internal/instrument/apply_file.go` was introduced (via #1095, following #1069) to remove `//go:build ignore` constraint lines from a file rule source before injecting it into the target package at compile time.

However, `stripBuildIgnoreTag` currently checks lines using `constraint.IsGoBuild(line)`:

```go
func stripBuildIgnoreTag(content string) string {
	lines := strings.Split(content, "\n")
	for i, line := range lines {
		if constraint.IsGoBuild(line) {
			lines[i] = ""
		}
	}
	return strings.Join(lines, "\n")
}
```

Per the standard library documentation for `go/build/constraint.IsGoBuild`:
> *IsGoBuild reports whether the line of text is a “//go:build” constraint. It only checks the prefix of the text, not that the expression itself parses.*

Because it only checks the prefix rather than parsing the constraint expression:
1. **It strips all `//go:build` constraints indiscriminately.** If a file rule carries platform, architecture, or Go version constraints (e.g. `//go:build linux`, `//go:build !windows`, or `//go:build cgo`), `stripBuildIgnoreTag` blanks them out entirely. The injected file is then compiled on every platform, violating the intended build constraints.
2. **It fails to strip legacy `// +build ignore` directives.** Legacy build constraints starting with `// +build` return `false` from `constraint.IsGoBuild(line)`, leaving the `ignore` directive in the injected file, which causes the Go compiler to skip compiling it.

### Environment

- OS: Any
- Go Version: 1.25+
- Repository Version: `main`

### To Reproduce

```go
input := `//go:build ignore
//go:build linux

package main

func main() {}
`
got := stripBuildIgnoreTag(input)
```

**Observed behavior:**
Both lines are blanked out: `//go:build linux` is stripped from the output.

Similarly, with legacy syntax:
```go
input := `// +build ignore

package main

func main() {}
`
got := stripBuildIgnoreTag(input)
```
**Observed behavior:**
`// +build ignore` is not stripped and remains in the output.

### Expected behavior

Only genuine build constraint lines specifying `ignore` should be stripped:
- Modern `//go:build ignore` should be stripped.
- Legacy `// +build ignore` should be stripped.
- Other build constraints (such as `//go:build linux`, `//go:build !windows`, or `//go:build darwin || linux`) must be preserved.
- Directives inside string literals or comment prose must continue to be preserved (as established in #1069).

### Proposed fix

Use `constraint.Parse` to inspect whether the build constraint directive expression specifies `"ignore"`:

```go
func isBuildIgnoreLine(line string) bool {
	trimmed := strings.TrimSpace(line)
	if !constraint.IsGoBuild(trimmed) && !constraint.IsPlusBuild(trimmed) {
		return false
	}
	expr, err := constraint.Parse(trimmed)
	if err != nil {
		return false
	}
	return expr.String() == "ignore"
}
```

I have a patch with comprehensive regression unit tests ready and will submit a PR.

Aryanbhargava18:
## Description

Updates `stripBuildIgnoreTag` in `tool/internal/instrument/apply_file.go` to use `constraint.Parse` and recursive AST expression tree walking to identify and remove only build constraint directives that contain an `ignore` tag.

Key changes:
- Evaluates parsed expressions via `constraint.Parse` and walks the `constraint.Expr` tree (`containsIgnoreTag`) across `TagExpr`, `AndExpr`, `OrExpr`, and `NotExpr` to detect `ignore` regardless of compound expressions (e.g. `//go:build ignore && linux`) or nesting depth.
- Preserves all valid standalone build constraints (e.g. `//go:build linux`, `//go:build !windows`, `//go:build darwin || linux`, `//go:build cgo`) when `ignore` is not part of the directive.
- Correctly strips legacy `// +build ignore` directives, which were previously missed by `constraint.IsGoBuild`.
- Preserves directive text inside string literals and comment prose (retaining the regression fix from #1069).
- Adds unit tests in `apply_file_test.go` covering combined tag directives, legacy directive stripping, standalone constraint preservation, and end-to-end AST processing in `TestApplyFileRule_CombinedIgnoreAndPlatformTag`.

## Motivation

As reported in #1357, `stripBuildIgnoreTag` previously relied solely on `constraint.IsGoBuild(line)`. Per Go standard library specifications, `constraint.IsGoBuild` checks only for the `//go:build` line prefix without evaluating the expression itself. Consequently:
1. All `//go:build` directives were stripped indiscriminately, even when no `ignore` tag was present.
2. Legacy `// +build ignore` lines were skipped and left in the generated file, causing the Go compiler to ignore the injected file.

Fixes #1357


tool/internal/instrument/apply_file.go
```diff
@@ -17,21 +17,35 @@ import (
...
+func isBuildIgnoreLine(line string) bool {
+	trimmed := strings.TrimSpace(line)
+	if !constraint.IsGoBuild(trimmed) && !constraint.IsPlusBuild(trimmed) {
+		return false
+	}
+	expr, err := constraint.Parse(trimmed)
+	if err != nil {
+		return false
+	}
+	return expr.String() == "ignore"
```
darkraider01: This misses valid combined constraints like `//go:build ignore && linux`. In that case the parsed expression is `ignore && linux`, so the `ignore` term is left untouched. The tests should cover the valid combined form instead of separate `//go:build` lines.

tool/internal/instrument/apply_file_test.go
```diff
@@ -84,6 +84,66 @@ package hooks
 
 // Every file rule source must carry //go:build ignore at the top.
 const marker = "//go:build ignore"
+`,
+		},
+		{
+			name: "preserves other go:build directives such as linux or !windows",
+			input: `//go:build ignore
+//go:build linux
+
```
darkraider01, line 94: This isn't a valid Go build header ,  Go rejects multiple  //go:build  directives in the same file. Please use a valid combined expression like `//go:build ignore && linux`; that also exposes the current `expr.String() == "ignore"` issue.

tool/internal/instrument/apply_file_test.go
```diff
@@ -228,3 +288,44 @@ func SubHelper() {}
...
+	err := ip.applyFileRule(t.Context(), fileRule, "targetpkg")
+	require.NoError(t, err)
+
+	outPath := filepath.Join(workDir, "otelc.helper.go")
+	require.FileExists(t, outPath)
+
+	outData, err := os.ReadFile(outPath)
+	require.NoError(t, err)
+	assert.Contains(t, string(outData), "//go:build linux")
```
darkraider01, line 467: This only verifies that the build tag survives in the generated source, not that it is actually respected. Since the generated file is later added directly to the compiler args, can we test that this file is not included for a non-matching platform instead?

tool/internal/instrument/apply_file.go
Aryanbhargava18: Makes sense. Replaced the exact string match with `containsIgnoreTag(expr)`, which walks the `constraint.Expr` tree recursively (`TagExpr`, `AndExpr`, `OrExpr`, `NotExpr`). That catches `ignore` regardless of compound expressions or nesting depth.

tool/internal/instrument/apply_file_test.go
Aryanbhargava18: Updated the tests to use valid combined directives like `//go:build ignore && linux` and `//go:build ignore && (darwin || linux)`. Also added standalone directive test cases to confirm they remain untouched when `ignore` isn't present.

tool/internal/instrument/apply_file_test.go
Aryanbhargava18: Right, because `applyFileRule` feeds the file straight into `compileArgs`, `go tool compile` compiles it anyway since it doesn't evaluate build tags on explicit CLI arguments (unlike `cmd/go`).

As you suggested on #1357, I scoped this PR to getting the ignore-stripping right so compound/legacy directives don't break. Updated this test to verify the compound directive line is removed cleanly from the output, and noted the compile-time platform gating as a follow-up.

tool/internal/instrument/apply_file.go
```diff
@@ -17,21 +18,99 @@ import (
...
+		if e.Tag == "ignore" {
+			return nil
+		}
+		return e
+	case *constraint.NotExpr:
+		sub := removeIgnore(e.X)
+		if sub == nil {
+			return nil
+		}
+		return &constraint.NotExpr{X: sub}
```
ADITYA-CODE-SOURCE, line 82: Pruning `ignore` as `nil` does not preserve Boolean semantics across `&&`, `||`, and `!`.

For example, `ignore && windows` is always false, but it simplifies to `windows`. Conversely, `!ignore || windows` is always true, but it also simplifies to `windows`. This could cause an `add_file` rule to be included or skipped on the wrong platforms.

Could we simplify `ignore` to `false` and `!ignore` to `true`, and add tests covering these AND/OR cases?

PR #1360, opened 2026-09-14
fix(tool/internal/setup): add_file rules ignore where predicates

### LINKED ISSUE #1347: bug(tool/internal/setup): where predicates are ignored for add_file rules

krsatyamthakur-droid:
## Description

A `where` block on an `add_file` rule parsed fine and then did nothing. File rules got pulled
out in `runMatch` before any `where` filter was built, so they skipped both matching and
validation — even a malformed `where` went through without an error.

Fixing that on its own wasn't enough though. Under `go test`, `target: main` only matches the
generated test main (the package under test compiles as `-p example.com/app`), and the dry run
lists `_testmain.go` before it exists. `findGoSources` drops files that aren't on disk, so the
test main showed up with no sources and looked like a normal build, meaning `is_test: false`
would still have let the file in.

So `Dependency` now records `IsTest` from every Go file the command lists, and file rules get
their `where` evaluated where they're collected. The clause is checked against the package as a whole — `has_struct: Mock` holds if any file
declares it, so `not` and `all-of` behave the way they read — and units with no sources get an
empty package so `is_test` still works there.
Added a short note to `docs/rules.md` too.

## Tests

Unit tests cover test-build detection for the not-yet-written `_testmain.go`, every `is_test`
and `has_func` case on file rules including the no-sources unit, and a malformed `where`
erroring out. They fail on the original `match.go`.

Also ran rarguelloF's repro against both binaries:

| where | command | before | after |
|---|---|---|---|
| `is_test: true` | `go build` | added | **not added** |
| `is_test: false` | `go test` | added | **not added** |

The other two cases are unchanged. `is_test: true` under `go test` still hits the empty package
clause from #1348, which #1356 fixes.

## Motivation

Without this there's no way to keep an injected `func init()` out of test builds, which is what
rarguelloF was after.

Fixes #1347


tool/internal/setup/match.go
```diff
@@ -165,6 +172,37 @@ func (sp *setupPhase) runMatch(
...
+	for _, source := range dep.Sources {
+		if err = ctx.Err(); err != nil {
+			return false, err
+		}
+		tree, parseErr := ast.ParseFileFast(source)
+		if parseErr != nil {
+			return false, parseErr
+		}
+		if f.Match(&matchContext{IsTest: isTest, SourceFile: source, AST: tree}) {
+			return true, nil
```
ADITYA-CODE-SOURCE: The full predicate appears to be evaluated independently against each source file, returning `true` when any single file matches. Could this lead to incorrect package-level behavior? For example, `not: {has_struct: Mock}` would match when one file lacks `Mock`, even if another file declares it. Similarly, an `all-of` predicate requiring declarations spread across multiple files would fail. Since `add_file` operates on the entire package, should the predicate be evaluated against the complete package source set instead?

ADITYA-CODE-SOURCE (changes requested):

tool/internal/setup/match.go
krsatyamthakur-droid: thanks, Checking files one at a time broke `not` and `all-of` at the package level.

It now checks the clause against the **whole package**, so `not {has_struct: Mock}` stays out if any file has `Mock`, and an `all-of` split across files works.

Added a test with both your examples. They **fail on the old code** and pass now.

om7057 (commented):
I pulled this into a worktree and ran the full suite (`go vet ./...`, `go build ./...`, `go test ./...` across the whole `tool` module). Everything passes clean.

To check the fix actually does something, I reverted `match.go` and `filter.go` back to `main` while keeping this PR's new test files (`match_test.go`, `find_test.go`), which still compiled since the `Dependency.IsTest` plumbing lives in `find.go`. With `main`'s matching logic, the new tests fail exactly where you'd expect. `TestRunMatch_FileRuleWhere/is_test_true_skips_a_normal_build`, `is_test_false_skips_the_generated_test_main`, both `has_func` cases, `TestRunMatch_FileRuleWhereBuildError`, `TestPreciseMatching_IsTestFromDependency`, and two of the `TestRunMatch_FileRuleWherePackageLevel` cases all fail with "Should be empty, but was [add_to_main]". That's the file rule firing when `where` should have excluded it, the exact bug from #1347. Restoring the two files makes them all pass again.

I also tried reproducing the CLI-level repro from #1347 in a scratch workspace, two modules, `otelc go build` vs `otelc go test`, but hit module-wiring issues getting my scratch `myrules` module recognized as a dependency at all. That's on my repro setup, not the PR. The unit tests already isolate the exact bug from the issue so I didn't chase it further.

Two things I'd flag:
- `TestRunMatch_FileRuleWhereBuildError` looks like it checks that a malformed `where` on a file rule now errors at load time. Worth confirming that's what actually ships, since the PR description says both the error-at-load-time and evaluate-it-properly paths landed.
- The package-level `has_struct`/`not`/`all-of` semantics (checking against the union of all files in the package, not per-file) is a real behavior choice, not just a bug fix. `not: {has_func: X}` ends up stricter than a per-file reading would suggest, worth calling that out explicitly since it's not obvious from the issue.

Test coverage looks good, and `docs/rules.md` got a short note.

krsatyamthakur-droid:
I confirmed that malformed  `where.file`  predicates on ` add_file`  rules now return an error before the rule is added. This is covered by  `TestRunMatch_FileRuleWhereBuildError` . The predicate is compiled during setup matching rather than at initial YAML loading, so the invalid rule fails before instrumentation continues.

The package-level behavior is intentional for  `add_file` , since the rule adds a file to the package rather than matching an individual source file. Therefore,  `has_func`  and  `has_struct`  match when any file in the package satisfies the predicate,  not  negates that package-level result, and  all-of  conditions can be satisfied by different files. I documented this in  `docs/rules.md`  and covered it with  `TestRunMatch_FileRuleWherePackageLevel` .

I also added a small guard so an explicitly empty  where: {}  behaves like no predicate instead of dereferencing a nil filter. The full Go test suite passes.

tool/internal/setup/match.go
```diff
@@ -165,6 +172,32 @@ func (sp *setupPhase) runMatch(
...
+	f, err := build(where)
+	if err != nil {
+		return false, ex.Wrapf(err, "build where filter for rule %q", fr.GetName())
+	}
+	trees := make([]*dst.File, 0, len(dep.Sources))
+	for _, source := range dep.Sources {
+		if err = ctx.Err(); err != nil {
+			return false, err
+		}
+		tree, parseErr := ast.ParseFileFast(source)
```
amazingakai: Is there a real world usecase for `where` predicates on file rules? This PR adds the additional overhead of parsing the entire package multiple times. `fileRuleApplies` calls `ast.ParseFileFast(source)` for every file in the package in a loop for *every* file rule. Then `preciseMatching` parses the package all over again. If we really need this feature, the ASTs should be parsed lazily and cached once per dependency to avoid the massive performance hit.

docs/rules.md
```diff
@@ -1244,6 +1244,10 @@ This rule adds a new Go source file to the target package.
 - Introducing new functionalities or APIs to an existing package.
 
 **Selectors:** none. File rules apply to the target package as a whole and have no point selector.
+`where.file` predicates are checked against the package as a whole: `has_func: X` holds if any
+file declares `X`, `not` negates that, and the branches of `all-of` can be satisfied by different
+files. Units the toolchain generates, such as the test main, have no source files, so `is_test`
+still applies there but no file-content predicate matches.
```
amazingakai, line 1250: Using `where.file` to match against the whole package sounds confusing from an API perspective since "file" implies a single file. We should probably rethink this naming or add a package-level selector

amazingakai (changes requested):

tool/internal/setup/match.go
krsatyamthakur-droid: Thanks for raising this. There is a real use case for predicates on file rules:  add_file  rules may need to inject a file only into production builds, or only into test builds. This was the issue behind #1347, where an injected  init()  function needed to be excluded from test builds using  is_test .
I also fixed the performance concern you pointed out. The dependency source files are now parsed once, and the cached ASTs are reused for both  add_file   where.file  evaluation and precise matching. This avoids reparsing the entire package for every file rule.

docs/rules.md
krsatyamthakur-droid: That’s a fair concern. The current implementation treats  add_file  differently because the rule adds one file to the target package, so evaluating the predicate against the package is what allows cases such as  is_test  to work for generated, source-less units.
I’ve documented the current package-level behavior, but I agree that the  where.file  name can be confusing in this context. I’d prefer not to introduce a new selector or rename the existing schema as part of this bug fix without first agreeing on the API direction. Would you prefer that we:
1. keep  where.file  and explicitly document that  add_file  predicates are package-level, or  
2. introduce a separate package-level selector and reserve  where.file  for predicates on individual source files?
The parsing overhead raised in the other comment has been addressed in  17b5e7d ; source files are now parsed once per dependency and the ASTs are reused.

Issue #1361, opened 2026-09-14
bug(tool): duplicated go:linkname definition when more than one package in a link unit receives otelc.runtime.go


rarguelloF:
### Describe the bug

The setup phase writes a generated `otelc.runtime.go` into every package selected on the command line. Each of those files declares `//go:linkname` variables for every matched hook package:

```go
//go:linkname _getstack0 example.com/otelc-ext-hooks.OtelGetStackImpl
var _getstack0 = _otel_debug.Stack

//go:linkname _printstack0 example.com/otelc-ext-hooks.OtelPrintStackImpl
var _printstack0 = func(bt []byte) { _otel_log.Print(string(bt)) }
```

These are push linknames, so each one defines the target symbol. When two packages in the same link unit both receive a runtime file, both define `<hook package>.OtelGetStackImpl` and the link fails:

```text
link: duplicated definition of symbol example.com/otelc-ext-hooks.OtelGetStackImpl,
  from example.com/otelc-ext/lib (type SDATA size 8) and example.com/otelc-ext (type SDATA size 8)
```

What drives this is package selection, not rule matching. `generateRuntimePerPackage` in `tool/internal/setup/setup.go` loops over every selected package and hands the full matched rule set to `addDeps`, so a package that no rule targets still receives a complete runtime file. In the reproduction below, nothing matches `lib`, yet its generated file is identical to the application package's:

```go
// lib/otelc.runtime.go
// This file is generated by the opentelemetry-go-compile-instrumentation tool. DO NOT EDIT.
package lib

import _ "example.com/otelc-ext-hooks"
import _otel_log "log"
import _otel_debug "runtime/debug"
import _ "unsafe"

//go:linkname _getstack0 example.com/otelc-ext-hooks.OtelGetStackImpl
var _getstack0 = _otel_debug.Stack

//go:linkname _printstack0 example.com/otelc-ext-hooks.OtelPrintStackImpl
var _printstack0 = func(bt []byte) { _otel_log.Print(string(bt)) }
```

**When it happens.** Any `otelc go build ./...` or `otelc go test ./...` where two or more selected packages end up in the same binary. The second package does not need to be instrumented. It only needs to be reachable from the first, so that the linker pulls both into one link unit. A module with a single package is unaffected, which is why the documented `./...` commands work on small examples.

The wildcard is not required either. Naming two packages in one invocation, where one imports the other, fails the same way:

```sh
[...]

PR #1362, opened 2026-09-14
fix(tool): avoid self-import in otelc.runtime.go for local hooks

rarguelloF:
## Description

`otelc` writes a generated `otelc.runtime.go` into every package selected on the command line, and each of those files blank-imports the hook packages of all matched rules. The import list did not depend on which package received the file, so a hook package ended up with a generated file that imports itself:

```text
package example.com/otelc-self-import/hooks
        imports example.com/otelc-self-import/hooks from otelc.runtime.go:
        import cycle not allowed
```

Rules are now filtered against the package receiving the file, and no file is written when nothing is left. The instrumented package still gets the hook import, so the hook still runs.

Two related cases are fixed too. A build that names files instead of packages loads a synthetic `command-line-arguments` package, so the filter could not tell that the hook package was itself, and a glob rule target could still produce the self-import. And when filtering leaves nothing to write, an `otelc.runtime.go` from an earlier setup stayed on disk with its old imports and linkname declarations active. That second case also happens on `main` when the new rules match nothing.

This does not make every `./...` build work. A module with more than one package in the same link unit still fails, because each generated file defines the same hook linkname. That is pre-existing and tracked in #1361, and the mixed unit test added here carries a comment saying so.

Also moves `writeTestFile` and `runOtelcCommand` out of `cgo_vet_test.go` into a shared `helpers_test.go`.

## Motivation

Hooks have to live in a separate package, but nothing requires them to live in a separate Go module. When they sit in the application module, `otelc go test ./...` selects the hook package too and the build fails before any test runs. That is the command the docs recommend, so an application with local hooks could not be tested with it. The in-repo instrumentations are unaffected because each one is its own module.

The workarounds were to name a single package (`otelc go test .`) or to move the hooks into a nested module.


tool/internal/setup/add.go
```diff
@@ -113,12 +122,24 @@ func buildOtelcRuntimeAst(decls []dst.Decl, packageName string) *dst.File {
...
 	for _, m := range matched {
-		funcRules = append(funcRules, m.AllFuncRules()...)
-		fileRules = append(fileRules, m.FileRules...)
+		for _, funcRule := range m.AllFuncRules() {
+			if funcRule.Path != pkg.importPath {
+				funcRules = append(funcRules, funcRule)
+			}
+		}
+		for _, fileRule := range m.FileRules {
+			if fileRule.Path != pkg.importPath {
```
darkraider01, line 139: for named .go file targets, go/packages uses the synthetic `PkgPath` `command-line-arguments`, so this comparison would never identify a local hook package as itself. I think that case is outside the go test ./... regression being fixed here, but is that intentionally out of scope?

tool/internal/setup/add.go
```diff
@@ -113,12 +122,24 @@ func buildOtelcRuntimeAst(decls []dst.Decl, packageName string) *dst.File {
...
+				fileRules = append(fileRules, fileRule)
+			}
+		}
 	}
 	if len(funcRules) == 0 && len(fileRules) == 0 {
 		return nil
```
ADITYA-CODE-SOURCE: Could this leave a stale `otelc.runtime.go` file behind? For example, a previous standalone otelc setup may have generated the file, but the rules later change so that the package contains only self-importing rules. This branch returns without tracking, removing, or replacing the existing file, leaving its old imports and `linkname` directives active. Should this path detect and remove an existing generated file, with a regression test covering two successive setup states?

tool/internal/setup/add.go
rarguelloF: You're right, and it's not out of scope. With an exact target nothing matches on a file build, so that case is harmless. But a `**` target does match `command-line-arguments`, and then naming a file inside the hook package brings the self-import back.

Fixed on the current branch, `runtimeImportPath` now resolves the real import path for file targets.

tool/internal/setup/add.go
rarguelloF: Yes, it can, and nothing cleans it up (`Cleanup` only runs from `otelc go build` or the explicit cleanup command). It already happens on main when the new rules match nothing, this branch just makes it easy to hit.

The early return now deletes the file, tracked first so a revert can restore it, with a test for two successive setups.

PR #1367, opened 2026-09-16
feat(instrumentation/echo): add labstack/echo/v4 route enrichment

### LINKED ISSUE #1366: New Instrumentation Request for github.com/labstack/echo/v4

### Package

`github.com/labstack/echo/v4`

### Version

v4.0.0 and later. v5 is out of scope.

### Environment

Go HTTP services built with `otelc`. Kubernetes Deployments and local binaries.

### Use-case

APIs served by Echo. `Echo.Start` uses `*http.Server`, so net/http already creates a SERVER span. The span stays named after the method. The route template never reaches `req.Pattern`, so `http.route` is empty. Same gap gin already fills.

### Telemetry

Enrich that span. Do not start a second one. No metrics.

- After `(*Router).Find`, set `http.route` from `Context.Path()` and rename the span to `METHOD {template}` (e.g. `GET /hello/:name`).
- On `c.Error(err)` and handler `return err` (5xx / plain error): set span status Error and record the exception.
- 404 / 405: keep the method-only name. Do not set `http.route`. Do not set Error status.

Echo has no `Context.Next`. Hook `Find` After, plus `(*context).Error` and `DefaultHTTPErrorHandler` Before.

### Anything else?

Datadog Orchestrion wraps `echo.New`. Longsuite: `dd-trace-go/internal/orchestrion/_integration/echo.v4`.

atharvamhaske:
## Description

This PR adds compile-time instrumentation for `github.com/labstack/echo/v4`.
The hook does not start a span. It enriches the existing `net/http` SERVER span.
After `(*Router).Find`, the hook sets `http.route` from `Context.Path()`.
Then the hook renames the span to `METHOD {template}`.
The hook records `c.Error(err)` and a handler `return err` that is a 5xx or a plain error.
A 404 or 405 keeps the method-only name.

Echo has no `Context.Next`. The rules hook `Find` After.
The rules also hook `(*context).Error` and `DefaultHTTPErrorHandler` Before.

The floor is v4.0.0. v5 is out of scope.

## Motivation

`Echo.Start` uses `*http.Server`. As a result, `net/http` already creates a SERVER span.
The span stays named after the method. The route template does not reach `req.Pattern`.
This is the same gap that gin already fills.

Fixes #1366


instrumentation/github.com/labstack/echo/v4/otelc.yaml
```diff
@@ -0,0 +1,41 @@
+version: "v1.0.0"
+
+# Echo has no Context.Next. ServeHTTP matches the route with Router.Find,
+# then runs the handler chain. After Find, Context.Path() is the template.
+echo_router_find_hook:
+  target: github.com/labstack/echo/v4
+  version: v4.0.0
```
darkraider01, line 7: Are we sure v4.0.0 is a valid lower bound here? matchedRoute() uses echo.ContextKeyHeaderAllow, which doesn't seem to be available in the earlier v4 releases. If we want to keep v4.0.0 support, could we avoid relying on that symbol here, maybe by checking MethodNotAllowedHandler directly?

instrumentation/github.com/labstack/echo/v4/otelc.yaml
```diff
@@ -0,0 +1,41 @@
...
+        path: "go.opentelemetry.io/otelc/instrumentation/github.com/labstack/echo/v4"
+
+# Handler `return err` is forwarded by ServeHTTP to HTTPErrorHandler. The
+# default implementation is this method. 4xx (including 404/405) is skipped
+# in the hook so those stay unset.
+echo_default_http_error_handler_hook:
+  target: github.com/labstack/echo/v4
+  version: v4.0.0
+  where:
+    func: DefaultHTTPErrorHandler
```
darkraider01, line 36: What happens here when the application replaces `e.HTTPErrorHandler` with a custom handler? `ServeHTTP` calls the configured `HTTPErrorHandler` field, so in that case `DefaultHTTPErrorHandler` wouldn't run and a handler `return err` wouldn't get recorded by this instrumentation. Should we cover that case as well?

instrumentation/github.com/labstack/echo/v4/hook.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	recordRequestError(c, err)
+}
+
+func recordRequestError(c echo.Context, err error) {
+	if err == nil || c == nil || c.Request() == nil {
+		return
+	}
+	var he *echo.HTTPError
+	if errors.As(err, &he) && he.Code < 500 {
```
ADITYA-CODE-SOURCE, line 85: This checks only the outer `*echo.HTTPError`, but Echo’s `DefaultHTTPErrorHandler` uses `he.Internal` when it contains another `*echo.HTTPError`.

For example, an outer 500 wrapping an internal 404 produces a 404 response, while this hook records an exception and sets the span status to `Error`. Conversely, an outer 404 wrapping an internal 500 is skipped even though Echo produces a 500 response.

Could we resolve the effective `HTTPError` in the same way as Echo and add coverage for nested HTTP errors in both directions?

atharvamhaske: Hey can you please let maintainers review this at their pace?
Ik what drives you to review this but can you please hold on?

amazingakai: @atharvamhaske Community peer reviews are a core part of an open-source project and are always welcome. I appreciate @darkraider01 and @ADITYA-CODE-SOURCE taking the time to review this PR and raise architectural and correctness questions. Please address the points they have raised.

I also need to address your response: "Hey can you please let maintainers review this at their pace? Ik what drives you to review this but can you please hold on?"

This is unnecessarily aggressive and goes against the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/main/code-of-conduct.md), which expects respectful communication and assuming good intent.

We want everyone to feel welcome to contribute and review. Please ensure future communication remains respectful and focused entirely on the technical merits of the code.

atharvamhaske: hey that was settled over slack
ik this was my mistake but reason i posted is i was busy in some meetings and my gmail notifications suddenly popped up with 15+ comments so he deleted those by adding one single comment 


i hope there's not an issue now
also ik code of conduct and even i do review multiple prs in other cncf project

atharvamhaske: cc @amazingakai

PR #1369, opened 2026-09-16
fix(instrumentation/redis): skip Conn hook after go-redis v9.9.0

### LINKED ISSUE #1286: bug(instrumentation/redis): Client.Conn commands emit duplicate spans
### Describe the bug

Commands executed through `redis.Client.Conn()` produce two Redis client spans instead of one.

The parent `redis.Client` receives an OpenTelemetry hook when it is created. go-redis then clones the parent's hooks when `Client.Conn()` creates a dedicated connection. However, the compile instrumentation also injects `afterClientConnV9`, which adds another OpenTelemetry hook to the connection.

As a result, both hooks process the same command. The additional span also has a malformed endpoint because `afterClientConnV9` passes `client.String()` instead of a `host:port` address.

  ### Environment

  - OS: Windows (`windows/amd64`)
  - Go version: `go1.25.5`
  - Repository version: commit `3d3d71e1786fb29120a0ce41913ae688fefe266d`
  - go-redis version: `v9.22.0`

  ### To Reproduce

  Add this temporary test under `instrumentation/github.com/redis/go-redis/v9`:

  ```go
  package v9

  import (
        "context"
        "errors"
        "net"
        "sync"
        "testing"

        "github.com/redis/go-redis/v9"
        "github.com/stretchr/testify/require"
  )

  func TestClientConnEmitsSingleSpan(t *testing.T) {
        initOnce = *new(sync.Once)
        t.Setenv("OTEL_GO_ENABLED_INSTRUMENTATIONS", "redis")
        sr := setupTestTracer(t)

        client := redis.NewClient(&redis.Options{
                Addr:       "localhost:6379",
                MaxRetries: -1,
                Dialer: func(context.Context, string, string) (net.Conn, error) {
                        return nil, errors.New("dial failed")
                },
        })
        t.Cleanup(func() { _ = client.Close() })

        afterNewRedisClientV9(nil, client)

        conn := client.Conn()
        afterClientConnV9(nil, conn)
        _ = conn.Get(context.Background(), "key").Err()

        require.Len(t, sr.Ended(), 1)
  }
  ```

  Run:

  ```console
  cd instrumentation/github.com/redis/go-redis/v9
  go test -run '^TestClientConnEmitsSingleSpan$' -count=1 -v .
  ```

  The assertion fails because two spans are recorded:

  ```text
  expected 1 item, got 2
  ```

  The recorded endpoints are:

  ```text
  server.address = "Redis<localhost:6379 db:0>"
  server.address = "localhost"
  server.port = 6379
  ```

  ### Expected behavior

  One Redis client command executed through `Client.Conn()` should produce one client span with:

  ```text
  server.address = "localhost"
  server.port = 6379
  ```

  ### Additional context

`afterNewRedisClientV9` installs the hook on the parent client. In go-redis, `Client.Conn()` passes the parent's hook mixin to `newConn`, which clones those hooks. The separate `redis_client_hook` rule then injects `afterClientConnV9` and adds a second hook.

A post-merge review comment on #292 already noted that `Conn.String()` is not a valid endpoint, but I could not find an issue or pull request tracking the duplicate spans.

Would limiting the explicit `Client.Conn()` hook rule to versions before go-redis v9.9.0 be the appropriate compatibility boundary?

 ### Proposed implementation

  I have a local implementation ready that limits the explicit `Client.Conn()` hook rule to go-redis versions before v9.9.0.

Starting with go-redis v9.9.0, `Client.Conn()` clones the parent client's hooks, so injecting another hook after `Conn()` creates the duplicate span. Versions through v9.8.x do not inherit those hooks and still require the explicit rule.

The implementation adds the `v9.0.0,v9.9.0` version boundary, exercises `Client.Conn()` in the Redis integration application, and asserts that one `GET` command produces exactly one span with the correct `server.address` and `server.port`.

  I will wait for maintainer confirmation on this version-gated approach before opening a PR.

atharvamhaske:
## Description

The `redis_client_hook` rule now matches only `v9.0.0,v9.9.0`.

From go-redis `v9.9.0`, `Client.Conn()` copies the parent hooks. A second `afterClientConnV9` hook then added another hook. One command made two spans. That extra hook also set `server.address` from `Conn.String()`, which is not a host:port value.

Older go-redis releases still get the Conn hook. Newer releases keep the parent hook only.

## Motivation

`Client.Conn()` made duplicate command spans after go-redis `v9.9.0`.

Fixes #1286


instrumentation/github.com/redis/go-redis/v9/otelc.yaml
```diff
@@ -29,6 +29,7 @@ redis_hook_newsentinelclient:
 
 redis_client_hook:
   target: github.com/redis/go-redis/v9
+  version: v9.0.0,v9.9.0
```
darkraider01, line 32: Could you also add a test with a pre-v9.9.0 version, maybe v9.8.x? Since we're introducing the version boundary here, I think it'd be good to make sure the legacy `Conn()` hook propagation still works below it too.

atharvamhaske: sure

instrumentation/github.com/redis/go-redis/v9/hook_test.go
```diff
@@ -552,3 +552,44 @@ func TestDialHook_Error(t *testing.T) {
...
+	client := redis.NewClient(&redis.Options{
+		Addr:     "localhost:6379",
+		Protocol: 2,
+		Dialer: func(context.Context, string, string) (net.Conn, error) {
+			return nil, errors.New("dial disabled")
+		},
+	})
+	t.Cleanup(func() { require.NoError(t, client.Close()) })
+
+	afterNewRedisClientV9(nil, client)
```
ADITYA-CODE-SOURCE: This test verifies go-redis’s native hook inheritance, but it does not exercise the version gate added in `otelc.yaml`. It installs the constructor hook manually and never runs the compile-instrumentation pipeline.

As a result, the test would still pass if the new version restriction were removed and `afterClientConnV9` were incorrectly injected for v9.9+.

Could the post-v9.9 regression test build an application through `otelc` so that it verifies the `Conn` rule is actually skipped?

y1yang0 (approved):

Issue #1370, opened 2026-09-17
[LFX Mentorship 2026 Term 3] Project plan: zero-code GenAI observability adapters for otelc


NameHaibinZhang:
## Summary

This issue tracks the LFX Mentorship 2026 Term 3 project:

> **OpenTelemetry: Zero-code AI Agent observability for otelc**

The project will use the **Generative AI semantic conventions registry** ([open-telemetry/semantic-conventions-genai](https://github.com/open-telemetry/semantic-conventions-genai)) as the source of truth and make GenAI the first adopter of the repository's Weaver-based semantic-convention pipeline.

The work is sequenced behind two prerequisite efforts:
- #728 — Generate Go semconv instrument registration from the Weaver registry.
- #745 — Centralize semconv telemetry contracts under `schemas/otelc/groups/`.

Once those foundations land, the mentee will migrate the existing OpenAI and Anthropic instrumentations onto generated conventions, then add Gemini, MCP, and LangChain-style integrations by implementing already-declared conventions rather than inventing new ones.

## Background and motivation

Manual instrumentation code tends to drift from the declared semantic conventions. The repository's direction, captured in #728 and #745, is to make the registry the source of truth:

- **Emit** → generated `*conv` types from the Weaver registry.
- **Declare** → only the attributes we invent, validated by the Weaver registry check.
- **Verify** → real emitted telemetry matched against the Weaver registry (live-check).

The GenAI domain is ideal as the first adopter because:

- `semantic-conventions-genai` already defines `openai.inference.client`, `anthropic.inference.client`, and `model/mcp/` conventions.
- OpenAI instrumentation from #604 is already merged and provides a concrete migration target.
- Anthropic instrumentation exists but is not yet aligned with the generated convention model.
- Gemini, MCP, and LangChain integrations become "implement declared conventions" instead of "design new conventions."

## Goals

- [ ] Land #745 schema organization and the Weaver-based contract pipeline.
- [ ] Land #728 codegen that turns the Weaver registry into usable Go semconv types and verification tooling.
- [ ] Migrate OpenAI instrumentation from #604 to emit generated `gen_ai.*` / `openai.*` conventions without changing observable semantics.
- [ ] Migrate Anthropic instrumentation to the same generated convention model.
- [ ] Add Gemini instrumentation by implementing conventions already declared in `semantic-conventions-genai`.
[...]

PR #1371, opened 2026-09-17
docs: add GenAI registry alignment ADR and compatibility matrix

vedant21-oss:
## Description

Adds two design documents for the LFX Term 3 project in #1370:

- `docs/adr/0006-genai-adapter-contract.md` proposes a provider-neutral GenAI
  adapter contract, so the shared span lifecycle, usage mapping, streaming
  handover, error mapping and privacy gate live in one place instead of being
  copied per provider.
- `docs/genai-compatibility-matrix.md` records what the OpenAI and Anthropic
  instrumentation supports today, and the open package choices for Gemini, MCP
  and a LangChain-style framework.

## Motivation

`semconv/genai.go` currently exists four times: three OpenAI copies (v2 and v3
are byte-identical) and an Anthropic copy that has drifted. The middleware
lifecycle is duplicated with small differences in each module, so adding Gemini,
MCP and LangChain this term would mean four more copies.

The matrix is the design artifact #1370 asks for before implementation starts.
The package choices for Gemini, MCP and LangChain are still open, and #1153 may
already cover the Gemini target. Feedback on the shape of the contract is very
welcome before I start on the adapter itself.

Opened as a draft because it is a design proposal, not finished work.

Related #1370


docs/genai-compatibility-matrix.md
```diff
@@ -0,0 +1,95 @@
...
+
+| | |
+| --- | --- |
+| Entry point | `NewClient`, via a `before` hook that appends `option.WithMiddleware` |
+| Transport | HTTP client middleware |
+| Operations | `chat` (`/v1/messages`), `count_tokens` (`/v1/messages/count_tokens`) |
+| Request attributes | `gen_ai.system`, `gen_ai.operation.name`, `gen_ai.request.model`, `gen_ai.provider.name`, `gen_ai.request.max_tokens`, `gen_ai.request.temperature`, `gen_ai.request.top_p`, `gen_ai.request.top_k`, `gen_ai.request.is_stream` |
+| Response attributes | `gen_ai.response.id`, `gen_ai.response.model`, `gen_ai.response.finish_reasons` |
+| Usage | `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, `gen_ai.usage.total_tokens` (computed as input + output), plus `gen_ai.usage.cache_read.input_tokens` and `gen_ai.usage.cache_creation.input_tokens` |
+| Streaming | Partial. `gen_ai.request.is_stream` is set; the stream itself is not instrumented |
```
ADITYA-CODE-SOURCE: This does not match the current Anthropic behavior. When a request contains `"stream": true`, the middleware returns before creating a GenAI span, so `gen_ai.request.is_stream` is not emitted.

Could we describe this as:

> **Not instrumented:** Declared streaming requests pass through without a GenAI span.

The attribute is only set in the defensive case where a nominally non-streaming request unexpectedly receives an SSE response.

vedant21-oss: Good catch, thanks. You're right, I misread it. if isStream { return next(req) } happens before the span is created, so a declared streaming request gets no span at all. Fixed the row to say not instrumented, and added a note that is_stream only shows up in the SSE fallback where a non-streaming request gets an event stream back.

PR #1372, opened 2026-09-17
fix(tool): omit duplicate go:linkname definitions in multi-package link units

### LINKED ISSUE #1361: bug(tool): duplicated go:linkname definition when more than one package in a link unit receives otelc.runtime.go

Babul422:
## Description

When multiple packages in the same link unit received `otelc.runtime.go` during multi-package builds or tests (for example, `otelc go test ./...`), `genVarDecl` emitted the same push `//go:linkname` declarations (`OtelGetStackImpl` and `OtelPrintStackImpl`) into multiple packages. At link time, `cmd/link` failed with duplicate symbol errors (Issue #1361).

## Solution

Following the investigation and maintainer feedback:

- Removed `genVarDecl` and the push `//go:linkname` declarations from `tool/internal/setup/add.go`.
- Removed the imports that were only needed by those declarations (`runtime/debug`, `log`, and `unsafe`).
- Simplified `otelc.runtime.go` generation so instrumented packages only receive blank imports (`_ "<hook-path>"`) for matched hook packages. These imports are still needed for hook registration.
- Removed the package dependency/ownership logic that was only needed to determine where the linkname declarations should be emitted.
- Left `tool/internal/instrument/impl.tmpl` unchanged. `OtelGetStackImpl` and `OtelPrintStackImpl` remain there with their existing nil checks, pending separate maintainer consensus on whether they should be removed or retained.

## Panic Recovery

Added a regression integration test in `test/integration/hook_panic_test.go` covering deliberately panicking `Before` and `After` hooks.

The test verifies that:

- hook panics are recovered by the trampoline;
- the existing `failed to exec ... hook` diagnostic and panic error are printed;
- the instrumented target function continues execution;
- the host process exits successfully without crashing.

The test does not make any assumption about stack trace output.

## Testing

- `go test -count=1 ./tool/...` — **PASS** (all 13 packages)
- `go -C test test -v -timeout=15m -count=1 -tags integration -run 'TestHookPanicRecovery|TestMultiPackageLinkUnit' ./integration/...` — **PASS** (all 4 tests)
- `git diff --check` — **CLEAN**
- `make lint/license-header` — **PASS**
- `golangci-lint run --config .tools/golangci.yml` — exits with status 1 because of 42 pre-existing issues in untouched repository files; no issues were reported in the changed/new files.

Fixes #1361

## AI Assistance

Assisted by Claude Opus 4.6.


darkraider01:
the issue owner already raised a PR against the issue https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/1361

Babul422:
Hi @darkraider01, thanks for reviewing!

Just to clarify, I believe PR #1362 and PR #1372 address two different issues that both show up during `./...` runs.

PR #1362 fixes the compile-time self-import cycle (`import cycle not allowed`) when a local hook package generates an `otelc.runtime.go` that imports itself.

PR #1372 addresses #1361, which is a link-time duplicate symbol error. This happens when multiple packages in the same link unit generate the same push `//go:linkname` definitions for `OtelGetStackImpl` and `OtelPrintStackImpl`.

I also noticed that the description of #1362 mentions that multi-package link units can still fail due to these duplicate linkname definitions, which are tracked separately in #1361.

If #1362 is merged first, I'm happy to rebase #1372 on top of it and resolve any overlap in `setup.go` and `add.go`.

darkraider01:
the issue owner already raised a PR against the issue #1362

tool/internal/setup/setup.go
```diff
@@ -303,22 +292,117 @@ func (sp *setupPhase) generateRuntimePerPackage(
...
+	importedByOther := findImportedPackages(pkgs)
+
 	for _, pkg := range pkgs {
 		pkgDir := pkgload.PackageDir(pkg)
 		if pkgDir == "" {
 			sp.Warn("skipping package without Go files", "package", pkg.PkgPath)
 			continue
 		}
 
+		emitLinknames := !importedByOther[pkg.PkgPath]
```
ADITYA-CODE-SOURCE: This selects linkname ownership globally, but `go build ./...` can produce multiple independent link units.

For example, if both `cmd/a` and `cmd/b` import an instrumented `lib`, both command packages may be classified as roots and emit the same hook linknames, even though they are linked into separate binaries.

Could ownership be determined and tested independently for each resulting link unit instead?

ADITYA-CODE-SOURCE (changes requested):
The current approach chooses ownership across the full package graph rather than per link unit, so builds producing multiple binaries may still assign duplicate linkname owners. I left an inline example.

tool/internal/setup/setup.go
Babul422: Thanks for the feedback regarding package-graph vs. per-link-unit ownership.
While investigating how ownership could be scoped per link unit, I looked more closely at how the symbols generated by `genVarDecl` are defined and consumed across the codebase.

What I found is:
1. `tool/internal/setup/add.go` generates push linknames targeting the hook package:
   `//go:linkname _getstack%d <hookPkg>.OtelGetStackImpl`
   `//go:linkname _printstack%d <hookPkg>.OtelPrintStackImpl`

2. `tool/internal/instrument/impl.tmpl` defines `OtelGetStackImpl` and `OtelPrintStackImpl` as package-local variables in the instrumented target package, initialized to `nil`, without corresponding `//go:linkname` directives.

3. I couldn't find any runtime references from `pkg/`, `instrumentation/`, or `test/` that resolve to the hook-package symbols generated by `genVarDecl`.

4. I also verified this with a minimal reproduction and `go tool nm`. The target package's variables remain `nil`, while the hook-package symbol generated by `genVarDecl` is stripped as dead code when there is a single definition. When multiple packages in the same link unit emit it, the linker reports the duplicate-symbol error.

Removing these linkname declarations allowed the reproduction to link successfully, with the same observed runtime behavior.

Based on this, I'm wondering whether the DAG-based ownership tracking in this PR is actually necessary, or whether the duplicate-symbol problem is coming from linkname declarations that are no longer connected to the runtime path.

A simpler approach could be:

- remove `genVarDecl`;
- have `otelc.runtime.go` generate only blank imports of the matched hook packages;
- remove the dependency-graph ownership logic that exists solely to decide where the linknames should be emitted;
- retain the multi-package integration tests.

Before I update the PR, could you confirm whether removing `genVarDecl` aligns with the intended architecture, or whether `OtelGetStackImpl` / `OtelPrintStackImpl` have an intended role that we should preserve?

ADITYA-CODE-SOURCE: > Thanks for the feedback regarding package-graph vs. per-link-unit ownership. While investigating how ownership could be scoped per link unit, I looked more closely at how the symbols generated by `genVarDecl` are defined and consumed across the codebase.
> 
> What I found is:
> 
> 1. `tool/internal/setup/add.go` generates push linknames targeting the hook package:
>    `//go:linkname _getstack%d <hookPkg>.OtelGetStackImpl`
>    `//go:linkname _printstack%d <hookPkg>.OtelPrintStackImpl`
> 2. `tool/internal/instrument/impl.tmpl` defines `OtelGetStackImpl` and `OtelPrintStackImpl` as package-local variables in the instrumented target package, initialized to `nil`, without corresponding `//go:linkname` directives.
> 3. I couldn't find any runtime references from `pkg/`, `instrumentation/`, or `test/` that resolve to the hook-package symbols generated by `genVarDecl`.
> 4. I also verified this with a minimal reproduction and `go tool nm`. The target package's variables remain `nil`, while the hook-package symbol generated by `genVarDecl` is stripped as dead code when there is a single definition. When multiple packages in the same link unit emit it, the linker reports the duplicate-symbol error.
> 
> Removing these linkname declarations allowed the reproduction to link successfully, with the same observed runtime behavior.
> 
> Based on this, I'm wondering whether the DAG-based ownership tracking in this PR is actually necessary, or whether the duplicate-symbol problem is coming from linkname declarations that are no longer connected to the runtime path.
> 
> A simpler approach could be:
> 
> * remove `genVarDecl`;
> * have `otelc.runtime.go` generate only blank imports of the matched hook packages;
> * remove the dependency-graph ownership logic that exists solely to decide where the linknames should be emitted;
> * retain the multi-package integration tests.
> 
> Before I update the PR, could you confirm whether removing `genVarDecl` aligns with the intended architecture, or whether `OtelGetStackImpl` / `OtelPrintStackImpl` have an intended role that we should preserve?

Thanks for tracing this further. I confirmed that the generated variables referenced by `impl.tmpl` belong to the instrumented target package, while `genVarDecl` defines symbols in the hook package. There is no pull `//go:linkname` connecting them, so the target package variables appear to remain `nil` today.

Given that, removing `genVarDecl` may be cleaner than introducing an ownership mechanism, while retaining the blank hook imports.

Since these declarations appear to be intended for panic stack reporting, it would be helpful for a maintainer to confirm whether that behavior is still required. A test using a deliberately panicking hook could also establish the expected behavior before the declarations are removed.

Babul422: Thanks for suggesting the panicking hook test. I tried this with a small reproduction using both a panicking Before hook and a panicking After hook.

The behavior I see is:

- The trampoline recovers the panic and prints the `failed to exec ... hook` message along with the panic error.
- The instrumented function continues normally, so the panic doesn't reach the host application.
- No stack trace is printed. `OtelGetStackImpl` and `OtelPrintStackImpl` in the target package remain nil, while the symbols generated by `genVarDecl` in the hook package are stripped by the linker.

I also compared the reproduction with and without the linknames generated by `genVarDecl`. The output and execution flow were the same in both cases.

I checked the existing tests and `docs/implementation.md` as well. What seems to be documented is panic isolation/recovery; I couldn't find anything requiring a stack trace to be printed when a hook panics.

So, at least with the current implementation, removing `genVarDecl` doesn't appear to change the panic handling behavior, and it would also remove the duplicate-symbol issue from #1361.

I think it would be useful to add a small regression test for panicking Before/After hooks so we keep the current recovery behavior covered.

Before I rework #1372 based on this, one thing I'd like to confirm: should we also remove the `OtelGetStackImpl` / `OtelPrintStackImpl` variables and their nil checks from `impl.tmpl`, or are those intentionally being kept for a future stack-reporting mechanism?

Happy to update the PR once I know which direction you prefer.

ADITYA-CODE-SOURCE: Good question. Since these variables currently appear disconnected, I would avoid deciding whether to remove the corresponding impl.tmpl code without maintainer confirmation. My suggestion is to keep this PR focused on removing genVarDecl, retain the blank imports, and add the panic recovery regression test you described. The unused template variables can then be removed in the same change if a maintainer confirms stack reporting is no longer intended, or handled separately otherwise.

PR #1374, opened 2026-09-17
feat: add method_call rule

hannahkm:
## Description

Adds a new `method_call` rule that allows for matching at call sites. Supports both plain identifiers (`logger.Info(...)`) and chained selectors (`http.DefaultClient.Do(...)`).

## Motivation

Currently, the `function_call` selector matches only a callee qualified by a package alias (i.e. `net/http.Get`). There is no receiver selector, so `logger.Info(...)` or `http.DefaultClient.Do(...)` cannot be matched at the call site.

Fixes #issue or Related #issue


tool/internal/instrument/apply_call.go
```diff
@@ -278,6 +287,226 @@ func matchesCallRule(call *dst.CallExpr, r *rule.InstCallRule, importAliases map
...
+// "*" on the type name marks a pointer receiver.
+//
+//nolint:revive // confusing-results conflicts with nonamedreturns
+func (pi *methodCallPackageInfo) methodReceiver(file string, line, col int) (string, string, bool) {
+	selection, found := pi.selByPos[methodCallPosition{file: file, line: line, col: col}]
+	if !found {
+		return "", "", false
+	}
+
+	fn, isFunc := selection.Obj().(*types.Func)
```
ADITYA-CODE-SOURCE, line 430: This also accepts `types.MethodExpr`, so a rule for `pkg.Logger.Info` matches both `logger.Info(...)` and the method expression `Logger.Info(logger, ...)`.

The method expression has a different call shape because the receiver is passed explicitly. This appears to be outside the documented `method_call` behavior and may particularly affect `append_args`.

Could we require `selection.Kind() == types.MethodVal` here and add a regression test covering a method expression?

hannahkm: Good point, fixing in [ad13a62](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/pull/1374/commits/ad13a626427daef5f0ca8a46bb0c570a8c451b76)

PR #1377, opened 2026-09-18
fix(instrumentation/log): avoid false trace_id correlation matches

### LINKED ISSUE #1285: bug(instrumentation/log): messages mentioning trace_id skip correlation injection
### Describe the bug

The standard-library `log` instrumentation does not append correlation fields when a log message merely contains the text `trace_id`.

`BeforeLogOutput` currently uses `strings.Contains(string(b), "trace_id")` to determine whether a trace ID is already present. Therefore, a message such as `could not parse trace_id header` is treated as though it already contains a `trace_id=<value>` field. Both the active trace ID and span ID are omitted.

  ### Environment

  - OS: Windows (`windows/amd64`)
  - Go Version: `go1.25.5`
  - Version: commit `3d3d71e1786fb29120a0ce41913ae688fefe266d`

  ### To Reproduce

  1. Add the following temporary test to `instrumentation/log/log_hook_test.go`:

  ```go
  func TestBeforeLogOutput_TraceIDMention(t *testing.T) {
        runtime.RegisterTraceAndSpanIDFunc(func() (string, string) {
                return "abc123", "def456"
        })
        t.Cleanup(func() {
                runtime.RegisterTraceAndSpanIDFunc(func() (string, string) {
                        return "", ""
                })
        })

        ictx := hooktest.NewMockHookContext()
        originalAppend := func(b []byte) []byte {
                return append(b, []byte("could not parse trace_id header\n")...)
        }

        BeforeLogOutput(ictx, nil, 0, 0, originalAppend)

        wrapped := ictx.GetParam(3).(func([]byte) []byte)
        result := string(wrapped(nil))

        assert.Contains(t, result, "trace_id=abc123")
        assert.Contains(t, result, "span_id=def456")
  }
  ```

  2. Run:

  ```console
  cd instrumentation/log
  go test -run '^TestBeforeLogOutput_TraceIDMention$' -count=1 .
  ```

  3. The test fails because the result is only:

  ```text
  could not parse trace_id header
  ```

  ### Expected behavior

  A message that only mentions `trace_id` as ordinary text should still receive the active correlation fields:

  ```text
  could not parse trace_id header trace_id=abc123 span_id=def456
  ```

 Injection should only be skipped when the output already contains an actual trace correlation field, rather than whenever the `trace_id` substring occurs.

  ### Additional context

  The early return occurs in `instrumentation/log/log_hook.go`:

  ```go
  if strings.Contains(string(b), traceIDKey) {
        return b
  }
  ```
The existing `TestBeforeLogOutput_AlreadyContainsTraceID` test covers a real `trace_id=existing` field but does not cover an ordinary message containing the same substring.

 ### Proposed implementation

I have a local prototype that narrows the existing check from the bare `trace_id` substring to the field marker `trace_id=`. This preserves the current behavior for messages that already contain a trace ID while allowing ordinary text such as `could not parse trace_id header` to receive the active `trace_id` and `span_id`.

The prototype also adds a regression test next to `TestBeforeLogOutput_AlreadyContainsTraceID` covering a message that mentions `trace_id` without containing a correlation field.

I noticed that #1215 proposes an allocation-focused refactor of the same hook and states that its behavior should remain unchanged. Could a maintainer confirm whether this behavioral correction should be submitted separately before #1215, or coordinated with that work? I will wait for maintainer confirmation on the preferred direction before opening a PR.

Babul422:
## Description

The standard-library `log` instrumentation treated any occurrence of
`trace_id` in a log message as an existing correlation field.

For example:

could not parse trace_id header

caused `BeforeLogOutput` to skip correlation injection, so the active
`trace_id` and `span_id` were not appended.

This changes the detection logic to look for the actual `trace_id=`
correlation marker while preserving existing behavior for real
correlation fields.

## Changes

- Added `hasTraceID` to detect `trace_id=` without matching ordinary
  mentions of `trace_id`.
- Added identifier-boundary handling so fields such as
  `other_trace_id=...` do not suppress correlation injection.
- Added regression coverage for:
  - plain-text `trace_id` mentions
  - `other_trace_id=...`
  - bracketed `trace_id=...`
  - `trace_id=...` at the beginning of the output
- Verified that the new detection path performs zero heap allocations.

## Testing

- `go test -C instrumentation/log -v -count=1 ./...`
- `go test -C instrumentation/log/slog -v -count=1 ./...`
- `go test -C instrumentation/github.com/sirupsen/logrus -v -count=1 ./...`
- `go test -C instrumentation/go.uber.org/zap -v -count=1 ./...`
- `./.bin/golangci-lint run --config .tools/golangci.yml --new-from-rev=upstream/main`
- `git diff --check`

All checks passed.

Fixes #1285

## AI Assistance

This contribution was developed with AI assistance.


ADITYA-CODE-SOURCE (approved):
Reviewed the marker detection, identifier-boundary handling, standard logger formatting paths, and regression coverage. The change narrows duplicate-correlation detection without affecting ordinary trace_id mentions, and the tests cover the important boundary cases. Looks good to me.

instrumentation/log/log_hook.go
```diff
@@ -15,8 +16,31 @@ const (
 	instrumentationKey = "logs/log"
 	traceIDKey         = "trace_id"
 	spanIDKey          = "span_id"
+	traceIDMarker      = traceIDKey + "="
```
atharvamhaske, line 19: This looks for `trace_id=` and ignores a match inside a longer name such as other_trace_id=. A line like
  could not parse trace_id header still receives trace_id=. The tests cover these cases.

  The helper does not find JSON or slog fields of the form "`trace_id":"...".` That form has no =. This is
  correct for `log.Logger` only. If you reuse this helper for zap or slog, those lines will not match.

Babul422: Thanks for calling that out! Yes, that's the intent — this helper is scoped to the standard log.Logger output. Structured formats like slog or zap are outside the scope of this PR and would need separate handling if supported in the future.

instrumentation/log/log_hook.go
```diff
@@ -15,8 +16,31 @@ const (
...
+	}
+	return false
+}
+
+func isIdentChar(c byte) bool {
+	return (c >= 'a' && c <= 'z') ||
+		(c >= 'A' && c <= 'Z') ||
+		(c >= '0' && c <= '9') ||
+		c == '_'
+}
```
amazingakai, line 25: IMO the `isIdentChar` boundary checking is counterintuitive. Since the instrumentation injects the `trace_id` field preceded by a space (`" trace_id="`), we can keep the check simple by checking for `" trace_id="`. Even if a user has manually formatted their log as `[trace_id=123]`, having our instrumentation append `" trace_id=123 span_id=456"` at the end is differentiable. WDYT?

Babul422: Make sense, since the instrumentation always injects with a leading space, so matching on " trace_id=" is easier and no extra boundary logic needed.

I’ve modified hasTraceID to work on bytes. isIdentChar with " trace_id=" removed Contains and tests accordingly affected. Also I checked that all tests in instrumentation/log pass.

instrumentation/log/log_hook_test.go
```diff
@@ -214,3 +310,35 @@ func TestBeforeLogOutput_PreservesLineEnding(t *testing.T) {
...
+}
+
+func BenchmarkHasTraceID_Match(b *testing.B) {
+	msg := []byte("2026/09/18 10:00:00 standard log message trace_id=abcdef123456\n")
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_ = hasTraceID(msg)
+	}
+}
```
amazingakai: Could you please remove the `BenchmarkHasTraceID_NoMatch` and `BenchmarkHasTraceID_Match` functions? We don't run instrumentation-level micro-benchmarks in CI (the CI only runs compiler benchmarks in `test/bench`), so these will just end up as dead code. I  
 noticed we have some micro-benchmarks in some other places as well that are never run and can be deleted in a folllow-up PR.

amazingakai:
The bundle also needs to be regenerated

instrumentation/log/log_hook_test.go
Babul422: Done! Removed both benchmarks and regenerated the bundle archive. Thanks for the review!

instrumentation/log/log_hook.go
```diff
@@ -15,8 +16,13 @@ const (
 	instrumentationKey = "logs/log"
 	traceIDKey         = "trace_id"
 	spanIDKey          = "span_id"
+	traceIDMarker      = " " + traceIDKey + "="
 )
 
+func hasTraceID(b []byte) bool {
+	return bytes.Contains(b, []byte(traceIDMarker))
```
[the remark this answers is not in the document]
amazingakai: Copilot has a good point about the start-of-output case. I dont think we need the full boundary-aware check, but could you update `hasTraceID` to also treat a leading `trace_id=` as an existing correlation field? Other than that, this looks good to me.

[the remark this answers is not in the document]
Babul422: @amazingakai done !

